PR #20, opened 2019-08-01
feat: ComponentHeader component, component overview fixes, more

TheSisb:
## Structural changes

- [x] Swap the default theme border-color light/dark token to match Starch
- [x] Shuffle around the folder and files in the paste-website package.  Having folders of just index files is more cumbersome than just putting the file in that parent folder directly.
- [x] Removed website building from `yarn build` command, but added it to CircleCI.  When we run `yarn build` locally, we usually don't want to build the website, so this makes it faster.
- [x] Our pre-push git hook now doesn't run build anymore.  It took way too long.
- [x] Added `yarn type-check` to CircleCI and on `pre-push` git hook.
- [x] Fix some type-check errors.

## Feature changes

- [x] Add breadcrumb component for doc site
- [x] Fixed border issues on Sidebar
- [x] Keep sidebar category open when navigating to subpages
- [x] Add `ComponentHeader` shortcode component
- [x] Improve `ComponentOverviewTable` design


packages/paste-theme-tokens/src/sendgrid/index.ts
```diff
@@ -24,6 +24,7 @@ const lineHeights = {
   lineHeight50: '1.75rem', // #28px
   lineHeight60: '2rem', // #32px
   lineHeight70: '2.25rem', // #36px
+  lineHeight80: '2.5rem', // #40px
```
TheSisb: Adding this newly needed token

packages/paste-website/src/components/shortcodes/ComponentHeader.tsx
```diff
@@ -0,0 +1,92 @@
...
+  }
+  const {description, status, name: packageName, version} = data[0].node;
+
+  return (
+    <>
+      <ComponentHeaderBasic name={name} />
+      <Text fontSize="fontSize30" mb="space50">
+        {description}
+      </Text>
+      <Box mb="space100">
```
SiTaggart: Description list for the key-value pairs?

packages/paste-website/src/components/Input.tsx
```diff
@@ -2,18 +2,16 @@ import * as React from 'react';
 import styled from '@emotion/styled';
```
SiTaggart: I'm generally of the opinion that every component should eventually have some tests, even if it's just a snapshot.

Is it worth moving these to named files, given we'll create a `__test__` folder for each under it's own directory in components?

TheSisb: Hmmm.. we can create sibling `Input.spec.ts` files in the same folder.  Thoughts?

SiTaggart: kinda messy, and isn't the convention `__test__` folders?

packages/paste-website/src/components/Breadcrumb.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+const CrumbSlash: React.FC<{}> = () => <CrumbSlashStyled>/</CrumbSlashStyled>;
+
+const Breadcrumb: React.FC<{}> = ({children}) => {
+  return (
+    <>
+      {React.Children.map(children, child => (
+        <>
+          {child}
+          <CrumbSlash />
```
richbachman: Can we add a check to see not show this `<CrumbSlash />` after the last breadcrumb. Not a blocker, just a thought.

packages/paste-website/src/components/Input.tsx
TheSisb: It isn't too bad.  Kind of looks like https://github.com/modulz/radix/tree/master/packages/radix/src/components

Or we can still have a __tests__ folder with all the component filenames repeated like
```
| __tests__
   -| Input.tsx
   -| ...
| Input.tsx
```

packages/paste-website/src/components/Breadcrumb.tsx
TheSisb: This is what the design wanted /shrug

packages/paste-website/src/components/shortcodes/ComponentHeader.tsx
TheSisb: Nice, added.

packages/paste-website/src/components/Input.tsx
TheSisb: Reverted the changes in this PR for now.

packages/paste-design-tokens/formatters/sketchpalette.ts
```diff
@@ -17,6 +17,7 @@ interface ColorRGB {
 }
 
 const convertToSketchPaletteColor = (input: string): {[key: string]: number} => {
+  // @ts-ignore FIXME
```
TheSisb: We should fix this, the typing is way off.  Didn't want it to block the PR though.

packages/paste-website/src/components/tokens-list/index.tsx
```diff
@@ -106,8 +106,8 @@ export const TokensList: React.FC<TokensListProps> = props => {
                   <thead>
                     <Tr>
                       <Th>Token</Th>
-                      <Th width="250px">Value</Th>
-                      <Th width="250px">Example</Th>
+                      <Th style={{width: '250px'}}>Value</Th>
```
TheSisb: I realized we could revert the changes we made to the table component and use an inline-style instead.

packages/paste-core/utilities/text/src/index.tsx
```diff
@@ -16,7 +16,6 @@ import {ThemeShape} from '@twilio-paste/theme-tokens';
 
 interface MarginProps {
   m?: keyof ThemeShape['space'];
-  margin?: keyof ThemeShape['space'];
```
SiTaggart: Curious why you chose to remove just `margin` from the interface?

TheSisb:
Yeah, I totally agree.  I hadn't realized at the time that it would balloon this much.  I could still split it up if you'd like.

packages/paste-core/utilities/text/src/index.tsx
TheSisb: I was getting a typescript error about it... but I am not anymore.  Re-added it

packages/paste-theme-tokens/package.json
```diff
@@ -13,7 +13,7 @@
     "dist"
   ],
   "scripts": {
-    "build": "yarn clean && yarn compile",
```
SiTaggart: This one doesn't need a clean?

TheSisb: Good catch.  I think this is an artifact from when I was debugging the js files being built in the src folder.

packages/paste-website/src/components/breadcrumb/index.tsx
```diff
@@ -0,0 +1,35 @@
...
+        </>
+      ))}
+    </>
+  );
+};
+
+const BreadcrumbItem = styled(Link)`
+  height: 24px;
+  font-size: 14px;
+  line-height: 1.71;
```
SiTaggart: Can you write a TODO here to convert these to tokens? line-heights are coming and font-size should already cover it.

Also is the height going to cause a problem if you bump the base font size for the page? Will the font scale (when using rems from tokens) and the fixed height cause issues?

PR #31, opened 2019-08-13
Landing page

richbachman:
- setup landing page structure
- renamed icons folder to sync up with another PR
- added terminal, brush, window-layout icons
- moved `StyledHr` to component file
- added `StyledHr` to mdx provider
- removed unused header from mdx provider

**Contributing to Twilio**


packages/paste-website/src/pages/index.tsx
```diff
@@ -1,10 +1,111 @@
...
+import {Box} from '@twilio-paste/box';
+import {Text} from '@twilio-paste/text';
 import {SiteWrapper} from '../components/default-site-wrapper';
+import {Grid} from '../components/grid';
+import {StyledHr} from '../components/StyledHr';
+import {Brush} from '../components/icons/Brush';
+import {Terminal} from '../components/icons/Terminal';
+import {WindowLayout} from '../components/icons/WindowLayout';
+
+const LandingContainer = styled(Box)`
```
richbachman: We need a way to use `auto` for `margin-right` and `margin-right`. Once we have that we can get rid of this styled component.

packages/paste-website/src/pages/index.tsx
```diff
@@ -1,10 +1,111 @@
...
+  margin-right: auto;
+  margin-left: auto;
+`;
+
+const StyledButton = styled(Button)`
+  margin-right: ${themeGet('space.space40')};
+  text-decoration: none;
+`;
+
+const IconCircle = styled(Box)`
```
richbachman: I was mostly able to use the box component, but still needed the `justify-content` and `align-items` values to style this to match the comp. I also kept the `width` and `height` values here because the we're using our `size` tokens for those props, and we currently don't have a size token for `60px`.

packages/paste-website/src/pages/index.tsx
SiTaggart: It doesn't work? I kind of was working under the assumption it was working

packages/paste-website/src/pages/index.tsx
SiTaggart: We need a flexbox component so bad right now

packages/paste-website/src/pages/index.tsx
```diff
@@ -1,10 +1,111 @@
...
+
+const LandingContainer = styled(Box)`
+  margin-right: auto;
+  margin-left: auto;
+`;
+
+const StyledButton = styled(Button)`
+  margin-right: ${themeGet('space.space40')};
+  text-decoration: none;
+`;
```
SiTaggart: This is interesting, I was not expecting this to work _at all_ as we don't expose the classname prop I don't think. If that's true you can extend any component from the system... I'm not sure how I feel about that 🤔

packages/paste-website/src/pages/index.tsx
richbachman: @SiTaggart I'm getting the following typescript error when trying to use `auto`:

```
Type '"auto"' is not assignable to type '"space0" | "space10" | "space100" | "space110" | "space120" | "space130" | "space140" | "space150" | "space160" | "space170" | "space180" | "space190" | "space20" | "space200" | "space30" | "space40" | "space50" | "space60" | "space70" | "space80" | "space90" | undefined'.
```

packages/paste-website/src/pages/index.tsx
richbachman: I noticed it wasn't working as well on the preview. That might be happening because of the button styles bug @TheSisb was looking into. We're getting a lot of `undefined` styles on the docs site for some reason. Once that is fixed, the `margin-right` might fix as well.

I'll go ahead and change the `text-decoration` at the button level.

packages/paste-website/src/pages/index.tsx
SiTaggart: ah, typings, right. You can add it to the typings if you wish. Otherwise this is totally fine.

packages/paste-website/src/pages/index.tsx
SiTaggart: yeah the undefined's are all autoprefixer based, everything that's not being auto prefixed seems to work just fine. Due to `Button` not exposing `classname` I full expect this particular code to not work though

richbachman: Right. That makes sense. Adding a nested style to the outer container to get the margin I want.

packages/paste-website/src/pages/index.tsx
richbachman: I'm going to leave as is for now. I wasn't sure how to add the type to the key without updating the `ThemeShape`

SiTaggart: Perfectly fine. I created a JIra ticket for it

packages/paste-website/src/pages/index.tsx
SiTaggart: It'll potentially break in the future, I was thinking more like:

```
<StyledButtonRow marginTop="space80" marginBottom="space80">
  <Box mr="space40">
    <Button as="a" href="/" size="default" variant="primary">
      Install Paste 0.1
    </Button>
  </Box>
  <Button as="a" href="/components" size="default" variant="secondary">
    View Components
  </Button>
</StyledButtonRow>```

richbachman: Ah I see. Worth changing now, or should I just run with what I have?

SiTaggart: Entirely up to you, but we should change it eventually

richbachman: Easy fix, so got it in. Had to add a `display` prop as well to make it all inline.

PR #34, opened 2019-08-14
feat: change the z-index scale to 10s

SiTaggart:
Update the z-index scale to use 10s as per @TheSisb's suggestion


TheSisb:
In the future I could see this being a breaking change

SiTaggart:
Yup, agree

PR #35, opened 2019-08-14
feat(tokens): Add box-shadow and line-height tokens

SiTaggart:
Adding some missing tokens:

- Focus ring box-shadow should ideally be the same style across all focusable elements. This is currently just using the one we use for Button (and the yet to be built new Input design by Sarah) 
- Line height tokens. These are tightly coupled to the font-size due to our use of unit less line-height values. I think this is fine. Using the font-size*1.5 to the nearest 4 rule that came out of the Typography work recently. @two24studios double check you're OK with the values for SendGrid based on that calculation
- Fix some references to the old font size token naming convention

### DIFF
--- a/packages/paste-design-tokens/tokens/global/line-height.yml
+++ b/packages/paste-design-tokens/tokens/global/line-height.yml
@@ -3,16 +3,16 @@ global:
   type: number
 props:
   line-height-10:
-    value: "1.67" ##12px * 1.67 = 20px
+    value: "1.666666667" ##12px * 1.666666667 = 20px
     comment: "Constant line-height token for font size 10"
   line-height-20:
-    value: "1.71" ##14px * 1.71 = 24px
+    value: "1.714285714" ##14px * 1.714285714 = 24px
     comment: "Constant line-height token for font size 20"
   line-height-30:
     value: "1.5" ##16px * 1.5 = 24px
     comment: "Constant line-height token for font size 30"
   line-height-40:
-    value: "1.56" ##18px * 1.56 = 28px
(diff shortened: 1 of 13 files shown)


packages/paste-design-tokens/tokens/global/line-height.yml
```diff
@@ -0,0 +1,34 @@
+global:
+  category: line-height
+  type: number
+props:
+  line-height-10:
+    value: "1.67" ##12px * 1.67 = 20px
```
TheSisb: The ordering of these is weird.  Is that because they map 1:1 to fontSizes?

TheSisb: Would this be a breaking change?

SiTaggart: Yeah, because we use unitless values, we have to have a 1:1 mapping to the corresponding font size otherwise we won't get whole numbers for line-height. The line-height scale is based on font-size*1.5 rounded _up_ to the nearest 4

SiTaggart: It's not a breaking change in that we never had any, and the values being replaced in the hard coded theme object were based on the old typographic scale, but the new ones aren't too far off

TheSisb: We had some in paste-theme-tokens/src/default/index.ts

image

Which are being used already like in the website breadcrumb component

SiTaggart: Fixed up @TheSisb. Each font size now uses the correct line-height token in this branch.

PR #38, opened 2019-08-14
feat(website): website footer

two24studios:
Footer for the Paste docs website.

**Contributing to Twilio**

### DIFF
--- a/packages/paste-website/src/components/Footer.tsx
+++ b/packages/paste-website/src/components/Footer.tsx
@@ -5,7 +5,7 @@ import {Link} from 'gatsby';
 
 import {Anchor} from '@twilio-paste/anchor';
 import {TWILIO_BLUE} from '../constants';
-import TwilioIcon from './svg/twilio-icon';
+import {TwilioIcon} from './icons/TwilioIcon';
 
 const StyledFooter = styled.footer`
   margin-top: 96px;
--- a/packages/paste-website/src/components/Footer.tsx
+++ b/packages/paste-website/src/components/Footer.tsx
@@ -120,17 +120,23 @@ const Footer: React.FC<FooterProps> = () => {
         <Styled4Column>
           <div>
             <h4>Support</h4>
-            <p>Vinyl venmo gochujang letterpress pabst drinking vinegar, franzen bitters schlitz fam.</p>
+            <p>
+              If you need support, please open a new issue in our GitHub repository. Please try to provide as much
+              detail as possible in your issue.
+            </p>
             <StyledLinkList>
-              <Anchor href="#">Ask a Question</Anchor>
-              <Anchor href="#">Report a Bug</Anchor>
+              <Anchor href="https://github.com/twilio-labs/paste/issues">Ask a Question</Anchor>
+              <Anchor href="https://github.com/twilio-labs/paste/issues">Report a Bug</Anchor>
             </StyledLinkList>
           </div>
           <div>
-            <h4>Support</h4>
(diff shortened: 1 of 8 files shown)


packages/paste-website/src/components/Footer.tsx
```diff
@@ -0,0 +1,146 @@
...
+import {Link} from 'gatsby';
+
+import {Anchor} from '@twilio-paste/anchor';
+import {TWILIO_BLUE} from '../constants';
+import TwilioIcon from './svg/twilio-icon';
+
+const StyledFooter = styled.footer`
+  margin-top: 96px;
+  padding: 120px ${themeGet('space.space200')} 120px;
+  background-color: #f9fafb;
```
SiTaggart: This could be `$color-background-row-striped` (`backgrounds.colorBackgroundRowStriped`) or potentially `$color-gray-10` (`colors.colorGray10`) tokens. I'm not sure how I feel about the grey-scale tokens and was potentially going to remove them. 

I kind of wanted very specific tokens for specific purposes, but this might (re)convince me that the any purpose gray tokens are _"OK_"?

packages/paste-website/src/components/Footer.tsx
```diff
@@ -0,0 +1,146 @@
...
+`;
+
+const StyledFooterGrid = styled.div`
+  display: grid;
```
SiTaggart: Are you ok with a stacked fall back for IE 11 on these columns?

packages/paste-website/src/components/Footer.tsx
```diff
@@ -0,0 +1,146 @@
...
+  padding: 0;
+  list-style: none;
+`;
+
+const StyledFooterLink = styled(Link)`
+  color: #465672;
+  text-decoration: none;
+
+  &:hover {
+    color: #08152c;
```
SiTaggart: Token `$color-text` color?

packages/paste-website/src/components/Footer.tsx
two24studios: I think for 0.1 I'm ok with that. We should define some guidelines around browser support for the docs site in a future sprint. I think since most of initial users will be internal and likely on macOS we'll be ok for the interim. But, knowing we are open source and likely will have visitors from all different types of browsers and operating systems, we should clearly define what we will support.

packages/paste-website/src/components/Footer.tsx
```diff
@@ -0,0 +1,152 @@
...
+    color: ${themeGet('textColors.colorText')};
+  }
+`;
+
+const Styled4Column = styled.div`
+  grid-column: 9 / span 4;
+`;
+
+const StyledLinkList = styled.p`
+  a {
```
TheSisb: cc @SiTaggart this kind of makes me think all components should accept margin tokens.  Thoughts?

TheSisb:
I made some tweaks for the footer to look like this 
image
Feel free to undo my commit if you don't like it.

packages/paste-website/src/components/site-wrapper/SiteFooter.tsx
```diff
@@ -0,0 +1,65 @@
...
+    <StyledFooter>
+      <Box
+        css={{
+          margin: '0 auto',
+          maxWidth: '640px',
+        }}
+      >
+        <Flex>
+          <Box mr="space60">
+            <h4>Support</h4>
```
SiTaggart: can you use the `Heading` and `P` components here at all for site consistency?

PR #39, opened 2019-08-14
chore(domain): alias the now production to the domain

SiTaggart:
Add domain aliasing to the production deployments.

### DIFF
--- a/now.json
+++ b/now.json
@@ -8,5 +8,8 @@
       "use": "@now/static-build",
       "config": {"distDir": "packages/paste-website/public"}
     }
+  ],
+  "alias": [
+    "paste.twilio.design"
   ]
 }


now.json
```diff
@@ -8,5 +8,8 @@
       "use": "@now/static-build",
       "config": {"distDir": "packages/paste-website/public"}
     }
+  ],
+  "alias": [
```
TheSisb: Is this needed?  According to the docs this gets ignored if configured through the dashboard https://zeit.co/docs/v2/advanced/configuration/#alias

SiTaggart: I added this as a “just in case” the domain didn’t alias automatically (I had some issues with both my sites previously). I can close this now

PR #63, opened 2019-08-19
fix(website): changed p to Text

richbachman:
Error was occurring because a div is used in `TwilioIcon`.

Changed the `p` tag in the footer to use `Text` and changed `TwilioIcon` to use a `span` instead of a `div`. I also added a `display` prop to `TwilioIcon`.

**Contributing to Twilio**

### DIFF
--- a/packages/paste-website/src/components/icons/TwilioIcon.tsx
+++ b/packages/paste-website/src/components/icons/TwilioIcon.tsx
@@ -11,7 +11,7 @@ export interface TwilioIconProps {
 }
 
 const TwilioIcon = React.memo(
-  ({title = 'Twilio Icon', decorative = true, className, color, size, display = 'block'}: TwilioIconProps) => {
+  ({title = 'Twilio Icon', className, color, decorative = true, display, size}: TwilioIconProps) => {
     const uid = useUID();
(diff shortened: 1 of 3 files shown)


SiTaggart:
Would a better fix be changing the Icon wrappers to `span`s so that they can be placed inside of phrasing content?

richbachman:
@SiTaggart yeah I think that would be a better fix. We can leave the footer as is for now. @TheSisb any objections to changing the `div`s to `span`s in the `svg-to-react` repo?

TheSisb:
Yeah that's a good point.  I'll make a ticket to fix that.  

In paste-icons for example, we wrap with 
```
const IconWrapper = styled.span`
  ${({size = 24, color = 'currentColor'}: IconWrapperProps) => `
    width: ${size}px;
    height: ${size}px;
    color: ${color};
    display: inline-block;
  `}
`;
```
which uses a `display: inline-block` span wrapper.  I think the default "out of the box" template should be similar.  👍

richbachman:
I've manually set the Twilio icon to use a span for this one, and added a display prop. I also changed the `p` tag to use `Text`.

Issue #61, opened 2019-08-18
Utility shows up as "Component" on click


aayushpi:
Steps to reproduce:
1. Open Utilities List: `https://paste.twilio.design/utilities`
2. Click on any utility. For example, `Text` in `https://paste.twilio.design/utilities`
3. You will see that the path changes to `https://paste.twilio.design/components/text` instead of staying in "Utilities"

Also, the page title shows up as "404"

TheSisb:
Opened a PR https://github.com/twilio-labs/paste/pull/64 to fix this

PR #64, opened 2019-08-20
fix(website): component header and overview with utilities

TheSisb:
Fixes ComponentOverviewTable and ComponentHeader to work with `utility` components. I add ComponentHeader into the Box documentation page to test it out.

Fixes https://github.com/twilio-labs/paste/issues/61

**Contributing to Twilio**

### DIFF
--- a/packages/paste-website/src/components/shortcodes/component-header/index.tsx
+++ b/packages/paste-website/src/components/shortcodes/component-header/index.tsx
@@ -7,11 +7,16 @@ import {SidebarCategoryRoutes} from '../../../constants';
 import {P} from '../../Typography';
 import {Heading} from '../../Heading';
 
-const ComponentHeaderBasic: React.FC<{name: string}> = ({name}) => (
+const ComponentHeaderBasic: React.FC<{
+  name: string;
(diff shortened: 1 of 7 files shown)


packages/paste-website/src/components/shortcodes/component-header/index.tsx
```diff
@@ -7,11 +7,16 @@ import {SidebarCategoryRoutes} from '../../../constants';
 import {P} from '../../Typography';
 import {Heading} from '../../Heading';
 
-const ComponentHeaderBasic: React.FC<{name: string}> = ({name}) => (
+const ComponentHeaderBasic: React.FC<{
+  name: string;
```
richbachman: Just a question. Why set types here over using `interface`?

TheSisb: So a couple thoughts:
1- The interface doesn't get re-used, so I just define it inline.
2- Naming interfaces takes effort, and the name is somewhat meaningless here.  I equate this to using the `css` prop rather than naming and creating a styled component.

richbachman: 👍 thanks for the explanation. Appreciate it.

PR #82, opened 2019-09-05
feat(website): add theme switcher and theme switching 

SiTaggart:
Add

- theme switcher control to the site header
- ability to switch the token values based on the theme switcher on the token page
- ability to show a live demo component based on the site theme switcher

**Contributing to Twilio**


packages/paste-website/src/components/site-wrapper/index.tsx
```diff
@@ -13,15 +14,17 @@ interface SiteWrapperProps {
...
-      <Global styles={globalStyles} />
-      <SiteBody>
-        <Sidebar />
-        <SiteMain>
-          <SiteStickyHeader />
-          <SiteMainInner>{props.children}</SiteMainInner>
-          <SiteFooter />
-        </SiteMain>
-      </SiteBody>
+      <SiteThemeProvider>
```
SiTaggart: this is a confusing name

packages/paste-website/src/components/table/index.tsx
```diff
@@ -26,7 +26,7 @@ export const Tbody = styled.tbody(props => ({
 }));
 
 export const Tr = styled.tr(props => ({
-  '&:nth-child(even)': {
+  '&:nth-of-type(even)': {
```
SiTaggart: clearing a console error locally

packages/paste-website/src/context/SiteThemeContext.tsx
```diff
@@ -0,0 +1,47 @@
+import * as React from 'react';
+import {Themes} from '../constants';
+
+export type SiteTheme = 'default' | 'sendgrid';
```
SiTaggart: I would love to create this type from the constant object `Themes` but that seems to be beyond my Typescript ability right now. 🤷‍♂

packages/paste-website/src/components/ThemeSwitcher.tsx
```diff
@@ -1,13 +1,77 @@
...
+  display: 'inline-block',
+  padding: `${themeGet('space.space20')(props)} ${themeGet('space.space40')(props)}`,
+  '&:hover': {
+    textDecoration: 'underline',
+  },
+}));
+
+const StyledThemeSwitcherRadio = styled.input<{}>(props => ({
+  opacity: 0,
+  position: 'absolute',
```
SiTaggart: using opacity and pos absolute instead of a more traditional "off screen" solution due to bug with Dragon Naturally speaking (most common dictation software) that we found at Salesforce where it couldn't find checkboxes or radio buttons if they were visually hidden any other way than opacity.

packages/paste-website/src/components/callout/types.ts
```diff
@@ -8,13 +9,6 @@ export interface CalloutTitleProps {
...
 
-export interface CalloutProps {
-  marginTop?: string;
-  marginBottom?: string;
-  paddingTop?: string;
-  paddingRight?: string;
-  paddingBottom?: string;
-  paddingLeft?: string;
-  children?: React.ReactNode;
+export interface CalloutProps extends BoxProps {
```
SiTaggart: Typescript suddenly started complaining about this. Dunno why

packages/paste-website/src/context/SiteThemeContext.tsx
TheSisb: Kind of a mouthful but:
```
const Themes = {DARK: 'dark', LIGHT: 'light'} as const // add `as const`

type ValueOf<O> = O[keyof O]
type Theme = ValueOf<typeof Themes> // 'dark' | 'light'
```

SiTaggart: Niiiiiiiice 😎

packages/paste-website/src/components/ThemeSwitcher.tsx
```diff
@@ -1,13 +1,79 @@
...
+    backgroundColor: themeGet('backgroundColors.colorBackgroundPrimary')(props),
+    borderRadius: themeGet('radii.borderRadius10')(props),
+    color: themeGet('textColors.colorTextInverse')(props),
+  },
+}));
+
 export const ThemeSwitcher: React.FC<ThemeSwitcherProps> = () => {
+  const {theme, updateActiveSiteTheme} = useActiveSiteTheme();
+
+  const handleChange = (event: React.FormEvent<HTMLInputElement>): void => {
```
TheSisb: Been doing some reading on Typescript and I wanted to offer an alternative typing approach similar to how we type components (`React.FC<>`):

```js
const handleChange: React.ChangeEventHandler<HTMLInputElement> = event => {
```



> The first method uses an inferred method signature (e: React.FormEvent<HTMLInputElement>): void and the second method enforces a type of the delegate provided by @types/react. So React.ChangeEventHandler<> is simply a "blessed" typing by @types/react, whereas you can think of the inferred method as more... artisanally hand-rolled. Either way it's a good pattern to know.

Source: https://github.com/typescript-cheatsheets/react-typescript-cheatsheet#forms-and-events

packages/paste-website/src/components/ThemeSwitcher.tsx
```diff
@@ -1,13 +1,79 @@
...
+    <Box
+      as="form"
+      borderColor="colorBorderLight"
+      borderRadius="borderRadius20"
+      borderStyle="solid"
+      borderWidth="borderWidth10"
+      padding="space10"
+    >
+      <Box as="fieldset" borderWidth="borderWidth0" padding="space0" margin="space0">
+        <legend>
```
TheSisb: Could this be `<ScreenReaderOnly as="legend">Change the site theme</ScreenReaderOnly>`?  If not, should the legend tag be inside the `ScreenReaderOnly` element?

packages/paste-website/src/components/ThemeSwitcher.tsx
```diff
@@ -1,13 +1,79 @@
...
+          onChange={handleChange}
+          type="radio"
+          value={Themes.CONSOLE}
+        />
+        <StyledThemeSwitcherLabel htmlFor="console">Console</StyledThemeSwitcherLabel>
+        <StyledThemeSwitcherRadio
+          checked={theme === Themes.SENDGRID}
+          id="sendgrid"
+          name="sitetheme"
+          onClick={handleChange}
```
TheSisb: Do we need these `onClick`?

packages/paste-website/src/components/ThemeSwitcher.tsx
```diff
@@ -1,13 +1,79 @@
...
+          onChange={handleChange}
+          type="radio"
+          value={Themes.CONSOLE}
+        />
+        <StyledThemeSwitcherLabel htmlFor="console">Console</StyledThemeSwitcherLabel>
+        <StyledThemeSwitcherRadio
+          checked={theme === Themes.SENDGRID}
+          id="sendgrid"
```
TheSisb: We should use react-uid for these as these `id`s are very generic and easy to clash with.

packages/paste-website/src/components/ThemeSwitcher.tsx
richbachman: Good to know.

packages/paste-website/src/components/ThemeSwitcher.tsx
SiTaggart: Let me try the screenreaderonly as approach, that should do it. `legend` must be the first child of `fieldset` sop I can't wrap it.

packages/paste-website/src/components/ThemeSwitcher.tsx
SiTaggart: I dig it

packages/paste-website/src/components/ThemeSwitcher.tsx
SiTaggart: This worked, I just needed to add the `as` prop to `screenReaderOnly`

SiTaggart:
Ready for re-review @TheSisb @richbachman

packages/paste-core/utilities/screen-reader-only/src/index.tsx
```diff
@@ -1,6 +1,11 @@
 import styled from '@emotion/styled';
 
-export const ScreenReaderOnly = styled.span<{}>`
+interface ScreenReaderOnlyProps {
+  // as prop isn't typed correctly from styled package https://github.com/emotion-js/emotion/issues/1137
+  as?: keyof JSX.IntrinsicElements;
+}
+
+export const ScreenReaderOnly = styled.div<ScreenReaderOnlyProps>`
```
SiTaggart: Ah crap, I need to swap this back to a span from my testing why the as prop wasn’t typing properly.

PR #88, opened 2019-09-09
feat(website): hyperlink doc site headers

TheSisb:
This PR modified the Headers in the doc site to prefix an anchor hyperlink and `id` if children is a simple string type.  Navigating to a page with a hash in the URL will scroll to the appropriate section automatically.

- Refactored site-wrapper layout and css
- Added scrollIntoView mechanism
- Added two new helper functions to RouteUtils
- Moved magic CSS values into constants file


packages/paste-website/src/components/site-wrapper/sidebar/Header.tsx
```diff
@@ -3,8 +3,8 @@ import {Text} from '@twilio-paste/text';
 import styled from '@emotion/styled';
```
TheSisb: I moved a bunch of files related to `site-wrapper` into that folder to make it easier to work on this stuff.

richbachman:
This all lgtm, and thanks for the explanation comments. Really helps with the context.

packages/paste-website/src/components/Heading.tsx
```diff
@@ -67,15 +65,47 @@ function getHeadingStyles(headingStyle?: HeadingStyle): {} {
...
+    text-decoration: underline;
+  }
+`;
+
+const AnchoredHeading: React.FC<HeadingProps> = props => {
+  // Only generate slugs for headings where children is 'string'
+  if (typeof props.children === 'string') {
+    const id = slugify(props.children);
+    return (
+      <Text {...props} id={id} textColor="colorText" {...getHeadingStyles(props.headingStyle)}>
```
SiTaggart: nit: I guess you could have just used `Heading` here instead to potentially save any future double refactoring.

```
<Heading {...props} id={id} headingStyle={props.headingStyle}>
  {props.children}
  <SteyledAnchorHyperlink href={...}>#</StyledAnhorHyperlink>
</Heading>
```

packages/paste-website/src/pages/index.tsx
```diff
@@ -39,7 +39,7 @@ const IndexPage: React.FC<{}> = (): React.ReactElement => {
         <meta name="description" content={SiteMetaDefaults.DESCRIPTION} />
       </Helmet>
       <LandingContainer maxWidth="size60" marginTop="space200" marginBottom="space200">
-        <Heading as="h1" headingStyle="headingStyle10">
+        <Heading as="h2" headingStyle="headingStyle10">
```
SiTaggart: What's the reasoning behind bumping the heading Heirarchy down one? There's now no `h1` on the landing page.

TheSisb: I noticed that component pages were using `h1` for the component header so this one would clash with that.  Thinking on it, I figured the component page h1 is more meaningful as it changes from page to page.  I don't think setting h1 on `Paste` is very meaningful semantically, and the SEO should be fine as `Paste` appears in the URL and pretty much everywhere a bunch already.

SiTaggart: So, we need one per page. Even if we change the main title from Paste to something else. Even if we don't think it's a good title for the page. You could even hide some text inside of it to make it more meaningful but keep the visuals the same, but we do need one. Lighthouse complains and it'll be the first thing someone picks up on from an Accessibility audit.

TheSisb: You're totally right 😅 I mistook this page for the `Paste` in the sidebar.  I'll revert this.

SiTaggart: ooooooooooooohhhhhh 🤣  Yeah, that being an `h2` also bugs me, but not in the scope of this work

TheSisb:
@SiTaggart ready for re-review

PR #91, opened 2019-09-09
fix(anchor): set children as non-nullable, remove child null error

richbachman:
- [x] Change anchor children to non-nullable type
- [x] Remove child null error

**Contributing to Twilio**

### DIFF
--- a/packages/paste-core/components/anchor/src/index.tsx
+++ b/packages/paste-core/components/anchor/src/index.tsx
@@ -8,7 +8,7 @@ const EXTERNAL_REL_DEFAULT = 'noreferrer noopener';
 
 const isExternalUrl = (url: string): boolean => EXTERNAL_LINK_REGEX.test(url);
 
-const handlePropValidation = ({href, tabIndex, children}: AnchorProps): void => {
+const handlePropValidation = ({href, tabIndex}: AnchorProps): void => {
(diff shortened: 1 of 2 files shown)


packages/paste-core/components/anchor/src/index.tsx
```diff
@@ -18,10 +18,6 @@ const handlePropValidation = ({href, tabIndex, children}: AnchorProps): void =>
     );
   }
 
-  if (children == null) {
```
TheSisb: For the time being, let's leave this error in.  Some consumers of this component might be in a pure JS ecosystem, so they won't get the typescript error.  I'm considering doing a TS -> React.propTypes thing at some point.

packages/paste-core/components/anchor/src/types.ts
```diff
@@ -9,7 +9,7 @@ export interface AnchorProps {
   target?: AnchorTargets;
   rel?: string;
 
-  children: React.ReactNode;
+  children: NonNullable<React.ReactNode>;
```
TheSisb: 👌

packages/paste-core/components/anchor/src/index.tsx
richbachman: I added that back in.

PR #96, opened 2019-09-12
chore(website): change casing of components in sidebar

TheSisb:
Simple tweak:

*Before*
image

*After*
image

### DIFF
--- a/packages/paste-website/src/components/component-overview-table/index.tsx
+++ b/packages/paste-website/src/components/component-overview-table/index.tsx
@@ -2,7 +2,7 @@ import * as React from 'react';
 import {Link} from 'gatsby';
 import {Table, Thead, Tbody, Tr, Th, Td} from '../table';
 import {SidebarCategoryRoutes, PackageStatus} from '../../constants';
-import {getPackagePath, getComponentNameFromPackageName} from '../../utils/RouteUtils';
+import {getPackagePath, getHumanizedNameFromPackageName} from '../../utils/RouteUtils';
 
 interface ComponentNode {
   node: {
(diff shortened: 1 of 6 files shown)


SiTaggart:
What do we think about sentence case instead? cc @aayushpi

TheSisb:
> What do we think about sentence case instead? 

The reason I went this route is so that it reads like how you would use the component in React

`<AspectRatio>` -> `AspectRatio` in sidebar

richbachman:
> > What do we think about sentence case instead?
> 
> The reason I went this route is so that it reads like how you would use the component in React
> 
> i.e.: `<AspectRatio>` -> `AspectRatio` in sidebar
> 
> It's the real component name.

Thats solid reasoning. I've seen that naming convention used on other design system doc sites. Specifically the Chakra Design System we all love.

SiTaggart:
I think my only concern is that our pages aren't designed just for an engineering audience. They are for designers too and have design guidelines etc, and marking page in the nav based on the react component name seems a little... engineer first. 

Sentence case feels a little more inclusive and plain english to me

aayushpi:
Nope. I prefer before.

serifluous:
@aayushpi Why the earlier version?

serifluous:
@TheSisb Separately, this casing will apply to _all_ components, right? Not just Utilities?

two24studios:
I don't mind the casing, I prefer it over the dashes and because it maps to the component naming. In a perfect world, I think I'd rather see something like `Aspect Ratio`, `Media Object`, etc. with a space between words just to make it easier to read.

TheSisb:
@serifluous yes all React Components (components, utilities, primitives).

I think having a consistent design language between Sketch/Abstract/Figma -> Engineering might be valuable if we end up using those tools to compose React Components and talking about the same things.

aayushpi:
I think sentence case is easier to read

serifluous:
+1 to sentence case.

I could see a case for title case because they're Names of Things, but sentence case would be easier to read.

TheSisb:
Fixed to look like this now:
image


@aayushpi @two24studios @SiTaggart @serifluous

PR #99, opened 2019-09-16
fix(website): automatically update website version from package

TheSisb:
This makes it so that the version number visible on the doc site is pulled directly from the package.json file.  Couldn't figure out [how to || if there was any benefit to] using the GraphQL system to pull this info. This seemed like a reasonable approach so I went with it.  Open to corrections :D

### DIFF
--- a/packages/paste-website/gatsby-config.js
+++ b/packages/paste-website/gatsby-config.js
@@ -42,9 +42,7 @@ module.exports = {
       options: {
         name: 'pages',
         path: `${__dirname}/src/pages`,
-        ignore: [
-          '**/components/**/*',
-        ],
+        ignore: ['**/components/**/*'],
       },
     },
     {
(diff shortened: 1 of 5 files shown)


packages/paste-website/gatsby-config.js
```diff
@@ -58,14 +56,8 @@ module.exports = {
...
         name: 'websiteCore',
-        path: `${__dirname}/src/../`,
-        ignore: [
-          '**/.cache/**',
-          '**/public/**',
-          '**/src/**/*',
-          '**/static/**',
-          '**/types/**',
-        ],
+        path: `${__dirname}/`,
```
TheSisb: The whole file is prettier except this line, which I just simplified.

tsconfig.json
```diff
@@ -43,6 +43,7 @@
     "noUnusedParameters": true,
     "noUnusedLocals": true,
     "skipLibCheck": true,
+    "resolveJsonModule": true,
```
TheSisb: Cool feature I turned on in Typescript [that automatically types values from imported JSON files](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-9.html).

richbachman: 🔥

richbachman:
Implementing GraphQL for the version number seems a bit overkill. I dig the method you've used.

SiTaggart:
Yeah, it's one of those "wrestle data into graphyQL and wrestle data out of graphyQL" or just link to the damn file. 👍

PR #104, opened 2019-09-23
feat(website): aspect-ratio docs page

richbachman:
- [x] Added Aspect Ratio doc page content
- [x] Fix changelog margin on `anchor` and `button` doc pages
- [x] Remove storybook link in API section of `anchor` and `button` doc pages
- [x] Add iframe, embed, object, and video child styles to Aspect Ratio

`Gatsby-Image` doesn't play well with our `LivePreview` component, so I opted to not use the `LivePreview` for the image/component example.

**Contributing to Twilio**


richbachman:
`Gatsby-Image` doesn't play well with our `LivePreview` component, so I opted to not use the `LivePreview` for the image/component example.

packages/paste-website/src/pages/utilities/aspect-ratio/index.mdx
```diff
@@ -0,0 +1,172 @@
...
+  borderStyle="solid"
+  borderWidth="borderWidth10"
+>
+  <AspectRatio ratio="4:3">
+    <Absolute backgroundColor="colorBackgroundBrand" />
+  </AspectRatio>
+</Box>`}
+</LivePreview>
+
+### Anatomy
```
TheSisb: Should we omit this section if there's no relevant info?

packages/paste-website/src/pages/utilities/aspect-ratio/index.mdx
```diff
@@ -0,0 +1,172 @@
...
+<content>
+
+## Usage Guide
+
+### API
+
+#### Installation
+
+```bash
+yarn add @twilio-paste/anchor
```
TheSisb: wrong package :P

packages/paste-website/src/pages/utilities/aspect-ratio/index.mdx
```diff
@@ -0,0 +1,172 @@
...
+```
+
+#### Usage
+
+```js
+import {AspectRatio} from '@twilio-paste/aspect-ratio';
+
+<AspectRatio ratio="4:3">
+  ... Embedded element
```
TheSisb: Should just put `<Absolute>Test</Absolute>` for more of an out-of-box copy-paste experience :)

packages/paste-website/src/pages/utilities/aspect-ratio/index.mdx
```diff
@@ -0,0 +1,172 @@
...
+</AspectRatio>
+```
+
+#### Storybook
```
TheSisb: We should kill this section because the link is now part of the header.

packages/paste-website/src/pages/utilities/aspect-ratio/index.mdx
```diff
@@ -0,0 +1,172 @@
...
+
+[View Aspect Ratio in Storybook](https://twilio-labs.github.io/paste/?path=/story/utilities-aspect-ratio--default)
+
+
+#### Props
+
+| Prop           | Type                  | Description                                                                                                              | Default   |
+| -------------- | --------------------- | ------------------------------------------------------------------------------------------------------------------------ | --------- |
+| ratio?         | string                | Determines the aspect ratio of the element. Use a colon separated number pattern (width:height). Required                | 4:3       |
```
TheSisb: If this is required why is there a `?` near ratio? 
Maybe the `?` should be next to string.

packages/paste-website/src/pages/utilities/aspect-ratio/index.mdx
```diff
@@ -0,0 +1,172 @@
...
+
+<DoDont>
+  <Do title="Do" body="Use to embed a component or other HTML element." preview={false} />
+  <Dont title="Don't" body="Don’t use without any child elements, because this is only a container element." preview={false} />
+</DoDont>
+
+### Examples
+
+**Image and embedded component use case:**<br/>
+(Using an Anchor component do/don't for this example).
```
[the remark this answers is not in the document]
richbachman: I think in the future we could add this. I worry about adding it now, since we haven't taken a pass at mobile/responsiveness yet and the site starts breaking when resizing.

packages/paste-website/src/pages/utilities/aspect-ratio/index.mdx
richbachman: `?` was a failed copy/paste and not looking closely at it.

packages/paste-website/src/pages/utilities/aspect-ratio/index.mdx
richbachman: Agreed. Seems redundant now. Removing from `anchor` and `button` doc pages as well.

packages/paste-website/src/pages/utilities/aspect-ratio/index.mdx
```diff
@@ -0,0 +1,172 @@
...
+
+<DoDont>
```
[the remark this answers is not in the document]
richbachman: I changed the wording a bit on this. Maybe it helps with the confusion.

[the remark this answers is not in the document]
richbachman: Nuking the do/dont example.

packages/paste-website/src/pages/utilities/aspect-ratio/index.mdx
```diff
@@ -0,0 +1,151 @@
...
+
+The Aspect Ratio component dynamically modifies the height of an element based on its
+current width to match a height-to-width ratio. For example, a 4:3 ratio would make
+an element with a width of 400px have a height of 300px. This updates in real time
+when you resize the screen.
+
+### Accessibility Information
+
+Be considerate of smaller devices when using this component. If necessary, you can
+modify the aspect ratio property at certain media query breakpoints.
```
SiTaggart: > modify the aspect ratio property at certain media query breakpoints

How?

richbachman: Someone could write a function that changes the prop value a certain breakpoints. Do I need to mention that here? Or leave it open ended?

TheSisb: So this is a thing where I mentioned to Rich we should experiment to see if we can use the styled-system thing of `ratio={['4:3', '5:3']}`

Alternatively, the consumer's wrapping code can just pass in a different prop based on window size, if they have some kind of window resize listener.

SiTaggart: > So this is a thing where I mentioned to Rich we should experiment to see if we can use the styled-system thing of ratio={['4:3', '5:3']}

❤️ 

I love that I drop a question and you both come to where I was aiming for.

I would LOVE to get the styled-system type arrays to work for this.

richbachman: Go team!
I'll create a ticket in the backlog for that investigation.

SiTaggart: Cool, I agree with that. My only other concern is that considering small screens and using breakpoints really has nothing to do with accessibility. Can we change this heading?

richbachman: Yeah. That heading doesn't really make sense anymore. What do you think about just dropping the heading, and the text drops down to a second paragraph under Composition Notes?

packages/paste-website/src/pages/utilities/aspect-ratio/index.mdx
```diff
@@ -0,0 +1,151 @@
...
+</DoDont>
+
+<DoDont>
+  <Do title="Do" body="Use to embed a component or other HTML element." preview={false} />
+  <Dont title="Don't" body="Don’t use without any child elements, because this is only a container element." preview={false} />
+</DoDont>
+
+### Examples
+
+<LivePreview scope={{Absolute, AspectRatio, Box}} language="jsx">
```
SiTaggart: You can then drop this, dangling example

packages/paste-website/src/pages/utilities/aspect-ratio/index.mdx
SiTaggart: > text drops down to a second paragraph under Composition Notes?

I buy it

packages/paste-website/src/pages/utilities/aspect-ratio/index.mdx
richbachman: I moved the color swatch example under the paragraph that discusses it.

packages/paste-website/src/components/site-wrapper/index.tsx
```diff
@@ -20,6 +20,14 @@ const globalStyles = css`
     margin: 0;
     font-size: 14px;
   }
+
+  iframe {
```
richbachman: I added this here for now as global style. It makes iframe embeds actually work with Aspect Ratio. We should probably just move this to the default styles for Aspect Ratio. Thoughts?

packages/paste-website/src/pages/utilities/aspect-ratio/index.mdx
```diff
@@ -0,0 +1,151 @@
...
+### Accessibility Information
+
+Be considerate of smaller devices when using this component. If necessary, you can
+modify the aspect ratio property at certain media query breakpoints.
+
+## Composition Notes
+
+Aspect Ratio must contain exactly one child element. The child element should be sized
+fully to the Aspect Ratio bounding box so that it resizes correctly. One way of achieving
+this is through the <Link to="/utilities/absolute">Absolute</Link> component.
```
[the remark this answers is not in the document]
TheSisb: I *think* if we don't use the Gatsby Link, history push state isn't updated correctly and the browser back button breaks.  This was reported on Slack if you land on the homepage, click the View Components button, click a component url, then hit back.  We should find a way to style the Gatsby Link component and be sure to use it.

packages/paste-website/src/components/site-wrapper/index.tsx
TheSisb: I don't like this as a global style, it doesn't scale to non-AspectRatio iframes.

I think you just have to make a new component and put the styles there.

SiTaggart: hmmm I feel like the expectation from reading the docs is that it should "just work", so maybe it belongs to the Aspect Ratio. Do you need to do the same thing for an image? What's the least amount of work for a consumer to do for all cases?

packages/paste-website/src/pages/utilities/aspect-ratio/index.mdx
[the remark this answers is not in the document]
richbachman: It looks like the browser back button works as expected on the deployed branch. At least from Aspect Ratio -> Absolute anchor -> Back. I do think we need to figure out a way to style `Link`. There are some other performance gains we get using it. I'll create a ticket.

packages/paste-website/src/components/site-wrapper/index.tsx
TheSisb: It's probably safe to do something like `> :first-child` here.

packages/paste-website/src/pages/utilities/aspect-ratio/index.mdx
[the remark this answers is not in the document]
SiTaggart: I _think_ we can do something like:

```
import {Anchor} from '@twilio-paste/anchor'
import {Link} from 'gatsby'

const SiteLink = (props) => (
  <Anchor {...props} as={Link}>{props.children}</Anchor>
);
```

And then use `SiteLink` everywhere

packages/paste-website/src/components/site-wrapper/index.tsx
richbachman: @SiTaggart you're right. It should "just work". I'm going to move that style over to the Aspect Ratio where it should be. I think if that component was kicked off properly, we would have caught that. Image is much simpler:

```
<AspectRatio ratio="4:3">
    <img src="" />
</AspectRatio>
```

richbachman: Added embed, iframe, object, video styles to Aspect Ration in the latest commit.

packages/paste-website/src/pages/utilities/aspect-ratio/index.mdx
```diff
@@ -0,0 +1,175 @@
...
+***
+
+<content>
+
+## Guidelines
+
+### About Aspect Ratio
+
+The Aspect Ratio component dynamically modifies the height of an element based on its
+current width to match a height-to-width ratio. For example, a 4:3 ratio would make
```
serifluous: "width-to-height" instead of "height-to-width"? to match the ratio we use

richbachman: 👍

PR #110, opened 2019-10-01
feat(website): add slugs to mdx GraphQL

richbachman:
- [x] Added slugs to GraphQL node field values for Mdx files
- [x] Also added example query to anchor doc page

If you want to view the queried data, here is the quick example code I was using to display the anchor links:

```
<div>
  {props.data.mdx.headings.filter(heading => heading.depth !== 1).map(({ value, depth }) => {
    const headingLink = '#' + value.replace(/\./g, '').split(' ').join('-').toLowerCase();
    return (
      <div key={value}>
        <Anchor href={headingLink}>
          {value}
        </Anchor>
      </div>
    )
  })}
</div>
```

**Contributing to Twilio**


packages/paste-website/gatsby-node.js
```diff
@@ -0,0 +1,13 @@
+const {createFilePath} = require(`gatsby-source-filesystem`);
+
+exports.onCreateNode = ({node, getNode, actions}) => {
```
TheSisb: Can you add some comments here explaining what this is or linking to some docs please

richbachman: @TheSisb I've added a comment explaining the what and why.

PR #113, opened 2019-10-02
chore(design-tokens): better typings

TheSisb:
- [x] Adds missing `minHeights` and `maxHeights` to the theme object.
- [x] More explicit typings for design-tokens.  (from "string" to the actual value).

<details>
  <summary>Before</summary>

image

image

</details>

<details>
  <summary>After</summary>

image

image

</details>


packages/paste-design-tokens/formatters/d.ts.ts
```diff
@@ -5,13 +5,24 @@ import {getTokenCategories} from '../utils/getTokenCategories';
...
 export const tokenTemplate = ({name, value}: {name: string; value: string}): string =>
   `export declare const ${lodash.camelCase(name)} = "${value}";`;
 
 const categoryTemplate = (categoryName: string, props: Token[]): string => `export declare const ${lodash.camelCase(
   categoryName
 )}: {
-${props.map(prop => `  ${lodash.camelCase(prop.name)}: 'string';`).join('\n')}
+${props
+  .map(prop => {
+    const value = isNumeric(prop.value) ? prop.value : `"${prop.value}"`;
```
SiTaggart: Would be cool to add at least one token in the fixtures to test this logic and get it in the snapshot

TheSisb: I pushed up z-index fixtures. Let me know if this is what you meant.

SiTaggart: Perfect 👍

PR #115, opened 2019-10-02
chore(types): add a typescript types package

TheSisb:
Currently, we redefine the same typescript types in many packages. This PR moves those repeated definitions into a single package.  Now you only need to define system-level types once and can import them when needed.  I went through and updated Box, Media Object, and Text to use the new package. 

I'm looking for feedback on if there's a better way to achieve this in a Typescript monorepo.


richbachman:
lgtm 🔥

PR #124, opened 2019-10-07
feat: theme switching fully on the docsite

SiTaggart:
- [x] Implement a full site theme switcher
- [x] Fix any out of the box styling issues with Paste
- [x] Fix TOC styling
- [x] Fix small sized inline code snippets
- [x] Fix small sized code blocks


packages/paste-core/components/button/src/styles.ts
```diff
@@ -9,70 +9,75 @@ import {ButtonWrapperProps, ButtonChildrenProps} from './types';
...
+  font-size: 100%;
   /* To fix abnormal button padding-bottom */
   line-height: unset;
 `;
 const sizeSmall = (props: ButtonWrapperProps): SerializedStyles => css`
   padding: ${themeGet('space.space10')(props)} ${themeGet('space.space30')(props)};
   border-radius: ${themeGet('radii.borderRadius10')(props)};
   font-size: ${themeGet('fontSizes.fontSize30')(props)};
-  line-height: 24px;
+  line-height: ${themeGet('lineHeights.lineHeight30')};
```
SiTaggart: Missing props

packages/paste-core/components/heading/src/index.tsx
```diff
@@ -13,7 +13,7 @@ export type HeadingStyle =
 
 interface Heading {
   as: asTags;
-  id?: never;
+  id?: string;
```
SiTaggart: I'm not sure why we wouldn't allow `id`. I need it for the site headings

richbachman: Oh right. That makes sense.

packages/paste-core/components/heading/__tests__/__snapshots__/heading.test.tsx.snap
```diff
@@ -3,7 +3,8 @@
 exports[`Heading it should render an H1 at fontSize90 1`] = `
 .emotion-1 {
   font-family: 'Whitney SSm A','Whitney SSm B','Helvetica Neue',Helvetica,Arial,sans-serif;
-  line-height: 1.15;
+  font-size: 0.875rem;
```
richbachman: Shouldn't these font-size tests be different font-sizes? Looks like they're all `0.875rem`.

packages/paste-website/src/components/empty-state/NotBuilt.tsx
```diff
@@ -1,10 +1,12 @@
 import * as React from 'react';
 import {Anchor} from '@twilio-paste/anchor';
-import {Text} from '@twilio-paste/text';
 import {Box} from '@twilio-paste/box';
+import {Paragraph} from '@twilio-paste/paragraph';
+import {Heading} from '@twilio-paste/heading';
```
richbachman: Should we test `import {Heading, Paragraph} from '@twilio-paste/typography';` to verify its working as expected?

packages/paste-core/components/heading/__tests__/__snapshots__/heading.test.tsx.snap
SiTaggart: It's the base font size from the theme provider which we wrap every component in. It threw me too for an hour or so until I realised what I'd changed.

packages/paste-website/src/components/empty-state/NotBuilt.tsx
SiTaggart: smart

packages/paste-core/components/heading/__tests__/__snapshots__/heading.test.tsx.snap
richbachman: Right. I see that now.

packages/paste-website/src/components/site-wrapper/sidebar/SidebarNavigation.styles.ts
```diff
@@ -18,7 +18,7 @@ interface SiteNavNestListProps {
 }
 
 export const SiteNavNestList: React.FC<SiteNavNestListProps> = styled(SiteNavList)`
-  max-height: ${props => (props.isOpen ? '3000px' : '0')};
+  display: ${props => (props.isOpen ? 'block' : 'none')};
```
SiTaggart: Makes the nav accessible

packages/paste-core/components/button/src/styles.ts
```diff
@@ -9,70 +9,75 @@ import {ButtonWrapperProps, ButtonChildrenProps} from './types';
...
  */
-const baseButtonWrapper = css`
+const baseButtonWrapper = (props: ButtonWrapperProps): SerializedStyles => css`
   /* Hide default browser styles */
   appearance: none;
   border: none;
   display: inline-block;
   outline: none;
   background: none;
-  font-size: 100%;
```
SiTaggart: Moving this to where it's needed. The ordering was weird so even though this is added first for each button, 100% was coming last and overriding the font sizes

packages/paste-theme/src/themeProvider.tsx
```diff
@@ -10,11 +10,15 @@ const pasteGlobalStyles = css`
   html {
     font-size: 100%;
   }
+  body {
+    margin: ${themeGet('space.space0')};
+    font-size: ${themeGet('fontSizes.fontSize30')};
+  }
```
SiTaggart: So, these might seem risky, but I've tried them in my custom css plugin in console, flex and marketing campaigns and it seems totally fine to do so. 

Font size not on the body results in some elements like `code` rendering incorrectly, even when given an explicit font size. It looks for the body instead of the html element for it's base. And we still want a good out of the box experience size wise.

packages/paste-theme/src/themeProvider.tsx
```diff
@@ -10,11 +10,15 @@ const pasteGlobalStyles = css`
...
 `;
 
 export const StyledBase = styled.div`
   font-family: ${themeGet('fonts.fontFamilyText')};
-  line-height: 1.15;
+  line-height: ${themeGet('lineHeights.lineHeight30')};
```
SiTaggart: this on the body is too risky

PR #125, opened 2019-10-07
feat(box): polish and fix

TheSisb:
So other than the MDX doc site file, this is ready for a review pass.  Sorry, this is a big one 🌊 

## Changes
- [x] Made some significant changes to the new `@twilio-paste/types` package.  It now exports zIndex tokens and also all typings support responsive props (so 'space20' or ['space10', 'space20']).  Unfortunately for my previous example, `width` doesn't work in practice due to a bug in Styled System v4.  This will be resolved when we upgrade to v5.  
- [x] Updated Box component to correct typings and props taken.  Box now takes boxShadow, zIndex, overflow, and all the missing sizing props.
- [x] Fixed a type bug in the Absolute component after the Box updates.
- [x] Added dummy breakpoints to Storybook so we can test responsive style props.
- [x] Updated the Box default story to add all the missing props and added a responsive story.
- [x] Refactored the `@twilio-paste/theme` package to export all the theme variants (Console/Sendgrid/Flex/Default).  We were duplicating this all over and it was hard keeping everything updated correctly.
- [x] Midway through my work, the storybook folks decided to nuke a DefinitelyTyped package, so I fix that too.
- [x] Added the `sizes` key to `@twilio-paste/theme-tokens` so we can use responsive sizes in props.
- [x] **BREAKING CHANGE**: I removed the shortand `m=` and `p=` etc spacing props in favor of the full word variants.  I updated all areas where these were used for the long version. 
- [x] Updated the Box documentation page.

Ordinarily I would pull this out into smaller PRs, but since we all jammed on the extraneous stuff together I left it all as one piece.


.storybook/addons/theme-switcher-addon/index.js
```diff
@@ -26,8 +26,10 @@ class ThemeWrapper extends React.Component {
   };
 
   render() {
+    // These are dummy breakpoints for testing
+    const breakpoints = ['40em', '52em', '64em', '80em'];
```
SiTaggart: reminder to remove these

packages/paste-theme/src/types.ts
```diff
@@ -0,0 +1 @@
+export type ValueOf<O> = O[keyof O];
```
SiTaggart: Should this go in the types package as some kind of utility as we'd got it twice now?

packages/paste-types/src/index.ts
```diff
@@ -1,2 +1,9 @@
 export {SendGridThemeShape, ThemeShape} from '@twilio-paste/theme-tokens';
-export * from './tokenProps';
+
+export * from './tokens/backgroundColor';
```
SiTaggart: Minor nit, these aren't token types, these are style prop types. (Just in case we ever move design token types into this package in the future)

packages/paste-website/src/pages/components/button/index.mdx
```diff
@@ -229,17 +231,19 @@ their default size.
...
+  <Box display="inline-block" marginRight="space30">
     <Button variant="primary" size="small">
       Submit
     </Button>
   </Box>
 
-  <Box display="inline-block" mr="space30">
-    <Button variant="secondary" size="small">Submit</Button>
-  </Box>
+<Box display="inline-block" marginRight="space30">
```
SiTaggart: What are you using to auto-format these MDX files? This is going to keep happening if we aren't all using the same thing

TheSisb: One (or both?) of these VS Code addons 
image

SiTaggart: Hmm, I have both of those, and if it was prettier I'd expect the build to fail as we run prettier on our files in circle so it can't be that either... mystery 🤔

.storybook/addons/theme-switcher-addon/index.js
SiTaggart: Ooooh wait, the logic here is wrong https://github.com/twilio-labs/paste/blob/master/packages/paste-theme/src/themeProvider.tsx#L47

`breakpoints` is on the theme object already, I just did a dumb. https://github.com/twilio-labs/paste/blob/master/packages/paste-theme-tokens/src/default/index.ts#L25


I don't think it was working for another reason that got fixed.

SiTaggart: Wait, is it wrong? Maybe it's not...

Anyway, regardless we don't need them here

packages/paste-website/src/pages/components/button/index.mdx
SiTaggart: Mine doesn't do any auto-formatting at all?! I'm missing out on so much!

SiTaggart: Have you both got some kind of local prettier config going on, as our config ignores mdx files entirely https://github.com/twilio-labs/paste/blob/master/.prettierignore#L2

TheSisb: In my settings (cmd + ,) I have
image

image

and 
image

.storybook/addons/theme-switcher-addon/index.js
TheSisb: Ok I'll remove it

packages/paste-website/src/pages/components/button/index.mdx
SiTaggart: Does VS Code auto-format a JSON file for you?

richbachman: I'm running the exact same settings as @TheSisb.

richbachman: Looks like VS Code is formatting JSON on my end.

SiTaggart: It very much shouldn't be. Your VS Code is ignoring our `.prettierignore` file

TheSisb: I think the `.prettierignore` file is just for CLI prettier (so our yarn scripts and CI) but user level configs override those.

.storybook/addons/theme-switcher-addon/index.js
TheSisb: Oh the issue was because initially we were passing `customBreakpoints={[]}` which set breakpoints to `[]`.  I removed the prop altogether so we will now get theme breakpoints.

packages/paste-website/src/pages/components/button/index.mdx
richbachman: Its seems like thats whats happening.

SiTaggart: I'm going to un ignore those files in another branch

richbachman:
💯

PR #127, opened 2019-10-08
feat(website): Table of Contents Component

richbachman:
- [x] Added table of contents component excluding `h1` elements
- [x] Added react-scrollspy and types
- [x] Added TOC to doc pages
- [x] Added `contentwrapper` to `PasteMDXProvider`

**Contributing to Twilio**

### DIFF
--- a/packages/paste-website/src/components/shortcodes/table-of-contents/index.tsx
+++ b/packages/paste-website/src/components/shortcodes/table-of-contents/index.tsx
@@ -21,10 +21,19 @@ const TableOfContents: React.FC<TableOfContentsProps> = data => {
               .split(' ')
               .join('-')
               .toLowerCase()}`;
+
             const depthLevel = depth.toString();
+
+            let active;
+            if (window.location.hash === headingLink) {
+              active = true;
+            }
+
             return (
-              <StyledListItem key={value} depth={depthLevel}>
-                <Anchor href={headingLink}>{value}</Anchor>
+              <StyledListItem key={value} depth={depthLevel} active={active}>
+                <Anchor href={headingLink}>
+                  {value} {active}
+                </Anchor>
               </StyledListItem>
             );
           })}
--- a/packages/paste-website/src/components/shortcodes/table-of-contents/index.tsx
+++ b/packages/paste-website/src/components/shortcodes/table-of-contents/index.tsx
@@ -4,15 +4,15 @@ import {TableOfContentsProps} from './types';
 import {StyledWrapper, StyledScrollSpy, StyledListItem} from './styles';
 import {slugify} from '../../../utils/RouteUtils';
 
-const TableOfContents: React.FC<TableOfContentsProps> = data => {
-  if (data == null || data.data == null) {
+const TableOfContents: React.FC<TableOfContentsProps> = headings => {
+  if (headings == null || headings.data == null) {
     return null;
   }
 
-  const headings = data.data;
+  const allHeadings = headings.data;
 
   // Get Array of heading anchors. Excluding h1 elements.
-  const headingsList = headings
+  const headingsList = allHeadings
     .filter(heading => heading.depth !== 1)
     .map(({value}) => {
       const headingAnchor = slugify(value);
@@ -26,7 +26,7 @@ const TableOfContents: React.FC<TableOfContentsProps> = data => {
     <StyledWrapper marginTop="space40" marginLeft="space140">
       <StyledScrollSpy items={headingsList} currentClassName="is-current" rootEl="#site-main">
         {// Get heading anchors and convert to #anchor format. Excluding h1 elements.
-        headings
+        allHeadings
           .filter(heading => heading.depth !== 1)
           .map(({value, depth}) => {
             const headingLink = `#${slugify(value)}`;
--- a/packages/paste-website/src/components/shortcodes/table-of-contents/index.tsx
+++ b/packages/paste-website/src/components/shortcodes/table-of-contents/index.tsx
@@ -1,42 +1,7 @@
 import * as React from 'react';
-import styled from '@emotion/styled';
-import {themeGet} from 'styled-system';
 import {Anchor} from '@twilio-paste/anchor';
-
-interface TableOfContentsProps {
-  data?: [
-    {
-      value: string;
-      depth: number;
-    }
-  ];
-}
-
-interface TableOfContentsListItemProps {
-  depth: string;
-  key: string;
-}
-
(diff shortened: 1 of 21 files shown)


SiTaggart:
Is there any way to get this to magically appear on every MDX page via a template, rather than having to manually add it yourself?

richbachman:
I'll dig into that. Adding to all page is kind of a pain, and I clearly missed adding this everywhere.

aayushpi:
That seems like a separate ticket

richbachman:
@aayushpi Agreed. Its going to be a heavier lift to alter the templates.

richbachman:
I also added a TODO to figure out how to handle the changelogs on the various pages. Since those are imported, the headings don't show up in the `mdx` header query. So we need to work out a way to only show the changelog TOC menu item on pages that have a changelog. I've added a ticket to figure that out.

SiTaggart:
If it’s not too much work let’s do it. I don’t want us creating a bunch of work for ourselves in our future or for day to day maintenance

richbachman:
@SiTaggart I think its definitely something we need to do. I think it may be more work than this ticket covers. We'll need to:
- Restructure the `mdx` template
- Rework the `pageQuery` so its dynamic by slug or something like that and make sure it works with the `ComponentHeader` and `TableOfContents`
- Rework to Do/Dont component images so they pull from somewhere other than the `pageQuery`
- Not break anything

I propose creating a new ticket for that, and prioritize so we're only reconfiguring the pages we have now.

Thoughts?

packages/paste-website/src/components/shortcodes/table-of-contents/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+import {TableOfContentsProps} from './types';
+import {StyledWrapper, StyledScrollSpy, StyledListItem} from './styles';
+import {slugify} from '../../../utils/RouteUtils';
+
+const TableOfContents: React.FC<TableOfContentsProps> = data => {
+  if (data == null || data.data == null) {
+    return null;
+  }
+
+  const headings = data.data;
```
TheSisb: `data.data` is kind of weird.  Does the first `data` have a more accurate name by any chance?

packages/paste-website/src/pages/tokens/design-tokens-package.mdx
```diff
@@ -91,8 +101,10 @@ As you use Paste, you'll likely encounter things that _don't seem right_. Please
 
 ***
 
-<Box marginTop="space90">
+<Box marginTop="space90" id="changelog">
```
TheSisb: Why does this need an id?

packages/paste-website/src/components/shortcodes/table-of-contents/index.tsx
SiTaggart: This is props, right?

packages/paste-website/src/components/shortcodes/table-of-contents/styles.ts
```diff
@@ -0,0 +1,63 @@
...
+  margin-top: ${marginTopPartial};
+  margin-left: ${marginLeftPartial};
+  padding-left: ${themeGet('space.space30')};
+  font-size: ${fontSizePartial};
+  color: ${themeGet('colors.colorGray80')};
+  border-left-width: 2px;
+  border-left-style: solid;
+  border-left-color: transparent;
+
+  a {
```
SiTaggart: Let's avoid descendants like this. You are trying to override the styles of the Anchor component and not honouring component encapsulation.

I'd actually like to request a design change @two24studios, our links / clickable thing are blue, can we keep them blue so it's consistent? Then we don't even need to override the styles

packages/paste-website/src/components/shortcodes/table-of-contents/styles.ts
```diff
@@ -0,0 +1,63 @@
...
+    color: ${themeGet('colors.colorGray80')};
+  }
+
+  &.is-current {
```
SiTaggart: This is kind of grim, and it's a fault of the library. Generally, components modifying their children is not a good idea. 

We don't have a choice here, but I'm just voicing my ick 😉

SiTaggart:
> I propose creating a new ticket for that, and prioritize so we're only reconfiguring the pages we have now.

OK, I buy it

packages/paste-website/src/components/shortcodes/table-of-contents/styles.ts
TheSisb: As long as it is the exception - and isolated - I guess it's safe enough.  And I totally agree haha

packages/paste-website/src/pages/tokens/design-tokens-package.mdx
richbachman: Left over `id` from when I was adding the changelog anchor to the TOCs. Just forgot to remove.

packages/paste-website/src/components/shortcodes/table-of-contents/styles.ts
richbachman: I'm on board with that. I'll chat with @two24studios about looking at the design. Removing the style overrides for now.

packages/paste-website/src/components/shortcodes/table-of-contents/index.tsx
```diff
@@ -0,0 +1,47 @@
...
+import {TableOfContentsProps} from './types';
+import {StyledWrapper, StyledScrollSpy, StyledListItem} from './styles';
+import {slugify} from '../../../utils/RouteUtils';
+
+const TableOfContents: React.FC<TableOfContentsProps> = headings => {
+  if (headings == null || headings.data == null) {
+    return null;
+  }
+
+  const allHeadings = headings.data;
```
SiTaggart: oh, so what I meant was change

```
<TableOfContents data={props.data.mdx.headings} />
```

to

```
<TableOfContents headings={props.data.mdx.headings} />
```

then you can destructure it from props

```
const TableOfContents: React.FC<TableOfContentsProps> = ({headings}) => {
  if (headings == null) {
    return null;
  }

  // Get Array of heading anchors. Excluding h1 elements.
  const headingsList = headings
    .filter(heading => heading.depth !== 1)
    .map(({value}) => {
      const headingAnchor = slugify(value);
      return headingAnchor;
    });
```

richbachman: 👍 Got it.

PR #129, opened 2019-10-09
feat(icons): delete old icons, add new streamline icons

TheSisb:
Our icons package needed some serious love.  This PR is massive, here's what happened:

### High level user-facing stuff:
- [x] Nuked all our old inherited icons.  We're starting with a fresh slate using Streamline icons.
- [x] I added two new icons (thx @serifluous): PlusIcon and LoadingIcon.  
- [x] Updated the Spinner and Button components to use our new LoadingIcon.
- [x] Our icons now are typed correctly with our tokens.  No more arbitrary number passing (in TS)!
- [x] Our icon size tokens have been updated to [16px 24px and 32px](https://github.com/twilio-labs/paste/pull/129/files#diff-d75e2b5954e0457648ea13560f2f1b86R52)
- [x] I updated the Spinner story: it now uses tokens and has a responsive demo.
- [x] I made our icons package npm publishable again by adding the missing `publishConfig`.
- [x] **Note**: Color is now passed as `iconColor` to icons, rather than simply `color`. [code](https://github.com/twilio-labs/paste/pull/129/files#diff-b84635347b37d884c052d11a547feff4R5-R12)
- [x] Our `TextColor` typing now also accepts `currentColor`. [code](https://github.com/twilio-labs/paste/pull/129/files#diff-b84635347b37d884c052d11a547feff4R5-R12)

### Icons package janitorial work
- [x] IconWrapper is cleaned up and now uses tokens.
- [x]  Previously, we had these folders: `src/svg` and `src/react`.  I moved the raw SVGs out of src, so it's just `svg`, and moved react directly into src.  This helped consolidate this package's architecture to match other packages in the monorepo.
- [x] I swapped out `tsc` for `rollup` to run our icon builds.  This along with some of the other tweaks helped to eliminate the monorepo issue where built files would be generated into incorrect directories.
- [x] Rollup by default bundles all the icons into a single `index` file.  We don't want consumers to have to download a massive bundle of icons if they're only using a handful.  In the rollup config, we have to explicitly specificy each input file if we want them to be separate.  Rather than updating it by hand each time we add a new icon, I generate a cache of icons in the system during the `list-icons` step.  This file is commit to the repo so builds work for everyone.
- [x] A whole slew of cleanup to our icons package build scripts and tooling.  Mostly refactoring to make it more robust and easier to read.

### Other infra work
- [x] I added an `iconSizes` export from our token page, for easier typing. [code](https://github.com/twilio-labs/paste/pull/129/files#diff-b84635347b37d884c052d11a547feff4R5-R12)
- [x] Our core-bundle kept generating a diff every rebuild, so I added sorting to the generated dependencies and imports.  This removes any meaningless changes when running builds because the output stays consistent.
- [x] A small thing, but I noticed all our rollup files were declaring format as `es` instead of `esm`, so I updated them all to say `esm` which matches the rollup docs.

### Future work

I'd like to add more documentation to this package, to make it easier to navigate for future maintainers.


SiTaggart:
The spinner isn't perfectly centered so it moves around slightly by like 2 pixels 



Can you add an example of loading buttons to the button docs too with this?

SiTaggart:
Should we change the PR title to be `BREAKING CHANGE: ...` too, as the merge commit needs to be that for lerna to do a major bump?

TheSisb:
> Should we change the PR title to be BREAKING CHANGE: ... too, as the merge commit needs to be that for lerna to do a major bump?

I was thinking about that on my Box PR as well.  Github topics can only be a single line though.  Conventional commits dictates that the title be "fix/etc" and "BREAKING CHANGE" be placed on another line beneath.

Looking at how my Box PR merged, it looks like it worked itself out: 
image


If I hit "Squash and merge" on this PR it also does the correct thing:
image

richbachman:
Just echoing the off center spinner here. Also, more of a visual thing, but I feel like the weight of the icons is off, especially when used alongside text:

image

serifluous:
> Just echoing the off center spinner here. Also, more of a visual thing, but I feel like the weight of the icons is off, especially when used alongside text:
> 
> image

After working with @TheSisb, we realized the 1px stroke width means centered icons like the plus icon don't align to full pixels. We're upping icon strokes to 2px.

TheSisb:
@SiTaggart @serifluous @richbachman Ready for re-review.

serifluous:
I can't see any of the changes. Can you add the loading button example to the website?

TheSisb:
I can't until we fix another issue with the release icons package on NPM.  For now the only way to test is to pull this locally and run storybook

serifluous:
I’ll try pulling it locally. But don’t let that hold this up. It looked good to me when we looked at it together yesterday.

SiTaggart:
super crispy 👌

packages/paste-core/components/spinner/src/index.tsx
```diff
@@ -12,7 +13,14 @@ const rotate = keyframes`
   }
 `;
 
-const SpinningWrapper = styled.div(
+interface SpinnerWrapperProps {
+  width: number;
+  height: number;
```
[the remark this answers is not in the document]
TheSisb: Yea, they should.

packages/paste-icons/src/react/helpers/IconWrapper.tsx
```diff
@@ -1,18 +1,35 @@
 import styled from '@emotion/styled';
+import {compose, size, system} from 'styled-system';
+import {IconSize, TextColor} from '@twilio-paste/types';
+
+// I can't name this simply `color` because TS + Emotion doesn't like it
+const iconColor = system({
```
TheSisb: kind of like how Text needs to accept `textColor` instead of `color`.  Would love if we can figure this out because `color` is more ergonomic

packages/paste-icons/src/react/helpers/IconWrapper.tsx
```diff
@@ -1,18 +1,35 @@
...
+  },
+});
 
 export interface IconWrapperProps {
-  as?: string;
-  size?: number;
-  color?: string;
+  as?: keyof JSX.IntrinsicElements;
+  size?: IconSize;
+  iconColor?: TextColor | 'currentColor';
```
TheSisb: I believe we need to add `currentColor` into our tokens list OR I can update `TextColor` in the types package maybe.  TBD

packages/paste-theme-tokens/src/default/index.ts
```diff
@@ -35,6 +36,11 @@ export const DefaultTheme = {
   maxHeights: sizings,
   minHeights: sizings,
   sizes: sizings,
+  iconSizes: {
+    sizeIcon10: sizings.sizeIcon10,
+    sizeIcon20: sizings.sizeIcon20,
+    sizeIcon30: sizings.sizeIcon30,
+  },
```
TheSisb: I tried automating this with an `Object.keys().filter().forEach()` but the resulting typings weren't as clean.

packages/paste-core/components/typography/rollup.config.js
```diff
@@ -2,20 +2,19 @@ import typescript from 'rollup-plugin-typescript2';
 import babel from 'rollup-plugin-babel';
 import resolve from 'rollup-plugin-node-resolve';
 import commonjs from 'rollup-plugin-commonjs';
-import {
-  terser
-} from 'rollup-plugin-terser';
+import {terser} from 'rollup-plugin-terser';
```
SiTaggart: is it weird that prettier suddenly thought this wasn't OK?

packages/paste-core/components/spinner/stories/index.stories.tsx
```diff
@@ -1,12 +1,26 @@
...
+import {DefaultTheme, ThemeShape} from '@twilio-paste/theme-tokens';
 import {Spinner} from '../src';
 
+const IconColorOptions = Object.keys(DefaultTheme.textColors);
+const SizeOptions = Object.keys(DefaultTheme.iconSizes);
+
 storiesOf('Components|Spinner', module)
   .addDecorator(withKnobs)
   .add('Default', () => {
+    const iconColorValue = select('iconColor', IconColorOptions, 'currentColor') as keyof ThemeShape['textColors'];
```
SiTaggart: You can use the types package here, can't you?

```
import {TextColor, IconSize} from '@twilio-paste/types'

...

const iconColorValue = select('iconColor', IconColorOptions, 'currentColor') as TextColor;
```

packages/paste-core/components/typography/rollup.config.js
TheSisb: It is lol.  I don't really question prettier.

packages/paste-core/components/spinner/stories/index.stories.tsx
TheSisb: Good point, fixed.

packages/paste-core/components/spinner/src/index.tsx
```diff
@@ -22,18 +29,21 @@ const SpinningWrapper = styled.div(
...
   title: string;
+  size?: IconSize;
 }
 
-const Spinner: React.FC<SpinnerProps> = props => (
-  <SpinningWrapper>
-    <SpinnerIcon {...props} />
+const Spinner: React.FC<SpinnerProps> = ({size, ...props}) => (
+  <SpinningWrapper size={size}>
+    <LoadingIcon size={size} {...props} />
```
SiTaggart: Should we spread props here? Is it only title and size as valid props?

*EDIT:* better question, do we want a whole bunch of different spinners in our apps?

TheSisb: I made it explicit.  Good call. https://github.com/twilio-labs/paste/pull/129/commits/ebb040a26e2078448fb6f394a8f915ee4e3bbae4

PR #135, opened 2019-10-14
feat(website): SiteLink Component, replace Link with SiteLink

richbachman:
- [x] SiteLink component
- [x] Replaced instances of `Link` with `SiteLink`
- [x] Fix `focus` and `active` `box-shadow` style so it now uses the `shadowFocus` token

I attempted to use Anchor `as` Link, but that caused conflicts with the `href` and `to` props respectively. If we used this method, we also lost the prefetching Gatsby provides with Link. This being one of the larger benefits we get get with Gatsby, and makes for a faster user experience on the site.

I decided the best path forward was to create a styled component that wrapped Link and used the token styles from Anchor.

I also found a minor bug where we're not using the `shadowFocus` token in Anchor and Button. Anchor is fixed, and I've created a ticket to fix Button since that has more complex styles.

### DIFF
--- a/packages/paste-website/src/pages/index.tsx
+++ b/packages/paste-website/src/pages/index.tsx
@@ -50,7 +50,13 @@ const IndexPage: React.FC<{}> = (): React.ReactElement => {
         </P>
         <Box marginTop="space80" marginBottom="space80">
           <Box display="inline" marginRight="space40">
-            <Button as="a" href="/getting-started/" size="default" variant="primary">
+            <Button
+              size="default"
+              variant="primary"
+              onClick={() => {
+                navigate('/getting-started/');
+              }}
+            >
               Get Started
             </Button>
           </Box>
@@ -59,13 +65,19 @@ const IndexPage: React.FC<{}> = (): React.ReactElement => {
               size="default"
               variant="secondary"
               onClick={() => {
-                navigate('/components');
+                navigate('/components/');
               }}
             >
               View Components
             </Button>
           </Box>
-          <Button as="a" href="/roadmap" size="default" variant="secondary">
+          <Button
+            size="default"
+            variant="secondary"
+            onClick={() => {
+              navigate('/roadmap/');
+            }}
+          >
             Roadmap
           </Button>
         </Box>
--- a/packages/paste-website/src/pages/index.tsx
+++ b/packages/paste-website/src/pages/index.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import {Helmet} from 'react-helmet';
+import {navigate} from 'gatsby';
 import styled from '@emotion/styled';
 import {themeGet} from 'styled-system';
 import {Button} from '@twilio-paste/button';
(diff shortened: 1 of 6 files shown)


TheSisb:
This is great stuff!

One of the issues that brought this up was that if you go to the index page and click "View components", then click on "Anchor", then hit your browser back button you go back to the index page.  That "View Components" link is actually a Button so it doesn't use the gatsby link behavior.  

I think we may also need a SiteButtonLink that uses https://www.gatsbyjs.org/docs/gatsby-link/#how-to-use-the-navigate-helper-function

richbachman:
@TheSisb Right. That was in another ticket, but its actually a pretty easy fix. Use can just use our standard buttons with an `onClick` event like this:

```
<Button
    size="default"
    variant="secondary"
    onClick={() => {
        navigate('/components');
    }}
>
    View Components
</Button>
```

I tested it using the home -> view components -> anchor -> back button flow mentioned above. Feel free to check it out here: https://paste-68n1xdjmq.now.sh/components

Do you think will work, or do you think we need to add a `SiteButtonLink` link still?

packages/paste-website/src/pages/index.tsx
```diff
@@ -54,7 +55,13 @@ const IndexPage: React.FC<{}> = (): React.ReactElement => {
             </Button>
           </Box>
           <Box display="inline" marginRight="space40">
-            <Button as="a" href="/components" size="default" variant="secondary">
+            <Button
+              size="default"
+              variant="secondary"
+              onClick={() => {
```
TheSisb: We need the same treatment on the other buttons above and below this one.

TheSisb:
I think what you've done is fine 👍

packages/paste-website/src/pages/index.tsx
richbachman: Yup. Just pushed the other two fixes.

packages/paste-website/src/pages/index.tsx
```diff
@@ -49,16 +50,34 @@ const IndexPage: React.FC<{}> = (): React.ReactElement => {
         </P>
         <Box marginTop="space80" marginBottom="space80">
           <Box display="inline" marginRight="space40">
-            <Button as="a" href="/getting-started/" size="default" variant="primary">
+            <Button
```
SiTaggart: 😟 We now have `<button />`s that act as links on our homepage. Is there a way we can still use the navigate API and switch out the element back to an anchor `<a />`?

richbachman: @SiTaggart We could revert back to an `<a />`, prevent the default `href` action, and then use navigate in an `onClick`. What do you think of that approach?

SiTaggart: yeah, just use the `as` prop, add a matching `href` so if I command-click it still loads the url in a new tab, and use the `onClick` to do client side routing for gatsby. Use :allthethings:

richbachman: @SiTaggart done here: https://github.com/twilio-labs/paste/pull/146

PR #137, opened 2019-10-15
feat(typography): typography paragraph component

richbachman:
- [x] Initial typography paragraph component
- [x] Paragraph story
- [x] Paragraph test


richbachman:
@TheSisb this is still a WIP, but will you take a peak and let me know if I'm on the right track. I feel like I should be doing something different with the `index.js` file.

packages/paste-core/components/typography/src/index.tsx
```diff
@@ -0,0 +1,3 @@
+import {Paragraph} from './Paragraph';
```
TheSisb: Looks good! 
You can shorthand this as `export {Paragraph} from './Paragraph'`
Also let's make it a rule to always export the props for every component too
 `export {Paragraph, ParagraphProps} from './Paragraph'`

packages/paste-core/components/typography/src/Paragraph.tsx
```diff
@@ -0,0 +1,27 @@
...
+import {Text} from '@twilio-paste/text';
+import {ParagraphProps} from './types';
+
+const handlePropValidation = ({children}: ParagraphProps): void => {
+  if (children == null) {
+    throw new Error(`[Paste: Typography Paragraph] Must have non-null children.`);
+  }
+};
+
+const Paragraph: React.FC<ParagraphProps> = props => {
```
TheSisb: I'd like us to also start giving our components a displayName manually.  This makes it so you can retrieve the component type while iterating through react children.  

So right before the export just add
`Paragraph.displayName = 'Paragraph';`

React.FC types it correctly so it should just work.

richbachman: Had to look up what adding a `displayName` does, but now I know. Added it.

packages/paste-core/components/typography/src/index.tsx
richbachman: Changed the exports here.

I kept types in a separate file for now so we can place all typography types there for Heading, Paragraph, and Lists.

packages/paste-core/components/typography/src/Paragraph.tsx
```diff
@@ -0,0 +1,28 @@
...
```
SiTaggart: The name of the component here worthy of a poll of some sort? 

`Paragraph` might seem verbose to some, `P` is related to the HTML tag but might be not as clear.

richbachman: Polling the team. I found a few other design systems/ui-component libraries using `Paragraph`, which is part of why I went that route. I also thought using `P` could be confused with the standard `p`.

SiTaggart: `Paragraph` it is then 👍

SiTaggart:
A Typography package seems sensible to me, where or how do you envisage the component docs?

A single Typography page with all the components in, or a page per component from the package. The latter fits better with how the component page is designed

richbachman:
@SiTaggart Yeah. I think the latter for the actual docs pages. I think there could be a main typography doc page with just a quick overview of whats contained in the package.

packages/paste-core/components/typography/package.json
```diff
@@ -0,0 +1,46 @@
...
+  },
+  "peerDependencies": {
+    "@emotion/core": "^10.0.10",
+    "@emotion/styled": "^10.0.10",
+    "react": "^16.8.6",
+    "react-dom": "^16.8.6",
+    "styled-system": "^4.1.0"
+  },
+  "devDependencies": {
+    "@twilio-paste/text": "^0.4.1",
```
TheSisb: All twilio paste packages need to be added to both `peerDeps` and `devDeps`.

packages/paste-core/components/typography/stories/index.stories.tsx
```diff
@@ -0,0 +1,23 @@
...
+      <>
+        <Paragraph>I am paragraph text.</Paragraph>
+        <Paragraph>
+          Kale chips distillery authentic, portland etsy cloud bread vinyl gentrify drinking vinegar viral meh hot
+          chicken bitters fashion axe palo santo. Chillwave fixie sustainable <i>helvetica</i> etsy.
+        </Paragraph>
+        <Paragraph>
+          Prism whatever ethical, gochujang <strong>edison bulb</strong> put a bird on it kitsch. Pop-up 90&apos;s la
+          croix tumeric, palo santo chia try-hard direct trade tote bag roof party scenester kitsch stumptown
+          intelligentsia. Literally heirloom blue bottle etsy.
```
TheSisb: ctrl+f "etsy"
_not enough found_

😆 How did you generate this?

packages/paste-core/components/typography/src/types.ts
```diff
@@ -0,0 +1,5 @@
+export interface ParagraphProps {
```
TheSisb: Just playing out this file structure in my mind, the eventuality would be something like

```
export {Paragraph} from './Paragraph';
export {List} from './List';
export {Heading} from './Heading';
export {ParagraphProps, ListProps, HeadingProps} from './types';
```

Two thoughts:
1- I kind of wonder (since these components are so minimal) if we should move the types into the same file? 
2- We can actually name them the same thing as the component.

So it becomes
```
export {Paragraph} from './Paragraph';
export {List} from './List';
export {Heading} from './Heading';
```

You are allowed to name an interface and a value the same thing, and Typescript knows how to infer which one you mean in code.

```
interface Paragraph {}
const Paragraph: React.FC<Paragraph> = () => {};
export {Paragraph}
```
should work.

https://stackoverflow.com/questions/49798156/typescript-variable-name-as-interface-type

packages/paste-core/components/typography/__tests__/typography.test.tsx
```diff
@@ -0,0 +1,17 @@
+import * as React from 'react';
```
TheSisb: We should rename this file to `paragraph.test.tsx` and have different test files for each component for cleanliness. 👍

packages/paste-core/components/typography/stories/index.stories.tsx
richbachman: From @two24studios: https://hipsum.co

packages/paste-core/components/typography/src/types.ts
SiTaggart: Love this 👍

packages/paste-core/components/typography/src/Paragraph.tsx
```diff
@@ -0,0 +1,33 @@
...
+  if (children == null) {
+    throw new Error(`[Paste: Typography Paragraph] Must have non-null children.`);
+  }
+};
+
+const Paragraph: React.FC<Paragraph> = props => {
+  handlePropValidation(props);
+
+  return (
+    <Text
```
SiTaggart: Should we explicitly set the `as` prop to set the element to be `p`, even though it's default? _Just in case?_

Or are we happy any change/regression would be caught by the snapshot test?

richbachman: Might as well be safe, and explicitly set it.

PR #138, opened 2019-10-15
feat(core): a core bundle

TheSisb:
This PR adds a "Core bundle".  This was added to make it easier to start using Paste by combining all of our packages into a single import.  

The general approach I took to making this work was by using Lerna to fetch all the packages in our monorepo.  I then [filter some out](https://github.com/twilio-labs/paste/commit/6e13bc726b3a75bde6791a1bfcc363beb6e9ee7e#diff-286239e76cbb6baf1f6897e7d3832a27R30) and proceed to generate the `package.json` and `index.tsx` files.

## Open questions:
- [x] Should any non-core packages be included? (i.e.: theme?)
- [x] Is where I put the paste-core folder (/packages/paste-core/core-bundle) appropriate?


packages/paste-core/core-bundle/tools/generate.js
```diff
@@ -0,0 +1,62 @@
...
+    const isCorePackage = item.location.includes('/paste-core/');
+    const isReleased = !item.private && (item.version !== '0.0.1' || item.version !== '0.0.0');
+    const isNotBlacklisted = !BLACKLIST.includes(item.name);
+
+    return isCorePackage && isReleased && isNotBlacklisted;
+  });
+}
+
+// Main
+(async () => {
```
TheSisb: Reading this function explains the main procedure that happens right before a build is triggered on this package.

packages/paste-core/core-bundle/rollup.config.js
```diff
@@ -0,0 +1,34 @@
+import typescript from 'rollup-plugin-typescript2';
+import babel from 'rollup-plugin-babel';
+import resolve from 'rollup-plugin-node-resolve';
+import commonjs from 'rollup-plugin-commonjs';
+// import {terser} from 'rollup-plugin-terser';
```
richbachman: Do we need this commented line for something in the future?

TheSisb: Oops, I forgot to un-comment it after testing my dist output. Thanks!

packages/paste-core/core-bundle/tools/generate.js
richbachman: Using private, version, and blacklisted seems like a good way to handle that filter. I like it.

richbachman:
The `/packages/paste-core/core-bundle` placement makes sense to me.

PR #145, opened 2019-10-18
feat(card): initial card implementation

GiantRobots:
**Contributing to Twilio**


package.json
```diff
@@ -43,6 +43,7 @@
   "dependencies": {
     "@emotion/core": "^10.0.15",
     "@emotion/styled": "^10.0.15",
+    "@types/react": "16.8.23",
```
SiTaggart: what's the reasoning for making this an exact dependency over a non-locked dev dependency?

packages/paste-core/components/card/src/index.tsx
```diff
@@ -1,25 +1,39 @@
...
-  // eslint-disable-next-line jsx-a11y/no-static-element-interactions, jsx-a11y/click-events-have-key-events, jsx-a11y/no-noninteractive-tabindex
-  <aside onClick={onClick} tabIndex={onClick != null ? 0 : undefined} role={onClick != null ? 'button' : undefined}>
+const Card: React.FunctionComponent<CardProps> = ({children, ...attributes}) => (
+  <Box
+    borderWidth="borderWidth20"
+    borderColor="colorBorder"
+    borderStyle="solid"
+    borderRadius="borderRadius20"
+    padding="space60"
+    {...attributes}
```
SiTaggart: Not a huge fan of spreading props like this because I can now change everything about the component outside of the systems control. 

Our components should only be modifiable within the confines of the API that we define through variants. If you don't think the component works for you, you can drop down to the `Box` and `Text` layer and create your custom experience.

SiTaggart:
Should the card have a set backgorund color? Right now they are transparent, but I'm wondering if it should be set white? Question for Sean, really.

packages/paste-core/components/card/src/index.tsx
GiantRobots: This is definitely a subject to talk about. The current implementation if you're using Typescript will only allow attributes that are defined and extended. This works great if you're using typescript...

If you're not using Typescript you can put anything in there and that sucks. +1

package.json
GiantRobots: Solid point I think it was an oversight because I installed that as a specific dependency to match what was in devDependencies. I'll update.

packages/paste-core/components/card/src/index.tsx
GiantRobots: If we don't spread then we have to pass through all of the properties (all aria, all events, all etc) which will that hard to maintain Aria, and event passthrough.

GiantRobots:
```typescript 
if (procces.env.TypeChecking) {
    {...attributes}
} else {
  filteredAttributes = filterAttrs(attributes);
  {...filteredAttributes}
}
```

packages/paste-core/components/card/stories/index.stories.tsx
```diff
@@ -1,11 +1,35 @@
 import * as React from 'react';
 import {storiesOf} from '@storybook/react';
-import {action} from '@storybook/addon-actions';
-import {Card, CardTitle, CardBody} from '../src';
+// import {action} from '@storybook/addon-actions';
```
TheSisb: Can just remove this line

packages/paste-core/utilities/absolute/package.json
```diff
@@ -28,7 +28,7 @@
   },
   "peerDependencies": {
     "@emotion/styled": "^10.0.10",
-    "@twilio-paste/box": "^0.2.4",
+    "@twilio-paste/box": "^0.3.1",
     "react": "^16.8.6",
     "react-dom": "^16.8.6",
     "styled-system": "^4.1.0"
```
TheSisb: We merged the styled-system v5 upgrade PR, can you make sure all these "styled-system" deps are on version 5.1.3 please?  https://github.com/twilio-labs/paste/pull/18

packages/paste-types/src/DomTypes.ts
```diff
@@ -0,0 +1,26 @@
...
+
+// Not a fan of having static interactiveX wish we could do
+// interactive<T> = React.ComponentProps<T>; but couldn't make it work ¯\_(ツ)_/¯
+export interface InteractiveDiv extends Omit<React.ComponentProps<'div'>, AttributesToDrop> {}
+export interface InteractiveButton extends Omit<React.ComponentProps<'button'>, AttributesToDrop> {}
+export interface InteractiveLabel extends Omit<React.ComponentProps<'label'>, AttributesToDrop> {}
+export interface InteractiveAnchor extends Omit<React.ComponentProps<'a'>, AttributesToDrop> {}
+export interface InteractiveInput extends Omit<React.ComponentProps<'input'>, AttributesToDrop> {}
+export interface InteractiveImage extends Omit<React.ComponentProps<'img'>, AttributesToDrop> {}
+export interface InteractiveSelect extends Omit<React.ComponentProps<'select'>, AttributesToDrop> {}
```
TheSisb: These are great! 👍

packages/paste-core/components/card/src/index.tsx
SiTaggart: Explain the above code snippet to me in more detail, please

GiantRobots: So if a user opts into TypeChecking by setting up Typescript or fork-ts-checker-webpack-plugin to ensure they have valid typings they can also set an env flag at the startup of the application like...

```js
new webpack.DefinePlugin({
  TypeChecking: JSON.stringify(true),
});
```

once a user has "opted in" to Typechecking then we can take the "easy path" we can use Types as a source of truth and drop a lot of the runtime checking and just passthrough attributes that match closer to the types that are coming from the React types package.

Another thought we should probably drop runtime checking when `process.env.NODE_ENV === 'production'` to mirror React's functionality and dropping development warnings when run in production mode

SiTaggart: OK, so given no one really has a good answer to solve this right now, we're making the decision not to solve it just yet, in favour of being explicit. 

We as component authors should have a reasonable idea of what is appropriate to add to an element from semantic and accessibility point of view given the intended use case of the component/design. This should inform us to declare a set of props a consumer would likely need. 

This will be plenty good enough to unblock us and get rolling.

package.json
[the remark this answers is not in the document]
GiantRobots: I did it because using types as a utility and interface seems to be in line with you need this for development. Unless I have dependencies/devDependencies wrong.

packages/paste-core/utilities/absolute/package.json
GiantRobots: Any guidance on how to do this merge better? I'd be happy to rebase or anything else.

packages/paste-core/components/card/src/card.test.tsx
```diff
@@ -0,0 +1,57 @@
...
+      .create(
+        <Theme.Provider>
+          <Card />
+        </Theme.Provider>
+      )
+      .toJSON();
+    expect(tree).toMatchSnapshot();
+  });
+
+  it('it should render default values', (): void => {
```
TheSisb: This should probably say something else

packages/paste-core/components/card/src/index.tsx
```diff
@@ -1,25 +1,51 @@
...
+};
+
+interface CardProps extends SpaceProps, StaticDiv {}
+
+const Card: React.FunctionComponent<CardProps> = ({children, ...attributes}) => {
+  errorOnBadProps(attributes, cardProps);
+
+  return (
+    <Box
+      borderWidth="borderWidth20"
```
SiTaggart: Still missing setting the `backgroundColor="colorBackgroundBody"`

I'm also tempted to set the `as` prop to be an `<article />`. I'm kind of hoping they are used as distinct pieces of content on the page and articles can provide nice navigation helpers to JAWS users

packages/paste-website/src/pages/components/card/index.mdx
```diff
@@ -0,0 +1,160 @@
...
+Card is an extremely flexible container that does not require specific components inside of it, 
+allowing an implementer to compose as needed, but elements such as Title, Body, 
+and Button or Anchor are commonly used. See Examples for common Card patterns. 
+A Card does not currently handle any interactive events such as hover, click or focus, 
+though children composed inside of it may commonly have event handlers.
+
+Acknowledging the flexibility of Card component, there are several non-negotiable 
+properties of a Card that differentiate it from a more basic page-layout element, such as Box:
+
+* A background color of $background-color-body.
```
SiTaggart: This token is called `$color-background-body`. Might want to double check the others too if this slipped through spec review.

packages/paste-core/components/card/stories/index.stories.tsx
```diff
@@ -1,11 +1,34 @@
...
 
-storiesOf('Components|Card', module).add('Default', () => (
-  <Card onClick={action('clicked')}>
-    <CardTitle>Title</CardTitle>
-    <CardBody>Body</CardBody>
-  </Card>
-));
+storiesOf('Components|Card', module)
+  .add('Default', () => (
+    <Card>
```
SiTaggart: Would it be possible to hook up the first story to be able to modify the props with the knobs feature of storybook?

Kind of like https://github.com/twilio-labs/paste/blob/master/packages/paste-core/utilities/text/stories/index.stories.tsx#L68

packages/paste-website/src/pages/components/card/index.mdx
```diff
@@ -0,0 +1,160 @@
...
+| Prop         	| Type             	| Description                              	| Default        	|
+|--------------	|------------------	|------------------------------------------	|----------------	|
+| children?    	| React.ReactNode  	| Child components to render into the card 	| undefined         |
+| borderWidth  	| borderWidths     	| Width of the border                      	| borderWidth20  	|
+| borderColor  	| BorderColors     	| Color of the border                      	| colorBorder    	|
+| borderStyle  	| BorderStyleProps 	| Style of the border                      	| solid          	|
+| borderRadius 	| Radii            	| Radius of the border                     	| borderRadius20 	|
+| padding      	| Spacing          	| Internal spacing of the card             	| space60        	|
+
+### Examples
```
SiTaggart: We usually place examples before the Usage Guidelines.

packages/paste-core/components/card/src/index.tsx
```diff
@@ -1,25 +1,51 @@
...
+      borderColor="colorBorder"
+      borderStyle="solid"
+      {...attributes}
+    >
+      {children}
+    </Box>
+  );
+};
+
+interface CardProps extends SpaceProps, StaticDiv {}
```
SiTaggart: Can we only use the `PaddingProps` from space for now?

packages/paste-website/src/pages/components/card/index.mdx
```diff
@@ -0,0 +1,160 @@
...
+### API
+
+| Prop         	| Type             	| Description                              	| Default        	|
+|--------------	|------------------	|------------------------------------------	|----------------	|
+| children?    	| React.ReactNode  	| Child components to render into the card 	| undefined         |
+| borderWidth  	| borderWidths     	| Width of the border                      	| borderWidth20  	|
+| borderColor  	| BorderColors     	| Color of the border                      	| colorBorder    	|
+| borderStyle  	| BorderStyleProps 	| Style of the border                      	| solid          	|
+| borderRadius 	| Radii            	| Radius of the border                     	| borderRadius20 	|
+| padding      	| Spacing          	| Internal spacing of the card             	| space60        	|
```
SiTaggart: This prop table is out of date a little bit, you only allow "Global HTML attributes" and `PaddingProps` now

packages/paste-core/components/card/stories/index.stories.tsx
```diff
@@ -1,12 +1,40 @@
 import * as React from 'react';
 import {storiesOf} from '@storybook/react';
 import {Heading} from '@twilio-paste/heading';
+import {withKnobs, select} from '@storybook/addon-knobs';
 import {Text} from '@twilio-paste/text';
 import {Card, CardFooter} from '../src';
 
+import {Padding, Space} from '@twilio-paste/types';
+
+const space = [
```
TheSisb: You can replace this with 
```js
import {DefaultTheme, ThemeShape} from '@twilio-paste/theme-tokens';

const spaceOptions = Object.keys(DefaultTheme.space);

....

const paddingValue = select('padding', spaceOptions, 'space10') as keyof ThemeShape['space'];

... 

<Card padding={paddingValue}>
```

so that this list doesn't go stale and we maintain a single source of truth.  (Box' story does this)

TheSisb: `'@twilio-paste/theme-tokens'` doesn't need to go into the package.json because Storybook has access to it.

packages/paste-core/components/card/src/utilities.tsx
```diff
@@ -0,0 +1,19 @@
+const marginProps = ['margin', 'marginTop', 'marginRight', 'marginBottom', 'marginLeft'];
+const paddingProps = ['padding', 'paddingTop', 'paddingRight', 'paddingBottom', 'paddingLeft'];
```
SiTaggart: For clarity do you mind just removing the ones we're not actually using anywhere?

Issue #54, opened 2019-08-16
Underlining on buttons seems like overkill


Matjaxon:
I personally find the added underlining on hovered buttons to be unnecessary and jarring.  The buttons already change color and have a pointer change.  I understand that the underlining was added for accessibility but wouldn't that already be sufficiently covered by the pointer change?

serifluous:
Thanks so much for the feedback! We'll take this into account when we iterate next on buttons and discuss with product design teams across Twilio.

For more accessibility context, the pointer change itself won't be sufficient when you're navigating a page with a keyboard only, for example. If you'd like to learn more about our accessibility considerations, definitely let us know. We love talking about accessibility with teams.

You're welcome at office hours any time to discuss this more, too.

Matjaxon:
I took the screenshot below after tabbing through the paste docs page.  Chrome (I'm assuming?) already seems to offer outlining of the active element.

<img width="781" alt="Screen Shot 2019-08-21 at 9 43 21 AM" src="https://user-images.githubusercontent.com/7808060/63451043-303b2900-c3f8-11e9-94a7-157a3716001f.png">

I would assume there's similar browser support or plugins to address this need better than us adding additional underlining.

TheSisb:
Hey @Matjaxon, a delayed update but we are going to remove the underlines from Buttons.  Will circle back when that PR is in.

Innovator007:
I would like to work on this if its open

SiTaggart:
Thanks for your interest @Innovator007, but I'll be tackling this today

PR #147, opened 2019-10-22
fix: remove the underline from button states

SiTaggart:
Because we have now fixed the contrast of the button hover and focus states with the outline color, we can now remove the text underline. 

Fixes https://github.com/twilio-labs/paste/issues/54

### DIFF
--- a/packages/paste-core/components/button/src/styles.ts
+++ b/packages/paste-core/components/button/src/styles.ts
@@ -55,13 +55,8 @@ const baseEnabled = css([
     text-decoration: none;
     cursor: pointer;
 
-    &:hover {
-      text-decoration: underline;
-    }
-
     &:focus,
     &:active {
-      text-decoration: underline;
       box-shadow: 0 0 0 4px rgba(0, 117, 195, 0.5);
     }
   `,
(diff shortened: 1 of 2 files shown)


richbachman:
It looks like that change killed the hover underline on the reset variant: https://paste-git-feat-remove-button-underline.twilio-dsys.now.sh/components/button#examples

Original here for comparison: https://paste.twilio.design/components/button#examples

SiTaggart:
> It looks like that change killed the hover underline on the reset variant: 

So, this was by accident, but it does raise the question of; should a `reset` button have any styles like underlines on hover?

serifluous:
I keep asking @TheSisb again and again about what `reset` is for, and I keep forgetting. Can someone remind me? 😬

As a stab in the dark, I'm going to say it should still keep the underline on hover.

TheSisb:
The Button component has a`reset` for `size` and a `reset` for `variant`.

The `variant reset` simply removes all the default button styles.  We added this in the first pass of the component but I'm no longer convinced we should have it.  I think it's safe to remove it knowing what we now know.

The `size reset` omits adding the Button font-size, line-height, border-radius, font-weight, and padding that other sizes set.  This was useful for buttons as links, where the additional padding seemed off.

Reflecting on those decisions, I think we should change our variants into a tuple shaped as `level_look`.  This allows us to add new variants with relative ease.

Example:
```
- primary_filled
- primary_outlined
- primary_link
- destructive_filled
- destructive_outlines
- destructive_link
```

TheSisb:
I approved the PR because this change is fine.  We should follow it up with some additional Button work.

richbachman:
@TheSisb thanks for the explanation and background.

PR #148, opened 2019-10-23
chore: get the applitools github checks API running

SiTaggart:
Try and get the Applitools GitHub checks running fully.

Goal is to see something like:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -64,11 +64,11 @@ jobs:
           keys:
             - v1-dependencies-{{ checksum "package.json" }}
             - v1-dependencies-
-      - attach_workspace:
-          at: ~/
+      - run: yarn
+      - run: yarn bootstrap
       - run:
-          name: Set applitools api key
-          command: "export APPLITOOLS_BATCH_ID=`echo $CIRCLE_SHA1`"
+          name: Run package builds
+          command: yarn build
       - run:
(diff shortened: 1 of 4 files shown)


SiTaggart:
This is better. We now have the checks coming in correctly, but I don't think it's 100%. I'm not sure it needs to run on master first for it to detect the changes correctly as it hasn't populated a master branch baseline. 

Who's willing to get this merged and try and see if it's better?

.circleci/config.yml
```diff
@@ -95,9 +95,7 @@ workflows:
       - test:
           requires:
             - build
-      - applitools:
-          requires:
-            - build
+      - applitools
```
SiTaggart: Making this run in parallel to build as it might be long-running, also we get parallel jobs now 🎉

applitools.config.js
```diff
@@ -1,3 +1,5 @@
 module.exports = {
   serverUrl: 'https://twilioeyesapi.applitools.com',
+  batchId: process.env.CIRCLE_SHA1,
```
SiTaggart: Setting the batchID here instead of the circleCi conig (which is what the docs say to do but it's wrong)

Making it exit with `0` so the circle job doesn't error if there are differences, we want Applitools to error so that resolving the diffs in Applitools marks the checks as green. (you'd have to run circleCI again to get it to be green otherwise)

PR #149, opened 2019-10-23
feat(typography): heading component

richbachman:
- [x] Move Paragraph to separate component
- [x] Refactored Heading component
- [x] Refactored font-size tokens to sync with design
- [x] Adjusted font-size of Paragraph to match refactored tokens
- [x] Adjust any font-size tokens used in Paste components, utilities, and primitives
- [x] Adjust any font-size tokens used in website
- [x] Import Heading and Paragraph into Typography component


packages/paste-core/components/typography/src/Heading.tsx
```diff
@@ -0,0 +1,114 @@
...
+function getHeadingStyles(headingStyle?: HeadingStyle): {} {
+  switch (headingStyle) {
+    case 'headingStyle60':
+      return {
+        marginBottom: 'space70',
+        fontSize: 'fontSize90',
+        fontWeight: 'fontWeightSemibold',
+        lineHeight: 'lineHeight60',
+      };
+    case 'headingStyle50':
```
SiTaggart: I _think_ you can do 

```
case 'headingStyle50':
default:
```

and delete the default from the bottom. This way we always know a default is a defined heading style.

packages/paste-website/src/components/callout/index.tsx
```diff
@@ -4,13 +4,13 @@ import {StyledCallout} from './styles';
 import {CalloutProps, CalloutTitleProps, CalloutTextProps} from './types';
 
 const CalloutTitle: React.FC<CalloutTitleProps> = ({as, children}) => (
-  <Text as={as} marginBottom="space40" fontSize="fontSize20">
+  <Text as={as} marginBottom="space40" fontSize="fontSize30">
```
SiTaggart: if the default of `Text` is `fontSize="fontSize30"` you can probably remove these font size settings if the idea is to just get the default body text size. Otherwise, if we shift the scale again, you'll be changing these all over again.

richbachman: Makes total sense.

packages/paste-core/components/typography/__tests__/heading.test.tsx
```diff
@@ -0,0 +1,79 @@
+import * as React from 'react';
+import renderer from 'react-test-renderer';
+import {Theme} from '@twilio-paste/theme';
+import {Heading} from '../src';
+
+describe('Heading', () => {
+  it('it should render an H1 with 32px font-size', (): void => {
```
TheSisb: The size totally depends on the scale implemented in the theme object and tokens.

packages/paste-core/components/typography/src/Heading.tsx
```diff
@@ -0,0 +1,108 @@
...
+  | 'headingStyle40'
+  | 'headingStyle50'
+  | 'headingStyle60';
+
+interface Heading {
+  as: asTags;
+  id?: never;
+  className?: never;
+  children: NonNullable<React.ReactNode>;
+  headingStyle?: HeadingStyle;
```
TheSisb: Should we call this `variant` so it's consistent with other components?

packages/paste-core/components/typography/src/Heading.tsx
```diff
@@ -0,0 +1,108 @@
...
+        marginBottom: 'space30',
+        fontSize: 'fontSize20',
+        fontWeight: 'fontWeightSemibold',
+        lineHeight: 'lineHeight10',
+      };
+  }
+}
+
+const handlePropValidation = ({as, children}: Heading): void => {
+  const hasAs = as;
```
TheSisb: This const seems kind of useless.  Also because you're setting a default prop, won't you always expect an `as`?

packages/paste-core/components/typography/src/Heading.tsx
```diff
@@ -0,0 +1,108 @@
...
+export type asTags = 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6' | 'div' | 'label' | 'span';
+export type HeadingStyle =
+  | 'headingStyle10'
+  | 'headingStyle20'
+  | 'headingStyle30'
+  | 'headingStyle40'
+  | 'headingStyle50'
+  | 'headingStyle60';
+
+interface Heading {
```
TheSisb: Let's make it a habit to export these interfaces as well so that they can be extended by consumers.

packages/paste-core/components/typography/src/Heading.tsx
```diff
@@ -0,0 +1,108 @@
...
+    </Text>
+  );
+};
+
+Heading.defaultProps = {
+  as: 'h2',
+  headingStyle: 'headingStyle50',
+};
+
+export {Heading};
```
TheSisb: Please add the `Heading.displayName = 'Heading'` line too, for consistency

packages/paste-core/components/typography/src/Heading.tsx
```diff
@@ -0,0 +1,108 @@
...
+      return {
+        marginBottom: 'space30',
+        fontSize: 'fontSize20',
+        fontWeight: 'fontWeightSemibold',
+        lineHeight: 'lineHeight10',
+      };
+  }
+}
+
+const handlePropValidation = ({as, children}: Heading): void => {
```
TheSisb: I'm now of the belief that we should use propTypes to do this work for us.  

```
Heading.propTypes = {
  as: PropTypes.oneOf(['h1', 'h2', etc]),
  variant: PropTypes.oneOf(['headingStyle60', etc..]),
}
```
What do you think?

packages/paste-core/components/typography/stories/index.stories.tsx
```diff
@@ -20,4 +30,13 @@ storiesOf('Components|Typography', module)
         </Paragraph>
       </>
     );
+  })
+  .add('Heading', () => {
+    const asOptions = text('as', 'h2') as asTags;
+    const headingStyleValue = select('headingStyle', headingStyleOptions, 'headingStyle50') as HeadingStyle;
+    return (
+      <Heading as={asOptions} headingStyle={headingStyleValue}>
+        I am a Very Large Heading
```
TheSisb: Can you add an example with an icon in the heading or some other component?

packages/paste-core/components/typography/src/Heading.tsx
SiTaggart: Oooooh I dig it

packages/paste-core/components/typography/src/Heading.tsx
SiTaggart: No idea if it's possible, but I created a ticket to automate linting this

packages/paste-core/components/typography/src/Heading.tsx
SiTaggart: and a ticket to automate this too

packages/paste-core/components/typography/src/Heading.tsx
SiTaggart: I'm still concerned about not being able to actively stop people doing bad things, if you're not using Typescript or you choose to ignore the proptype warnings (folks do). But on a scale of 1 to 10 on how I care, if runtime checks are  _really_ so much worse, I'm a 4.

packages/paste-design-tokens/tokens/themes/sendgrid/global/font-size.yml
```diff
@@ -4,34 +4,3 @@ global:
...
-    comment: "Constant typography token for font size 70"
-  font-size-80:
-    value: "{!font-size-80}"
-    comment: "Constant typography token for font size 80"
-  font-size-90:
-    value: "{!font-size-90}"
-    comment: "Constant typography token for font size 90"
-  font-size-100:
-    value: "{!font-size-100}"
-    comment: "Constant typography token for font size 100"
```
SiTaggart: 🎉 Unification 🎉

packages/paste-core/components/typography/__tests__/heading.test.tsx
SiTaggart: What do you mean?

packages/paste-core/components/typography/src/Heading.tsx
richbachman: I wondered if that was the case, but erred on the side of caution and added as to the prop validation. I'll remove it.

packages/paste-core/components/typography/src/Heading.tsx
richbachman: Making the Heading props exportable means I have to change the name back to HeadingProps because the exportable name Heading is taken by the component export. Going that route for now.

packages/paste-core/components/typography/src/Heading.tsx
richbachman: I took a stab at adding these propTypes to Paragraph and Heading. Let me know if its done right.

packages/paste-core/components/typography/__tests__/heading.test.tsx
TheSisb: Just that 32px in this string doesn't map for sure to our token value based on the theme provided.  32px isn't future proof.  We should use the design language and use the token name here.

packages/paste-core/components/typography/stories/index.stories.tsx
richbachman: We may need to work with design a little on how to handle icons in the headers, i.e. size restrictions for the icons. Did you have another component in mind to use as an example here? Anchor?

packages/paste-core/components/typography/src/Heading.tsx
```diff
@@ -0,0 +1,97 @@
...
+export type asTags = 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6' | 'div' | 'label' | 'span';
+export type HeadingStyle =
+  | 'headingStyle10'
+  | 'headingStyle20'
+  | 'headingStyle30'
+  | 'headingStyle40'
+  | 'headingStyle50'
+  | 'headingStyle60';
+
+export interface HeadingProps {
```
TheSisb: I'm a doof, remove export and renaming this to `Heading`

Then when you do `export {Heading}` at the bottom it will export both this and the const.

packages/paste-core/components/typography/src/Paragraph.tsx
```diff
@@ -1,26 +1,19 @@
 import * as React from 'react';
+import * as PropTypes from 'prop-types';
 import {Text} from '@twilio-paste/text';
 
-interface Paragraph {
+export interface ParagraphProps {
```
TheSisb: This should have been fine the old way, no?

packages/paste-core/components/typography/src/Heading.tsx
richbachman: Reverted that change.

packages/paste-core/components/typography/src/Paragraph.tsx
richbachman: Reverted that change.

TheSisb:
This is looking good to me! Did you finalize those two checkboxes from your PR description?

richbachman:
@TheSisb Not yet. I'll be going through the website today and knocking that out. How do you think we should handle deprecating the old header? As part of this PR or another?

richbachman:
@TheSisb @SiTaggart I've moved Paragraph out to a separate component, and kept Heading separate. I'm just importing those into the Typography component now. All tests are with their respective components, but I kept the stories with the Typography component. I debated moving those as well. Let me know any feedback there.

packages/paste-core/components/typography/package.json
```diff
@@ -29,13 +29,15 @@
...
+    "@twilio-paste/heading": "0.3.4",
+    "@twilio-paste/paragraph": "^0.0.1",
     "react": "^16.8.6",
     "react-dom": "^16.8.6",
     "styled-system": "^5.1.2"
   },
   "devDependencies": {
-    "@twilio-paste/text": "^0.4.1",
+    "@twilio-paste/heading": "0.3.4",
+    "@twilio-paste/paragraph": "^0.0.1",
```
richbachman: Move to deps

packages/paste-core/components/typography/package.json
```diff
@@ -29,13 +29,15 @@
   "peerDependencies": {
     "@emotion/core": "^10.0.10",
     "@emotion/styled": "^10.0.10",
-    "@twilio-paste/text": "^0.4.1",
+    "@twilio-paste/heading": "0.3.4",
+    "@twilio-paste/paragraph": "^0.0.1",
```
richbachman: Remove.

packages/paste-core/components/heading/stories/index.stories.tsx
```diff
@@ -1,25 +0,0 @@
-import * as React from 'react';
```
SiTaggart: We should keep these stories for `Heading`

richbachman: @SiTaggart I moved the stories under the Typography component, for both Heading and Paragraph. Now that I'm thinking about it though, that doesn't make sense. I'll move those back to their respective components.

packages/paste-core/components/heading/src/index.tsx
```diff
@@ -10,66 +11,86 @@ export type HeadingStyle =
...
-    case 'headingStyle10':
-      return {
+        marginBottom: 'space30',
         fontSize: 'fontSize30',
-        fontWeight: 'fontWeightSemiBold',
+        fontWeight: 'fontWeightSemibold',
         lineHeight: 'lineHeight20',
       };
-    default:
+    case 'headingStyle10':
```
TheSisb: Did you forget to add the default case here?

packages/paste-core/components/paragraph/package.json
```diff
@@ -2,6 +2,47 @@
...
+    "prepublishOnly": "yarn build",
+    "type-check": "tsc --noEmit"
+  },
+  "peerDependencies": {
+    "@emotion/core": "^10.0.10",
+    "@emotion/styled": "^10.0.10",
+    "@twilio-paste/text": "^0.4.1",
+    "prop-types": "^15.7.2",
+    "react": "^16.8.6",
+    "react-dom": "^16.8.6",
```
TheSisb: You can remove the following peerDeps
```
"@emotion/core": "^10.0.10",
"react-dom": "^16.8.6",
"styled-system": "^4.1.0"
```

packages/paste-core/components/heading/package.json
```diff
@@ -30,6 +30,7 @@
     "@emotion/core": "^10.0.10",
     "@emotion/styled": "^10.0.10",
     "@twilio-paste/text": "^0.3.1",
+    "prop-types": "^15.7.2",
     "react": "^16.8.6",
     "react-dom": "^16.8.6"
```
TheSisb: You can remove the following peerDeps
```
 "@emotion/core": "^10.0.10",
 "react-dom": "^16.8.6"
```

packages/paste-core/components/heading/src/index.tsx
richbachman: Doesn't this say that `case 'headingStyle50` is also the `default`?

```
case 'headingStyle50':
default:
  return {
    marginBottom: 'space60',
    fontSize: 'fontSize70',
    lineHeight: 'lineHeight50',
  };
```

packages/paste-core/components/typography/stories/index.stories.tsx
TheSisb: Maybe just a `Text` then?

packages/paste-core/components/heading/src/index.tsx
richbachman: I tried removing the default, but that causes a lint error (default required). For now I'm moving `headingStyle50` and `default` to the bottom, and adding a comment about why the heading styles are out of order.

packages/paste-core/components/typography/stories/index.stories.tsx
richbachman: I added a couple more tests to `Paragraph` and `Heading`.
- italics and bold children in `Paragraph`
- italics children in `Heading`

packages/paste-core/components/heading/src/index.tsx
```diff
@@ -10,66 +11,90 @@ export type HeadingStyle =
...
 }
 
-const Heading: React.FC<HeadingProps> = ({as, headingStyle, children}) => (
-  <Text as={as} textColor="colorText" {...getHeadingStyles(headingStyle)}>
-    {children}
-  </Text>
-);
+const Heading: React.FC<Heading> = ({as, children, variant}) => {
+  return (
+    <Text as={as} textColor="colorText" {...getHeadingStyles(variant)}>
```
SiTaggart: We'll have to set `display="block"` here. If you set the as to a `label` or `span` the default display property from headings supplied by the browser will disappear and may cause undesired or unexpected styling issues for the consumer.

packages/paste-core/components/heading/src/index.tsx
```diff
@@ -10,66 +11,86 @@ export type HeadingStyle =
   | 'headingStyle50'
   | 'headingStyle60';
 
-interface HeadingProps {
+interface Heading {
   as: asTags;
-  headingStyle?: HeadingStyle;
+  id?: never;
+  className?: never;
+  children: NonNullable<React.ReactNode>;
```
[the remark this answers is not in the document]
TheSisb: Yea, you only need to put children if you're making it required or want it to be a different type.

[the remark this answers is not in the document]
richbachman: Just pushed that change.

TheSisb:
🚢

PR #152, opened 2019-10-26
fix: improved way of handling box props

SiTaggart:
Dealing with the `as` prop and spreading props onto components that extend Box:

## Dealing with `as`

I just retype the component based on what element I'm rendering as, being mindful to omit the props from `BoxProps` as they might clash with the input HTML attributes. The `ref` is a little weird, it clashes, haven't got my head around that yet.

```ts
type Input = Omit<React.ComponentProps<'input'>, keyof BoxProps | 'ref'>;
```

In essence I can now use `<Input />` which is an extension of Box, a styled div, and place any input HTML attribute on it and typescript not complain.

## Spreading those props

Now we can add HTML element attrs to our component we need to be able to spread them down to Box to apply them to the rendered HTML, _but_, I don't want you to have secret access to the Box styling props by doing so (or className or style).

### Solution

A function that filters out the `BoxProps` from your props

```tsx
import {Box, filterBoxPropsFromProps} from '@twilio-paste/box';

type Input = Omit<React.ComponentProps<'input'>, keyof BoxProps | 'ref'>;

export const Input: React.FC<Input> = props => {
  const filteredProps = filterBoxPropsFromProps(props)
  return (
    <Box
      padding="space40"
      {...filteredProps}
    />
  )
};
```

`filterBoxPropsFromProps` takes props, but also an optional array of "exceptions". Say you do want to expose padding to a component from Box, just pass in the props you _don't_ want to filter.

```ts
const filteredProps = filterBoxPropsFromProps(props, ['padding']);
```

Thoughts @TheSisb @GiantRobots @alanbsmith?

## TODO

- [ ] add className and style to props to filter list
- [ ] maybe a better name for the function?


alanbsmith:
Could we use [`shouldForwardProp`](https://emotion.sh/docs/styled#customizing-prop-forwarding) here?

packages/paste-core/utilities/box/src/FilterBoxPropsFromProps.ts
```diff
@@ -0,0 +1,101 @@
...
+  'borderLeftStyle',
+  'borderLeftColor',
+  'borderX',
+  'borderY',
+  'borderColor',
+  'boxShadow',
+  'zIndex',
+] as const;
+
+type Exceptions = typeof supportedBoxProps | string[];
```
TheSisb: Shouldn't this be `(typeof supportedBoxProps)[]` to force the array?  
Why the `string[]` too?

packages/paste-core/utilities/box/src/FilterBoxPropsFromProps.ts
```diff
@@ -0,0 +1,101 @@
...
+
+export const filterBoxPropsFromProps = (props: {}, exceptions: Exceptions = []): {} => {
+  const boxPropsToFilter = supportedBoxProps.filter(prop => {
```
TheSisb: Slight improvement:
```
let boxPropsToFilter = supportedBoxProps;
if (exceptions.length !== 0) {
   boxPropsToFilter = supportedBoxProps.filter(prop => {
      return !exceptions.includes(prop);
  });
}
return lodash.omit(props, boxPropsToFilter);
```

packages/paste-core/utilities/box/src/FilterBoxPropsFromProps.ts
```diff
@@ -0,0 +1,101 @@
...
+  'borderLeft',
+  'borderLeftWidth',
+  'borderLeftStyle',
+  'borderLeftColor',
+  'borderX',
+  'borderY',
+  'borderColor',
+  'boxShadow',
+  'zIndex',
+] as const;
```
GiantRobots: As we add components we will want to have a single source of truth so we can keep things correct I would prefer we compose this list with things like...

```js
const boxProps = [
...spacingProps,
...borderProps,
...positionProps,
...alignmentProps
]
```

That way we can control all of the props components need from a singular place and we can derive types from it as well.

SiTaggart:
> Could we use shouldForwardProp here?

Maybe even styled systems [@styled-system/should-forward-prop](https://github.com/styled-system/styled-system/tree/master/packages/should-forward-prop)?

packages/paste-core/utilities/box/src/FilterBoxPropsFromProps.ts
GiantRobots: I always like Set's for filtering work another implementation

```js
const validProps = new Set(['allThoseProps']);

const incomingAttributes = new Set(Object.entries(attributes));
const whiteList = intersection(validProps, exceptions);

return lodash.pick(attributes, whiteList)

```

Lodash is also removing omit because it's slow AF apparently and recommends pick.

https://github.com/lodash/lodash/issues/2930

SiTaggart:
> Developers have to remember to run this function in their components which is error prone.

Is this a big concern? It's system engineers only, which is us. I also would expect a higher level of intentionality to designing a component API for us too. So yes, it's error prone but are we likley to catch it more often than not? Can you lint it?

>This adds a performance overhead at runtime; components that rerender quickly will feel this the most.

Sure. I'm not a perf engineering, so I have little experience here. Is this why the styled-system approach memos this call in their filter? https://github.com/styled-system/styled-system/blob/master/packages/should-forward-prop/src/index.js#L40. Also, optimising for performance is important, but let's make sure we're not micro optimising yet. If it's seconds or hundreds of millliseconds then we should pay attention, if it's single or double-digit milliseconds... 🤷‍♂  

> This isn't something we can just run in dev and remove in prod; if we remove this in prod all the props will fall through unexpectedly.

The challenge is I guess just blocking them all together so you absolutely can't change things we don't intend for you to change. I'm not sure how far down a custom babel transform we'd want to go just yet. 

So the balance is, what's "good enough for now" to allow you to spread props and use `as`, versus being extremely specific in your component API and not spreading props in favour of having to write a few more lines of code.

packages/paste-core/utilities/box/src/FilterBoxPropsFromProps.ts
GiantRobots: Thinking about it more and what you said here 

> This isn't something we can just run in dev and remove in prod; if we remove this in prod all the props will fall through unexpectedly.

We should probably throw an error because the props are incorrect and we _want_ to strip this check in production. If it doesn't work locally in JS land I feel as if we have a pretty decent shot at making sure we're not letting bad props in.

```js
const validProps = new Set(['allThoseProps']);

const incomingAttributes = new Set(Object.entries(attributes));
const whiteList = intersection(validProps, exceptions);

if( lodash.omit(attributes, whiteList).length > 0 ){
throw new Error("THOSE AREN'T THE RIGHT PROPS!");
}
```

alanbsmith:
> Maybe even styled systems @styled-system/should-forward-prop?

@SiTaggart Just a note: The styled-system version is really nice to keep the props from hitting the lowest-level element, but the last time I tried to use it, it wasn't compatible with TS. 😦

packages/paste-core/utilities/box/src/FilterBoxPropsFromProps.ts
alanbsmith: Kind of makes me think of this: https://github.com/primer/components/blob/master/src/constants.js

TheSisb: This was also copied from Primer back into Styled System https://styled-system.com/api/

SiTaggart:
Closing.

Firstly on the `as` prop: Though it is not _perfect_, placing the burden of retyping the consuming component to be whatever the intended rendered HTML element will be when using `as`, it doesn't really seem like any sort of blocker. If anything it's an ergonomic optimisation we don't need to be worrying about in our infancy.

Secondly on spreading: No one has a _good_ solution, so lets' just not solve it right now. Being explicit for the time being a typing exactly what attributes each component supports has a few time consuming down points, but being explicit and predictable has a bunch of upsides to it too.

Conclusion: We will support the `as` props and retype the component based on the target as rendering, and we will not support prop spreading for the time being.

PR #153, opened 2019-10-28
fix(anchor): export types and add tests

richbachman:
- [x] Exported Anchor types
- [x] Added Anchor snapshot tests
- [x] Adjusted prop-types deps in core and website


packages/paste-core/components/anchor/src/index.tsx
```diff
@@ -1,48 +1,53 @@
...
-      onBlur={props.onBlur}
-      onClick={props.onClick}
-      onFocus={props.onFocus}
-      tabIndex={props.tabIndex}
-      target={isExternalUrl(props.href) && !props.target ? EXTERNAL_TARGET_DEFAULT : props.target}
-    >
-      {props.children}
-    </StyledLink>
-  );
+Anchor.defaultProps = {
```
SiTaggart: As these are optional I'm not sure we need to declare default values for them, given they are the browser defaults. 

Thoughts?

richbachman: @SiTaggart if i remove the default props, I get an lint error: https://github.com/yannickcr/eslint-plugin-react/blob/master/docs/rules/require-default-props.md

SiTaggart: huh, that lint rule isn't very sympathetic to HTML attributes. 

Leave it for now, just seems weird to apply `target` and `tabindex` to an anchor when there's really no need to.

TheSisb: I'd be into just removing the rule, this situation brings it's usefulness (or lack of) to light.

packages/paste-core/components/anchor/src/index.tsx
```diff
@@ -1,48 +1,53 @@
...
-  }
+const Anchor: React.FC<Anchor> = props => (
+  <StyledLink
+    href={props.href}
+    rel={isExternalUrl(props.href) && !props.rel ? EXTERNAL_REL_DEFAULT : props.rel}
+    onBlur={props.onBlur}
+    onClick={props.onClick}
+    onFocus={props.onFocus}
+    tabIndex={props.tabIndex}
+    target={isExternalUrl(props.href) && !props.target ? EXTERNAL_TARGET_DEFAULT : props.target}
```
TheSisb: If we don't remove the lint rule, we shouldn't even bother checking `!props.target` here because it's always set.

TheSisb: readability nit: `target={props.target || isExternalUrl(props.href) ? EXTERNAL_TARGET_DEFAULT : null}`

packages/paste-core/components/anchor/src/index.tsx
```diff
@@ -1,48 +1,53 @@
...
+  id?: never;
+  onBlur?(event: React.FocusEvent<HTMLElement>): void;
+  onClick?(event: React.MouseEvent<HTMLElement>): void;
+  onFocus?(event: React.FocusEvent<HTMLElement>): void;
+  rel?: string;
+  tabIndex?: AnchorTabIndexes;
+  target?: AnchorTargets;
+}
 
 const EXTERNAL_LINK_REGEX = /^(https?:)[^\s]*$/;
```
TheSisb: Since we're not using the word "Link" for this component, maybe we can rename this to `EXTERNAL_URL_REGEX`

packages/paste-core/components/anchor/src/index.tsx
```diff
@@ -1,48 +1,53 @@
...
-
-  if (children == null) {
-    throw new Error(`[Paste: Anchor] Must have non-null children.`);
-  }
-
-  if (hasTabIndex && !(tabIndex === 0 || tabIndex === -1)) {
-    throw new Error(`[Paste: Anchor] TabIndex must be 0 or -1.`);
-  }
+const Anchor: React.FC<Anchor> = props => (
+  <StyledLink
```
TheSisb: And rename this to `StyledAnchor`

packages/paste-core/components/anchor/src/index.tsx
richbachman: Going with `target={props.target || isExternalUrl(props.href) ? EXTERNAL_TARGET_DEFAULT : undefined}`

`null` causes a TS error: `Type 'string | null' is not assignable to type 'string | undefined'.`

packages/paste-core/components/anchor/src/index.tsx
richbachman: Turning off the rule.

PR #154, opened 2019-10-29
feat: update types package to styled-system v5 format

TheSisb:
Styled-system v5 has a new grouping of css properties (https://styled-system.com/api/).  [Composing these groupings](https://styled-system.com/api/#compose) is more performant.  For simplicity in development, I wanted to modify our types exports so that rather than importing any types from the styled-system package, we only import types from our package.  The benefit is that it is much easier to copy types you would ordinarily use in styled-system, but supercharged with our token typings.

I've also updated Box and Text to use the new typings, and in the process updated them to the v5 approach.

## Text changes

Text now takes `layout` and `typography` which adds the following under Layout:
```
  width?: Width;
  minWidth?: MinWidth;
  maxWidth?: MaxWidth;
  height?: Height;
  minHeight?: MinHeight;
  maxHeight?: MaxHeight;
  size?: Size;
  verticalAlign?: VerticalAlign;
  overflow?: Overflow;
  overflowX?: OverflowX;
  overflowY?: OverflowY;
```
 and these under Typography
```
  letterSpacing?: LetterSpacing;
  fontStyle?: FontStyle;
```


packages/paste-core/utilities/text/src/index.tsx
```diff
@@ -31,25 +12,25 @@ const textColor = system({
...
+    margin: 0,
+    padding: 0,
+  },
   compose(
-    display,
     space,
-    textColor,
-    textDecoration,
-    typography
+    layout,
```
SiTaggart: So yeah, re: slack, I don't think we need layout for `Text` right now.

PR #166, opened 2019-11-07
fix(button): add font-weight semibold to small buttons

TheSisb:
We updated the regular button to a semibold font-weight but forgot to update the small button.


richbachman:
💯

PR #167, opened 2019-11-07
fix(website): update button docs with loading and icon states

TheSisb:
Updating the button docs to show the loading state and Icon only buttons (now that icons are fixed).


packages/paste-website/src/pages/components/button/index.mdx
```diff
@@ -272,8 +297,52 @@ However, the loading state may make an action appear to take longer
 than it does and doesn’t communicate what’s preventing the action
 from completing. Use it when you can’t move the user to the next state.
 
-<LivePreview scope={{Button}} language="jsx">
-  {`<span>Example coming soon</span>`}
+<LivePreview scope={{Button, Box}} language="jsx">
+  {`<div>
+  <Box display="inline-block" marginRight="space30">
```
SiTaggart: _just_ realised we should apply `aria-busy="true"` when the button is in a loading state too when I was testing with voice over just now.

PR #174, opened 2019-11-11
fix(design-tokens): update lightest console color palette

richbachman:
- [x] changed lightest colors in Console color palette


serifluous:
lgtm!

Issue #75, opened 2019-08-28
Why do we use multipliers for line-height tokens?


serifluous:
Wondering why we use multipliers for line-height tokens instead of, say, rems. Is it an issue with scaling text?

https://github.com/twilio-labs/paste/blob/master/packages/paste-design-tokens/tokens/global/line-height.yml

SiTaggart:
We'll be switching to rem based line-height values for better mix-and-match usage with different font sizes.

PR #176, opened 2019-11-12
Feat(design tokens): Update the line-height token values

SiTaggart:
Update the line-height token values to match the new typography scale.

Switch from unitless to rem based values to allow for better reuse of line-heights to different font sizes. Unit less line heights have advantages when you aren't setting sizes on every node, but we do, so it's not necessary.

This will close https://github.com/twilio-labs/paste/issues/75


richbachman:
lgtm!

PR #179, opened 2019-11-12
chore: use yarn policy for a consistent yarn version in repo

TheSisb:
I added some yarn configuration to our repo to guarantee we are all on the same version.  I recommend reading through https://yarnpkg.com/lang/en/docs/cli/policies/.  I hope this fixes issues during install from machine to machine using different versions, as well as reducing some of the `yarn.lock` thrashing.


SiTaggart:
> I hope this fixes issues during install from machine to machine using different versions, as well as reducing some of the yarn.lock thrashing.

Maybe I'm missing something here, so you can clarify it for me: How would locking the yarn version help our install and lock file churn, when it's the versions of the packages we install that cause the problems and the churn as they get bumped to the latest versions whenever we do a `yarn clean`?

What am I missing?

TheSisb:
We spoke on Zoom but for other readers, the `yarn.lock` churn may not necessarily get fixed. The main purpose of this PR is to bring everyone to the same version of yarn so that we can avoid any potential future setup issues.  An example of this is when people have really old versions of yarn on their machines installed globally.

PR #180, opened 2019-11-12
test(button): adding tests

TheSisb:
Adding tests to the Button component.


packages/paste-core/components/button/__tests__/button.test.tsx
```diff
@@ -0,0 +1,211 @@
...
+
+  it('Throws an error when the user should use an Anchor component instead', () => {
+    expect(() =>
+      shallow(
+        <Button as="a" variant="link" href={HREF}>
+          Go to Paste
+        </Button>
+      )
+    ).toThrowError();
+  });
```
richbachman: Looks like this is a dupe of the test above (lines 164-172).

SiTaggart:
I think we may want some simple tests to make sure the attributes we expect to appear on a button when set, stay getting set. Same with method calls. Just to catch any regressions in future implementation changes.

PR #186, opened 2019-11-15
fix: correct button heights add new default icon size

SiTaggart:
- Correct the button heights by dropping the line-height by one
- Introduce a new icon size so the default icon makes a button the correct height

### DIFF
--- a/packages/paste-core/components/button/src/styles.ts
+++ b/packages/paste-core/components/button/src/styles.ts
@@ -23,14 +23,13 @@ const sizeSmall = (props: ButtonWrapperProps): SerializedStyles => css`
   padding: ${themeGet('space.space10')(props)} ${themeGet('space.space30')(props)};
   border-radius: ${themeGet('radii.borderRadius10')(props)};
   font-size: ${themeGet('fontSizes.fontSize30')(props)};
-  font-weight: ${themeGet('fontWeights.fontWeightSemibold')(props)};
   line-height: ${themeGet('lineHeights.lineHeight20')(props)};
 `;
 const sizeDefault = (props: ButtonWrapperProps): SerializedStyles => css`
   padding: ${themeGet('space.space30')(props)} ${themeGet('space.space50')(props)};
   border-radius: ${themeGet('radii.borderRadius20')(props)};
   font-size: ${themeGet('fontSizes.fontSize30')(props)};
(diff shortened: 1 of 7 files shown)


packages/paste-core/components/button/src/styles.ts
```diff
@@ -46,6 +45,7 @@ const baseButtonWrapper = (props: ButtonWrapperProps): SerializedStyles => css`
   background: none;
   transition: background-color 100ms ease-in, border-color 100ms ease-in;
   font-family: ${themeGet('fonts.fontFamilyText')(props)};
+  font-weight: ${themeGet('fontWeights.fontWeightSemibold')(props)};
```
SiTaggart: Even link style buttons should be semi bold and they are reset size by default https://share.goabstract.com/5343d021-abae-4f49-898e-c95df03667df?mode=design&sha=f133b3b4b51b7372f5805ed57bea98fce5e78d8d

packages/paste-core/components/button/stories/index.stories.tsx
```diff
@@ -6,7 +6,7 @@ import {PlusIcon} from '@twilio-paste/icons/esm/PlusIcon';
 import {Button} from '../src';
 import {ButtonVariants, ButtonSizes, ButtonTabIndexes} from '../src/types';
 
-const ButtonSizeOptions = ['default', 'small', 'icon', 'reset'];
+const ButtonSizeOptions = ['', 'default', 'small', 'icon', 'reset'];
```
SiTaggart: For non-required props its useful to have stories that default to blank. For example, link buttons default to reset size, but we were setting a size in storybook with knobs. 

Setting a size set the font weight to semibold. Reset size was normal.

We didn't have enough coverage for all the options so we didn't catch it. I've added an example of each below for applitools to check going forward

packages/paste-design-tokens/tokens/global/sizing.yml
```diff
@@ -49,8 +49,11 @@ props:
     value: "{!size-square-30}"
     comment: Icon sizing token
   size-icon-20:
-    value: "{!size-square-50}"
+    value: "{!size-square-40}"
```
SiTaggart: Shifting the icon size scale. New default is 20px, shift the last 2 up one in the scale

PR #188, opened 2019-11-15
chore: add the accessibilty plugin to storyboook

SiTaggart:
Add a plugin for a11y that we should have added a long time ago for quick checks

also give in storybook colorblindness simulation.


aayushpi:
I’m so excited for this

PR #189, opened 2019-11-15
fix: update destructive button styles to be filled

SiTaggart:
Update the destructive buttons to be ~outlined~ filled.

also get storybook serving colfax correctly for the SG theme


.storybook/addons/theme-switcher-addon/index.js
```diff
@@ -26,7 +51,12 @@ class ThemeWrapper extends React.Component {
   };
 
   render() {
-    return <Theme.Provider theme={this.state.theme}>{this.props.children}</Theme.Provider>;
+    return (
+      <Theme.Provider theme={this.state.theme}>
+        <Global styles={globalStyles} />
+        <Box padding="space40">{this.props.children}</Box>
```
TheSisb: Thank you 👍

package.json
```diff
@@ -14,7 +14,7 @@
     "prestart": "yarn prebuild",
     "start": "yarn build:tokens && concurrently \"yarn:start:tokens\" \"yarn:start:storybook\"",
     "start:all": "yarn build:tokens && concurrently \"yarn:start:tokens\" \"yarn:start:storybook\" \"yarn:start:website\"",
-    "start:storybook": "start-storybook",
+    "start:storybook": "start-storybook -s ./packages/paste-website/static -p 9001",
```
TheSisb: Not sure how I feel about this coupling.  Could we move these static assets out of the `paste-website` package?

SiTaggart: Seems like Gatsby is pretty fixed on needing that static directory in the root of the gatsby project so it know where to grab the static assets from. https://www.gatsbyjs.org/docs/static-folder/

I can duplicate them for storybook, but that seems not ideal either.

TheSisb: While I agree it isn't ideal, duplication here might be the better alternative.  As a rule we should try to follow, the only things that can reach into other parts of the monorepo are build tooling and npm packages (lerna).  The duplication helps prevent unintended and unexpected regressions.

SiTaggart: Fair, dumped a copy of it under the `.storybook` directory

PR #190, opened 2019-11-18
chore: switch storybook deploy to use gh-pages

SiTaggart:
- Removes the docs directory from the master branch
- Uses the gh-pages branch for storybook deploys
- Uses gh-pages package to automate deploying the storybook docs dorectory to the gh-pages branch


SiTaggart:
> Are we planning on calling release:storybook from a hook?

It's like you read my mind... once I write up my thoughts on the release process it'll explain where and how

PR #195, opened 2019-11-20
fix: remap console semibold to medium font weight

SiTaggart:
As console doesn't load semibold, we can map it to medium via aliases.


packages/paste-design-tokens/tokens/global/font-weight.yml
```diff
@@ -14,7 +14,7 @@ props:
     value: "{!font-weight-medium}"
     comment: Font weight for medium weight
   font-weight-semibold:
-    value: "{!font-weight-semibold}"
+    value: "{!font-weight-medium}"
```
TheSisb: This particular change makes me wish `console` was a theme rather than the base theme others inherit from.

SiTaggart: My quarterly goal is to do this very thing and create the default theme of paste as the output of BPM 7

TheSisb: That sounds awesome.  I'm totally behind that idea 👍

PR #192, opened 2019-11-19
docs: add the heading component documentation

richbachman:
- [x] Setup the `Heading` documentation page for docs site

PR #196, opened 2019-11-20
fix: change heading variants

richbachman:
- [x] Changed headingStyles to headingVariants
- [x] Mapped headingVariants to use our scale system -> 10 being largest and 60 being smallest.
- [x] Updated docs site to use new headingVariants


packages/paste-core/components/heading/src/index.tsx
```diff
@@ -3,63 +3,57 @@ import * as PropTypes from 'prop-types';
...
-  variant?: HeadingStyle;
+  variant?: HeadingVariants;
 }
 
-function getHeadingStyles(headingStyle?: HeadingStyle): {} {
-  switch (headingStyle) {
-    case 'headingStyle60':
+function getHeadingProps(headingVariant?: HeadingVariants): {} {
+  switch (headingVariant) {
+    case 'heading60':
```
TheSisb: Isn't this supposed to be heading10?  Or are we leaving it inverted?

richbachman: I decided to leave it inverted because it goes against the rest of the scaling we use in token. I know it's not a token, but it just felt weird being opposite those. I'm not 100% tied to that idea, so open to changing.

TheSisb: More words: I think as a user, component APIs !== design system core tokens / aliases.  I want the simplest component API.  As it is currently, I have to do `6 - hlevel` every time I want the size and the tag to match, which isn't intuitive.

richbachman: Yeah. I think that makes the most sense. I'll change real quick.

richbachman: Pushed those changes.

packages/paste-core/components/heading/src/index.tsx
```diff
@@ -3,63 +3,57 @@ import * as PropTypes from 'prop-types';
...
       return {
         marginBottom: 'space30',
         fontSize: 'fontSize20',
         fontWeight: 'fontWeightSemibold',
         lineHeight: 'lineHeight20',
       };
     /**
-     * headingStyle50 is out of order because its also default.
+     * heading50 is out of order because its also default.
```
SiTaggart:
```suggestion
     * heading20 is out of order because its also default.
```

SiTaggart:
How are you going to keep the docs up to date with this new API change,m if the docs are in another PR?

richbachman:
@SiTaggart the plan is to merge this PR, then go back and edit the docs in https://github.com/twilio-labs/paste/pull/192 so they align.

PR #200, opened 2019-11-21
feat: add the flex theme to design tokens

SiTaggart:
Create a Flex theme for the design tokens package

- [x] Aliases
- [x] Tokens
- [x] Theme
- [x] Theme Switcher


packages/paste-theme/src/themeProvider.tsx
```diff
@@ -3,13 +3,15 @@ import styled from '@emotion/styled';
...
 import {ThemeProvider as StyledThemeProvider} from 'emotion-theming';
-import {DefaultTheme, SendGridTheme} from '@twilio-paste/theme-tokens';
+import {DefaultTheme, SendGridTheme, FlexTheme} from '@twilio-paste/theme-tokens';
 import {ThemeVariants} from './constants';
 
 interface GlobalStyleProps {
   theme: {};
 }
 const pasteGlobalStyles = (props: GlobalStyleProps): SerializedStyles => css`
+  @import url('https://fonts.googleapis.com/css?family=Open+Sans:400,400i,600,600i,700,700i&display=swap');
```
TheSisb: Do we have plans to load these only for the currently selected theme in the future?

packages/paste-design-tokens/tokens/themes/flex/aliases/color-palette.yml
```diff
@@ -0,0 +1,147 @@
...
+  palette-pink-60: "#BC0959"
+  palette-pink-70: "#8D0743"
+  palette-pink-80: "#5E042C"
+  palette-pink-90: "#2F0216"
+  palette-pink-100: "#000000"
+
+  ## Grays
+  palette-gray-0: "#FFFFFF"
+  palette-gray-10: "#EEEEEE"
+  palette-gray-20: "#DDDDDD"
```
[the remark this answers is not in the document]
SiTaggart: Yup

packages/paste-theme/src/themeProvider.tsx
SiTaggart: Certainly can. It only loads the font files for the font-face that is currently in use though too right? So it's not _dire_ right now

SiTaggart: I'm actually almost tempted to remove entirely and let font / asset handling be a product problem... just have a page on _what_ and _how_ to load them 🤔

TheSisb: So this line in particular still hits the googleapis.com domain then loads that css.  The font files won't be loaded until use, correct.  Theoretically this means you can use any theme's fonts in any other theme, because the font-face rules are defined. Not a big deal, just an observation mainly.

SiTaggart:
Closing, we won't be needing this.

PR #204, opened 2019-11-26
feat(Flex): add Flex component package

richbachman:
- [x] `Flex` component package
- [x] `Flex` stories
- [x] `Flex` unit and snapshot tests


packages/paste-core/utilities/flex/src/index.tsx
```diff
@@ -0,0 +1,110 @@
+import * as React from 'react';
+import * as PropTypes from 'prop-types';
+import {Box} from '@twilio-paste/box';
+import {FlexboxProps} from '@twilio-paste/types';
+
+export type displayOptions = 'flex' | 'inline-flex';
+export type vAlignOptions = 'top' | 'center' | 'bottom';
```
TheSisb: Missing: stretch

packages/paste-core/utilities/flex/src/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+
+const Flex: React.FC<Flex> = props => {
+  return (
+    <Box {...getFlexStyles(props)} display={props.display} order={props.order}>
+      {props.children}
+    </Box>
+  );
+};
+
+Flex.propTypes = {
```
TheSisb: 👍

packages/paste-core/utilities/flex/src/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+    return basis + suffix;
+  }
+
+  return 'auto'; // default
+}
+
+function getFlexStyles(props: Flex): FlexboxProps {
+  const {column, wrap, vAlignContent, hAlignContent} = props;
+
+  function alignPropToFlex(align: Flex['vAlignContent'] | Flex['hAlignContent']) {
```
TheSisb: Move this function outside the `getFlexStyles` function for perf.  By defining it here, it will be rebuilt each re-render ;)

packages/paste-core/utilities/flex/src/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+  }
+
+  if (basis && basis !== 'auto') {
+    return 0;
+  }
+
+  return 1; // default
+}
+
+function getBasis(props: Flex): string {
```
TheSisb: nit: `getBasis({basis} FlexProps): number`

packages/paste-core/utilities/flex/src/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+  if (typeof grow === 'number') {
+    return grow;
+  } else if (grow) {
+    return 1;
+  }
+
+  return 0; // default
+}
+
+function getShrink(props: Flex): number {
```
TheSisb: nit `getShrink({shrink, basis}: FlexProps): number`

packages/paste-core/utilities/flex/src/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+  vAlignContent?: vAlignOptions;
+  hAlignContent?: hAlignOptions;
+  grow?: boolean | number;
+  shrink?: boolean | number;
+  basis?: string | number;
+  order?: number;
+  wrap?: boolean;
+}
+
+function getGrow(props: Flex): number {
```
TheSisb: similar nit to other functions

packages/paste-core/utilities/flex/src/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+      case 'right':
+        return 'flex-end';
+    }
+  }
+
+  return {
+    flex: `${getGrow(props)} ${getShrink(props)} ${getBasis(props)}`,
+    flexDirection: column ? 'column' : 'row',
+    flexWrap: wrap ? 'wrap' : 'nowrap',
+    justifyContent: alignPropToFlex(column ? vAlignContent : hAlignContent),
```
TheSisb: It's a nice touch when the component makes it so you don't have to mentally swap these lines based on the direction 👍

TheSisb: I think it would be better overall to not apply some of these CSS values if none are passed in props.  This would improve perf on two fronts: less work generating these values by bypassing some called functions, and less CSS for the browser to paint. For example
```js
const styles = {};

if (grow || shrink || basis) {
  styles.flex = `${getGrow(props)} ${getShrink(props)} ${getBasis(props)}`;
}
if (column) {
  styles.flexDirection = column ? 'column' : 'row';
}

return styles;
```

packages/paste-core/utilities/flex/stories/index.stories.tsx
```diff
@@ -0,0 +1,41 @@
...
+import {Paragraph} from '@twilio-paste/paragraph';
+import {Flex, displayOptions, vAlignOptions, hAlignOptions} from '../src';
+
+const flexDisplayOptions = ['flex', 'inline-flex'];
+const flexVerticalAlignOptions = ['top', 'center', 'bottom'];
+const flexHorizontalAlignOptions = ['left', 'center', 'right'];
+
+storiesOf('Utilities|Flex', module)
+  .addDecorator(withKnobs)
+  .add('Default', () => {
```
TheSisb: Please make sure to add example stories other than the knob story here.  It's a new direction we're moving in, similar to how Sendgrid has done it. Since our VRT relies on storybook stories to operate and it can't really toggle the knob values, we should have stories that provide real usage examples.  Beyond just VRT, use-case based examples here will help consumers learn and be inspired by our components.

packages/paste-core/utilities/flex/src/index.tsx
richbachman: So fresh and so clean!

packages/paste-core/utilities/flex/src/index.tsx
```diff
@@ -0,0 +1,110 @@
...
+    alignItems: alignPropToFlex(column ? hAlignContent : vAlignContent),
+  };
+}
+
+const Flex: React.FC<Flex> = props => {
+  return (
+    <Box {...getFlexStyles(props)} display={props.display} order={props.order}>
```
SiTaggart: I'm worried about supporting order as when it is used to change the visual appearance from then natural flow of the DOM, it will cause accessibility failures. Visually elements should appear in the same order as you would progress through the DOM with a screen reader or keyboard focus, in the direction that is suited to language.

So in English pages should visually traverse left to right, top to bottom as you tab through the page. Covered under [logical focus order](https://www.w3.org/TR/UNDERSTANDING-WCAG20/navigation-mechanisms-focus-order.html)

packages/paste-core/utilities/flex/src/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+      {props.children}
+    </Box>
+  );
+};
+
+Flex.propTypes = {
+  display: PropTypes.oneOf(['flex', 'inline-flex']),
+  column: PropTypes.bool,
```
SiTaggart: Would `column` be better as `direction=column|row`?

packages/paste-core/utilities/flex/src/index.tsx
```diff
@@ -0,0 +1,123 @@
...
+    case 'stretch':
+      return 'stretch';
+    case 'around':
+      return 'space-around';
+    case 'between':
+      return 'space-between';
+  }
+}
+
+function getFlexStyles(props: FlexProps): FlexboxProps {
```
SiTaggart: Returning `FlexboxProps` is tricky, as we're not returning responsive values, so the typings are wrong

packages/paste-core/utilities/flex/src/index.tsx
TheSisb: bike shedding: what about `horizontal, vertical`?

packages/paste-core/utilities/flex/src/index.tsx
TheSisb: Good point.  I've never really used it myself in live code and always thought of it as a crutch for bad markup or SEO hackery which doesn't matter in 2019.  I'm +1 for removing it.

richbachman: Agreed on this. Removing.

packages/paste-core/utilities/flex/src/index.tsx
richbachman: I think I like `vertical`. Keeping it a `boolean`. So if `vertical` then `flex-direction = column`. See latest.

TheSisb: I'm into it

packages/paste-core/utilities/flex/src/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+import * as PropTypes from 'prop-types';
+import {Box} from '@twilio-paste/box';
+import {FlexboxProps} from '@twilio-paste/types';
+
+export type displayOptions = 'flex' | 'inline-flex';
+export type vAlignOptions = 'top' | 'center' | 'bottom' | 'stretch';
+export type hAlignOptions = 'left' | 'center' | 'right' | 'around' | 'between';
+
+export interface FlexProps {
+  display?: displayOptions;
```
TheSisb: `import {ResponsiveValue} from 'styled-system';`
```
ResponsiveValue<DisplayOptions>;
```
similar treatment to the other types here

packages/paste-core/utilities/flex/src/index.tsx
```diff
@@ -0,0 +1,128 @@
...
+  if (basis) {
+    const suffix = typeof basis === 'number' || String(parseInt(basis as string, 10)) === basis ? 'px' : '';
+    return basis + suffix;
+  }
+
+  return 'auto'; // default
+}
+
+function alignPropToFlex(align: FlexProps['vAlignContent'] | FlexProps['hAlignContent']): {} {
+  switch (align) {
```
TheSisb: then here check is `Array.isArray(align)` to see if you have to loop the responsive values and generate an array return value

packages/paste-core/utilities/flex/__tests__/flex.test.tsx
```diff
@@ -0,0 +1,317 @@
...
+              <Box backgroundColor="colorBackgroundBrand" minWidth="size20" minHeight="size10" />
+            </Flex>
+            <Flex>
+              <Box backgroundColor="colorBackgroundBrandHighlight" minWidth="size20" minHeight="size10" />
+            </Flex>
+          </Flex>
+        </Theme.Provider>
+      )
+      .toJSON();
+    expect(tree).toMatchSnapshot();
```
TheSisb: We should replace `toMatchSnapshot` with an actual check to see if the property was set.  you can check the button tests to see how

packages/paste-core/utilities/flex/src/index.tsx
```diff
@@ -0,0 +1,191 @@
...
+import {Box} from '@twilio-paste/box';
+import {FlexboxProps} from '@twilio-paste/types';
+
+export type DisplayOptions = ResponsiveValue<'flex' | 'inline-flex'>;
+type VerticalAlignOptions = 'top' | 'center' | 'bottom' | 'stretch';
+export type VerticalAlign = ResponsiveValue<VerticalAlignOptions>;
+type HorizontalAlignOptions = 'left' | 'center' | 'right' | 'around' | 'between';
+export type HorizontalAlign = ResponsiveValue<HorizontalAlignOptions>;
+export type VerticalOptions = ResponsiveValue<boolean | 'column' | 'row'>;
+export type GrowOptions = ResponsiveValue<boolean | number>;
```
TheSisb: Let's remove the `Options` suffix from these types for consistency

packages/paste-core/utilities/flex/src/index.tsx
```diff
@@ -0,0 +1,191 @@
...
+  vAlignContent?: VerticalAlign;
+  hAlignContent?: HorizontalAlign;
+  grow?: GrowOptions;
+  shrink?: ShrinkOptions;
+  basis?: BasisOptions;
+  wrap?: WrapOptions;
+}
+
+const getGrow = ({grow}: FlexProps): {} => {
+  if (typeof grow === 'number' || Array.isArray(grow)) {
```
TheSisb: what if grow is `[true, false, true]`? Will this still work or do we need to convert it into a number?  (note: we should test for this if the current tests didn't catch the error)

packages/paste-core/utilities/flex/src/index.tsx
```diff
@@ -0,0 +1,191 @@
...
+    return grow;
+  }
+
+  if (grow) {
+    return 1;
+  }
+
+  return 0; // default
```
TheSisb: nit: I don't think these comments provide much value, they could be safely removed.

packages/paste-core/utilities/flex/src/index.tsx
```diff
@@ -0,0 +1,191 @@
...
+    return shrink;
+  }
+
+  if (shrink) {
+    return 1;
+  }
+
+  if (shrink === false) {
+    return 0;
+  }
```
TheSisb: I think this would be more clear as 
```
if (typeof shrink === 'boolean') {
  return shrink ? 1 : 0;
}
```

packages/paste-core/utilities/flex/src/index.tsx
```diff
@@ -0,0 +1,191 @@
...
+
+  if (grow) {
+    return 1;
+  }
+
+  return 0; // default
+};
+
+const getShrink = ({shrink, basis}: FlexProps): {} => {
+  if (typeof shrink === 'number' || Array.isArray(shrink)) {
```
TheSisb: Same question here about if `[true, false]` would work?

packages/paste-core/utilities/flex/src/index.tsx
```diff
@@ -0,0 +1,191 @@
...
+
+  if (basis && basis !== 'auto') {
+    return 0;
+  }
+
+  return 1; // default
+};
+
+const getSuffix = (item: BasisOptions): {} => {
+  const suffix = typeof item === 'number' || String(parseInt(item as string, 10)) === item ? 'px' : '';
```
TheSisb: We should check if this would work with `rem`, `%` or other types of values.  Would be nice to pull this out into it's own testable function too.

packages/paste-core/utilities/flex/src/index.tsx
```diff
@@ -0,0 +1,191 @@
...
+  return item + suffix;
+};
+
+const getBasis = ({basis}: FlexProps): {} => {
+  if (Array.isArray(basis)) {
+    return (basis as BasisOptions[]).map((value: BasisOptions) => {
```
TheSisb: `BasisOptions[]` means an array of `ResponsiveValue<>`, which isn't quite right.  You should apply the same treatment here as we did on horizontal/vertical typings.

packages/paste-core/utilities/flex/src/index.tsx
```diff
@@ -0,0 +1,191 @@
...
+
+  if (basis) {
+    return getSuffix(basis);
+  }
+
+  return 'auto'; // default
+};
+
+const getVertical = ({vertical}: FlexProps): {} => {
+  if (Array.isArray(vertical)) {
```
TheSisb: Same question about if this `[true, false]` or something else.  Need tests.

packages/paste-core/utilities/flex/src/index.tsx
```diff
@@ -0,0 +1,191 @@
...
+
+  if (hAlignContent) {
+    return RemapedHorizontalAlignments[hAlignContent as HorizontalAlignOptions];
+  }
+
+  return 'flex-start'; // default
+};
+
+const getFlexStyles = (props: FlexProps): FlexboxProps => {
+  const {basis, vertical, grow, shrink, wrap, hAlignContent, vAlignContent} = props;
```
TheSisb: We can simplify this by passing props into every function directly

richbachman: Similar. I think I have it right. Let me know.

packages/paste-core/utilities/flex/src/index.tsx
richbachman: I think I have that setup right now. Let me know if not.

packages/paste-core/utilities/flex/src/index.tsx
richbachman: Agreed. Added some tests for `px`, `rem`, `em`, and `%`.

packages/paste-core/utilities/flex/__tests__/flex.test.tsx
richbachman: I'm seeing `toMatchSnapshot` in the button tests as well. What am I missing there?

packages/paste-core/utilities/flex/src/index.tsx
richbachman: Right. That did break `grow`, but I've converted the true/false values to a number. Works now. Number values also work, a user can enter `[true, 2, false]` if they want. Added a test as well.

packages/paste-core/utilities/flex/src/index.tsx
richbachman: Did a similar thing here so `[0, 1, true]` works.

packages/paste-core/utilities/flex/src/index.tsx
richbachman: For this one, I changed the function so `[row, false, column]` will work, as well as `[true, false, true]`

packages/paste-core/utilities/flex/__tests__/flex.test.tsx
TheSisb: Scroll way down lol, there's a different method we're using on aria- attributes

richbachman: Unit tests added.

PR #206, opened 2019-12-02
fix(spinner): export the prop types with the component

SiTaggart:
Each component should export it's prop types so a consumer can extend them or use them in dynamic prop setting situations.

### DIFF
--- a/packages/paste-core/components/spinner/src/index.tsx
+++ b/packages/paste-core/components/spinner/src/index.tsx
@@ -29,7 +29,7 @@ const SpinningWrapper = styled.div<SpinnerWrapperProps>(
   })
 );
 
-interface SpinnerProps extends LoadingIconProps {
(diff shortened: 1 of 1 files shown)


packages/paste-core/components/spinner/src/index.tsx
```diff
@@ -45,4 +45,4 @@ Spinner.defaultProps = {
 };
 
 Spinner.displayName = 'Spinner';
-export {Spinner};
+export {Spinner, SpinnerProps};
```
richbachman: Could we change `SpinnerProps` to `Spinner` like we've done in other components like `Header` and `Paragraph`?

SiTaggart: I think we decided going forward we'd revert back to naming them `${Component}Props` just in case we do any components with props and state.

PR #218, opened 2019-12-04
fix(console): change "color-text-inverse-weak" for a11y

serifluous:
- [x] In console theme, `color-text-inverse-weak` changed to `palette-gray-50` to pass AA contrast with `color-background-inverse`

It didn't originally pass AA as its original value of `palette-gray-60`.


packages/paste-design-tokens/tokens/global/text-color.yml
```diff
@@ -16,7 +16,7 @@ props:
     value: "{!palette-gray-0}"
     comment: Inverse text color for dark backgrounds
   color-text-inverse-weak:
-    value: "{!palette-gray-60}"
+    value: "{!palette-gray-50}"
```
TheSisb: Is the -60 value still fine on the Sendgrid theme?

packages/paste-design-tokens/tokens/global/text-color.yml
```diff
@@ -16,7 +16,7 @@ props:
...
     comment: Weak inverse text color for dark backgrounds
```
TheSisb: Are there any notes or gotchas related to this change we should add in this comment line?

packages/paste-design-tokens/tokens/global/text-color.yml
serifluous: Yep, it passes AA (https://www.aremycoloursaccessible.com/?background=%2308152c&colorCombos=%5Bobject%20Object%5D&colorCombos=%5Bobject%20Object%5D&isLight=false&textColor=%239ca7ba)

packages/paste-design-tokens/tokens/global/text-color.yml
serifluous: Do you mean whether we should change the comment to, e.g., "Weak inverse text color for dark backgrounds. Must pass AA color contrast with color-background-inverse"?

Or did you have something else in mind?

TheSisb: Nothing specific in mind, just wanted to make sure to ask.  What you wrote above seems good to me!

packages/paste-design-tokens/tokens/global/text-color.yml
serifluous: ok actually, just noticed `color-background-inverse` isn't defined in the SendGrid theme tokens. I'll add it.

PR #219, opened 2019-12-04
chore: fix the lockfile discrepencies introduced by dependabot

SiTaggart:
As dependabot doesn't support `lerna bootstrap` it was changing the lockfile to include linked internal deps as dependencies. This results in the lockfile being always different from our local installs and CI installs. Meaning releases would be impossible as the git status wouldn't be clean.

Putting it back to the way it was


package.json
```diff
@@ -80,9 +80,7 @@
     "@types/react": "^16.8.23",
     "@types/react-dom": "^16.8.4",
     "@types/react-test-renderer": "^16.8.2",
-    "@types/storybook__addon-actions": "^3.4.3",
```
SiTaggart: actions is deprecated and removed

package.json
```diff
@@ -80,9 +80,7 @@
...
-    "@types/storybook__addon-info": "^4.1.2",
-    "@types/storybook__react": "^4.0.2",
```
SiTaggart: react is deprecated and removed

PR #221, opened 2019-12-05
feat(list): create the list component package

SiTaggart:
Create the list component package

- [x] Create the package
- [x] Create stories
- [x] Tests


packages/paste-core/components/list/src/index.tsx
```diff
@@ -0,0 +1,72 @@
+import * as React from 'react';
+import {Text, NULLIFY_TEXT_PROPS} from '@twilio-paste/text';
+
+interface BaseListProps extends React.OlHTMLAttributes<''> {
```
SiTaggart: This is now effectively typed as an HTML list (there is no UlHTMLAttributes, I assume they are identical). So you can apply any HTML attribute that is relevant to being on a list and we can safely spread those onto `Text`

packages/paste-core/components/list/src/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+  className?: never;
+  style?: never;
+  as: 'ol' | 'ul';
+}
+
+const List: React.FC<BaseListProps> = ({as, children, ...props}) => {
+  return (
+    <Text
+      {...props}
+      {...NULLIFY_TEXT_PROPS}
```
SiTaggart: Nullify any sneaky style props people might be forcing through, or able to pass when not using Typescript.

packages/paste-core/components/list/src/index.tsx
```diff
@@ -0,0 +1,72 @@
...
+const UnorderedList: React.FC<UnorderedListProps> = ({children, ...props}) => {
+  return (
+    <List {...props} as="ul">
+      {children}
+    </List>
+  );
+};
+UnorderedList.displayName = 'UnorderedList';
+
+export interface ListItemProps extends React.LiHTMLAttributes<''> {
```
SiTaggart: Same deal, legit typed as an `<li />` element, totally safe to pass any prop to. Styles won't change

packages/paste-core/utilities/text/src/index.tsx
```diff
@@ -30,7 +30,12 @@ const Text = styled.span<Text>(
     typography,
     textColor
   )
-);
+  // we do this because the default typings of emotion styled
+  // means Text gets typed as a span, and can't be extended
+  // correctly to utilise the as prop. The new HTML element attributes
+  // always clash with the span html attributes. To override this,
+  // we retype as a basic functional component which is easy to extend
+) as React.FC<Text>;
```
SiTaggart: This is the fun part

packages/paste-style-props/package.json
```diff
@@ -0,0 +1,45 @@
+{
+  "name": "@twilio-paste/style-props",
```
SiTaggart: I know we discussed shoving these into types, but honestly it just didn't feel right, and this feels more appropriate. Even pulling in the types for style-props into here felt better.

packages/paste-style-props/src/border.ts
```diff
@@ -0,0 +1,43 @@
+export const BORDER_RADIUS_PROPS = {
```
SiTaggart: These are all based on the typings. So in the same way you can compose groups of these to nullify different design primtives.

packages/paste-core/utilities/text/src/NullifyTextProps.ts
```diff
@@ -0,0 +1,12 @@
+import {SPACE_PROPS, TYPOGRAPHY_PROPS, OVERFLOW_PROPS} from '@twilio-paste/style-props';
+
+export const NULLIFY_TEXT_PROPS = {
+  ...SPACE_PROPS,
+  ...TYPOGRAPHY_PROPS,
+  ...OVERFLOW_PROPS,
+  display: undefined,
+  textColor: undefined,
+  verticalAlign: undefined,
+  className: undefined,
```
SiTaggart: Still need these to stop doing bad things.

packages/paste-core/utilities/text/src/index.tsx
```diff
@@ -1,9 +1,9 @@
 import styled from '@emotion/styled';
 import {compose, space, display, verticalAlign, overflow, typography, system} from 'styled-system';
-import {SpaceProps, Display, VerticalAlign, OverflowProps, TypographyProps} from '@twilio-paste/types';
+import {SpaceProps, Display, VerticalAlign, OverflowProps, TypographyProps} from '@twilio-paste/style-props';
 
-interface Text extends SpaceProps, OverflowProps, TypographyProps {
-  as?: keyof JSX.IntrinsicElements;
+interface Text extends React.AllHTMLAttributes<''>, SpaceProps, OverflowProps, TypographyProps {
```
SiTaggart: Need to allow basic HTML attributes on `Text`, stuff like id doesn't come for free.

SiTaggart: This may or may not need to be `React.AllHTMLAttributes<any>` as it is literally _any element_ that text can be. Just not sure if `''` is _proper_ typescript. It works 🤷‍♂

packages/paste-core/utilities/text/src/SafelySpreadProps.ts
```diff
@@ -0,0 +1,17 @@
+import {SPACE_PROPS, TYPOGRAPHY_PROPS, OVERFLOW_PROPS} from '@twilio-paste/style-props';
+
+export const PROPS_TO_NULLIFY = {
```
SiTaggart: This is just an object of all the props `Text` can take and set to undefined.

packages/paste-core/utilities/text/src/SafelySpreadProps.ts
```diff
@@ -0,0 +1,17 @@
...
+  ...TYPOGRAPHY_PROPS,
+  ...OVERFLOW_PROPS,
+  display: undefined,
+  textColor: undefined,
+  verticalAlign: undefined,
+  className: undefined,
+  style: undefined,
+};
+
+export const safelySpreadProps = (props: {}): typeof props & typeof PROPS_TO_NULLIFY => ({
```
SiTaggart: This is the magic. To safely spread any props onto `Text`, we take `props` and merge with all the `Text` undefined props. This way if you've tried to restyle one of our style props, we reset it back to undefined, before setting the real styles in the component itself.

This isn't typed properly yet, but basically I want to type the return as the type of the props coming in and the `PROPS_TO_NULLIFY` as a combined object. I'm sure you can do it, but I suck at Typescript (seemingly like 99% of the rest of the internet)

packages/paste-core/utilities/text/src/SafelySpreadProps.ts
```diff
@@ -0,0 +1,17 @@
...
+
+export const PROPS_TO_NULLIFY = {
+  ...SPACE_PROPS,
+  ...TYPOGRAPHY_PROPS,
+  ...OVERFLOW_PROPS,
+  display: undefined,
+  textColor: undefined,
+  verticalAlign: undefined,
+  className: undefined,
+  style: undefined,
```
SiTaggart: Still need to unset classname and style just in case you're doing "bad things" with our styled components.

richbachman:
The List package looks good. 

@TheSisb gave me a high level overview of the spreading props issue during our working session. The solution you have here looks like a good solution.

packages/paste-core/utilities/box/src/SafelySpreadProps.ts
```diff
@@ -0,0 +1,30 @@
...
+  ...SHADOW_PROPS,
+  ...POSITION_PROPS,
+  ...FLEXBOX_PROPS,
+  'backgroundColor',
+  'borderColor',
+  'className',
+  'style',
+];
+
+export const safelySpreadBoxProps = (props: {}, compName: string): void => {
```
TheSisb: I'm not a fan of the compName argument here because it complicates the API. When we throw, the stacktrace should show with file the error originated from.

packages/paste-core/components/list/package.json
```diff
@@ -2,6 +2,45 @@
   "name": "@twilio-paste/list",
   "version": "0.0.2",
   "category": "typography",
-  "status": "backlog",
-  "private": true
+  "status": "beta",
+  "description": "The paragraph component is used to render blocks of text, with a preset amount of space beneath it to distinguish from adjacent content.",
```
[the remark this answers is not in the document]
SiTaggart: This an outdated diff you're commenting on

packages/paste-core/utilities/box/src/SafelySpreadProps.ts
```diff
@@ -0,0 +1,30 @@
...
+  const badProps = Object.keys(props).filter(prop => BOX_PROPS_TO_BLOCK.includes(prop));
+  if (badProps.length > 0) {
+    throw new Error(`${badProps.join(', ')} cannot be restyled for the ${compName} component`);
```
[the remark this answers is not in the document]
SiTaggart: How would silently swallowing prevent the consumer from spreading bad props onto the Text or Box component?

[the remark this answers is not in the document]
TheSisb: So consumer passes `props = {className: 'apple', borderColor: 'red', variant: 'default'}`
this function takes that and returns `props = {variant: 'default'}`

No errors are thrown, nothing bad gets passed down.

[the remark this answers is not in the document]
SiTaggart: OK, so back to mutating the object. That works for me

packages/paste-core/components/list/stories/index.stories.tsx
```diff
@@ -0,0 +1,68 @@
...
+import {storiesOf} from '@storybook/react';
+import {withKnobs} from '@storybook/addon-knobs';
+import {Paragraph} from '@twilio-paste/paragraph';
+import {OrderedList, UnorderedList, ListItem} from '../src';
+
+storiesOf('Components|List', module)
+  .addDecorator(withKnobs)
+  .add('Ordered List', () => {
+    return (
+      <>
```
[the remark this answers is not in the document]
SiTaggart: We do, they are sibling jsx elements with no single parent element.

packages/paste-core/utilities/box/src/SafelySpreadProps.ts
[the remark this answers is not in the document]
SiTaggart: Updated to pass back a filtered props object

packages/paste-core/utilities/box/src/SafelySpreadProps.ts
```diff
@@ -0,0 +1,35 @@
...
+  ...POSITION_PROPS,
+  ...FLEXBOX_PROPS,
+  'backgroundColor',
+  'borderColor',
+  'className',
+  'style',
+];
+
+export const safelySpreadBoxProps = (props: {}): {} => {
+  // https://www.measurethat.net/Benchmarks/Show/6642/0/for-in-vs-reduce-vs-pick#latest_results_block
```
TheSisb: Nice!

PR #232, opened 2019-12-16
docs: add typography section to heading

richbachman:
Proposed solution for not having a `Typography` component page on the docs site. Page is unnecessary at this point since it mainly just links to the `Heading` and `Paragraph` pages.

Similar text will need to be added to `Paragraph` and `List` pages when ready.

- [x] Added Typography component paragraph to `Heading`
- [x] Removed `Typography` component from left nav and component list


richbachman:
@aayushpi @serifluous @two24studios will you all take a look at this copy I added to the `Heading` page referencing the `Typography` component, and let me know if that works? Thanks!

serifluous:
+1 to @two24studios's edit.

This section could make more sense placed between the "Installation" and "Usage" sections.

richbachman:
@serifluous What about as part of Composition notes? I'm mentioning Typography component install instructions in the install and usage sections.

serifluous:
It's more that the API section could use more explanation about what the Typography component is, because right now it just shows how to install and doesn't explain what's included.

richbachman:
@serifluous I added it after the install section of API.

packages/paste-website/src/components/component-overview-table/index.tsx
```diff
@@ -29,7 +29,7 @@ const ComponentOverviewTable: React.FC<ComponentOverviewTableProps> = ({category
     .filter(({node}) => node.status === PackageStatus.BACKLOG)
     .sort(sortNodeByName);
   const sortedComponentsList = componentsList
-    .filter(({node}) => node.status !== PackageStatus.BACKLOG)
+    .filter(({node}) => node.status !== PackageStatus.BACKLOG && node.name !== '@twilio-paste/typography')
```
TheSisb: This line made me notice that the typography meta-package is in the components folder.  It should probably be a sibling to it, like the core package.  No? @SiTaggart thoughts appreciated here too

SiTaggart: Yeah, potentially actually.

Would core still include typography, heading, paragraph and list? Or (my thinking) should typography just be it's own package outside of core, and core is just a bundle of the individual components?

SiTaggart: Further reasoning is I noticed there's a bit of an exported name clash when you build and deploy core. 

```
core
  exports typography
    exports heading
    exports paragraph
  exports heading
  exports paragraph
```
so heading and paragraph clash (or something, forget the exact warning)

TheSisb: I'm starting to feel like the typography package is unnecessary complexity.  `core` and individual packages should be sufficient.

SiTaggart: > I'm starting to feel like the typography package is unnecessary complexity.

Now we're talking, simplify. I lean towards removing too

richbachman: I agree on this one. May have one been a good idea, but unnecessary now with `core`. So we need a ticket to deprecate the `typography` package? Any others?

TheSisb: I think just Typography.  We need to remove any references to it from other packages/docs too

richbachman: Jira ticket created. I'm going to close this PR and open another one to remove references from other packages/docs.

PR #234, opened 2019-12-17
fix: add prop-types package, add description, bump types

richbachman:
- [x] Add missing prop-types package
- [x] Bump types version
- [x] Add description


SiTaggart:
Kind of weird that it built... 🤔

richbachman:
I thought so as well. Didn't catch it until I was importing into the docs site, and it was blowing that build up.

richbachman:
Talked to @TheSisb about this. The original build worked because Prop-Types was already in the node_modules from other component builds. Once I imported into the website project, Prop-Types didn't come with it.

PR #254, opened 2020-01-09
feat(alert): create alert component

SiTaggart:
- New status tokens to support the alert
- update media-object to use box so it can be used in alert
- create the alert component
- add alert to core package

### DIFF
--- a/packages/paste-core/components/alert/__tests__/index.spec.tsx
+++ b/packages/paste-core/components/alert/__tests__/index.spec.tsx
@@ -1,0 +1,186 @@
+import * as React from 'react';
+import renderer from 'react-test-renderer';
+import {render} from 'react-dom';
+import {ReactWrapper, mount} from 'enzyme';
+import {axe} from 'jest-axe';
+import {Theme} from '@twilio-paste/theme';
+import {Alert} from '../src';
+
+const onDismissMock: jest.Mock = jest.fn();
+
+describe('Alert', () => {
+  describe('Dismiss button', () => {
+    it('Should add a dismiss button when onDismiss is passed as a function to call', () => {
+      const eventHandlerMock: jest.Mock = jest.fn();
+      const wrapper: ReactWrapper = mount(
+        <Alert onDismiss={eventHandlerMock} variant="neutral">
+          This is an alert
+        </Alert>
+      );
+      expect(wrapper.find('button').length).toEqual(1);
+    });
+
+    it('Should call the onDismiss event handler when close button clicked', () => {
+      const eventHandlerMock: jest.Mock = jest.fn();
+      const wrapper: ReactWrapper = mount(
+        <Alert onDismiss={eventHandlerMock} variant="neutral">
+          This is an alert
+        </Alert>
+      );
+      wrapper.find('button').simulate('click');
+      expect(eventHandlerMock).toHaveBeenCalledTimes(1);
+    });
+  });
+
+  describe('Aria roles', () => {
+    it('Should add the role of status to the neutral alert', () => {
+      const wrapper: ReactWrapper = mount(<Alert variant="neutral">This is an alert</Alert>);
+      expect(wrapper.exists('[role="status"]')).toEqual(true);
+    });
+
+    it('Should add the role of alert to the error alert', () => {
+      const wrapper: ReactWrapper = mount(<Alert variant="error">This is an alert</Alert>);
+      expect(wrapper.exists('[role="alert"]')).toEqual(true);
+    });
+
+    it('Should add the role of alert to the warning alert', () => {
+      const wrapper: ReactWrapper = mount(<Alert variant="warning">This is an alert</Alert>);
+      expect(wrapper.exists('[role="alert"]')).toEqual(true);
+    });
+
+    it('Should add the provided role to the alert', () => {
+      const wrapper: ReactWrapper = mount(
+        <Alert role="tab" variant="error">
+          This is an alert
+        </Alert>
+      );
+      expect(wrapper.exists('[role="tab"]')).toEqual(true);
+    });
+  });
+
+  describe('Variant neutral', () => {
+    it('Should render a neutral alert', (): void => {
+      const tree = renderer
+        .create(
+          <Theme.Provider theme="console">
+            <Alert variant="neutral">This is an alert</Alert>
+          </Theme.Provider>
+        )
+        .toJSON();
+      expect(tree).toMatchSnapshot();
+    });
+
+    it('Should render a neutral alert with dismiss button', (): void => {
+      const tree = renderer
+        .create(
+          <Theme.Provider theme="console">
+            <Alert onDismiss={onDismissMock} variant="neutral">
+              This is an alert
+            </Alert>
+          </Theme.Provider>
+        )
+        .toJSON();
+      expect(tree).toMatchSnapshot();
+    });
+
+    it('Should have no accessibility violations', async () => {
+      const container = document.createElement('div');
+      document.body.append(container);
+      render(
+        <Theme.Provider theme="console">
+          <Alert variant="neutral">This is a neutral alert</Alert>
+          <Alert onDismiss={onDismissMock} variant="neutral">
+            This is a neutral alert
+          </Alert>
+        </Theme.Provider>,
+        container
+      );
+      const results = await axe(document.body);
+      expect(results).toHaveNoViolations();
+    });
+  });
+
+  describe('Variant error', () => {
+    it('Should render an error alert', (): void => {
+      const tree = renderer
+        .create(
+          <Theme.Provider theme="console">
+            <Alert variant="error">This is an error alert</Alert>
+          </Theme.Provider>
+        )
+        .toJSON();
+      expect(tree).toMatchSnapshot();
+    });
+
+    it('Should render an error alert with dismiss button', (): void => {
+      const tree = renderer
+        .create(
+          <Theme.Provider theme="console">
+            <Alert onDismiss={onDismissMock} variant="error">
+              This is an error alert
+            </Alert>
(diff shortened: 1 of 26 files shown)


packages/paste-core/components/alert/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,186 @@
...
+      const wrapper: ReactWrapper = mount(<Alert variant="error">This is an alert</Alert>);
+      expect(wrapper.exists('[role="alert"]')).toEqual(true);
+    });
+
+    it('Should add the role of alert to the warning alert', () => {
+      const wrapper: ReactWrapper = mount(<Alert variant="warning">This is an alert</Alert>);
+      expect(wrapper.exists('[role="alert"]')).toEqual(true);
+    });
+
+    it('Should add the provided role of alert to the alert', () => {
```
TheSisb: `'Should add the provided role of alert to the alert` -> `'Should add the provided role to the alert`?

packages/paste-core/components/alert/src/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+export type BackgroundColors = ValueOf<typeof BACKGROUND_COLORS>;
+export type BorderColors = ValueOf<typeof BORDER_COLORS>;
+export type Roles = ValueOf<typeof ROLES>;
+
+export interface AlertProps {
+  id?: never;
+  className?: never;
+  children: NonNullable<React.ReactNode>;
+  onDismiss?: () => void;
+  role?: string;
```
TheSisb: Just curious, this can't be Roles because there are possibly other types of roles that would work here?

packages/paste-core/components/alert/src/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+    return ROLES.error;
+  }
+  return ROLES.neutral;
+};
+
+const Alert: React.FC<AlertProps> = ({children, onDismiss, variant, role, ...props}) => {
+  return (
+    <Box
+      {...safelySpreadBoxProps(props)}
+      backgroundColor={getBackgroundColor(variant)}
```
TheSisb: Since we share the same keys among all the objects defined above on lines 12-31, we can do something like this and remove all these functions/switches

```js
backgroundColor={BACKGROUND_COLORS[variant]}
borderColor={BORDER_COLORS[variant]}
role={role != null ? role : ROLES[variant]}
```
Typescript should make this safe to do, but additionally we can define the objects like this
```js
const BackgroundColors = {
 [ERROR]: 'colorBackgroundErrorLightest'
  ...
}
```
in JS land to make sure the keys matched.  We don't need that here though

packages/paste-core/components/alert/src/index.tsx
```diff
@@ -0,0 +1,126 @@
...
+import {ValueOf} from '@twilio-paste/types';
+import {Box, safelySpreadBoxProps} from '@twilio-paste/box';
+import {MediaObject, MediaFigure, MediaBody} from '@twilio-paste/media-object';
+import {Button} from '@twilio-paste/button';
+import {CloseIcon} from '@twilio-paste/icons/esm/CloseIcon';
+import {ErrorIcon} from '@twilio-paste/icons/esm/ErrorIcon';
+import {InformationIcon} from '@twilio-paste/icons/esm/InformationIcon';
+import {WarningIcon} from '@twilio-paste/icons/esm/WarningIcon';
+
+export const ROLES = {
```
TheSisb: mega nit: these constants should be PascalCase and the Keys all UpperCased for consistency (paste-website/src/constants.ts)

packages/paste-core/core-bundle/src/index.tsx
```diff
@@ -1,4 +1,5 @@
 export * from '@twilio-paste/absolute';
+export * from '@twilio-paste/alert';
```
TheSisb: I'm concerned `core` may not work this way anymore, since we're exporting `VARIANTS` from the Alert package, and undoubtedly another component will export Variants too.  Either we prefix all exports with the component name (i.e.: `AlertVariants`) in the component packages, or we manually maintain this file to do the renaming.  Thoughts?

packages/paste-core/utilities/media-object/src/index.tsx
```diff
@@ -1,50 +1,45 @@
...
+  verticalAlign?: 'center' | 'top';
 }
 
-const MediaObject = styled.div<MediaObjectProps>`
-  align-items: ${(props: MediaObjectProps) => (props.isCentered ? 'center' : 'flex-start')};
-  display: flex;
-`;
+const MediaObject: React.FC<MediaObjectProps> = ({children, verticalAlign = 'top'}) => {
+  return (
+    <Box display="flex" alignItems={verticalAlign === 'top' ? 'flex-start' : 'center'}>
```
TheSisb: Should we use the Flex component here? 🤔

packages/paste-core/utilities/media-object/src/index.tsx
```diff
@@ -1,50 +1,45 @@
...
-const getMargin = (props: MediaFigureProps): string => {
-  const direction = props.align === 'end' ? 'left' : 'right';
-  const spacing = themeGet(`space.${props.spacing}`)(props);
-  return `margin-${direction}: ${spacing}`;
+const MediaFigure: React.FC<MediaFigureProps> = ({children, align, spacing}) => {
+  return (
+    <Box
+      display="flex"
+      flexShrink={0}
+      marginLeft={align === 'end' ? spacing : undefined}
```
TheSisb: I don't think the Flex component currently allows margins, which may be an oversight after looking at this..

packages/paste-design-tokens/tokens/global/background-color.yml
```diff
@@ -27,6 +27,9 @@ props:
     comment: Background color of the main page body
 
   # status
+  color-background-neutral-lightest:
+    value: "{!palette-blue-10}"
+    comment: Lightest neutral background color
```
TheSisb: Should we be more explicit about these tokens, saying that they're used in Alerts and that they are paired with certain other colors?

packages/paste-design-tokens/tokens/global/text-color.yml
```diff
@@ -55,16 +55,22 @@ props:
...
     value: "{!palette-orange-60}"
     comment: Color for warning text.
+  color-text-warning-dark:
+    value: "{!palette-orange-70}"
+    comment: Color for dark warning text.
 
   #icon
   color-text-icon:
-    value: "{!palette-gray-100}"
+    value: "{!palette-gray-70}"
```
TheSisb: Just curious about this change, `color-text` still looks to be set to `gray-100`.  Why was this changed?

packages/paste-core/components/alert/src/index.tsx
SiTaggart: Yup, it's basically a free for all is you really want to do still things other than making alerts and status'

packages/paste-core/utilities/media-object/src/index.tsx
SiTaggart: Thought about it, but there reason I didn't was that Alert would require Media-object as a peer, then that has a peer on Flex, and that has a peer on Box, and I use _so little_ of Flex it just didn't seem worth it

packages/paste-core/components/alert/src/index.tsx
SiTaggart: oooh yeah, that's much better

packages/paste-core/core-bundle/src/index.tsx
SiTaggart: Switching all the exports, for sure this is going to be a problem

packages/paste-design-tokens/tokens/global/background-color.yml
SiTaggart: We should, and we have a ticket in the backlog to look at token meta data and pairing information. I'd punt to making these more desccriptive and detailed then so we can decide what we are expecting for all tokens

packages/paste-design-tokens/tokens/global/text-color.yml
SiTaggart: Good catch, total oversight.

packages/paste-design-tokens/tokens/themes/console/global/text-color.yml
```diff
@@ -56,16 +56,22 @@ props:
...
     value: "{!palette-orange-60}"
     comment: Color for warning text.
+  color-text-warning-dark:
+    value: "{!palette-orange-70}"
+    comment: Color for dark warning text.
 
   #icon
   color-text-icon:
-    value: "{!palette-gray-100}"
+    value: "{!palette-gray-70}"
```
TheSisb: And this one too?

packages/paste-design-tokens/tokens/themes/sendgrid/global/text-color.yml
```diff
@@ -53,16 +53,22 @@ props:
...
     value: "{!palette-orange-60}"
     comment: Color for warning text.
+  color-text-warning-dark:
+    value: "{!palette-orange-70}"
+    comment: Color for dark warning text.
 
   #icon
   color-text-icon:
-    value: "{!palette-gray-100}"
+    value: "{!palette-gray-70}"
```
TheSisb: 70 -> 100 too?

packages/paste-design-tokens/tokens/themes/console/global/text-color.yml
SiTaggart: Oh, no these are to spec for both console and sendgrid. So I'm going to put it back for the global theme

richbachman:
🚢

PR #255, opened 2020-01-09
feat(grid): create grid component

richbachman:
- [x] Create `Grid` component
- [x] Add `Grid` stories to storybook
- [x] Update `Flex` component with support for margins, padding, min-width, and width
- [x] Add `Grid` to `Core` package

### DIFF
--- a/packages/paste-core/utilities/flex/src/index.tsx
+++ b/packages/paste-core/utilities/flex/src/index.tsx
@@ -1,8 +1,8 @@
 import * as React from 'react';
 import * as PropTypes from 'prop-types';
 import {ResponsiveValue} from 'styled-system';
-import {Box} from '@twilio-paste/box';
-import {FlexboxProps} from '@twilio-paste/style-props';
+import {Box, safelySpreadBoxProps} from '@twilio-paste/box';
+import {FlexboxProps, LayoutProps, MarginProps, PaddingProps} from '@twilio-paste/style-props';
 
 type DisplayOptions = 'flex' | 'inline-flex';
 export type Display = ResponsiveValue<DisplayOptions>;
@@ -21,7 +21,7 @@ export type Basis = ResponsiveValue<BasisOptions>;
 type WrapOptions = boolean;
 export type Wrap = ResponsiveValue<WrapOptions>;
 
-export interface FlexProps {
+export interface FlexProps extends LayoutProps, MarginProps, PaddingProps {
   display?: Display;
   vertical?: Vertical;
   vAlignContent?: VerticalAlign;
@@ -187,11 +187,51 @@ const getFlexStyles = (props: FlexProps): FlexboxProps => {
   return styles;
 };
 
-const Flex: React.FC<FlexProps> = props => {
-  const FlexStyles = React.useMemo(() => getFlexStyles(props), [props]);
+const Flex: React.FC<FlexProps> = ({
+  basis,
+  children,
+  display,
+  hAlignContent,
+  grow,
+  marginTop,
+  marginRight,
+  marginBottom,
+  marginLeft,
+  paddingTop,
+  paddingRight,
+  paddingBottom,
+  paddingLeft,
+  maxWidth,
+  minWidth,
+  shrink,
+  vertical,
+  vAlignContent,
+  width,
+  wrap,
+  ...props
+}) => {
+  const FlexStyles = React.useMemo(
+    () => getFlexStyles({basis, hAlignContent, grow, shrink, vertical, vAlignContent, wrap}),
+    [basis, hAlignContent, grow, shrink, vertical, vAlignContent, wrap]
+  );
   return (
-    <Box {...FlexStyles} display={props.display}>
-      {props.children}
+    <Box
+      {...FlexStyles}
+      {...safelySpreadBoxProps(props)}
+      display={display}
+      marginTop={marginTop}
+      marginRight={marginRight}
+      marginBottom={marginBottom}
+      marginLeft={marginLeft}
+      paddingTop={paddingTop}
+      paddingRight={paddingRight}
+      paddingBottom={paddingBottom}
+      paddingLeft={paddingLeft}
+      maxWidth={maxWidth}
+      minWidth={minWidth}
+      width={width}
+    >
+      {children}
     </Box>
   );
 };
--- a/packages/paste-core/utilities/grid/src/Column.tsx
+++ b/packages/paste-core/utilities/grid/src/Column.tsx
@@ -1,0 +1,138 @@
+import * as React from 'react';
+import * as PropTypes from 'prop-types';
+import styled from '@emotion/styled';
+import {compose, layout, space, flexbox, ResponsiveValue} from 'styled-system';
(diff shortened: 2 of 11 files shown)


packages/paste-core/utilities/grid/src/index.tsx
```diff
@@ -0,0 +1,75 @@
...
+
+const getColumns = ({children}: GridProps): number => {
+  return React.Children.count(children);
+};
+
+const Grid: React.FC<GridProps> = props => {
+  const Count = getColumns(props);
+
+  const GridColumns = React.Children.map(props.children, child =>
+    React.isValidElement(child) ? React.cloneElement(child, {count: Count, gutter: props.gutter}) : child
```
SiTaggart: Generally speaking being reliant on, inspecting, modifying components another component doesn't own isn't something we wanna do.

So, why shouldn't I be concerned about this?

richbachman: `React.cloneElement` is pretty safe from a performance standpoint. Its basically just copying an object and passing props. For a more in depth answer: https://stackoverflow.com/questions/54922160/react-cloneelement-in-list-performance

packages/paste-core/utilities/flex/src/index.tsx
```diff
@@ -190,7 +190,7 @@ const getFlexStyles = (props: FlexProps): FlexboxProps => {
 const Flex: React.FC<FlexProps> = props => {
   const FlexStyles = React.useMemo(() => getFlexStyles(props), [props]);
   return (
-    <Box {...FlexStyles} display={props.display}>
+    <Box {...FlexStyles} display={props.display} marginLeft={props.marginLeft} marginRight={props.marginRight}>
```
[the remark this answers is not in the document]
SiTaggart: Yup, I think we should.

packages/paste-core/utilities/grid/src/Column.tsx
```diff
@@ -0,0 +1,119 @@
...
+type OffsetOptions = number;
+export type Offset = ResponsiveValue<OffsetOptions>;
+type SpanOptions = number;
+export type Span = ResponsiveValue<SpanOptions>;
+
+export interface ColumnProps extends Box {
+  count?: number;
+  gutter?: ResponsiveValue<keyof ThemeShape['space']>;
+  offset?: Offset;
+  span?: Span;
```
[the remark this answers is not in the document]
SiTaggart: I don't _hate_ it. I'd be cautious on `colspan` as it's very table specific, though.

packages/paste-core/utilities/grid/src/Grid.tsx
```diff
@@ -0,0 +1,70 @@
...
+    marginRight: getNegativeMargin(props, theme),
+  };
+
+  return marginStyles;
+};
+
+const Grid: React.FC<GridProps> = props => {
+  const count = React.Children.count(props.children);
+
+  const GridColumns = React.Children.map(props.children, child =>
```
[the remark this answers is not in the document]
richbachman: Yeah. I think you're right. Once set the children wouldn't change very often.

packages/paste-core/utilities/grid/src/Column.tsx
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  if (typeof offset === 'number') {
+    return `${(offset / 12) * 100}%`;
+  }
+};
+
+const getColumnStyles = (props: ColumnProps): LayoutProps => {
+  const columnStyles: FlexboxProps & LayoutProps & MarginProps & PaddingProps = {
+    width: getSpan(props),
```
richbachman: Tried to use `Basis` here instead, but it didn't really work the `minWidth` coming from `Box`. So I passed in `width` and `minWidth` props in `Flex`.

packages/paste-core/utilities/grid/src/Grid.tsx
```diff
@@ -0,0 +1,73 @@
...
+  const marginStyles: MarginProps = {
+    marginLeft: getNegativeMargin(props, theme),
+    marginRight: getNegativeMargin(props, theme),
+  };
+
+  return marginStyles;
+};
+
+const Grid: React.FC<GridProps> = ({children, gutter, marginTop, marginBottom, vertical, ...props}) => {
+  const count = React.Children.count(children);
```
TheSisb: You should move this inside the useMemo function, so we don't call it every re-render when GridColumns doesn't need to update.

A good rule of thumb is to just put everything related to hooks (useEffect, useMemo, useCallback, etc) inside the hook function.  It makes the code much more resilient because it's clear what the dependencies are for it to work simply due to code proximity.  I think it's recommended on the React website somewhere too.

packages/paste-core/utilities/grid/src/Grid.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+  const GridColumns = React.useMemo(
+    () =>
+      React.Children.map(children, child =>
+        React.isValidElement(child) ? React.cloneElement(child, {count, gutter, vertical}) : child
+      ),
+    [children]
+  );
+
+  const GridStyles = React.useMemo(() => getGridStyles({children, gutter, vertical}), [{children, gutter, vertical}]);
```
TheSisb: Same issue on the second argument here (object wrapping)

TheSisb: Other than `gutter`, I don't think `children` and `vertical` are being used here?

packages/paste-core/utilities/grid/src/Grid.tsx
```diff
@@ -0,0 +1,73 @@
...
+
+  if (gutter) {
+    return `-${theme.space[gutter as string]}`;
+  }
+
+  return 'auto';
+};
+
+const getGridStyles = (props: GridProps): MarginProps => {
+  const theme = useTheme();
```
TheSisb: For clarity, let's make theme a required argument to this function and put this line inside the function on line 50

packages/paste-core/utilities/grid/src/Grid.tsx
richbachman: Since `children` is a required prop, if I omit from here I get TS errors.

packages/paste-core/utilities/grid/src/Column.tsx
richbachman: For `Grid`, we need to use an absolute width. `Flex-basis` would still allow the columns to grow if the content within them is wider. Basically `flex-basis` is not a guaranteed size.

packages/paste-core/utilities/grid/src/Column.tsx
```diff
@@ -0,0 +1,118 @@
...
+  }
+
+  if (vertical) {
+    return '100%';
+  }
+
+  return '0';
+};
+
+const getSpan = ({count, span}: ColumnProps): {} => {
```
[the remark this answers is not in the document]
richbachman: `ResponsiveValue<string>` is causing TS errors. I think we need to return as an object in order to work around not using our tokens for width and margins.

PR #271, opened 2020-01-23
fix(style-props): add non-responsive options

TheSisb:
Exports non-responsive typings so that we can properly type code, like this here https://github.com/twilio-labs/paste/pull/255/files#diff-065500df3a0119bd45665e09d97fe283R23

Helps unblock https://github.com/twilio-labs/paste/pull/255

### DIFF
--- a/packages/paste-style-props/src/types/flexbox.ts
+++ b/packages/paste-style-props/src/types/flexbox.ts
@@ -11,7 +11,8 @@ import {ResponsiveValue, TLengthStyledSystem} from 'styled-system';
  *
  * [MDN reference](https://developer.mozilla.org/en-US/docs/Web/CSS/align-items)
  */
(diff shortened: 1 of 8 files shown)


packages/paste-style-props/src/types/flexbox.ts
```diff
@@ -11,74 +11,100 @@ import {ResponsiveValue, TLengthStyledSystem} from 'styled-system';
...
+  | CSS.SelfPosition
+  | 'auto'
+  | 'baseline'
+  | 'left'
+  | 'normal'
+  | 'right'
+  | 'stretch'
+  | string;
 export type JustifySelf =
   | ResponsiveValue<CSS.Globals | CSS.SelfPosition | 'auto' | 'baseline' | 'left' | 'normal' | 'right' | 'stretch'>
```
SiTaggart: Should this line be `ResponsiveValue<JustifySelfOptions>`?

TheSisb: /shrug
Spoke in person

PR #272, opened 2020-01-23
chore(ci): use github actions for publishing packages

SiTaggart:
- Use GitHub actions for publishing packages via lerna on merge to master

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -23,7 +23,8 @@ jobs:
         env:
           GH_TOKEN: ${{ secrets.GH_TOKEN }}
           GH_USER_NAME: ${{ secrets.GH_USER_NAME }}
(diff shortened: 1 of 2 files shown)


.github/workflows/publish.yml
```diff
@@ -0,0 +1,40 @@
...
+        with:
+          node-version: ${{ matrix.node-version }}
+          registry-url: "https://registry.npmjs.org"
+          scope: twilio-paste
+      - name: "Checkout master"
+        env:
+          GH_TOKEN: ${{ secrets.GH_TOKEN }}
+          GH_USER_NAME: ${{ secrets.GH_USER_NAME }}
+          GH_EMAIL: ${{ secrets.GH_EMAIL }}
+        run: git remote rm origin && git remote add origin "https://$GH_USER_NAME:$GH_TOKEN@github.com/twilio-labs/paste.git" && git fetch && git config user.email $GH_EMAIL && git config user.name $GH_USER_NAME"
```
SiTaggart: The checkout action used to have issues with getting into a detacted head state, so I needed to do this manually. 

It may not be necessary but I would like something that works on merge (as you can't test on a branch very easily, withoout actually commiting and publishing)

PR #280, opened 2020-01-25
chore: play with github actions

SiTaggart:
Playing with using GitHub actions as the CI tool


SiTaggart:
Closing for now until such time as we need to switch, if ever. As least we know it's possible to migrate to actions

PR #282, opened 2020-01-27
chore: stop removing the lock file when we clean

SiTaggart:
We no longer need to remove the lockfile when we clean. It's also actively causing build breaks when packages update.

Also missing a yarn install before the bootstrap


SiTaggart:
> I thought yarn bootstrap already handled the yarn install command?

It only runs `lerna bootstrap`. I know in CI if you don't run yarn install first it doesn't install the root dependencies which are all the build tooling.

PR #285, opened 2020-01-31
chore(grid): rename and move functions, move types

richbachman:
- [x] setup helper file and moved functions
- [x] setup types file and moved types

### DIFF
--- a/packages/paste-core/utilities/grid/src/Column.tsx
+++ b/packages/paste-core/utilities/grid/src/Column.tsx
@@ -3,7 +3,7 @@ import * as PropTypes from 'prop-types';
 import styled from '@emotion/styled';
 import {compose, layout, space, flexbox} from 'styled-system';
 import {ColumnProps, ColumnStyleProps} from './types';
-import {getFlexDirection, getColumnOffset, getColumnSpan} from './helpers';
+import {getStackedColumns, getColumnOffset, getColumnSpan} from './helpers';
 
 const getColumnStyles = (props: ColumnProps): ColumnStyleProps => {
   const columnStyles: ColumnStyleProps = {
@@ -26,7 +26,7 @@ const getColumnStyles = (props: ColumnProps): ColumnStyleProps => {
   }
 
   if (props.vertical && !props.offset) {
-    columnStyles.minWidth = getFlexDirection(props.vertical);
+    columnStyles.minWidth = getStackedColumns(props.vertical);
     columnStyles.marginLeft = 'space0';
   }
 
--- a/packages/paste-core/utilities/grid/src/Column.tsx
+++ b/packages/paste-core/utilities/grid/src/Column.tsx
@@ -1,79 +1,13 @@
 import * as React from 'react';
 import * as PropTypes from 'prop-types';
 import styled from '@emotion/styled';
-import {compose, layout, space, flexbox, ResponsiveValue} from 'styled-system';
-import {FlexboxProps, LayoutProps, PaddingProps, Space} from '@twilio-paste/style-props';
-import {FlexProps, Vertical} from '@twilio-paste/flex';
+import {compose, layout, space, flexbox} from 'styled-system';
+import {ColumnProps, ColumnStyleProps} from './types';
+import {getFlexDirection, getColumnOffset, getColumnSpan} from './helpers';
 
-type ColumnMinWidth = ResponsiveValue<'100%' | '0'>;
-type ColumnWidthSpan = ResponsiveValue<string>;
-type ColumnOffsetOptions = number;
-export type ColumnOffset = ResponsiveValue<ColumnOffsetOptions>;
-type ColumnSpanOptions = number;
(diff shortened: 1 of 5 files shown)


SiTaggart:
You can probably get away with this not being a breaking change just now considering we haven't even advertised it's existance yet.

richbachman:
@SiTaggart I wondered about that. Also since these functions aren't used outside of `Grid`. Do I just need to amend my commit message to remove the breaking change part?

SiTaggart:
As you're only touch grid files, you can squash and merge and amend the commit message in the github UI

packages/paste-core/utilities/grid/src/Column.tsx
```diff
@@ -88,11 +22,11 @@ const getColumnStyles = (props: ColumnProps): ColumnStyles => {
   }
 
   if (props.offset) {
-    columnStyles.marginLeft = getOffset(props.offset);
+    columnStyles.marginLeft = getColumnOffset(props.offset);
   }
 
   if (props.vertical && !props.offset) {
-    columnStyles.minWidth = getVertical(props.vertical);
+    columnStyles.minWidth = getFlexDirection(props.vertical);
```
SiTaggart: This feels like it still might clash in the future potentially

packages/paste-core/utilities/grid/src/helpers.tsx
```diff
@@ -0,0 +1,71 @@
...
+import {
+  ColumnOffset,
+  ColumnOffsetOptions,
+  ColumnProps,
+  ColumnSpanOptions,
+  ColumnMinWidth,
+  ColumnWidthSpan,
+} from './types';
+
+export const getNegativeMargin = (theme: ThemeShape, gutter?: Space): Margin => {
```
SiTaggart: `getOuterGutterPull`? or something that describes why you need to do it?

richbachman: That works for me. It explains what it's doing in a better way.

packages/paste-core/utilities/grid/src/Column.tsx
richbachman: Yeah. I suspect that will be the case. What about `getStackedColumns`?

packages/paste-core/utilities/grid/src/helpers.tsx
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  if (gutter) {
+    return `-${theme.space[gutter as SpaceOptions]}` as SpaceOptions;
+  }
+
+  return 'auto';
+};
+
+export const getFlexDirection = (vertical: Vertical): ColumnMinWidth => {
```
SiTaggart: Even though this is exported from another file in the package, when it's part of the core bundle would another exported method called `getFlexDirection` from another package, when bundled together still cause a name  clash @TheSisb

packages/paste-core/utilities/grid/src/helpers.tsx
```diff
@@ -0,0 +1,71 @@
...
+import {
+  ColumnOffset,
+  ColumnOffsetOptions,
+  ColumnProps,
+  ColumnSpanOptions,
+  ColumnMinWidth,
+  ColumnWidthSpan,
+} from './types';
+
+export const getOuterGutterPull = (theme: ThemeShape, gutter?: Space): Margin => {
```
TheSisb: Let's give these functions some comments now that they aren't in proximity of where they're being used.

packages/paste-core/utilities/grid/src/helpers.tsx
TheSisb: This is safe because when `core` imports packages it does it like this:

image

When we import from a package name rather than from a file (i.e.: `from '@paste/icons/cjs/coolIcon.js';`), the resolver will check the `package.json` file for which file to look up:

image

Since we're importing from the index file, only the exports from that file are available.  This is a neat trick because you package can have a bunch of different files with exports, but none get exposed to the user unless you explicitly export them from the entry point.  This allows us to test internal functions easily.

TheSisb:
Since nothing changes to the end user*, there aren't any breaking changes.  Breaking changes should only be considered if projects that depend on this package can't blindly update anymore.  All (?) of these changes are just shuffling code around, so they're all refactors (chores).

*: If someone was depending on an export from index.js that we exposed but didn't intend to expose, it would be theoretically be breaking.  However undocumented APIs don't have to be declared as a breaking change when they're modified.

packages/paste-core/utilities/grid/src/helpers.tsx
richbachman: @TheSisb comments added.

packages/paste-core/utilities/grid/src/helpers.tsx
SiTaggart: Nice, thanks for the explanation

PR #288, opened 2020-02-04
test(grid): unit tests and snapshot tests

richbachman:
- [x] `Grid` unit tests
- [x] `Grid` snapshot tests

### DIFF
--- a/packages/paste-core/utilities/grid/__tests__/grid.test.tsx
+++ b/packages/paste-core/utilities/grid/__tests__/grid.test.tsx
@@ -3,7 +3,6 @@ import renderer from 'react-test-renderer';
 import {DefaultTheme} from '@twilio-paste/theme-tokens';
 import {Theme} from '@twilio-paste/theme';
 import {Space} from '@twilio-paste/style-props';
-import {Box} from '@twilio-paste/box';
 import {Grid, Column} from '../src';
 import {getOuterGutterPull, getStackedColumns, getColumnOffset, getColumnSpan} from '../src/helpers';
 
@@ -95,35 +94,13 @@ describe('Grid Unit Tests', () => {
 });
 
 describe('Grid Gutter Prop', () => {
-  it('it should not set a gutter property', (): void => {
-    const tree = renderer
-      .create(
-        <Theme.Provider theme="console">
-          <Grid>
-            <Column>
-              <Box backgroundColor="colorBackgroundPrimaryLight" width="100%" />
-            </Column>
-            <Column>
-              <Box backgroundColor="colorBackgroundPrimaryLight" width="100%" />
-            </Column>
-          </Grid>
-        </Theme.Provider>
-      )
-      .toJSON();
-    expect(tree).toMatchSnapshot();
-  });
-
   it('it should set a gutter property', (): void => {
     const tree = renderer
       .create(
         <Theme.Provider theme="console">
           <Grid gutter="space20">
-            <Column>
-              <Box backgroundColor="colorBackgroundPrimaryLight" width="100%" />
-            </Column>
-            <Column>
-              <Box backgroundColor="colorBackgroundPrimaryLight" width="100%" />
-            </Column>
+            <Column />
+            <Column />
           </Grid>
         </Theme.Provider>
       )
@@ -136,12 +113,8 @@ describe('Grid Gutter Prop', () => {
       .create(
         <Theme.Provider theme="console">
           <Grid gutter={['space10', 'space20', 'space30']}>
-            <Column>
-              <Box backgroundColor="colorBackgroundPrimaryLight" width="100%" />
-            </Column>
-            <Column>
-              <Box backgroundColor="colorBackgroundPrimaryLight" width="100%" />
-            </Column>
+            <Column />
+            <Column />
           </Grid>
         </Theme.Provider>
       )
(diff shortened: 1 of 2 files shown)


packages/paste-core/utilities/grid/__tests__/grid.test.tsx
```diff
@@ -0,0 +1,588 @@
...
+import {Theme} from '@twilio-paste/theme';
+import {Space} from '@twilio-paste/style-props';
+import {Box} from '@twilio-paste/box';
+import {Grid, Column} from '../src';
+import {getOuterGutterPull, getStackedColumns, getColumnOffset, getColumnSpan} from '../src/helpers';
+
+describe('Grid Unit Tests', () => {
+  it('it should return: -0.25rem', (): void => {
+    const mockGutter: Space = 'space20';
+    expect(getOuterGutterPull(DefaultTheme, mockGutter)).toStrictEqual('-0.25rem');
```
TheSisb: These tests seem.... fragile.  If we change the token, the test will break right?  I don't think returning `-0.25rem` matters to whether Grid works well or not, does it?  Can we test less precise things?  Essentially, any test that expects a token to be a specific value is fragile.

packages/paste-core/utilities/grid/__tests__/grid.test.tsx
```diff
@@ -0,0 +1,588 @@
...
+              <Box backgroundColor="colorBackgroundPrimaryLight" width="100%" />
+            </Column>
+            <Column>
+              <Box backgroundColor="colorBackgroundPrimaryLight" width="100%" />
+            </Column>
+          </Grid>
+        </Theme.Provider>
+      )
+      .toJSON();
+    expect(tree).toMatchSnapshot();
```
TheSisb: I don't think testing `it should not set a gutter property` should be done with a snapshot test because of two thoughts:
- This code is not deterministic (https://jestjs.io/docs/en/snapshot-testing#2-tests-should-be-deterministic)
- The same code should not be used for two different snapshot tests (this and line 154)

packages/paste-core/utilities/grid/__tests__/grid.test.tsx
```diff
@@ -0,0 +1,588 @@
...
+              <Box backgroundColor="colorBackgroundPrimaryLight" width="100%" />
+            </Column>
+            <Column>
+              <Box backgroundColor="colorBackgroundPrimaryLight" width="100%" />
+            </Column>
+            <Column>
+              <Box backgroundColor="colorBackgroundPrimaryLight" width="100%" />
+            </Column>
+            <Column>
+              <Box backgroundColor="colorBackgroundPrimaryLight" width="100%" />
```
TheSisb: Can we omit these child `Box`s from our tests?  The only functionality we're testing is the `Grid` and `Column` components really.  An empty div would work if a child is required `<div />`

packages/paste-core/utilities/grid/__tests__/grid.test.tsx
```diff
@@ -0,0 +1,588 @@
...
+            </Column>
+          </Grid>
+        </Theme.Provider>
+      )
+      .toJSON();
+    expect(tree).toMatchSnapshot();
+  });
+
+  it('it should render a 11 span, 1 span column grid', (): void => {
```
TheSisb: We should test:
- a `5span 3span` column grid (sum <12)
- a `11span 2span` column grid (sum >12)

packages/paste-core/utilities/grid/__tests__/grid.test.tsx
richbachman: Is it worth unit testing this function? Right it takes a token, and returns the negative value of that token. If anything other than a token else is passed to it, it returns undefined.

packages/paste-core/utilities/grid/__tests__/grid.test.tsx
richbachman: I'll add these tests. They don't return any errors, but the column math gets wonky. We should consider adding a console warning saying something about `span` totally out to 12. We are adding information to the spec doc about the 12 column support.

packages/paste-core/utilities/grid/__tests__/grid.test.tsx
SiTaggart: We should test the functionality, but you should mock the inputs. You always want to make sure that give a gutter, it returns the negative value of that gutter as a key from the space object on a theme. And given an array of gutters, it returns an array of values that match the keys on the theme object.

What theme doesn't matter, what gutters doesn't matter.

SiTaggart:
```ts
const mockTheme = {
  space: {
    space10: '2px',
    space20: '4px',
    space30: '6px',
    space40: '8px'
  }
}
const mockGutter = 'space20'
cosnt mockGutters = ['space10', 'space30', 'space20']

describe('Grid Unit Tests', () => {

  it('should return a single negative margin value', (): void => {
    expect(getOuterGutterPull(mockTheme, mockGutter)).toStrictEqual('-4px');
  }

  it('should return a responsive set of negative margin values', (): void => {
    expect(getOuterGutterPull(mockTheme, mockGutters)).toStrictEqual(['2px', '6px', '4px']);
  }

}
```

packages/paste-core/utilities/grid/__tests__/grid.test.tsx
```diff
@@ -0,0 +1,429 @@
...
+    const mockSpan = {
+      count: 12,
+      span: [3, 6, 3],
+    };
+    expect(getColumnSpan(mockSpan)).toStrictEqual(['25%', '50%', '25%']);
+  });
+});
+
+describe('Grid Gutter Prop', () => {
+  it('it should set a gutter property', (): void => {
```
SiTaggart: What does "properly" mean, exactly?

SiTaggart: What does "property" mean, exactly? What are we expecting to see in the snapshot

packages/paste-core/utilities/grid/__tests__/grid.test.tsx
```diff
@@ -0,0 +1,429 @@
...
+  });
+
+  it('it should return span: 8.333333333333332%', (): void => {
+    const mockSpan = {
+      count: 12,
+    };
+    expect(getColumnSpan(mockSpan)).toStrictEqual('8.333333333333332%');
+  });
+
+  it('it should return span: 8.333333333333332% (null span)', (): void => {
```
SiTaggart: Whats this testing? it's not immediately obvious as the test case is almost the same and the output is identical. Is there a more written english was to say what we're testing here?

packages/paste-core/utilities/grid/__tests__/grid.test.tsx
richbachman: This makes more sense now that I'm seeing the example. Thanks.

packages/paste-core/utilities/grid/__tests__/grid.test.tsx
richbachman: I've reworded most of the test descriptions so they're more readable, and easier to read whats being tested.

packages/paste-core/utilities/grid/__tests__/grid.test.tsx
richbachman: Test descriptions rewritten.

PR #291, opened 2020-02-06
feat(website): create the screen-reader-only documentation page

SiTaggart:
Create a page on the website to document the Screen Reader Only component

Set it to beta

Delete jank breaking change entry in the changelog


packages/paste-website/src/pages/utilities/screen-reader-only/index.mdx
```diff
@@ -0,0 +1,200 @@
+---
+title: Screen Reader Only - Utilities
+description: The Screen reader only component should be used to visually hide content that may provide extra context to assistive technology users.
```
serifluous: "Screen reader only" --> "Screen Reader Only"

packages/paste-website/src/pages/utilities/screen-reader-only/index.mdx
```diff
@@ -0,0 +1,200 @@
...
+When wrapped with the Screen Reader Only component, content is visually hidden but still available to assistive technology.
+
+This can be useful when context or relationships between UI controls are _visually_ clear, but in isolation to an assistive technology user, they may not be. By being able to visually hide some content, you are able to provide additional clarifying context that may only be clear to a sighted user.
+
+
+## Examples
+
+### Providing headings to sectioning content
+
+All sectioning elements, such as Cards, should contain a heading to provide a clear document outline and form of navigation to assistive technology users. By using the Screen Reader Only component you can provide a visually hidden heading as a means to give additional context and heirarchy.
```
serifluous: ~~as a means~~

packages/paste-website/src/pages/utilities/screen-reader-only/index.mdx
```diff
@@ -0,0 +1,200 @@
...
+  <ScreenReaderOnly as="div">
+    <Heading as="h2" variant="heading20">Some content</Heading>
+  </ScreenReaderOnly>
+  <Paragraph>
+    TODO: Find good example
+  </Paragraph>
+</Card>`}
+</LivePreview>
+
+### Repeatative buttons and hidden headers
```
serifluous: Repeatative --> Repetitive

packages/paste-website/src/pages/utilities/screen-reader-only/index.mdx
```diff
@@ -0,0 +1,200 @@
...
+
+Each column in a table is required to have a column header. Sometimes those column headers can be visually hidden from sighted users if desired. To do so use the Screen Reader Only component to hide the cell content for that column header.
+
+Another common usecase is for providing context to repetative actions. Visually the "Buy" button is associated with the phone number via the row highlighting, but to a screen reader user the "Buy" action might be unclear. To provide additional clarification we can add _what_ you would be buying as visually hidden text to the button.
```
serifluous: - usecase --> use case
- repetative --> repetitive

packages/paste-website/src/pages/utilities/screen-reader-only/index.mdx
```diff
@@ -0,0 +1,212 @@
...
+
+<content>
+
+## Guidelines
+
+### About Screen Reader Only
+
+<Paragraph>{props.pageContext.frontmatter.description}</Paragraph>
+
+When wrapped with the Screen Reader Only component, content is visually hidden but still available to assistive technology.
```
TheSisb: I wonder if we should always refer to it as `ScreenReaderOnly` (one word) to disambiguate from some abstract "Screen reader only" thing.  Reason being, I had to read over this sentence twice to follow along.

packages/paste-website/src/pages/utilities/screen-reader-only/index.mdx
```diff
@@ -0,0 +1,212 @@
...
+
+This can be useful when context or relationships between UI controls are _visually_ clear, but in isolation to an assistive technology user, they may not be. By being able to visually hide some content, you are able to provide additional clarifying context that may only be clear to a sighted user.
```
TheSisb: The first sentence is not super easy to read.  Might be easier if we flipped these two sentences as so:
> This component allows you to _visually_ hide something from the page that you only want a screen-reader to see.  Usually, these are things that are clear when seen visually but may need more context when navigating through sound.

packages/paste-website/src/pages/utilities/screen-reader-only/index.mdx
```diff
@@ -0,0 +1,212 @@
...
+## Examples
+
+### Providing headings to sectioning content
+
+All sectioning elements, such as Cards, should contain a heading to provide a clear document outline and form of navigation to assistive technology users. By using the Screen Reader Only component you can provide a visually hidden heading to give additional context and heirarchy.
+
+<LivePreview scope={{Card, ScreenReaderOnly, Heading, Paragraph, Button, Anchor, Flex}} language="jsx">
+  {`<Card>
+  <ScreenReaderOnly as="div">
+    <Heading as="h2" variant="heading20">Flex</Heading>
```
TheSisb: `Flex` -> `Twilio Flex`

packages/paste-website/src/pages/utilities/screen-reader-only/index.mdx
```diff
@@ -0,0 +1,212 @@
...
+Use the Screen Reader Only component when you are required to provide additional or contextual information to a screen reader user which a sighted user maybe able to infer and when you cannot do so via semantic HTML.
+
+<DoDont>
+  <Do
+    title="Do"
+    body="Use to provide additional text content to users of assistive technology"
+    preview={false}
+  />
+  <Dont title="Don't" body="Don't use to hide interactive content" preview={false} />
+</DoDont>
```
richbachman: Any worth to adding a Do/Dont about:
- Adding meaningful text that relates to the what Screen Reader Only is added to
- Dont add spammy text

GiantRobots:
Do we have tests that validate this component works? Is that something we can do easily or is it a lot of work? (Not really documentation related)

packages/paste-website/src/pages/utilities/screen-reader-only/index.mdx
```diff
@@ -0,0 +1,212 @@
...
+
+```jsx
+import {ScreenReaderOnly} from '@twilio-paste/screen-reader-only';
+
+<Button variant="secondary" size="small">
+  Buy <ScreenReaderOnly>+1 (484) 662-4088</ScreenReaderOnly>
+</Button>
+```
+
+All the regular HTML attributes (`role`, `aria-*`, `type`, and so on) including the following custom props:
```
GiantRobots: Should we have recommendations for `aria-describedBy` and screen reader only?

packages/paste-website/src/pages/utilities/screen-reader-only/index.mdx
```diff
@@ -0,0 +1,212 @@
...
+
+<LivePreview scope={{Table, Thead, Tbody, Tr, Td, Th, Button, Text, ScreenReaderOnly}} language="jsx">
+  {`<Table>
+  <Thead>
+    <Tr>
+      <Th>Number</Th>
+      <Th>Type</Th>
+      <Th>Monthly Fee</Th>
+      <Th>
+        <ScreenReaderOnly>Actions</ScreenReaderOnly>
```
GiantRobots: Should we be recommending this pattern? Actions should probably be visible in the table here.

packages/paste-website/src/pages/utilities/screen-reader-only/index.mdx
```diff
@@ -0,0 +1,212 @@
...
+      <Td>$1.00</Td>
+      <Td>
+        <Button variant="secondary" size="small">
+          Buy <ScreenReaderOnly>+1 (484) 662-4088</ScreenReaderOnly>
+        </Button>
+      </Td>
+    </Tr>
+  </Tbody>
+</Table>`}
+</LivePreview>
```
GiantRobots: I know this is a useful pattern but am struggling to know _when_ to use it more examples would definitely help

GiantRobots: A little research uncovered a few patterns 
https://webaim.org/techniques/css/invisiblecontent/#skipnavlinks

```html
<div class=”screen-reader-only”>These are screen reader user specific instructions.</div>
<div aria-hidden=”true”>These are instructions for everyone else.</div>
```

https://accessible360.com/accessible360-blog/use-aria-label-screen-reader-text/

GiantRobots:
Do we have any guidance on using media queries and screen readers? ie. a Table column disappears on smaller screen sizes so it shouldn't have a "screen reader" title when the rest of the content isn't visible?

GiantRobots:
Love this components, sorry for all the questions they don't all need resolved but just wanted to get them out in the air.

packages/paste-website/src/pages/utilities/screen-reader-only/index.mdx
SiTaggart: Does that work for headings though?

packages/paste-website/src/pages/utilities/screen-reader-only/index.mdx
```diff
@@ -0,0 +1,212 @@
...
+    </Anchor>
+  </Flex>
+</Card>`}
+</LivePreview>
+
+### Repetitive buttons and hidden headers
+
+Each column in a table is required to have a column header. Sometimes those column headers can be visually hidden from sighted users if desired. To do so use the Screen Reader Only component to hide the cell content for that column header.
+
+Another common use case is for providing context to repetitive actions. Visually the "Buy" button is associated with the phone number via the row highlighting, but to a screen reader user the "Buy" action might be unclear. To provide additional clarification we can add _what_ you would be buying as visually hidden text to the button.
```
aayushpi: I would write this as:

Another common use case is for providing correct context to repetitive actions. In the example below, the "Buy" button is clearly visually associated with the phone number via row highlighting. However, to a screen reader user, this association may not be as clear. To provide additional clarification we can add _what_ you would be buying (Buy <associated number>) as visually hidden text to the button. This way, we can keep the same level of context for all users.

packages/paste-website/src/pages/utilities/screen-reader-only/index.mdx
aayushpi: ^ this is great, we should potentially add this

packages/paste-website/src/pages/utilities/screen-reader-only/index.mdx
aayushpi: I like this, but we should add an example.

packages/paste-website/src/pages/utilities/screen-reader-only/index.mdx
SiTaggart: This is a direct copy of the flex card in console, so I'll probably leave it as it for now

packages/paste-website/src/pages/utilities/screen-reader-only/index.mdx
SiTaggart: I'm not sure I follow. What kind of recommendations are you thinking of?

packages/paste-website/src/pages/utilities/screen-reader-only/index.mdx
SiTaggart: _Ideally_ we should show table headers, but we do it a bunch in console, and as long as you do it right, it's not super bad...

This is a real example from console too

packages/paste-website/src/pages/utilities/screen-reader-only/index.mdx
SiTaggart: Yeah this one is a bit hard. I was scouring sendgrid and console for examples I could rebuild with the stuff we have in Paste at the moment and this is the best I could do. It's like a chicken and the egg situation

packages/paste-website/src/pages/utilities/screen-reader-only/index.mdx
```diff
@@ -0,0 +1,212 @@
+---
+title: Screen Reader Only - Utilities
+description: The Screen Reader Only component should be used to visually hide content that may provide extra context to assistive technology users.
```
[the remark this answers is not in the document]
SiTaggart: I don't think you need the second part, but the first part is good

packages/paste-website/src/pages/utilities/screen-reader-only/index.mdx
SiTaggart: My second take:

>This can be useful when information or relationships between UI controls are only communicated _visually_. By being able to visually hide some content, you are able to provide clarifying information to non-sighted users.

packages/paste-website/src/pages/utilities/screen-reader-only/index.mdx
SiTaggart: This might be difficult to do an example of. Are you suggesting it shouldn't be used for stuff like old school SEO keyword stuffing, @richbachman

richbachman: Yeah. Thats what I was thinking. Might be hard to show solid examples, so may not be necessary.

SiTaggart:
> Do we have any guidance on using media queries and screen readers? ie. a Table column disappears on smaller screen sizes so it shouldn't have a "screen reader" title when the rest of the content isn't visible?

@GiantRobots Not something I would use this component for, but it's an option. Ideally, responsive treatment would come at the component level and takes a little more nuance. For example when you transition content to a more condensed view on a small screen, the new rendering of visible content _maybe_ acceptable for screen readers, and hiding the rest fully with `hidden` or `display:none` would be totally fine. Kind of a case by case and potentially a more formal guide on handling responsive content in an accessible manor.

> Do we have tests that validate this component works? Is that something we can do easily or is it a lot of work? (Not really documentation related)

@GiantRobots It's basically impossible to validate it works with A.T. right now but we do have visual regression tests making sure the treatment continues to works as intended. Once the Accessibility Object Model becomes a real thing then we'd be able to programmatically test what was being exported as the accessible name via the accessibility tree. Then we'd be able to validate our intension, until then we're a bit stuck with manual testing.

packages/paste-website/src/pages/utilities/screen-reader-only/index.mdx
GiantRobots: Mostly when you should use aria-described by or the screen reader component, I can think of a few use cases for either but stating the do it like this would be really helpful (I personally am not very versed in the ways of accessibility)

SiTaggart: OK, I think I can think of some stuff, but I think we'll wait until we can create some more complex examples with Paste components to document it.

Issue #57, opened 2019-08-16
Button as Anchor


elainelylee:
The documentation states that the recommendation is "Try not to use anchors that look like buttons."  And, "Only one primary button should be used on each screen so users are clear about what the intended action is."

There will be use cases where the primary action navigates to another page. Traditionally, this is the call to action but the action is taken place on another page for example. Foreseeing that our console will have pages like these that seem like marketing pages or gates, what's the recommendation here regarding using a button or an anchor?

serifluous:
Thanks, Elaine! I'll be working on some guidelines over the next two weeks to address this. Definitely let us know if you notice anything else missing or confusing in them (even the smallest things).

serifluous:
Needs more refinement based on feedback during office hours, especially what a good alternative is.

PR #293, opened 2020-02-06
feat(website): add more detail on why not to use link-style buttons

serifluous:
- Closing out DSYS-1202 based on issue #57 
- Changed from this: "Try not to use anchors that look like buttons. Voice dictation users will struggle with these as they may say “click button” but the software can’t tell what the anchor looks like visually."

to this:

"Use anchors that look like buttons sparingly. Voice dictation users may encounter issues when trying to activate them. A user may say, "Click the read more button", but the dictation software won't respond since it can’t tell what the anchor looks like visually. Use an alternative where possible, such as an anchor with an icon or with a larger font size."

https://paste-git-feat-button-accessibility-docs-revision.twilio-dsys.now.sh/components/button#accessibility


richbachman:
Should we add anything about why using an icon or with a larger font size is better?

PR #295, opened 2020-02-07
chore: update some package statuses

TheSisb:
Button and Card are being bumped out of beta status because they're being used in prod.

### DIFF
--- a/packages/paste-core/components/anchor/package.json
+++ b/packages/paste-core/components/anchor/package.json
@@ -2,7 +2,7 @@
   "name": "@twilio-paste/anchor",
   "version": "1.0.4",
   "category": "interaction",
-  "status": "beta",
+  "status": "production",
   "description": "Twilio's Anchor component",
   "author": "Twilio Inc.",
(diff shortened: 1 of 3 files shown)


packages/paste-core/components/card/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@twilio-paste/card",
   "version": "1.3.3",
   "category": "data display",
-  "status": "beta",
+  "status": "card",
```
aayushpi: prod

aayushpi: Can you prod the anchor too?

packages/paste-core/components/anchor/package.json
```diff
@@ -2,7 +2,7 @@
   "name": "@twilio-paste/anchor",
   "version": "1.0.4",
   "category": "interaction",
-  "status": "beta",
+  "status": "prod",
```
SiTaggart: prod -> production

PR #298, opened 2020-02-07
fix(grid): add getColumnGutters function to process token value

richbachman:
Currently the `Column`'s aren't setting the padding value correctly. It's coming through as the token value, i.e. `space20`. I've added a function to process the token value into the correct padding values.

- [x] added `getColumnGutters` function
- [x] added test for function

Tested locally via the website.

### DIFF
--- a/packages/paste-core/utilities/grid/__tests__/grid.test.tsx
+++ b/packages/paste-core/utilities/grid/__tests__/grid.test.tsx
@@ -4,7 +4,7 @@ import {DefaultTheme} from '@twilio-paste/theme-tokens';
 import {Theme} from '@twilio-paste/theme';
 import {Space} from '@twilio-paste/style-props';
 import {Grid, Column} from '../src';
-import {getOuterGutterPull, getStackedColumns, getColumnOffset, getColumnSpan} from '../src/helpers';
+import {getOuterGutterPull, getStackedColumns, getColumnGutters, getColumnOffset, getColumnSpan} from '../src/helpers';
 
 describe('Grid Unit Tests', () => {
   const mockTheme = {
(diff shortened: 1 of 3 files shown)


packages/paste-core/utilities/grid/__tests__/grid.test.tsx
```diff
@@ -22,10 +22,18 @@ describe('Grid Unit Tests', () => {
...
+  it('it should return a responsive set of negative margin values', (): void => {
     expect(getOuterGutterPull(mockTheme, mockGutters)).toStrictEqual(['-2px', '-6px', '-4px']);
   });
 
+  it('it should return a single padding', (): void => {
+    expect(getColumnGutters(mockTheme, mockGutter)).toStrictEqual('4px');
+  });
+
+  it('it should return a responsive set of padding values', (): void => {
+    expect(getColumnGutters(mockTheme, mockGutters)).toStrictEqual(['2px', '6px', '4px']);
```
SiTaggart: Nice

packages/paste-core/utilities/grid/src/helpers.tsx
```diff
@@ -26,6 +26,21 @@ export const getOuterGutterPull = (theme: ThemeShape, gutter?: Space): Margin =>
...
+export const getColumnGutters = (theme: ThemeShape, gutter?: Space): Padding => {
+  if (Array.isArray(gutter)) {
+    return (gutter as SpaceOptions[]).map((value: SpaceOptions) => {
+      return `${theme.space[value]}` as SpaceOptions;
+    });
+  }
+
+  if (gutter) {
+    return `${theme.space[gutter as SpaceOptions]}` as SpaceOptions;
+  }
```
SiTaggart: This is identical to the code in `getOuterGutterPull`. It might be worth pulling it out of both into a shared function. Something like `getResponsiveValueFromTheme` or something.

The only difference is the returned default values of `auto` and `space0`.

packages/paste-core/utilities/grid/src/helpers.tsx
```diff
@@ -26,6 +26,21 @@ export const getOuterGutterPull = (theme: ThemeShape, gutter?: Space): Margin =>
...
+
+  return 'space0';
```
SiTaggart: How does this work? When there is a gutter you return the _value_ from the theme object. But here for the default, you're returning a token name.

packages/paste-core/utilities/grid/src/helpers.tsx
richbachman: This one also returns the positive value of the token instead of the negative value like `getOuterGutterPull`. Still worth pulling into a shared function?

SiTaggart: See updated edit

packages/paste-core/utilities/grid/src/helpers.tsx
richbachman: Oh right. That's wrong. Changed in my latest push. Returning the first value of space in the theme object now.

PR #299, opened 2020-02-10
feat(website): create website blockquote

katieporter:
Adds a Blockquote component for use on the Paste website. Closes out [DSYS-1621](https://issues.corp.twilio.com/browse/DSYS-1621). Also migrates research insights from Callouts to Blockquotes.

**Contributing to Twilio**


katieporter:
@SiTaggart I adjusted this a bit to add an optional URL to the citation in the Blockquote - not sure if how I structured it is the best approach, so I'd love your thoughts!

packages/paste-website/src/components/Blockquote.tsx
```diff
@@ -0,0 +1,42 @@
...
+  sourceUrl: string;
+}
+
+const Citation: React.FC<Props> = ({name, source, sourceUrl}) => {
+  return (
+    <Text as="footer" fontWeight="fontWeightSemibold" marginTop="space40">
+      &mdash; {name}
+      {source && ', '}
+      {source && (
+        <cite>{sourceUrl ? <Anchor href={sourceUrl}>{source}</Anchor> : <Text as="span">{source}</Text>}</cite>
```
TheSisb: Currently the typing of the `source` and `sourceUrl` props means that you cannot pass them as `null` or `undefined`.  Typescript will make sure they come through.  Therefore, these checks against `{source && ...}` will always be true.  In order to correct this, we first need to understand what we expect.  Do we expect the consumers of this component to use this a `source` and `sourceUrl`? Will this component be "correct" if they do? 

If we must always pass them, then simply remove the conditional check `source &&`.
If we don't need to always pass them, we can replace the typing as:
```
source?: string;
sourceUrl?: string;
```
and keep the checks.

packages/paste-website/src/components/Blockquote.tsx
```diff
@@ -0,0 +1,42 @@
...
+      paddingLeft="space60"
+      paddingTop="space30"
+      paddingBottom="space30"
+      marginTop="space60"
+      marginBottom="space60"
+    >
+      <Text as="p" fontSize="fontSize40" lineHeight="lineHeight50">
+        {children}
+      </Text>
+      {name && <Citation name={name} source={source} sourceUrl={sourceUrl} />}
```
TheSisb: Same comment for `name`.  Typescript always expects it to be passed.

PR #300, opened 2020-02-10
docs(website): update list spec page

richbachman:
- [x] Update list doc page with `marginTop` and `marginBottom` props
- [x] Update nested list example


richbachman:
@SiTaggart the do and dont examples are updated.

PR #303, opened 2020-02-11
fix(grid): remove column flex styles

richbachman:
@SiTaggart discovered we were using `flex` improperly in `Column`. `Flex` was interfering with children `block` elements an not allowing them to be full width. The `flex` styles are unnecessary because we're calculating a width for each `Column`.

This removes the `Column` `flex` styles, which allows children `block` elements to be 100% width.

### DIFF
--- a/packages/paste-core/utilities/grid/__tests__/__snapshots__/grid.test.tsx.snap
+++ b/packages/paste-core/utilities/grid/__tests__/__snapshots__/grid.test.tsx.snap
@@ -36,38 +36,10 @@ exports[`12 Column Options it should render a 5 span, 3 span column grid 1`] = `
 
 .emotion-2 {
   width: 25%;
-  -webkit-box-flex: 1;
-  -webkit-flex-grow: 1;
-  -ms-flex-positive: 1;
-  flex-grow: 1;
-  -webkit-flex-shrink: 1;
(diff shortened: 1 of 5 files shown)


SiTaggart:
Do you mind removing all the width 100% on the box components in your stories to prove the theory?

TheSisb:
Curious why this wasn't caught in any tests.  Can we add a test for it or a storybook story of the behavior?

richbachman:
@SiTaggart stories updated and theory proven. When I added the display flex rule, it set the default value of `flex-wrap` to `nowrap`, which overrides the width of `block` elements unless those elements have a `width` for `flex-basis` set.

richbachman:
@TheSisb it was an error on my part. In the tests, we're not adding any child elements to each `Column` so the error wouldn't really have shown up. Stories are now updated so we can visually see the block child elements are now full width of each `Column`.

SiTaggart:
Worth adding a couple of cards in grid stories? That would prove our compositions don’t affect our own components in adverse ways, rather than faking examples with box all the time?

Setting 100% width and not type checking mdx and storybook files is biting us a little

richbachman:
@SiTaggart I've added a couple of stories showing actual components.

PR #306, opened 2020-02-12
feat(media-object): add support for the as prop and margin top and bottom

SiTaggart:
In creating these alignment examples https://codesandbox.io/s/vertical-alignment-with-text-using-paste-t1bpn it dawned on me that you need the ability to change the element type to be able to create valid semantic markup.

For example you can't nest a `div` inside a heading element. By pass as down to box, you can now change the markup as you desire to support valid HTML. I also set the default to a span as it's the one element that will work in the most places.

Also adding support for setting margin top and bottom to the outer element

### DIFF
--- a/packages/paste-core/utilities/media-object/__tests__/index.spec.tsx
+++ b/packages/paste-core/utilities/media-object/__tests__/index.spec.tsx
@@ -1,0 +1,121 @@
+import * as React from 'react';
+import renderer from 'react-test-renderer';
+import {ReactWrapper, mount} from 'enzyme';
+import {Theme} from '@twilio-paste/theme';
+import {MediaObject, MediaFigure, MediaBody} from '../src';
+
+describe('MediaObject', () => {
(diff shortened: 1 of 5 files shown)


packages/paste-core/utilities/media-object/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,133 @@
...
+
+  it('should render as another HTML element', (): void => {
+    const tree = renderer
+      .create(
+        <Theme.Provider theme="console">
+          <MediaObject as="section">background single</MediaObject>
+        </Theme.Provider>
+      )
+      .toJSON();
+    expect(tree).toMatchSnapshot();
```
TheSisb: I kind of think an enzyme test would be better than a snapshot test for a bunch of these: https://airbnb.io/enzyme/docs/api/ShallowWrapper/name.html#examples

PR #307, opened 2020-02-12
docs(grid): add grid spec to docs site

richbachman:
- [x] Grid spec doc page on website

### DIFF
--- a/packages/paste-website/src/pages/utilities/grid/index.mdx
+++ b/packages/paste-website/src/pages/utilities/grid/index.mdx
@@ -1,0 +1,520 @@
+---
+title: Grid - Utilities
+description: The grid component controls the visual layout and spacing between elements on a page or in containers.
+---
+
+import {graphql} from 'gatsby';
+import {Box} from '@twilio-paste/box';
+import {Grid, Column} from '@twilio-paste/grid';
+import {Card} from '@twilio-paste/card';
+import {Heading} from '@twilio-paste/heading';
+import {Paragraph} from '@twilio-paste/paragraph';
+import {Text} from '@twilio-paste/text';
+import {SidebarCategoryRoutes} from '../../../constants';
+import {Callout, CalloutTitle, CalloutText} from '../../../components/callout';
+import {DoDont, Do, Dont} from '../../../components/DoDont';
+import {SiteLink} from '../../../components/SiteLink';
+import Changelog from '@twilio-paste/grid/CHANGELOG.md';
+import {TwilioIcon} from '../../../components/icons/TwilioIcon';
+import {TWILIO_RED} from '../../../constants';
+
+export const pageQuery = graphql`
+  {
+    allPasteUtility(filter: {name: {eq: "@twilio-paste/grid"}}) {
+      edges {
+        node {
+          name
+          description
+          status
+          version
+        }
+      }
+    }
+    mdx(fields: {slug: {eq: "/grid/"}}) {
+      headings {
+        depth
+        value
+      }
+    }
+  }
+`;
+
+<ComponentHeader
+  name="Grid"
+  categoryRoute={SidebarCategoryRoutes.UTILITIES}
+  githubUrl="https://github.com/twilio-labs/paste/tree/master/packages/paste-core/utilities/grid"
+  storybookUrl="https://twilio-labs.github.io/paste/?path=/story/utilities-grid--grid-12-column-and-gutter-support"
+  abstractUrl="https://share.goabstract.com/3e7528f5-7981-421c-9308-e06d2f341c70?"
+  data={props.data.allPasteUtility.edges}
+/>
+
+---
+
+<contentwrapper>
+
+<TableOfContents headings={props.data.mdx.headings} />
+
+<content>
+
+## Guidelines
+
+<Paragraph>{props.pageContext.frontmatter.description}</Paragraph>
+
+### About Grid
+
+Grid creates organized and consistent product layouts, by providing a standard for positioning elements. Grid speeds up the designer-to-developer workflow by improving the consistency of design implementations.
+
+Use a Grid to layout and align content by applying standard spacing between elements on a page or within a component.
+
+The standard desktop grid has 24-pixel gutters and can support up to 12 columns. The outside gutters are removed from the overall width of the grid.
+
+Grid is built using the [Flex component](/utilities/flex) and can be made responsive.
+
+## Examples
+
+#### Equal column widths
+
+Use Column to lay out equal-sized content, such as a feed of cards.
+
+<LivePreview scope={{Grid, Column, Card, Heading, Paragraph}} language="jsx">
+  {`<Grid gutter="space30">
+  <Column>
+    <Card padding="space70">
+      <Heading as="h4" variant="heading40">Card Heading</Heading>
+      <Paragraph>Chillwave fixie sustainable helvetica etsy.</Paragraph>
+    </Card>
+  </Column>
+  <Column>
+    <Card padding="space70">
+      <Heading as="h4" variant="heading40">Card Heading</Heading>
+      <Paragraph>Chillwave fixie sustainable helvetica etsy.</Paragraph>
+    </Card>
+  </Column>
+  <Column>
+    <Card padding="space70">
+      <Heading as="h4" variant="heading40">Card Heading</Heading>
+      <Paragraph>Chillwave fixie sustainable helvetica etsy.</Paragraph>
+    </Card>
+  </Column>
+</Grid>`}
+</LivePreview>
+
+#### Unequal column widths, using column span
+
+Use `span` to allow content to span multiple columns, such as a layout with a sidebar that takes up 1/4 of the screen and a content area that takes up 3/4 of the screen.
+
+<LivePreview scope={{Grid, Column, Card, Heading, Paragraph}} language="jsx">
+  {`<Grid gutter="space30">
+  <Column span={8}>
+    <Heading as="h4" variant="heading40">Content Heading</Heading>
+    <Paragraph>
+      Kale chips distillery authentic, portland etsy cloud bread vinyl gentrify drinking vinegar viral meh hot
+      chicken bitters fashion axe palo santo. Chillwave fixie sustainable <i>helvetica</i> etsy.
+    </Paragraph>
+  </Column>
+  <Column span={4}>
+    <Card padding="space70">
+      <Heading as="h4" variant="heading40">Card Heading</Heading>
+      <Paragraph>Chillwave fixie sustainable helvetica etsy.</Paragraph>
+    </Card>
+  </Column>
+</Grid>`}
+</LivePreview>
+
+<Callout variant="warning">
+  <CalloutTitle as="h4" variant="heading40">An Important Note About Span and Offset</CalloutTitle>
+  <CalloutText>When using the <strong>span</strong> or <strong>offset</strong> props, base your layout off a 12-column grid. For example, if you want to have two columns with one spanning 8 columns, one column would span 8 and the other would need to span 4.</CalloutText>
+</Callout>
+
+#### Offset column
+
+Use `offset` to push a column to the right by a certain number of columns. For example, a layout with centered columns.
+
+<LivePreview scope={{Grid, Column, Card, Heading, Paragraph}} language="jsx">
+  {`<Grid gutter="space30">
+  <Column span={8} offset={2}>
+    <Card padding="space70">
+      <Heading as="h4" variant="heading40">Card Heading</Heading>
+      <Paragraph>Chillwave fixie sustainable helvetica etsy.</Paragraph>
+    </Card>
+  </Column>
+</Grid>`}
+</LivePreview>
+
+#### Stacked columns in a responsive view
+
+Use `vertical` to stack columns vertically instead of horizontally. Use this when your layout needs to respond to a smaller view.
+
+<LivePreview scope={{Grid, Column, Box, Card, Heading, Paragraph}} language="jsx">
+  {`<Grid gutter="space30" vertical>
+  <Column>
+    <Box marginTop="space30">
+      <Card padding="space70">
+        <Heading as="h4" variant="heading40">Card Heading</Heading>
+        <Paragraph>Chillwave fixie sustainable helvetica etsy.</Paragraph>
+      </Card>
+    </Box>
+  </Column>
+  <Column>
+    <Box marginTop="space30">
+      <Card padding="space70">
+        <Heading as="h4" variant="heading40">Card Heading</Heading>
+        <Paragraph>Chillwave fixie sustainable helvetica etsy.</Paragraph>
+      </Card>
+    </Box>
+  </Column>
+  <Column>
+    <Box marginTop="space30">
+      <Card padding="space70">
+        <Heading as="h4" variant="heading40">Card Heading</Heading>
+        <Paragraph>Chillwave fixie sustainable helvetica etsy.</Paragraph>
+      </Card>
+    </Box>
+  </Column>
+</Grid>`}
+</LivePreview>
+
+#### Nested Grids
+
+A Grid can be nested inside another Grid. For example, you can split a column into multiple smaller columns in order to stack the outer Grid in a responsive view but not the inner Grid.
+
+<LivePreview scope={{Grid, Column, Box, Card, Heading, Paragraph}} language="jsx">
+  {`<Grid vertical>
+  <Column>
+    <Grid gutter="space30">
+      <Column>
+        <Card padding="space70">
+          <Heading as="h4" variant="heading40" variant="heading40">Card Heading</Heading>
+          <Paragraph>Chillwave fixie sustainable helvetica etsy.</Paragraph>
+        </Card>
+      </Column>
+      <Column>
+        <Card padding="space70">
+          <Heading as="h4" variant="heading40" variant="heading40">Card Heading</Heading>
+          <Paragraph>Chillwave fixie sustainable helvetica etsy.</Paragraph>
+        </Card>
+      </Column>
+    </Grid>
+  </Column>
+  <Column>
+    <Box marginTop="space60" marginBottom="space60">
+      <Card padding="space70">
+        <Heading as="h4" variant="heading40" variant="heading40">Card Heading</Heading>
+        <Paragraph>Chillwave fixie sustainable helvetica etsy.</Paragraph>
+      </Card>
+    </Box>
+  </Column>
+  <Column>
+    <Grid gutter="space30">
+      <Column>
+        <Card padding="space70">
+          <Heading as="h4" variant="heading40" variant="heading40">Card Heading</Heading>
+          <Paragraph>Chillwave fixie sustainable helvetica etsy.</Paragraph>
+        </Card>
+      </Column>
+      <Column>
+        <Card padding="space70">
+          <Heading as="h4" variant="heading40" variant="heading40">Card Heading</Heading>
+          <Paragraph>Chillwave fixie sustainable helvetica etsy.</Paragraph>
+        </Card>
+      </Column>
+    </Grid>
+  </Column>
+</Grid>`}
+</LivePreview>
+
+### Composition Notes
+
+Grid can be used to compose various types of layouts ranging from whole application pages and views down to individual component layouts.
+
+<Callout variant="secondary">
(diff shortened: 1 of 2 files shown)


packages/paste-website/src/pages/utilities/grid/index.mdx
```diff
@@ -0,0 +1,520 @@
...
+      <Heading as="h4" variant="heading40">Card Heading</Heading>
+      <Paragraph>Chillwave fixie sustainable helvetica etsy.</Paragraph>
+    </Card>
+  </Column>
+</Grid>`}
+</LivePreview>
+
+#### Unequal column widths, using column span
+
+Use `span` to allow content to span multiple columns, such as a layout with a sidebar that takes up 1/4 of the screen and a content area that takes up 3/4 of the screen.
```
SiTaggart: I think you need to explain to me here the span values. It might be confusing that you talk about 1/4 and 3/4 but the values are 8 and 4

packages/paste-website/src/pages/utilities/grid/index.mdx
```diff
@@ -0,0 +1,520 @@
...
+      <Heading as="h4" variant="heading40">Card Heading</Heading>
+      <Paragraph>Chillwave fixie sustainable helvetica etsy.</Paragraph>
+    </Card>
+  </Column>
+</Grid>`}
+</LivePreview>
+
+#### Stacked columns in a responsive view
+
+Use `vertical` to stack columns vertically instead of horizontally. Use this when your layout needs to respond to a smaller view.
```
SiTaggart: Could this example be slightly misleading? You call it a responsive view, but the layout doesn't actually respond to the browser. You're not using the responsive values in your props.

There's also no example showing me how to do responsive values

SiTaggart: hmmm I maybe miss understanding how responsive grid work, but I kind of expect

```
<Grid>
  <Column span={[12 , 6, 4]} />
  <Column span={[12 , 6, 4]} />
  <Column span={[12 , 12, 4]} />
</Grid>
```
to out put a 3 stack on mobile, 2 over 1 on the next breakpoint and 3 in a row after. But the 3 stacked doesn't work

packages/paste-website/src/pages/utilities/grid/index.mdx
TheSisb: I'm not sold on this, but an idea nonetheless: would it be helpful to use `12` as the denominator? i.e.:  1/4 = 3/12 and 3/4 = 9/12.
This way it's clear to people how we think of spanning over 12 columns, and if they want to simplify it into 1/4 and 3/4 they can in their mind, but that work isn't forced.

packages/paste-website/src/pages/utilities/grid/index.mdx
```diff
@@ -0,0 +1,520 @@
...
+    <Card padding="space70">
+      <Heading as="h4" variant="heading40">Card Heading</Heading>
+      <Paragraph>Chillwave fixie sustainable helvetica etsy.</Paragraph>
+    </Card>
+  </Column>
+</Grid>`}
+</LivePreview>
+
+<Callout variant="warning">
+  <CalloutTitle as="h4" variant="heading40">An Important Note About Span and Offset</CalloutTitle>
```
TheSisb: `An important note about the span and offset props`

TheSisb: then the next line can follow as 
> When using span or offset, base your layout...

packages/paste-website/src/pages/utilities/grid/index.mdx
```diff
@@ -0,0 +1,520 @@
...
+  <CalloutText>When using the <strong>span</strong> or <strong>offset</strong> props, base your layout off a 12-column grid. For example, if you want to have two columns with one spanning 8 columns, one column would span 8 and the other would need to span 4.</CalloutText>
```
TheSisb: `For example, if you want to have two columns with one spanning 8 columns, one column would span 8 and the other would need to span 4.` -> `For example, if you want to have two columns with one spanning twice as much as the other, then one column would span=8 and the other would span=4.`

packages/paste-website/src/pages/utilities/grid/index.mdx
```diff
@@ -0,0 +1,520 @@
...
+      <Heading as="h4" variant="heading40">Card Heading</Heading>
+      <Paragraph>Chillwave fixie sustainable helvetica etsy.</Paragraph>
+    </Column>
+  </Grid>
+</Card>`}
+</LivePreview>
+
+### When to use Grid
+
+Use Grid whenever you need to layout or align content on a page or within a component. Grid should also be used for responsive page layouts.
```
TheSisb: I don't know where or how, but I kind of feel like this might be a good place to say something like: "Since the grid is sized to the full width of its parent container, you may not want to use it for fixed width columns in a variable width container.

packages/paste-website/src/pages/utilities/grid/index.mdx
```diff
@@ -0,0 +1,520 @@
...
+
+### About Grid
+
+Grid creates organized and consistent product layouts, by providing a standard for positioning elements. Grid speeds up the designer-to-developer workflow by improving the consistency of design implementations.
+
+Use a Grid to layout and align content by applying standard spacing between elements on a page or within a component.
+
+The standard desktop grid has 24-pixel gutters and can support up to 12 columns. The outside gutters are removed from the overall width of the grid.
+
+Grid is built using the [Flex component](/utilities/flex) and can be made responsive.
```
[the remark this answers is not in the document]
richbachman: It can be made responsive using the vertical prop. Its not responsive out of the box since current products aren't.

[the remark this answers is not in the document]
TheSisb: But it is responsive in that it shrinks to the width of the screen.  Theoretically, a 12 column grid could still be used on mobile devices, couldn't it?

packages/paste-website/src/pages/utilities/grid/index.mdx
```diff
@@ -0,0 +1,520 @@
...
+  <Column>
+    <Card padding="space70">
+      <Heading as="h4" variant="heading40">Card Heading</Heading>
+      <Paragraph>Chillwave fixie sustainable helvetica etsy.</Paragraph>
+    </Card>
+  </Column>
+  <Column>
+    <Card padding="space70">
+      <Heading as="h4" variant="heading40">Card Heading</Heading>
+      <Paragraph>Chillwave fixie sustainable helvetica etsy.</Paragraph>
```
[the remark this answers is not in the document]
richbachman: I think its from https://hipsum.co, Jason showed that to me.

packages/paste-website/src/pages/utilities/grid/index.mdx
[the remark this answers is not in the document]
richbachman: Right. I updated the wording there.

packages/paste-website/src/pages/utilities/grid/index.mdx
richbachman: I rewrote this section. Let me know if it works better.

packages/paste-website/src/pages/utilities/grid/index.mdx
richbachman: Because our site isn't responsive, its hard to show actual responsive examples. However I did rewrite the vertical section so it explains the vertical prop more. I also added a responsive column example that links to the storybook story: https://github.com/twilio-labs/paste/pull/307/files#diff-031c2d993f08f8fad55f849848f24b5cR231

As for your example, I didn't account for the 2 over 1 on a breakpoint use case. That is something we should add. I can add a ticket to our backlog.

In order to stack on mobile, you need the vertical prop set on `Grid`. Something like `vertical={[true, false, false]}` in your example. The true on vertical sets `flex-direction: column` so all columns will to go full with and stack.

packages/paste-website/src/pages/utilities/grid/index.mdx
```diff
@@ -0,0 +1,533 @@
...
+
+Use a Grid to layout and align content by applying standard spacing between elements on a page or within a component.
+
+The standard desktop grid has 24-pixel gutters and can support up to 12 columns. The outside gutters are removed from the overall width of the grid.
+
+Grid is built using the [Flex component](/utilities/flex) and is responsive. Grid column widths grow and shrink based on the parent element width.
+
+## Examples
+
+#### Equal column widths
```
SiTaggart: You’re skipping a heading level for these examples. Make sure we adhere to a good document hierarchy

packages/paste-website/src/pages/utilities/grid/index.mdx
```diff
@@ -0,0 +1,533 @@
...
+import {Grid, Column} from '@twilio-paste/grid';
+
+<Grid>
+  <Column>
+    Foo
+  </Column>
+</Grid>
+```
+
+#### Props
```
SiTaggart: Is it worth splitting this props table into 2, I’m unsure which props are for which component at this stage.

You could have 

Grid props

Table 

Column props

Table

packages/paste-website/src/pages/utilities/grid/index.mdx
```diff
@@ -0,0 +1,533 @@
...
+
+#### Equal column widths
+
+Use Column to lay out equal-sized content, such as a feed of cards.
+
+<LivePreview scope={{Grid, Column, Card, Heading, Paragraph}} language="jsx">
+  {`<Grid gutter="space30">
+  <Column>
+    <Card padding="space70">
+      <Heading as="h4" variant="heading40">Card Heading</Heading>
```
SiTaggart: If you can replace some of these examples with real ones from console or sendgrid that would really set these docs off. I know it’s consistent feedback we get that Loren ipsum examples aren’t helpful.

packages/paste-website/src/pages/utilities/grid/index.mdx
```diff
@@ -0,0 +1,533 @@
...
+  </Column>
+</Grid>
+```
+
+#### Props
+
+| Prop      | Type                                         | Description                                                                                                                         | Default |
+| ----------| ---------------------------------------------| ----------------------------------------------------------------------------------------------------------------------------------- | ------- |
+| gutter?   | `ResponsiveValue<keyof ThemeShape['space']>` | Sets the spacing on both sides of a Column. Use a spacing token that represents half of the total space you need between 2 Columns. |         |
+| vertical? | `boolean, ResponsiveValue<boolean>`          | Stacks the Columns                                                                                                                  |         |
```
TheSisb: I never really was a fan of this prop name, reading it now I kind of wonder if it should be named `stack` or `stackColumns`.  Scope creep, I know :P

packages/paste-website/src/pages/utilities/grid/index.mdx
```diff
@@ -0,0 +1,533 @@
...
+| span?     | `number, ResponsiveValue<number>`            | Sets the width of Columns (how many columns a column can span)                                                                      |         |
```
TheSisb: I think we need two different props tables for this component.  One for `Grid` and another for `Column`

packages/paste-website/src/pages/utilities/grid/index.mdx
richbachman: @SiTaggart I've added some content from both console and sendgrid. Let me know if I should push those further.

serifluous:
In the top description, can you change "A flex based horizontal..." to "A flex-based horizontal..."?

richbachman:
@serifluous that's being pulled from our package description. I'll put in another that changes that, and then this will auto-update when everything is merged.

packages/paste-website/src/pages/utilities/grid/index.mdx
```diff
@@ -0,0 +1,567 @@
...
+    <Card padding="space70">
+      <Heading as="h4" variant="heading40">Expert Insights</Heading>
+      <Paragraph>Send better email with Expert Insights. Our detailed monthly reports with expert recommendations will give you deeper visibility into the health and performance of your email program.</Paragraph>
+      <Button>Add to Plan</Button>
+    </Card>
+  </Column>
+</Grid>`}
+</LivePreview>
+
+<Callout variant="warning">
```
serifluous: This could just be the default Callout variant since it's not exactly a warning about the content, or indicating that anything's wrong

packages/paste-website/src/pages/utilities/grid/index.mdx
```diff
@@ -0,0 +1,567 @@
...
+
+Grid can be used to compose various types of layouts ranging from whole application pages and views down to individual component layouts.
+
+<Callout variant="secondary">
+  <CalloutText>For spacing content in a layout wider than 1440 pixels, we recommend using a 24-pixel gutter. Use $space-40 to get a 24-pixel gutter.</CalloutText>
+</Callout>
+
+<Callout variant="secondary">
+  <CalloutText>For spacing content in a layout narrower than 1440 pixels, we recommend using a 16-pixel gutter. Use $space-30 to get a 16-pixel gutter.</CalloutText>
+</Callout>
```
serifluous: These could be combined into a single Callout in the default variant, and with a heading like "How to size gutters" or "Gutter sizing"

packages/paste-website/src/pages/utilities/grid/index.mdx
```diff
@@ -0,0 +1,567 @@
...
+
+### Basic page layout
+
+Let's look at a common use case where we need to layout a page with a sidebar and a grid within the page content area.
+
+<LivePreview scope={{Grid, Column, Button, Box, Card, Heading, Paragraph, Anchor, Text}} language="jsx">
```
serifluous: v cool

packages/paste-website/src/pages/utilities/grid/index.mdx
```diff
@@ -0,0 +1,567 @@
...
+
+### Component layout
+
+Grids can also be used for layouts within other components. For example, we can set up our card content to include 2 columns.
+
+<LivePreview scope={{Grid, Column, Card, Heading, Paragraph, TwilioIcon, TWILIO_RED, Button}} language="jsx">
+  {`<Card padding="space70">
+  <Grid gutter="space30">
+    <Column span={2}>
+      <TwilioIcon color={TWILIO_RED} display="block" size={108} />
```
serifluous: This isn't showing up with the gutter for me since the icon is larger than the gutter (at max width on a MBP 13"). Can you knock the size down a bit?

Also...is TwilioIcon one of ours? Should this be using a size token?

packages/paste-website/src/pages/utilities/grid/index.mdx
```diff
@@ -0,0 +1,567 @@
...
+</Card>`}
+</LivePreview>
+
+## When to use Grid
+
+Use Grid whenever you need to layout or align content on a page or within a component. Grid should also be used for responsive page layouts.
+Since the grid is sized to the full width of its parent container, you may not want to use it for fixed width columns in a variable width container.
+
+<DoDont>
+  <Do title="Do" body="Base the Column span and offset on a 12-columns.">
```
serifluous: "on a 12-columns" --> "on a 12-column layout"

packages/paste-website/src/pages/utilities/grid/index.mdx
```diff
@@ -0,0 +1,567 @@
...
+        <Column span={4} offset={8}>
+          <Card>
+            <Paragraph>Card content</Paragraph>
+          </Card>
+        </Column>
+      </Grid>
+    </Box>
+  </Do>
+  <Dont title="Don't" body="Don’t use Grid with only one column of content. This causes unnecessary elements. Box is what you’re looking for here.">
+    <Box padding="space60">
```
serifluous: Is it worth showing this in a Card, too? So it doesn't look like just some text?

serifluous: And is it "one column of full-width content"? Like people should still use Grid if they want to center something as in the "Offset column" example?

packages/paste-website/src/pages/utilities/grid/index.mdx
```diff
@@ -0,0 +1,567 @@
...
+#### Column Props
+
+| Prop      | Type                                         | Description                                                                                                                         | Default |
+| ----------| ---------------------------------------------| ----------------------------------------------------------------------------------------------------------------------------------- | ------- |
+| span?     | `number, ResponsiveValue<number>`            | Sets the width of Columns (how many columns a column can span)                                                                      |         |
+| offset?   | `number, ResponsiveValue<number>`            | Sets the left margin of a Column (how many columns a column is offset)                                                              |         |
+
+### Sketch
+
+Add Grid symbols from the “Paste components” project to your own artboards to help you align content.
```
serifluous: Can you add the Abstract link (https://share.goabstract.com/3e7528f5-7981-421c-9308-e06d2f341c70?) to "Grid symbols from the "Paste components" project"?

packages/paste-website/src/pages/utilities/grid/index.mdx
```diff
@@ -0,0 +1,567 @@
...
+
+Additionally, each Paste container component (*e.g.*, Card, Modal) symbol in Sketch has an invisible grid symbol named “📐 Grid” that you can replace with visible grid symbols.
```
serifluous: "replace" --> "override"

packages/paste-website/src/pages/utilities/grid/index.mdx
```diff
@@ -0,0 +1,567 @@
...
+
+We use grid symbols instead of Sketch Layout Settings for two reasons:
+
+1. Resizing a symbol better represents the behavior of a fluid grid in a responsive layout. You might be designing for multiple screen widths in a single Sketch file. Sketch Layout Settings, however, restricts you to a single setting of grid width, column width, gutter width, and number of columns for each file.
+2. Symbols allow you to place a grid within a component or container on an artboard, whereas Sketch Layout Settings can be applied to a whole artboard only.
```
serifluous: "can be applied to a whole artboard only" --> "can be applied only to a whole artboard"

packages/paste-website/src/pages/utilities/grid/index.mdx
```diff
@@ -0,0 +1,567 @@
...
+      <Heading as="h4" variant="heading40">Quality Screening</Heading>
+      <Paragraph>Every phone number in the inventory passes a rigorous 60-day screening process to verify capabilities and test for unwanted call traffic.</Paragraph>
+    </Card>
+  </Column>
+</Grid>`}
+</LivePreview>
+
+### Unequal column widths, using column span
+
+Use the `span` prop to allow content to span multiple columns. Span calculates the width of a column based off a 12 column grid. The value of span becomes the
```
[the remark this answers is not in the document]
richbachman: I reworded this, but kept some of what explains the column/12 info.

packages/paste-website/src/pages/utilities/grid/index.mdx
SiTaggart: Great example

packages/paste-website/src/pages/utilities/grid/index.mdx
```diff
@@ -0,0 +1,569 @@
...
+  <Column>
+    <Card padding="space70">
+      <Heading as="h4" variant="heading40">Quality Screening</Heading>
+      <Paragraph>Every phone number in the inventory passes a rigorous 60-day screening process to verify capabilities and test for unwanted call traffic.</Paragraph>
+    </Card>
+  </Column>
+</Grid>`}
+</LivePreview>
+
+### Unequal column widths, using column `span`
```
serifluous: Backticks can be removed here since it messes with the anchor on the heading.

packages/paste-website/src/pages/utilities/grid/index.mdx
```diff
@@ -0,0 +1,569 @@
...
+      <Grid>
+        <Column span={4} offset={8}>
+          <Card>
+            <Paragraph>Card content</Paragraph>
+          </Card>
+        </Column>
+      </Grid>
+    </Box>
+  </Do>
+  <Dont title="Don't" body="Don’t use Grid with only one column of full-width content. This causes unnecessary elements. Box is what you’re looking for here.">
```
serifluous: This is actually better worded as "one full-width column of content".

(being super nitpicky now 😅)

packages/paste-website/src/pages/utilities/grid/index.mdx
richbachman: Good catch.

serifluous:
> I'm really happy with how this turned out

+1 to this!

PR #308, opened 2020-02-12
chore(icons): simple convert command to handle icon additions, removals, and changes

TheSisb:
The intent of this PR is to reduce errors in the icon generation process caused by manual intervention and the original two-step process.

- [x] I added a `yarn convert` command that nukes the files in the `src` folder and regenerates them from scratch.  This is a very simple way of having one command that can handle all situations thrown at it. 
  - Added a file? Convert generates it. ✅ 
  - Removed a file? Convert will delete it and won't know to regenerate it. ✅ 
  - Changed a file?  Convert removes the old one and generates the new changes ✅ 
- [x] Updated the stale readme
- [x] Updated the doc site to have 1 less step.


packages/paste-icons/tools/actions/convertAllAction.js
```diff
@@ -0,0 +1,32 @@
...
+        });
+      });
+  });
+}
+
+async function convertAllAction() {
+  clearBuiltFiles();
+  await convertNewAction();
+  // filesystem operations take time, giving a delay to the next step so it doesn't fail
+  setTimeout(listIconsAction, 2000);
```
SiTaggart: This feels like it might be a little fragile. My async await knowledge it literally zero, don't awaits chain?

```
async function convertAllAction() {
  clearBuiltFiles();
  await convertNewAction();
  await listIconsAction();
}
```

packages/paste-website/src/pages/icon-system/how-to-add-an-icon.mdx
```diff
@@ -127,11 +127,10 @@ There are two ways for icons to be added into the `@twilio-paste/icons` code pac
 1. Fork the Paste repository because PRs can only be open against forks and not branches for security reasons.
 2. Clone your newly forked Paste repository: `git clone <url>`
 3. Navigate inside the newly cloned repository with `cd paste`
-4. For each new icon you would like to create, add an `svg` icon file into the `packages/paste-icons/svg/` folder.
-5. From the root Paste folder, run the following command in your terminal: `yarn workspace @twilio-paste/icons convert-new`
-6. Verify your new icons were added correctly by checking the `packages/twilio-icons/src` folder to see the newly  created source files.
-7. At this point, run the following command to update our documentation with the new icons: `yarn workspace @twilio-paste/icons list-icons`
-8. Commit your changes and submit a new PR on Github!
+4. Add all the new `svg` icon files into the `packages/paste-icons/svg/` folder.
```
SiTaggart: I think it would be worth mentioning the naming convention for the exported assets. I don’t recall seeing anywhere that we expect a strict SnakeCase file name.

SiTaggart: That goes for the symbol in sketch when committed to abstract too

packages/paste-icons/tools/actions/convertAllAction.js
TheSisb: It is fragile :(, that was my first attempt.  It looks like some part of svgr or svg-to-react isn't respecting await.  I'll dig into it.

packages/paste-website/src/pages/icon-system/how-to-add-an-icon.mdx
TheSisb: @serifluous Do you know where I should add this in the instructions? Seems like it will happen on the design side primarily since we just drop the icons into the package and run a script blindly.

serifluous: Good point. I'll write a blurb to add to the "Setting up an icon in Paste Icons" section in a bit.

packages/paste-icons/tools/actions/convertAllAction.js
SiTaggart: As per Slack, let's stick a pin in this for now. It's good enough

packages/paste-website/src/pages/icon-system/how-to-add-an-icon.mdx
serifluous: 1. Under the step "In the artboard names, replace "icon name" with the name of your icon.", add a callout that says:

**How to name icon artboards in Sketch**

Use camel case. For an interface icon, give the icon a name that describes its purpose rather than its content. For example, "24/edit" instead of "24/pencil".

If you're adding an icon used to brand a product, name it "product/[ProductName][SubproductName (optional)]". For example: "24/product/CodeExchangeCommunity".


2. Under the nested list of the step "Make sure the SVG code is clean:", add a list item (step 4) that says:

> If you're adding an icon used to brand a product, add the word "Product" to the beginning of the SVG file name. For example: "ProductCodeExchangeCommunity"

TheSisb: Looks okay? https://paste-git-icons-simple-convert.twilio-dsys.now.sh/icon-system/how-to-add-an-icon @serifluous

packages/paste-website/src/pages/icon-system/how-to-add-an-icon.mdx
```diff
@@ -57,6 +57,19 @@ Written for Sketch 59.1
 4. Turn the new artboards into symbols.
 5. In the artboard names, replace "icon name" with the name of your icon.
 
+
+<Callout>
+   <CalloutTitle>How to name icon artboards in Sketch</CalloutTitle>
+   <CalloutText>
+     Use camel case. For an interface icon, give the icon a name that describes its 
+     purpose rather than its content. For example, "24/linkExternal" instead of 
+     "24/arrowUpRight". If you're adding an icon used to brand a product, name it
```
serifluous: Can you add a line break before "If you're adding an icon..."

packages/paste-website/src/pages/icon-system/how-to-add-an-icon.mdx
```diff
@@ -107,6 +120,8 @@ Written for Sketch 59.1
    2. You shouldn’t see `transform`. If you do, flatten your shape.
    3. You should see `width=”24" height="24"`. If you don’t, check that the artboard you’re
       exporting is 24px × 24px and you’re exporting the artboard, not the layer.
+   4. If you're adding an icon used to brand a product, add the word "Product" to the
```
serifluous: My instructions here weren't very good lol.

Can you add this as step 4 to the parent list, i.e., "Commit your changes..." would become step 5.

packages/paste-website/src/pages/icon-system/how-to-add-an-icon.mdx
TheSisb: Like so? 
image

packages/paste-website/src/components/callout/index.tsx
```diff
@@ -15,36 +16,45 @@ const borderColorPartial = (variant?: CalloutVariants): BorderColor => {
...
 const CalloutTitle: React.FC<CalloutTitleProps> = ({as = 'h3', children}) => (
   <Heading as={as} variant="heading50">
     {children}
   </Heading>
 );
 
-const CalloutText: React.FC<CalloutTextProps> = ({as = 'p', children, marginTop, marginBottom}) => (
-  <Text as={as} marginTop={marginTop} marginBottom={marginBottom}>
+const CalloutText: React.FC = ({children}) => (
+  <Text as="p" marginTop="space30">
```
SiTaggart: You may still need the override for `as` here for the homepage (and anywhere else) where we're passing `CalloutText` a list. https://github.com/twilio-labs/paste/pull/308/files#diff-3434f70a811aebf8c3d7a2a7bf6559beR98

You can't nest a list in a paragraph in HTML.

TheSisb: Since `CalloutText` is intended as a Callout `p` wrapper, I removed it around the UnorderedList usage.

SiTaggart: Makes sense to me

PR #309, opened 2020-02-12
feat: add as support to flex and grid

SiTaggart:
In creating these alignment examples https://codesandbox.io/s/vertical-alignment-with-text-using-paste-t1bpn it dawned on me that you need the ability to change the element type to be able to create valid semantic markup.

In the example it would have been handy to change the rendered div of `Flex` to be a span so it can be nested in a heading correctly.

Decided to apply the same to `Grid` as I can see you may need to play with the rendered element to create valid HTML in your compositions

### DIFF
--- a/packages/paste-core/utilities/flex/__tests__/flex.test.tsx
+++ b/packages/paste-core/utilities/flex/__tests__/flex.test.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import renderer from 'react-test-renderer';
+import {ReactWrapper, mount} from 'enzyme';
 import {Theme} from '@twilio-paste/theme';
 import {Box} from '@twilio-paste/box';
 import {
@@ -202,6 +203,20 @@ describe('Flex Display', () => {
           <Flex>
             <Box padding="space30" backgroundColor="colorBackgroundBrand" width="100%" minHeight="size10" />
           </Flex>
+        </Theme.Provider>
+      )
+      .toJSON();
+    expect(tree).toMatchSnapshot();
+  });
+
+  it('it should render as any HTML element', (): void => {
+    const wrapper: ReactWrapper = mount(<Flex as="article">child</Flex>);
+    expect(wrapper.exists('article')).toEqual(true);
+
+    const tree = renderer
+      .create(
+        <Theme.Provider theme="console">
+          <Flex as="article">child</Flex>
         </Theme.Provider>
       )
       .toJSON();
--- a/packages/paste-core/utilities/flex/src/index.tsx
+++ b/packages/paste-core/utilities/flex/src/index.tsx
@@ -22,6 +22,7 @@ type WrapOptions = boolean;
 export type Wrap = ResponsiveValue<WrapOptions>;
 
 export interface FlexProps extends LayoutProps, MarginProps, PaddingProps {
+  as?: keyof JSX.IntrinsicElements;
   display?: Display;
   vertical?: Vertical;
   vAlignContent?: VerticalAlign;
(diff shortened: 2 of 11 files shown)


packages/paste-core/utilities/flex/src/index.tsx
```diff
@@ -237,6 +240,181 @@ const Flex: React.FC<FlexProps> = ({
 };
 
 Flex.propTypes = {
+  as: PropTypes.oneOf([
```
richbachman: Can we make these reusable somehow? So they can be used across `Flex`, `Grid`, and `Column`. Is that over-optimizing?

SiTaggart: Yes, but I'm not sure where. And I don't _think_ I want to create another package right now...

packages/paste-core/utilities/flex/__tests__/flex.test.tsx
```diff
@@ -207,6 +207,17 @@ describe('Flex Display', () => {
...
+
+  it('it should render as any HTML element', (): void => {
+    const tree = renderer
+      .create(
+        <Theme.Provider theme="console">
+          <Flex as="article">child</Flex>
+        </Theme.Provider>
+      )
+      .toJSON();
+    expect(tree).toMatchSnapshot();
```
TheSisb: Would prefer an enzyme test here to check the tag https://airbnb.io/enzyme/docs/api/ShallowWrapper/name.html#examples

packages/paste-core/utilities/flex/src/index.tsx
```diff
@@ -1,9 +1,10 @@
 import * as React from 'react';
 import * as PropTypes from 'prop-types';
 import {ResponsiveValue} from 'styled-system';
-import {Box, safelySpreadBoxProps} from '@twilio-paste/box';
+import {Box, BoxProps, safelySpreadBoxProps} from '@twilio-paste/box';
 import {FlexboxProps, LayoutProps, MarginProps, PaddingProps} from '@twilio-paste/style-props';
 
+export type AsProp = Pick<BoxProps, 'as'>;
```
TheSisb: I don't think the Flex package should be the one to export this.  It might make more sense to have this defined in the `style-props` package.

packages/paste-core/utilities/flex/src/index.tsx
TheSisb: This should just be a string imo

SiTaggart: makes it so much easier if it's a string

packages/paste-core/utilities/flex/src/index.tsx
SiTaggart: Honestly, even that seems like too much. 

I'm leaning towards just declaring as on the interface when ever you need it as `as?: keyof JSX.IntrinsicElements;`

packages/paste-core/utilities/flex/__tests__/flex.test.tsx
SiTaggart: Good call, went with both

richbachman:
lgtm!

packages/paste-core/utilities/flex/src/index.tsx
TheSisb: Works for me!

packages/paste-core/utilities/flex/src/index.tsx
```diff
@@ -237,6 +240,7 @@ const Flex: React.FC<FlexProps> = ({
 };
 
 Flex.propTypes = {
+  as: PropTypes.string as any,
```
TheSisb: Just had a thought, does `as keyof JSX.IntrinsicElements` work? 🤔

SiTaggart: Nope, because string !== keyof JSX.IntrinsicElements

PR #310, opened 2020-02-12
feat(heading): add marginBottom prop to remove margin

SiTaggart:
Based on a conversation in Slack, a use case was presented where removing the bottom margin on the `Heading` component would be useful.

This can be use when placing headings in contained areas or on top of other components with bounding boxes such as tables.

This PR introduces the `noMargin` prop, which removes margin from a rendered `Heading`

### DIFF
--- a/packages/paste-website/src/pages/components/heading/index.mdx
+++ b/packages/paste-website/src/pages/components/heading/index.mdx
@@ -6,6 +6,7 @@ description: Headings are words or phrases that are intended to introduce sectio
 
 import {graphql} from 'gatsby';
 import {Box} from '@twilio-paste/box';
+import {Card} from '@twilio-paste/card';
 import {Heading} from '@twilio-paste/heading';
 import {Paragraph} from '@twilio-paste/paragraph';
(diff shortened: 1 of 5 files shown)


packages/paste-website/src/pages/components/heading/index.mdx
```diff
@@ -199,6 +210,7 @@ import {Heading} from '@twilio-paste/heading';
 | Prop           | Type                  | Description                                                                        | Default          |
 | -------------- | --------------------- | ---------------------------------------------------------------------------------- | ---------------- |
 | as             | asTags                | 'h1', 'h2', 'h3', 'h4', 'h5', 'h6', 'div', 'label', 'span'                         | null             |
+| marginBottom   | string                | 'space0'                                                                           | null             |
```
TheSisb: I think you can put 'oneOf(['space0'])` as type here, and then leave a description about how we have default values but allow users to disable it using the `space0` token.

SiTaggart: Fair. Updated

PR #313, opened 2020-02-14
feat(modal-dialog-primitive): add package

TheSisb:
Adds the modal dialog primitive package.  

**Todo:**
- [x] Add the package
- [x] Make sure it builds
- [x] Write the story

### DIFF
--- a/packages/paste-core/primitives/modal-dialog/src/index.tsx
+++ b/packages/paste-core/primitives/modal-dialog/src/index.tsx
@@ -1,0 +1,19 @@
+/**
+ * If you’re wondering why we wrapped that package into our own, we reasoned that it
+ * would be best for our consumers’ DX, with reasons such as:
+ *
+ * - If we want to migrate the underlying nuts and bolts in the future, Twilio products
+ *   that depend on this primitive would need to replace all occurrences of
+ *   `import … from ‘@reach/dialog’` to `import … from ‘@some-new/package’`.
+ *   By wrapping it in `@twilio-paste/modal-dialog-primitive`, this refactor can be avoided.
+ *   The only change would be a version bump in the package.json file.
+ * - We can more strictly enforce semver and backwards compatibility than some of
+ *   our dependencies.
+ * - We can control when to provide an update and which versions we whitelist, to
+ *   help reduce potential bugs our consumers may face.
+ * - We can control which APIs we expose. For example, we may chose to enable or
+ *   disable usage of certain undocumented APIs.
+ */
+import {DialogContent as ModalDialogContent, DialogOverlay as ModalDialogOverlay} from '@reach/dialog';
+
+export {ModalDialogContent, ModalDialogOverlay};
--- a/packages/paste-core/primitives/modal-dialog/src/index.tsx
+++ b/packages/paste-core/primitives/modal-dialog/src/index.tsx
@@ -14,6 +14,9 @@
  * - We can control which APIs we expose. For example, we may chose to enable or
  *   disable usage of certain undocumented APIs.
  */
-import {DialogContent as ModalDialogContent, DialogOverlay as ModalDialogOverlay} from '@reach/dialog';
+import {
+  DialogContent as ModalDialogPrimitiveContent,
+  DialogOverlay as ModalDialogPrimitiveOverlay,
+} from '@reach/dialog';
 
-export {ModalDialogContent, ModalDialogOverlay};
+export {ModalDialogPrimitiveContent, ModalDialogPrimitiveOverlay};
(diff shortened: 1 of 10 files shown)


packages/paste-core/primitives/modal-dialog/src/index.tsx
```diff
@@ -0,0 +1,19 @@
...
+ * - We can more strictly enforce semver and backwards compatibility than some of
+ *   our dependencies.
+ * - We can control when to provide an update and which versions we whitelist, to
+ *   help reduce potential bugs our consumers may face.
+ * - We can control which APIs we expose. For example, we may chose to enable or
+ *   disable usage of certain undocumented APIs.
+ */
+import {DialogContent as ModalDialogContent, DialogOverlay as ModalDialogOverlay} from '@reach/dialog';
+
+export {ModalDialogContent, ModalDialogOverlay};
```
TheSisb: Not exporting `Dialog` from `reach` because for it to work, we'd need to import the CSS and I don't want to do that.  Just import the two parts and style them.  This is in spirit of a primitive.

packages/paste-core/primitives/modal-dialog/stories/index.stories.tsx
```diff
@@ -0,0 +1,63 @@
...
+  background: '#f4f5f6',
+  borderRadius: '5px',
+});
+
+interface BasicModalDialogProps {
+  isOpen: boolean;
+  handleClose: () => void;
+}
+
+const BasicModalDialog: React.FC<BasicModalDialogProps> = ({isOpen, handleClose}) => (
```
TheSisb: I split this up into a separate component to showcase how the code should be split up.  Lots of examples keep this JSX code inside the other component, but for real applications I think decoupling makes the most sense.

packages/paste-core/primitives/modal-dialog/src/index.tsx
SiTaggart: Nice, I actually prefer the names too

packages/paste-core/primitives/modal-dialog/stories/index.stories.tsx
```diff
@@ -0,0 +1,63 @@
...
+      </Button>
+      <BasicModalDialog isOpen={isOpen} handleClose={handleClose} />
+    </div>
+  );
+};
+
+storiesOf('Primitives|ModalDialog', module)
+  .addDecorator(withKnobs)
+  .add('Custom overlay and content', () => {
+    return <ModalActivator />;
```
SiTaggart: Here's a question: Should we create stories for the main feature of the reach dialog just as a means of knowing what features we need to continue to support if we ever swap out the 3rd party, or if they do a major update and the documentation for the (now, hopefully) widely used features in console etc suddenly disappear.

How much do we care about backwards compat going forward? how much "cover your ass" do we need to make sure consumers of paste and their features still work?

TheSisb: Good questions.  

> Should we create stories for the main feature of the reach dialog just as a means of knowing what features we need to continue to support if we ever swap out the 3rd party

I pushed up a slight change so that the story covers every prop on the components we export (Overlay/Content), excluding the normal spread.  I'm not sure which other stories we could add since this primitive is un-styled and intended to be used in only one way.  Let me know what you think may be missing.

>  or if they do a major update and the documentation for the (now, hopefully) widely used features in console etc suddenly disappear.

I've [forked the reach-ui repo](https://github.com/TheSisb/reach-ui) just in case, but we could also leverage git history to recover docs.  This is a fair point and a good reason why we should probably duplicate docs on our doc site rather than pointing to theirs.  I'll do that work in my docsite PR, it's out of scope for this one.

> How much do we care about backwards compat going forward? how much "cover your ass" do we need to make sure consumers of paste and their features still work?

So long as the documentation is viewable, we're fully covered in the fact that we're pinned to a specific version of `@reach/dialog`.  No one will get breaking changes unless we merge it to master.

packages/paste-core/primitives/modal-dialog/src/index.tsx
richbachman: 👍 on the names.

packages/paste-core/primitives/modal-dialog/stories/index.stories.tsx
SiTaggart: 👍

packages/paste-core/primitives/modal-dialog/src/index.tsx
SiTaggart: as this is the primitive, should we name the exports 

`export {ModalDialogPrimitiveContent, ModalDialogPrimitiveOverlay};`

So that when we create the actual styled modal package we don't get name clashes?

TheSisb: Fixed in https://github.com/twilio-labs/paste/pull/313/commits/e07cdec968eed6f52d308546ef1fa4e059e2f5f1

PR #314, opened 2020-02-14
Fix: missing peer deps

SiTaggart:
Add missing peer deps from some of our packages.

It was causing Grid and Aspect-ration to not compile correctly. The other were but I think that's only by accident due to the monorepo effect

The minimum I'm expecting every package to need is 

- react
- react-dom
- emotion/styled
- styled-system

### DIFF
--- a/packages/paste-core/components/card/package.json
+++ b/packages/paste-core/components/card/package.json
@@ -28,10 +28,13 @@
     "type-check": "tsc --noEmit"
   },
   "peerDependencies": {
+    "@emotion/styled": "^10.0.10",
     "@twilio-paste/box": "^2.1.8",
     "@twilio-paste/style-props": "^0.1.7",
-    "react": ">= 16.0.0",
-    "react-dom": ">= 16.0.0"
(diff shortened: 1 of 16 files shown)


packages/paste-core/components/card/package.json
```diff
@@ -28,10 +28,12 @@
     "type-check": "tsc --noEmit"
   },
   "peerDependencies": {
+    "@emotion/styled": "^10.0.10",
     "@twilio-paste/box": "^2.1.8",
     "@twilio-paste/style-props": "^0.1.7",
-    "react": ">= 16.0.0",
-    "react-dom": ">= 16.0.0"
```
TheSisb: We should give everything prop-types too.  Eventually we will add them back to these packages if they don't already have them, and we might miss this unless we find a more permanent solution.

SiTaggart: OK, that's a good point. Lemme do some copy pasta and gitting and I'll push them up soon, might as well get it all done in one go

PR #315, opened 2020-02-18
docs(website): add status legend to roadmap and overview pages

richbachman:
- [x] Added legend to roadmap page
- [x] Added legend to components overview page
- [x] Added legend to primitives overview page
- [x] Added legend to utilities overview page

### DIFF
--- a/packages/paste-website/src/components/StatusLegend.tsx
+++ b/packages/paste-website/src/components/StatusLegend.tsx
@@ -1,0 +1,23 @@
+import * as React from 'react';
+import {Card} from '@twilio-paste/card';
+import {Heading} from '@twilio-paste/heading';
+import {Text} from '@twilio-paste/text';
+
+export const StatusLegend: React.FC<{}> = () => (
+  <Card>
+    <Heading as="h4" variant="heading40">
+      Component Status Details
+    </Heading>
+    <Text as="p" marginBottom="space20">
+      <strong>Alpha</strong>: Component is a candidate for testing in products with guidance from the design systems
+      team. If {"you 're"} planning on using one of these components, check in with the Paste team first through{' '}
+      <strong>#help-design-system</strong> in Slack.
+    </Text>
+    <Text as="p" marginBottom="space20">
+      <strong>Beta</strong>: Component is a candidate for prototypes and beta products.
+    </Text>
+    <Text as="p" marginBottom="space20">
+      <strong>Production</strong>: The component is stable.
+    </Text>
+  </Card>
+);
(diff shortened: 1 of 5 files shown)


SiTaggart:
The offset is really playing with my eyes when the screen is this large

image

packages/paste-website/src/pages/components/index.mdx
```diff
@@ -21,8 +24,28 @@ import {Callout, CalloutTitle, CalloutText} from '../../components/callout';
...
+        team. If you're planning on using one of these components, check in with the Paste team first through{' '}
+        <strong>#help-design-system</strong> in Slack.
+      </Text>
+      <Text as="p" marginBottom="space20">
+        <strong>Beta</strong>: Component is a candidate for prototypes and beta products. If you're planning on using
+        one of these components, check in with the Paste team first through <strong>#help-design-system</strong> in
+        Slack.
+      </Text>
+      <Text as="p" marginBottom="space20">
+        <strong>Production</strong>: We shipped a component! Everyone should use it.
```
SiTaggart: I'm still concerned with the positioning of this text.

The point of the lifecycle was to enable us to be a little more nimble and not have to jump through all the hoops to release a component to production (actual product, end user facing). This was because we're a small team, and having to design, spec, build, internal usability test, external usability test and go through UX review before  anyone would place a component in production is an unrealistic expectation. 

Given that we version our components and work iteratively the idea of the lifecycle was to allow us to punt customer usability and ux review steps to a later date, but people feel comfortable still using the component in products before hand. Stating that we'll get to it, and it's perfectly safe to use, embracing the idea of a design system evolving component designs over time as we learn.

But the way this reads still really, strongly indicates that components aren't "ready" until they are "production". Given the amount of work that entails based on the life cycle that's going to slow us down and adoption down.

packages/paste-website/src/components/StatusLegend.tsx
```diff
@@ -0,0 +1,23 @@
...
+import {Text} from '@twilio-paste/text';
+
+export const StatusLegend: React.FC<{}> = () => (
+  <Card>
+    <Heading as="h4" variant="heading40">
+      Component Status Details
+    </Heading>
+    <Text as="p" marginBottom="space20">
+      <strong>Alpha</strong>: Component is a candidate for testing in products with guidance from the design systems
+      team. If {"you 're"} planning on using one of these components, check in with the Paste team first through{' '}
```
TheSisb: I'm not a fan of how dense `Alpha` is.  Would it be possible to omit the `If you're...` sentence and add a little blurb to the bottom of the card about reaching out about concerns and assistance instead?  I think it would look more visually balanced this way.

richbachman: Revised this based off the new doc from @SiTaggart

PR #328, opened 2020-02-24
chore: add a package size check action for PRs

SiTaggart:
Use the [compressed size action](https://github.com/preactjs/compressed-size-action) to track package size diffs on PRs

When you make a change it highlights how that changes affects the file size of the built and gzipped files. As an example I changed the grid package by excluding emotion/styled as a peer dep and this was the result to the built package. Pretty neat.


package.json
```diff
@@ -19,7 +19,7 @@
     "start:tokens": "yarn workspace @twilio-paste/design-tokens build:dev",
     "start:website": "yarn workspace @twilio-paste/website develop",
     "watch:icons": "yarn workspace @twilio-paste/icons watch",
-    "prebuild": "node ./tools/build/pre-build.js",
+    "prebuild": "yarn bootstrap && node ./tools/build/pre-build.js",
```
SiTaggart: The action just run `yarn install` but we need to bootstrap. It recommends a postinstall or prebuild hook to run any install commands you need for the build

TheSisb: I don't think we should add the bootstrap step in here because it will cause double runs with clean.  See my other comment

.github/workflows/package-size.yml
```diff
@@ -0,0 +1,16 @@
...
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        node-version: [10.16.x]
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v1
+      - uses: preactjs/compressed-size-action@v1
+        with:
+          repo-token: "${{ secrets.GITHUB_TOKEN }}"
```
TheSisb: you can add on line 17 `build-script: "build-bundle-size"` and make a new build script just for this that will run bootstrap

SiTaggart: mmmm good call, updated

richbachman:
Good find!

PR #330, opened 2020-02-25
feat: storybook responsive addon

richbachman:
- [x] Added responsive addon to Storybook

### DIFF
--- a/.storybook/config.js
+++ b/.storybook/config.js
@@ -3,26 +3,36 @@ import requireContext from 'require-context.macro';
 import {addReadme} from 'storybook-readme';
 import {withA11y} from '@storybook/addon-a11y';
 import withThemeSwitcher from './addons/theme-switcher-addon';
+import {DefaultTheme} from '@twilio-paste/theme-tokens';
+
+const mobileArray = DefaultTheme.breakpoints[0].split(/([0-9]+)/);
+const mobileSize = mobileArray[1];
+
+const tabletArray = DefaultTheme.breakpoints[1].split(/([0-9]+)/);
+const tabletSize = tabletArray[1];
+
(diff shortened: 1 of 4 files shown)


.storybook/config.js
```diff
@@ -4,6 +4,30 @@ import {addReadme} from 'storybook-readme';
 import {withA11y} from '@storybook/addon-a11y';
 import withThemeSwitcher from './addons/theme-switcher-addon';
 
+const customViewports = {
+  Mobile: {
+    name: 'Mobile',
+    styles: {
+      width: '24rem',
```
richbachman: Manually added these from the default theme. Subtracted 1 rem in order to get the correct viewport with the addon. 25rem wouldn't actually show the correct view for the media query.

TheSisb: Can you import these values directly from our tokens package?

richbachman: We can, but this add on creates an iframe with the width, so our media queries don't line up. The iframe with is 25rem using our smallest theme breakpoint, so `@media screen and (min-width: 25rem)` is still enabled in the styles. See screenshot:

image

If I set to 24rem, it works as expected. I could run some js to subtract 1 from our theme breakpoints. Thoughts?

TheSisb: I'd rather not have 2 sources of truth for theme values anywhere, so I'd lean towards some computation based on those values.  However the bigger question is: is this addon helpful if the breakpoints aren't the ones users expect? How big of a deal is that 1rem offset?

richbachman: Not a big deal to calculate that. Just pushed up that change.

richbachman: Screenshot showing expected mobile view:
image

richbachman:
Closing and opening revised PR.

PR #336, opened 2020-02-27
fix(button): change borders to box-shadow

richbachman:
- [x] Changed buttons to use box-shadow instead of border

### DIFF
--- a/packages/paste-core/components/button/__tests__/__snapshots__/button.test.tsx.snap
+++ b/packages/paste-core/components/button/__tests__/__snapshots__/button.test.tsx.snap
@@ -39,7 +39,7 @@ exports[`Button States Has a disabled state 1`] = `
   padding: 0.5rem 1rem;
   border-radius: 4px;
   font-size: 0.875rem;
-  line-height: 1.25rem;
+  line-height: 1.5rem;
   -webkit-appearance: none;
   -moz-appearance: none;
   appearance: none;
@@ -47,16 +47,14 @@ exports[`Button States Has a disabled state 1`] = `
   display: inline-block;
   outline: none;
   background: none;
-  -webkit-transition: background-color 100ms ease-in,border-color 100ms ease-in;
-  transition: background-color 100ms ease-in,border-color 100ms ease-in;
+  -webkit-transition: background-color 100ms ease-in,box-shadow 100ms ease-in;
+  transition: background-color 100ms ease-in,box-shadow 100ms ease-in;
   font-family: 'Whitney SSm A','Whitney SSm B','Helvetica Neue',Helvetica,Arial,sans-serif;
   font-weight: 500;
   cursor: not-allowed;
-  border-width: 2px;
-  border-style: solid;
   color: #ffffff;
-  border-color: #b3d3e9;
   background-color: #b3d3e9;
+  box-shadow: inset 0 0 0 0.125rem #b3d3e9;
   width: auto;
 }
 
@@ -73,6 +71,7 @@ exports[`Button States Has a disabled state 1`] = `
 .emotion-2:hover,
 .emotion-2:active {
   background-color: #b3d3e9;
+  box-shadow: inset 0 0 0 0.125rem #b3d3e9;
 }
 
 .emotion-6 {
@@ -80,7 +79,7 @@ exports[`Button States Has a disabled state 1`] = `
   padding: 0.5rem 1rem;
   border-radius: 4px;
   font-size: 0.875rem;
-  line-height: 1.25rem;
+  line-height: 1.5rem;
   -webkit-appearance: none;
   -moz-appearance: none;
   appearance: none;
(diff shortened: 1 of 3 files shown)


packages/paste-core/components/button/src/styles.ts
```diff
@@ -115,20 +113,26 @@ const variantPrimaryEnabled = (props: ButtonWrapperProps): SerializedStyles =>
...
+        box-shadow: inset 0 0 0 ${themeGet('borderWidths.borderWidth20')(props)}
+            ${themeGet('backgroundColors.colorBackgroundPrimaryDarker')(props)},
+          ${themeGet('shadows.shadowFocus')(props)};
       }
       &:active {
-        border-color: ${themeGet('borderColors.colorBorderPrimaryDarker')(props)};
         background-color: ${themeGet('backgroundColors.colorBackgroundPrimaryDark')(props)};
+        box-shadow: inset 0 0 0 ${themeGet('borderWidths.borderWidth20')(props)}
+            ${themeGet('backgroundColors.colorBackgroundPrimaryDarker')(props)},
+          ${themeGet('shadows.shadowFocus')(props)};
```
richbachman: Is there a better way to do this, since we can't set the individual parameters of `box-shadow` separately?

packages/paste-core/components/button/src/styles.ts
```diff
@@ -10,7 +10,7 @@ import {ButtonWrapperProps, ButtonChildrenProps} from './types';
  */
 const sizeReset = css`
   font-size: 100%;
-  padding: 0;
+  padding: 2px;
```
richbachman: No size token for this. We need padding here because removing the border makes the text right next to the edge of the background.

packages/paste-core/components/button/__tests__/__snapshots__/button.test.tsx.snap
```diff
@@ -110,25 +107,24 @@ exports[`Button States Has a disabled state 1`] = `
...
   font-weight: 500;
   cursor: not-allowed;
-  border-width: 2px;
-  border-style: solid;
   background-color: #e22525;
   color: #ffffff;
-  border-color: #ecb6b6;
+  box-shadow: inset 0 0 0 0.125rem #e22525;
   background-color: #ecb6b6;
+  box-shadow: inset 0 0 0 0.125rem #ecb6b6;
```
TheSisb: Looks like the code added two box-shadow rules with different hexes.  Is this a bug?

packages/paste-core/components/button/src/styles.ts
```diff
@@ -115,20 +113,26 @@ const variantPrimaryEnabled = (props: ButtonWrapperProps): SerializedStyles =>
...
         background-color: ${themeGet('backgroundColors.colorBackgroundPrimaryDarker')(props)};
+        box-shadow: inset 0 0 0 ${themeGet('space.space10')(props)}
+          ${themeGet('backgroundColors.colorBackgroundPrimaryDarker')(props)};
       }
       &:focus {
-        border-color: ${themeGet('borderColors.colorBorderPrimaryDarker')(props)};
         background-color: ${themeGet('backgroundColors.colorBackgroundPrimary')(props)};
+        box-shadow: inset 0 0 0 ${themeGet('space.space10')(props)}
+            ${themeGet('backgroundColors.colorBackgroundPrimaryDarker')(props)},
+          ${themeGet('shadows.shadowFocus')(props)};
```
TheSisb: `shadows.shadowFocus` does the whole rule: 
image
this might be why we're seeing two box shadows in the snapshot maybe

packages/paste-core/components/button/__tests__/__snapshots__/button.test.tsx.snap
richbachman: Good catch. I think I found the issue.

richbachman: Never mind. Confused now. We're rendering two different background colors there as well. https://github.com/twilio-labs/paste/pull/336/files#diff-56164225824111833313cb729bde7ffeL128

packages/paste-core/components/button/src/styles.ts
TheSisb: Misread the `,` - lgtm

packages/paste-core/components/button/__tests__/__snapshots__/button.test.tsx.snap
TheSisb: Per Slack, seems this gets resolved in prod builds.

packages/paste-core/components/button/src/styles.ts
```diff
@@ -8,9 +8,9 @@ import {ButtonWrapperProps, ButtonChildrenProps} from './types';
 /*
  * Sizes
  */
-const sizeReset = css`
+const sizeReset = (props: ButtonWrapperProps): SerializedStyles => css`
   font-size: 100%;
-  padding: 0;
+  padding: ${themeGet('space.space10')(props)};
```
TheSisb: This should be a hard 0 too, because no matter what theme is provided we want it to remain 0.

richbachman:
@TheSisb @SiTaggart in reference to the padding change on the rest variant. I added the 2px padding to add a little spacing in between the text and background color on the primary and secondary reset variants. We lost that when removing the border. See screenshot:

image

SiTaggart:
The problem with reset having a 2px padding to make up for the lack of border, means that when I legit don't want a border and background, I also don't want padding. I want nothing. So we may have to live with the text hitting the edges

serifluous:
No padding makes sense to me. People generally won't be using it in its secondary/primary forms anyway, right? It'll mostly be used unstyled?

richbachman:
No padding it is. Thanks!

PR #337, opened 2020-02-28
feat(design-tokens): consistent iconsize to line-height tokens

TheSisb:
Adds new line-height alias values and replaces the existing `lineHeight` and `sizeIcon` tokens to use the same alias values. This is handy because we often encounter issues sizing and aligning icons in the normal flow of text.  Now, icons and text can share the same sizing values. For example, if your content has `lineHeight50`, you can use the matching `sizeIcon50` for a perfect fit (just match the number).


serifluous:
Not sure why, but $size-icon-40 isn't taking the new value
Screen Shot 2020-02-28 at 12 21 59 PM

TheSisb:
@serifluous I caught that in the VRT test too.  There was a bug in the change, I just fixed and pushed it up.  Should be good now.

serifluous:
Is showing the calculation in the comment intentional? Looks a little strange out of context on the tokens page
Screen Shot 2020-02-28 at 4 13 17 PM

It'd help to add back the description that these map to the font-size tokens, or explain what each of the numbers in the calculation mean.

TheSisb:
@serifluous The old comment didn't really provide any value.  Can you remind me what the line-height rule was? I can add that instead.  1.55 for < 20px? 1.25 for ?

serifluous:
@TheSisb https://share.goabstract.com/6e8b72d4-fce3-4b2f-938c-8b845ad2e8a8?mode=design&sha=ddb5a6a06fc0caf9451f1a3d156d6c768c6a4ea7

richbachman:
Reviewed storybook locally. Looks like everything is looking as it should.

PR #340, opened 2020-03-03
feat(box): allow passing tokens via pseudo-class props

TheSisb:
Inspired from [Chakra-UI's PseudoBox](https://chakra-ui.com/pseudobox).

This component makes Box much more versatile by allowing consumers to use the token props on CSS pseudo-selectors.

### DIFF
--- a/packages/paste-core/primitives/box/package.json
+++ b/packages/paste-core/primitives/box/package.json
@@ -29,6 +29,7 @@
   },
   "peerDependencies": {
     "@emotion/styled": "^10.0.10",
+    "@styled-system/css": "^5.1.5",
     "@twilio-paste/design-tokens": "^4.2.0",
     "@twilio-paste/style-props": "^0.1.10",
     "@twilio-paste/theme-tokens": "^3.1.0",
--- a/packages/paste-core/primitives/box/src/index.tsx
+++ b/packages/paste-core/primitives/box/src/index.tsx
@@ -1,5 +1,6 @@
 import styled from '@emotion/styled';
 import {compose, layout, space, background, border, boxShadow, position, flexbox, system} from 'styled-system';
+import css from '@styled-system/css';
 import {
   LayoutProps,
   SpaceProps,
@@ -9,8 +10,32 @@ import {
   PositionProps,
   FlexboxProps,
 } from '@twilio-paste/style-props';
+import {
+  CursorProperty,
+  AppearanceProperty,
+  AnimationProperty,
+  TransformProperty,
+  TransformOriginProperty,
+  VisibilityProperty,
+  WhiteSpaceProperty,
+  UserSelectProperty,
+  PointerEventsProperty,
+  BoxSizingProperty,
+  ResizeProperty,
+  TransitionProperty,
+  ListStyleTypeProperty,
+  ListStylePositionProperty,
+  ListStyleImageProperty,
+  ObjectFitProperty,
+  ObjectPositionProperty,
+  BackgroundAttachmentProperty,
+  OutlineProperty,
+  FloatProperty,
+  WillChangeProperty,
+} from 'csstype';
+import {PseudoPropStyles} from './PseudoPropStyles';
 
-export interface BoxProps
+interface BaseBoxProps
   extends React.HTMLAttributes<any>,
     LayoutProps,
     SpaceProps,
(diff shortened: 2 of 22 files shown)


packages/paste-core/primitives/box/src/fakeCss.js
```diff
@@ -0,0 +1,217 @@
...
+  marginY: ['marginTop', 'marginBottom'],
+  paddingX: ['paddingLeft', 'paddingRight'],
+  paddingY: ['paddingTop', 'paddingBottom'],
+  size: ['width', 'height'],
+};
+
+const scales = {
+  color: 'colors',
+  backgroundColor: 'backgroundColors',
+  borderColor: 'borderColors',
```
TheSisb: I had to change what these mapped to from the normal theme keys in styled-system https://styled-system.com/css/#theme-keys because we differ.  Not sure if anything else needed changing.

packages/paste-core/primitives/box/src/index.tsx
```diff
@@ -55,5 +86,22 @@ const Box = styled.div(
...
+    return {};
+  }
+
+  const pseudoStyles = {};
+  for (const pseudoProp of pseudoProps) {
+    pseudoStyles[PseudoStyles[pseudoProp.replace('_', '')]] = props[pseudoProp];
+  }
+
+  console.log(pseudoStyles);
+  return css(pseudoStyles);
```
TheSisb: I should add comments but basically:
1. Find all props that start with a `_`, we will use this as our way to tell if the prop is a pseudo selector
2. Build an object of styles from those props where the key is retrieved from the PseudoStyles.tsx file, and the value is from the prop
3. Run this through a modified `@styled-system/css` so we can get the right token from a string value.

packages/paste-core/primitives/box/package.json
```diff
@@ -29,6 +29,7 @@
   },
   "peerDependencies": {
     "@emotion/styled": "^10.0.10",
+    "@styled-system/css": "^5.1.5",
```
SiTaggart: You might want to run `npx @yarnpkg/doctor`. My hunch is that you're going to need to add this as a peer dep to every package that uses box, and also add it as a dev dep to any package that has a peer dep of box.

packages/paste-theme-tokens/src/console/index.ts
```diff
@@ -19,12 +19,18 @@ const breakpoints = [sizings.size40, sizings.size100, sizings.size120];
 
 export const ConsoleTheme = {
   shadows: boxShadows,
-  backgroundColors,
-  borderColors,
   borderWidths,
   radii,
   breakpoints,
-  colors,
+  colors: {
```
richbachman: 🔥 Excited for this change. Makes the input styling way cleaner.

richbachman:
This is looking pretty solid. +1 to adding stories and a snapshot test or two.

packages/paste-core/primitives/box/src/index.tsx
```diff
@@ -20,8 +23,40 @@ export interface BoxProps
     PositionProps,
     FlexboxProps {
   as?: keyof JSX.IntrinsicElements;
+  content?: string;
+  cursor?: CursorProperty;
```
SiTaggart: I don't know how content is working, but for cursor and appearance you have compose a couple of `system` styles into the component for them to work by the looks of it. Similar to what Chakra are doing with this object https://github.com/chakra-ui/chakra-ui/blob/master/packages/chakra-ui/src/Box/config.js#L113

I created a little cut down object locally and it seemed to get it working as expected.

```
const config = {
  textTransform: true,
  animation: true,
  appearance: true,
  transform: true,
  transformOrigin: true,
  visibility: true,
  whiteSpace: true,
  userSelect: true,
  pointerEvents: true,
  wordBreak: true,
  overflowWrap: true,
  textOverflow: true,
  boxSizing: true,
  cursor: true,
  resize: true,
  transition: true,
  listStyleType: true,
  listStylePosition: true,
  listStyleImage: true,
  objectFit: true,
  objectPosition: true,
  outline: true,
  float: true,
  willChange: true,
};

const extraConfig = system(config);
```

then add to the compose

```
export const Box = styled.div(
  {
    boxSizing: 'border-box',
    minWidth: 0,
  },
  compose(
    space,
    layout,
    flexbox,
    background,
    backgroundColor,
    border,
    borderColor,
    boxShadow,
    position
  ),
  getPseudoStyles
) as React.FC<BoxProps>;
```

TheSisb: Added a bunch. Good catch

packages/paste-website/src/pages/primitives/box/index.mdx
```diff
@@ -196,55 +196,92 @@ yarn add @twilio-paste/box
...
+  backgroundColor="colorBackgroundBody" 
+  padding="space60" 
+  _hover={{
+    backgroundColor: "colorBackgroundPrimary"
+    padding: "space70"
+  }}
+>
   Foo
-</Box>;
+</Box>
```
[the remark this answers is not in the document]
TheSisb: Sure, added one.

PR #341, opened 2020-03-04
feat(form): setup base form package

richbachman:
- [x] FormInput
- [x] FormLabel
- [x] FormHelperText


packages/paste-core/components/form/src/FormInput.tsx
```diff
@@ -0,0 +1,107 @@
...
+    hasError={hasError}
+    id={id}
+    name={name}
+    placeholder={placeholder}
+    readOnly={readOnly}
+    required={required}
+    type={type}
+    value={value}
+  />
+);
```
SiTaggart: With this design, how are you going to handle the prefix and suffix children?

packages/paste-core/components/form/src/FormLabel.tsx
```diff
@@ -0,0 +1,47 @@
...
+    marginRight: 'space20',
+    backgroundColor: 'colorBackgroundDestructive',
+    borderRadius: '50%',
+  })
+);
+
+const FormLabel: React.FC<FormLabelProps> = props => {
+  return (
+    <Flex vAlignContent="center" marginBottom="space10">
+      {props.required ? <StyledRequiredDot /> : null}
```
SiTaggart: The only downside I can see with this approach is that the required indicator is no longer clickable when it's outside of the `label` element

packages/paste-core/components/form/stories/index.stories.tsx
```diff
@@ -0,0 +1,102 @@
...
+  .add('Input - Error', () => {
+    const uid = useUID();
+    return (
+      <>
+        <FormLabel for={uid}>Label</FormLabel>
+        <FormInput id={uid} type="text" placeholder="Placeholder" hasError />
+        <FormHelperText hasError>
+          <Flex>
+            <Box marginRight="space20">
+              <WarningIcon iconColor="colorTextError" decorative={false} title="warning: " size="sizeIcon30" />
```
SiTaggart: Should we handle this status level icon for the consumer? In the same way we handle it for alerts, the level should come with an icon for accessibility reasons, so we handle that for you. The content _next_ to it is up to you still.

https://github.com/twilio-labs/paste/blob/master/packages/paste-core/components/alert/src/index.tsx#L74

packages/paste-core/components/form/src/FormHelperText.tsx
```diff
@@ -0,0 +1,27 @@
+import * as React from 'react';
+import {Text, safelySpreadTextProps} from '@twilio-paste/text';
+
+export interface FormHelperTextProps extends React.HTMLAttributes<''> {
+  className?: never;
+  hasError?: boolean;
```
SiTaggart: For the sake of scaling the component in the future, lets imagine we add a warning help text variant, should this be a `variant` enum with just `error` as the only option for now.

I came across this article this week and it got me thinking about it https://kentcdodds.com/blog/stop-using-isloading-booleans

packages/paste-core/components/form/stories/index.stories.tsx
richbachman: I restructured this to include the warning icon.

packages/paste-core/components/form/src/FormHelperText.tsx
richbachman: I changed this so it has an error variant, like alerts. Do we need the `hasError` prop anymore?

packages/paste-core/components/form/src/FormLabel.tsx
richbachman: Good catch. Changes so the indicator is part of the label.

packages/paste-core/components/form/src/FormInput.tsx
richbachman: Good question. I'm open to ideas on this. My initial thought is something like this:
```
<FormInputGroup>
    <FormInputAddon />
    <FormInput variant="addon" />
</FormInputGroup>
```

packages/paste-core/components/form/src/FormHelperText.tsx
richbachman: Killed the `hasError` prop. Replaced with error variant.

packages/paste-core/components/form/src/FormInput.tsx
```diff
@@ -0,0 +1,136 @@
...
+    typeProps.type = 'text';
+    typeProps.inputmode = 'numeric';
+    typeProps.pattern = '[0-9]*';
+  }
+
+  return (
+    <StyledFormInput
+      aria-invalid={hasError}
+      aria-readonly={readOnly}
+      {...safelySpreadBoxProps(props)}
```
SiTaggart: Is the idea here that if you explicitly want to override what we set for invalid and readonly you can do so?

packages/paste-core/components/form/src/FormInput.tsx
SiTaggart: So would the `FormInputGroup` have the border? That way you don't need to start conditionally setting border radius on things, which might not treat us too well in the future if we ever need to change or add anything.

packages/paste-core/components/form/src/FormLabel.tsx
```diff
@@ -0,0 +1,59 @@
...
+    borderRadius: '50%',
+  })
+);
+
+const FormLabel: React.FC<FormLabelProps> = ({marginBottom, required, disabled, children, ...props}) => {
+  return (
+    <Text
+      {...safelySpreadTextProps(props)}
+      as="label"
+      display="inline-block"
```
SiTaggart: This should probably be `block` so that it spans the full width. That was I get the full row as a click target, no just the text

SiTaggart: It'll also get rid of this inline-block problem when adding the dot, which cause a few pixel jump in size of the parent.

packages/paste-core/components/form/src/FormInput.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+// eslint-disable-next-line emotion/syntax-preference
+const StyledFormInput = styled(Box)<FormInputProps>(({hasError}: FormInputProps) =>
+  css({
+    appearance: 'none',
+    outline: 'none',
+    resize: 'none',
+    width: '100%',
+    backgroundColor: 'colorBackgroundBody',
+    borderWidth: 'borderWidth10',
```
SiTaggart: Require here is that input height matches the button height. So we're going to have to remove that border and use a box-shadow to get the math to line up.

packages/paste-core/components/form/src/FormInput.tsx
SiTaggart: Mainly for reused purposes but also alignment. Textare and select all share the same border and prefix / suffix treatments, but select also has an addition icons.

image

image

packages/paste-core/components/form/src/FormInput.tsx
TheSisb: Yup!

packages/paste-core/components/form/src/FormLabel.tsx
richbachman: It could be `display: flex`, but `alignItems` isn't supported on `Text`.

richbachman: I worked out a different way to use `Flex` and `Text`.

packages/paste-core/components/form/src/FormInput.tsx
richbachman: I've setup new shadow tokens and replaced all of the borders with the box-shadows.

packages/paste-design-tokens/tokens/aliases/box-shadow.yml
```diff
@@ -1,8 +1,14 @@
...
 aliases:
   shadow-elevation-0: "none"
   shadow-elevation-10: "{!offset-0} {!offset-10}px {!offset-20}px {!offset-0} {!palette-gray-100-transparent-30}"
   shadow-elevation-20: "{!offset-0} {!offset-40}px {!offset-60}px {!offset-0} {!palette-gray-100-transparent-20}"
   shadow-elevation-30: "{!offset-0} {!offset-50}px {!offset-80}px {!offset-20} {!palette-gray-100-transparent-20}"
+  shadow-border-input: "{!offset-0} {!offset-0} {!offset-0} 1px {!palette-gray-60}"
+  shadow-border-input-hover: "{!offset-0} {!offset-0} {!offset-0} 1px {!palette-blue-90}"
+  shadow-border-input-error: "{!offset-0} {!offset-0} {!offset-0} 1px {!palette-red-60}"
+  shadow-border-input-error-hover: "{!offset-0} {!offset-0} {!offset-0} 1px {!palette-red-80}"
+  shadow-border-input-disabled: "{!offset-0} {!offset-0} {!offset-0} 1px {!palette-gray-20}"
```
serifluous: In the console and SG themes, this should be palette-gray-20, but it's correct here in the default theme.

packages/paste-design-tokens/tokens/global/border-color.yml
```diff
@@ -74,8 +77,8 @@ props:
 
   # destructive borders
   color-border-destructive-darker:
-    value: "{!palette-red-90}"
-    comment: Destructive hover border color
+    value: "{!palette-red-80}"
+    comment: Darker destructive hover border color
```
serifluous: This should stay red-90 in the default theme.

packages/paste-design-tokens/tokens/aliases/box-shadow.yml
serifluous: Should these match the same naming as the border tokens, so they can also be used in buttons? As in: shadow-border-input, shadow-border-primary-darker, shadow-border-error, etc.

packages/paste-design-tokens/tokens/global/border-color.yml
```diff
@@ -62,6 +62,9 @@ props:
   color-border-warning-lightest:
     value: "{!palette-orange-10}"
     comment: Lightest warning border color
+  color-border-error-darker:
+    value: "{!palette-red-80}"
+    comment: Darker Error border color
```
serifluous: This should be dark rather than darker in all themes + the form field error state.

packages/paste-design-tokens/tokens/global/border-color.yml
richbachman: Corrected, and fixed the comment.

PR #345, opened 2020-03-13
feat(form): textarea component

richbachman:
- [x] Added FormTextArea component
- [x] Added FormTextArea component tests
- [x] Changed InputWrapper to FieldWrapper
- [x] Moved wrapper, prefix, and suffix to shared folder


packages/paste-core/components/form/src/FormTextArea.tsx
```diff
@@ -0,0 +1,100 @@
...
+        </TextAreaElement>
+        <Suffix>{insertAfter}</Suffix>
+      </FieldWrapper>
+    );
+  }
+);
+
+FormTextArea.displayName = 'FormTextArea';
+
+FormTextArea.propTypes = {
```
TheSisb: please wrap in `if (process.env.NODE_ENV === 'development') {`

packages/paste-core/components/form/src/shared/FieldWrapper.tsx
```diff
@@ -0,0 +1,42 @@
...
+      _focusWithin={{
+        boxShadow: 'shadowFocus',
+      }}
+    >
+      {children}
+    </Box>
+  );
+};
+
+FieldWrapper.displayName = 'FieldWrapper';
```
TheSisb: missing propTypes

packages/paste-core/components/form/src/shared/Prefix.tsx
```diff
@@ -0,0 +1,25 @@
+import * as React from 'react';
+import {Box} from '@twilio-paste/box';
+
+const Prefix: React.FC = ({children}) => {
+  if (children == null) return null;
+  return (
+    <Box
+      display="flex"
+      alignItems="flex-start"
```
TheSisb: I think if we remove this line we should get the same outcome with less output CSS

packages/paste-core/components/form/src/shared/Suffix.tsx
```diff
@@ -0,0 +1,25 @@
...
+    <Box
+      display="flex"
+      alignItems="flex-start"
+      backgroundColor="colorBackground"
+      padding="space30"
+      borderLeftWidth="borderWidth10"
+      borderLeftColor="colorBorderLight"
+      borderLeftStyle="solid"
+      borderTopRightRadius="borderRadius20"
+      borderBottomRightRadius="borderRadius20"
```
TheSisb: Good fix 👍

packages/paste-core/components/form/__tests__/formTextarea.test.tsx
```diff
@@ -0,0 +1,240 @@
...
+
+  it('it should render with readOnly', (): void => {
+    const tree = renderer
+      .create(
+        <Theme.Provider theme="console">
+          <FormTextArea id="textarea" onChange={NOOP} readOnly />
+        </Theme.Provider>
+      )
+      .toJSON();
+    expect(tree).toMatchSnapshot();
```
TheSisb: prefer enzyme to validate these props over snapshots

packages/paste-core/components/form/src/shared/Prefix.tsx
richbachman: We do need to set that here, other wise in certain cases the child element will span the full height so the click area is huge. See focus in screenshot:

image

packages/paste-core/components/form/__tests__/formTextarea.test.tsx
SiTaggart: Potentially if you're doing the enzyme test down at https://github.com/twilio-labs/paste/pull/345/files#diff-60403f775a91aa64baef1b2aed75204fR124 you can probably just remove this snapshot test entirely

SiTaggart:
If you're including the auto height stuff then I'll hold off on approval. Otherwise this looks pretty great

packages/paste-core/components/form/__tests__/formTextarea.test.tsx
richbachman: Removed and just left the initial render snapshot.

richbachman:
@SiTaggart decided to move the height sizing to a new ticket so we figure out the proper solution.

packages/paste-core/components/form/stories/index.stories.tsx
```diff
@@ -190,4 +190,33 @@ storiesOf('Components|Form', module)
         <FormHelperText>Info that helps a user with this field.</FormHelperText>
       </>
     );
+  })
+  .add('Textarea Options', () => {
```
SiTaggart: Can we get a story for every visual variant and state a textarea can appear in please? Just for VRT purposes.

richbachman: @SiTaggart added VRT stories.

PR #347, opened 2020-03-16
fix: acorn dependency vulnerability notice

TheSisb:
Because this notice is not pretty and the automated PR doesn't solve the root issue:
image


packages/paste-core/components/alert/package.json
```diff
@@ -50,11 +50,8 @@
   },
   "devDependencies": {
     "@babel/core": "^7.8.4",
-    "@emotion/core": "^10.0.10",
```
SiTaggart: These were all added based on the output of the package doctor, and it kind of makes sense when you take into consideration that this package needs to build on its own. Peer dependencies are installed by a host app, but for the sake of building this component, the host app is itself, hence all the peer deps that are needed for the build are listed as dev deps

package.json
```diff
@@ -61,7 +61,7 @@
     "styled-system": "^5.1.2"
   },
   "devDependencies": {
-    "@applitools/eyes-storybook": "3.2.2",
```
SiTaggart: this is locked to 3.2.2 because later versions seem to stop working and I haven't bothered to debug why

PR #348, opened 2020-03-17
chore: bump acorn version to 5.7.4

SiTaggart:
Address security vulnerability in acorn

### DIFF
--- a/package.json
+++ b/package.json
@@ -120,7 +120,7 @@
     "git-branch-is": "^3.0.0",
     "husky": "^3.0.0",
     "jest": "^24.8.0",
-    "jest-axe": "^3.2.0",
+    "jest-axe": "3.3.0",
     "jest-emotion": "^10.0.14",
     "lerna": "^3.16.4",
     "lodash": "^4.17.14",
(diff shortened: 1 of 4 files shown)


package.json
```diff
@@ -120,7 +120,7 @@
     "git-branch-is": "^3.0.0",
     "husky": "^3.0.0",
     "jest": "^24.8.0",
-    "jest-axe": "^3.2.0",
+    "jest-axe": "3.3.0",
```
SiTaggart: `3.4.0` caused a bunch of errors I don't want to deal with right now.

packages/paste-website/src/components/shortcodes/table-of-contents/styles.ts
```diff
@@ -36,7 +36,7 @@ export const fontSizePartial = ({depth}: TableOfContentsListItemProps): string =
   return themeGet('fontSizes.fontSize30');
 };
 
-export const StyledListItem: React.FC<TableOfContentsListItemProps> = styled.li`
+export const StyledListItem = styled.li<TableOfContentsListItemProps>`
```
TheSisb: Derp, so that was the issue huh?

SiTaggart: Yeah... it was kind of frustrating. We do this in other components. I still think the function component route _should_ work, but what ever, this works for us now.

PR #351, opened 2020-03-17
chore: move utilities to layout components

richbachman:
- [x] Moved all utility components to layout directory
- [x] Update storybook
- [x] Update website nav and pages


SiTaggart:
I don't think I was explicit enough in the Jira ticket, but not all utilities are layout components. The assumption being we need to move some to components

richbachman:
@SiTaggart I wondered that as I was working on this. I'll restructure and get those updated. Thanks for the clarification.

richbachman:
@SiTaggart I confused myself trying to move everything around properly in the website, so I'm opening a new PR with everything moved around in the right spot.

PR #358, opened 2020-03-25
fix(form): change FormHelperText to FormHelpText

richbachman:
- [x] Renamed `FormHelperText` to `FormHelpText`
- [x] Updated stories
- [x] Updated tests
- [x] Updated input doc


packages/paste-core/components/form/src/FormHelpText.tsx
```diff
@@ -45,12 +45,12 @@ const FormHelperText: React.FC<FormHelperTextProps> = ({marginTop, children, var
   );
 };
 
-FormHelperText.displayName = 'FormHelperText';
+FormHelpText.displayName = 'FormHelperText';
```
SiTaggart: Missed the value here

richbachman: Fixed!

PR #359, opened 2020-03-25
fix(form): remove input type hidden box shadow

richbachman:
- [x] Added `box-shadow: none` rule for hidden input type
- [x] Added VRT story

### DIFF
--- a/packages/paste-core/components/form/src/FormInput.tsx
+++ b/packages/paste-core/components/form/src/FormInput.tsx
@@ -5,8 +5,7 @@ import css from '@styled-system/css';
 import {FieldWrapper} from './shared/FieldWrapper';
 import {Prefix} from './shared/Prefix';
 import {Suffix} from './shared/Suffix';
-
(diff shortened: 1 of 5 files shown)


packages/paste-core/components/form/src/FormInput.tsx
```diff
@@ -81,7 +80,7 @@ const FormInput = React.forwardRef<HTMLInputElement, FormInputProps>(
     }
 
     return (
-      <FieldWrapper readOnly={readOnly} disabled={disabled} hasError={hasError}>
+      <FieldWrapper {...typeProps} readOnly={readOnly} disabled={disabled} hasError={hasError}>
```
SiTaggart: I would favour just passing down `type` here. If it's ever a number, `typeProps` is an object, and not a string and that feels weirdly misleading for FieldWrapper.

richbachman: Passing `type` now.

PR #360, opened 2020-03-26
fix(box): removing min-width

SiTaggart:
Box having a default `min-width:0` property was causing undesirable effects when flex children were collapsing their containers.

Removing it as a default and only setting it when needed it a safer, more intentional fix.

TODO:
- [x] verify Flex needs it?


richbachman:
Doing some initial testing. I'm not sure `Flex` needs the min-width. Did you see anything stand out that made you think it was necessary to keep there?

PR #365, opened 2020-03-31
feat(form): add autoresizing to the textarea component

SiTaggart:
Adding resizing to the textarea component via react-autosize-textarea.

I have not added any tests here as enzyme doesn't seem to detect that styles applied by the library in the style prop. We have to rely on the fact that we're just directly using the library directly and that should have it's own tests. It's annoying, I'm getting annoyed with enzyme.

### DIFF
--- a/packages/paste-core/components/form/package.json
+++ b/packages/paste-core/components/form/package.json
@@ -58,5 +58,8 @@
     "rollup-plugin-typescript2": "^0.21.2",
     "typedoc": "^0.16.9",
     "typescript": "3.7.5"
+  },
+  "dependencies": {
+    "react-autosize-textarea": "^7.0.0"
   }
 }
--- a/packages/paste-core/components/form/package.json
+++ b/packages/paste-core/components/form/package.json
@@ -60,6 +60,6 @@
     "typescript": "3.7.5"
   },
   "dependencies": {
-    "react-autosize-textarea": "^7.0.0"
+    "react-autosize-textarea": "7.0.0"
   }
 }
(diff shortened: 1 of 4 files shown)


packages/paste-core/components/form/src/FormTextArea.tsx
```diff
@@ -66,6 +67,7 @@ const FormTextArea = React.forwardRef<HTMLTextAreaElement, FormTextAreaProps>(
           aria-invalid={hasError}
           aria-readonly={readOnly}
           {...props}
+          async
```
richbachman, line 70: I understand the `async` concept, but am curious what adding `async` here does for the styled component?

richbachman: I'm guessing it has something to do with adding the arrow function on line 28.

SiTaggart: No, it's actually a little more criptic than that. It's part of the API of the `TextareaAutosize` component https://github.com/buildo/react-autosize-textarea#props.

But because I'm using a styled version of it, you can't immediately tell it's actually part of that components API.

richbachman: Ok. I see that now. Thanks for the clarification. I pulled down locally and everything looks like its working. I like how it still adheres to our max-height.

packages/paste-core/components/form/package.json
```diff
@@ -58,5 +58,8 @@
     "rollup-plugin-typescript2": "^0.21.2",
     "typedoc": "^0.16.9",
     "typescript": "3.7.5"
+  },
+  "dependencies": {
+    "react-autosize-textarea": "^7.0.0"
```
richbachman, line 63: Just thought of this. Should we pin this version number?

SiTaggart: Excellent call!

PR #372, opened 2020-04-04
feat(modal): create the modal package

SiTaggart:
**Contributing to Twilio**


packages/paste-core/components/modal/package.json
```diff
@@ -0,0 +1,82 @@
+{
+  "name": "@twilio-paste/modal",
+  "version": "0.0.0",
+  "category": "overlay",
+  "status": "beta",
+  "description": "An alert notifies a user to high-priority or time-sensitive information.",
```
richbachman, line 6: I think this is the Alert package description.

packages/paste-core/components/modal/src/Modal.tsx
```diff
@@ -0,0 +1,97 @@
...
+}) => {
+  return (
+    <ModalDialogOverlay
+      isOpen={isOpen}
+      onDismiss={onDismiss}
+      allowPinchZoom={allowPinchZoom}
+      initialFocusRef={initialFocusRef}
+    >
+      <ModalContext.Provider value={{onDismiss}}>
+        <ModalDialogContainer aria-labelledby={ariaLabelledby} {...props} className={null} style={null} size={size}>
```
richbachman, line 77: Why are we setting `className={null} style={null}` here?

richbachman: Oh, is it so they don't pass down through `...prop`?

packages/paste-core/components/modal/src/ModalHeading.tsx
```diff
@@ -0,0 +1,24 @@
...
+import * as PropTypes from 'prop-types';
+import {Heading, HeadingProps} from '@twilio-paste/heading';
+
+export interface ModalHeadingProps extends Omit<HeadingProps, 'marginBottom' | 'variant' | 'as'> {
+  children: NonNullable<React.ReactNode>;
+  as?: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6';
+}
+const ModalHeading: React.FC<ModalHeadingProps> = ({children, ...props}) => {
+  return (
+    <Heading as="h2" {...props} marginBottom="space0" variant="heading30">
```
richbachman, line 11: Should this `as` be a default prop? It looks like you're setting it as an `h3` in the stories.

packages/paste-core/components/modal/package.json
SiTaggart: It is, good spot

packages/paste-core/components/modal/src/Modal.tsx
SiTaggart: Yup. Because we're spreading props, and not using box, I need to unset className and style so that people can't override our styles.

packages/paste-core/components/modal/src/ModalHeading.tsx
SiTaggart: Yeah, we could do I suppose.

packages/paste-core/components/modal/stories/index.stories.tsx
```diff
@@ -0,0 +1,214 @@
...
+
+storiesOf('Components|Modal', module)
+  .addDecorator(withKnobs)
+  .add('default', () => {
+    return <ModalTrigger size="default" />;
+  })
+  .add('wide', () => {
+    return <ModalTrigger size="wide" />;
+  })
+  .add('footer actions', () => {
```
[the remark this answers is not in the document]
SiTaggart: What's the concern? The only reason I have it, even though it's the same, is that it's an explicit visual regression test for footer action layout. If you change the alignment independently, or modify that component we have an explicit test for it.

packages/paste-core/components/modal/package.json
SiTaggart: Updated

packages/paste-core/components/modal/src/ModalHeading.tsx
SiTaggart: Updated with a default prop instead

packages/paste-core/components/modal/stories/index.stories.tsx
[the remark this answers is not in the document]
richbachman: No real concern. The VRT totally makes sense here. Just noticed the similarity so thought I'd call it out. I'm good to go forward as is.

richbachman:
This looks great!

PR #375, opened 2020-04-08
fix: dependencies in all packages

SiTaggart:
I notice the form package wasn't building correctly in the publish stage. Turned out to be some missing dependencies.

image

Modal was also missing deps.

Also realised that all packages needed an update whilst I was at it too due to the theme package update with theme tokens.

Also noticed modal wasn't part of core yet, so added that too.


SiTaggart:
Yarn package doctor no longer works until we upgrade to yarn 2 unfortunately. 

There is a pattern to follow though. In any package with a peer dependency, the peer dependencies of that peer, also need to be peer deps of your package, all the way down the dep tree. They also need to be dev dependencies of the package. 

So your package has a peer dep on Box, Box has a peer dep on theme so you need to hoist that up to your package. Theme should appear as a peer dep and a dev dep

richbachman:
Thanks for the pattern explanation. Until we upgrade to yarn 2, it sounds like we just need to keep a close on on deps, and review those with more scrutiny.

PR #376, opened 2020-04-08
fix(website): add overflow-y to do/dont Box

richbachman:
- [x] Added `overflow-y` to add scroll capabilities to the do/dont `Box` containers. Now longer do/dont example content is visible via scroll.


serifluous:
Can you make scrollbar visible when there's overflow? Hard to tell on the List doc page that there's overflow content.

richbachman:
@serifluous not without building a custom scrollbar. This uses the default browser scrollbar, so its the fancy hidden one for Macs, and I think its visible for most PC browsers.

PR #392, opened 2020-04-14
feat(form): add radio and checkbox component

richbachman:
**Contributing to Twilio**


packages/paste-core/components/form/src/FormRadio.tsx
```diff
@@ -0,0 +1,139 @@
...
+          _hover={{
+            borderColor: 'colorBorderPrimary',
+          }}
+          _focus={{
+            boxShadow: 'shadowFocus',
+          }}
+          // _active={{
+          //   borderColor: 'colorBorderPrimaryDarker',
+          //   backgroundColor: 'colorBackgroundPrimaryDarker',
+          // }}
```
richbachman, line 91: Questions for design. Active states are kind of funky. I'll review with design.

packages/paste-core/components/form/src/FormRadio.tsx
```diff
@@ -0,0 +1,139 @@
...
+            borderColor: 'colorBorderError',
+          }}
+          _checkedAndHover={{
+            borderColor: 'colorBorderPrimaryDarker',
+            backgroundColor: 'colorBackgroundPrimaryDarker',
+          }}
+          // _checkedAndActive={{
+          //   borderColor: 'colorBorderPrimaryDarker',
+          //   backgroundColor: 'colorBackgroundPrimaryDarker',
+          // }}
```
richbachman, line 109: Same here. Active states are kind of funky. I'll review with design.

packages/paste-core/components/form/src/FormRadio.tsx
```diff
@@ -0,0 +1,139 @@
...
+            borderColor: 'colorBorder', // need new token
+            backgroundColor: 'colorBackground', // need new token
+          }}
+        >
+          <Box
+            as="span"
+            // @ts-ignore
+            width="0.375rem"
+            // @ts-ignore
+            height="0.375rem"
```
richbachman, line 120: Probably a design question as well. Not sure what to do about this width/height here.

packages/paste-core/components/form/src/shared/SiblingBox.tsx
```diff
@@ -0,0 +1,67 @@
...
+  ({
+    type = 'checkbox',
+    _active,
+    _hover,
+    _invalid,
+    _disabled,
+    _focus,
+    _checked,
+    // _child = {opacity: 0},
+    // _checkedAndChild = {opacity: 1},
```
richbachman, line 20: I don't quite understand if we need these two.

richbachman: Understand these more now. Just need to work through TS errors.

packages/paste-core/components/form/src/FormRadio.tsx
```diff
@@ -0,0 +1,158 @@
...
+        {...safelySpreadBoxProps(props)}
+        as="label"
+        display="inline-flex"
+        verticalAlign="top"
+        htmlFor={id}
+        alignItems="flex-start"
+        width={isFullWidth ? '100%' : undefined}
+        cursor={isDisabled ? 'not-allowed' : 'pointer'}
+      >
+        <ScreenReaderOnly
```
SiTaggart, line 61: I meant wrap an input with it
```
<ScreenReaderOnly>
  <input {...props} />
</ScreenReaderOnly>
```

richbachman: Doing so wrap the `input` in a `span` which makes the adjacent pseudo styles like `input[type=${type}]:focus + &` not work.

SiTaggart: ooooooooooooooooooooooooooooooohhhhh, Ok we need different typings then

packages/paste-core/components/screen-reader-only/src/index.tsx
```diff
@@ -1,6 +1,6 @@
 import styled from '@emotion/styled';
 
-export interface ScreenReaderOnlyProps {
+export interface ScreenReaderOnlyProps extends React.InputHTMLAttributes<HTMLInputElement> {
```
SiTaggart, line 3: Not sure how I feel about this. Can you try something similar to what we do in `Box`? We recast the styled div as a Functional React component to blow away the div typings. https://github.com/twilio-labs/paste/blob/master/packages/paste-core/primitives/box/src/index.tsx#L166

richbachman: Reverted this change.

packages/paste-core/components/form/src/FormRadio.tsx
richbachman: Resolved with the square size tokens.

packages/paste-core/components/form/src/FormRadio.tsx
richbachman: Resolved by using a styled `input`

packages/paste-design-tokens/tokens/themes/sendgrid/global/background-color.yml
```diff
@@ -10,6 +10,12 @@ props:
   color-background:
     value: "{!palette-gray-20}"
     comment: Default background color for any container
+  color-background-dark:
```
SiTaggart, line 13: Are these not needed anymore? Seem like left overs from a refactor

richbachman: Nope. Just missed those. I'll delete them.

packages/paste-core/primitives/box/src/PseudoPropStyles.ts
```diff
@@ -28,4 +28,12 @@ export const PseudoPropStyles: {[key: string]: string} = {
   _after: '&:after',
   _focusWithin: '&:focus-within',
   _placeholder: '&::placeholder',
+  _child: '& > *',
```
SiTaggart, line 31: Do we need any of these?

packages/paste-core/primitives/box/src/index.tsx
```diff
@@ -67,6 +67,7 @@ interface BaseBoxProps
   backgroundAttachment?: BackgroundAttachmentProperty;
   outline?: OutlineProperty<string>;
   float?: FloatProperty;
+  htmlFor?: string;
```
SiTaggart, line 70: We don't need this anymore I don't think

packages/paste-core/primitives/box/src/index.tsx
```diff
@@ -95,6 +96,13 @@ interface PseudoStylesProps {
   _notFirst?: BaseBoxProps;
   _notLast?: BaseBoxProps;
   _placeholder?: BaseBoxProps;
+  _child?: BaseBoxProps;
```
SiTaggart, line 99: Same here, I don't _think_ we need them.

packages/paste-core/primitives/text/src/index.tsx
```diff
@@ -22,6 +22,7 @@ interface BaseTextProps
   content?: string;
   cursor?: CursorProperty;
   display?: Display;
+  userSelect?: UserSelectProperty;
```
SiTaggart, line 25: Needed anymore?

richbachman: Nope. Removed.

packages/paste-core/primitives/box/src/index.tsx
richbachman: Right. Removed.

packages/paste-core/primitives/box/src/index.tsx
richbachman: Will this come into play when someone wants to create custom radio buttons, using RadioGroup?

packages/paste-core/primitives/box/src/PseudoPropStyles.ts
richbachman: Same, I think we might need this when someone wants to create custom radios, using RadioGroup?

packages/paste-core/primitives/box/src/index.tsx
SiTaggart: We should just export the sibling box at that stage as it's own component, potentially in the box package or something

richbachman: Removed and moved to `SiblingBox`

packages/paste-core/primitives/box/src/PseudoPropStyles.ts
richbachman: Removed and moved to `SiblingBox`

packages/paste-core/components/form/src/Checkbox.tsx
```diff
@@ -0,0 +1,184 @@
...
+        </FormLabel>
+      </Box>
+    );
+  }
+);
+Checkbox.displayName = 'Checkbox';
+
+if (process.env.NODE_ENV === 'development') {
+  Checkbox.propTypes = {
+    children: PropTypes.node.isRequired,
```
richbachman, line 181: Need to add propTypes for:
  ```
checked
hasError
helpText
id
indeterminate
isSelectAll
isSelectAllChild
```

File: packages/paste-core/components/form/src/CheckboxGroup.tsx [hunk 13]
```diff
@@ -0,0 +1,34 @@
...
+              disabled,
+            })
+          : child;
+      })}
+    </InlineControlGroup>
+  );
+};
+
+CheckboxGroup.displayName = 'CheckboxGroup';
+
```
richbachman, line 33: Missing propType for `isSelectAll`.

SiTaggart:
I think this is an oversight. We should bake in error text into the checkbox if it's a valid option. If it's not and errors are only shown at the group level then we should remove the example. 

image

richbachman:
@SiTaggart I think they're only shown at the group level. @corinne-nagel @serifluous will you verify?

packages/paste-core/components/form/src/RadioGroup.tsx
```diff
@@ -0,0 +1,40 @@
...
+        return React.isValidElement(child)
+          ? React.cloneElement(child, {
+              name,
+              onChange: (event: React.ChangeEvent<HTMLInputElement>): void => {
+                if (onChange != null) {
+                  onChange(event.target.value);
+                }
+              },
+              checked: child.props.value === value,
+              disabled,
```
richbachman, line 23: @SiTaggart do we need to pass `hasError` and `required` down to the individual radios and checkboxes at the group level? Similar to what we're doing with `disabled`?

corinne-nagel:
Yep, that'll be the group

packages/paste-core/components/form/src/RadioGroup.tsx
SiTaggart: Potentially for error, maybe? If the group is in an error state, all the options are errored. 

As for required, I don't think so. Just becuase a group of checkboxes has a requirement, doesn't mean _all_ checkboxes should be checked, it might only be one. And it's not really applicable for radios either, as you can only pick one, meaning all the others are unchecked but still required individually.

I actually don't even think it's a thing that should be exposed for a single radio. If it is on their interface we should probably remove it.

SiTaggart:
Error messages are valid for checkboxes though, so we should add that

packages/paste-core/components/form/src/RadioGroup.tsx
richbachman: Just pushed the fix for passing `hasError`. We were never exposing `required`, so we're good there.

PR #399, opened 2020-04-17
feat(select): Select, Option, and OptionGroup

andipants12:
PR for `<Select />`, `<Option />`, and `<OptionGroup />`

**Contributing to Twilio**

Tasks:
- [x] Create components (styled where appropriate) for: `<Select />`, `<OptionGroup />`, and `<Option />`
- [x] Create stories with Knobs (titled "Select Options")
- [x] Create stories showing each render state
- [x] Complete unit tests


packages/paste-core/components/form/src/select/Select.tsx
```diff
@@ -0,0 +1,121 @@
+import * as React from 'react';
+import * as PropTypes from 'prop-types';
+import styled from '@emotion/styled';
+import css from '@styled-system/css';
+import ChevronDown from '@twilio-paste/icons/svg/ChevronDown.svg';
```
richbachman, line 5: We use a library to convert our icon svgs into react components. Changing the import path to: `@twilio-paste/icons/esm/ChevronDownIcon` will point to the react component version.

packages/paste-core/components/form/src/select/Select.tsx
```diff
@@ -0,0 +1,121 @@
...
+import {safelySpreadFormControlProps} from '../shared/Utils';
+
+export interface SelectProps extends React.SelectHTMLAttributes<HTMLSelectElement> {
+  id: string;
+  className?: never;
+  disabled?: boolean;
+  required?: boolean;
+  value?: string; // returns empty string by default if no value is selected
+  type?: SelectInputType;
+  helpText?: string; // not sure if this one is needed...
```
richbachman, line 19: I don't think we need the `helpText` here. We can use `FormHelpText`, which it looks like you're doing in the stories.

andipants12: Ah yes, that makes sense. I think I need to clean up some of these other props too. We chatted briefly about this in the doc I shared with some questions, but are there any others here I can omit from the interface? I'm thinking: `type`, `options`, and `label`?

andipants12: @SiTaggart Any input here as well? ☝️

packages/paste-core/components/form/src/select/Select.tsx
andipants12: Is that compatible with background image? I think I need to use an svg for the css background, or is there a way to use the icon?

packages/paste-core/components/form/src/select/Select.tsx
richbachman: I think we can drop `type` and `label`.

packages/paste-core/components/form/src/select/Select.tsx
richbachman: Hmm. Right. That might cause some issues. Let me think about that one a bit.

andipants12:
FYI @richbachman @SiTaggart here is the source code after my code review with Simon this morning. I plan to start on unit tests next. Thank you! 😸

packages/paste-core/components/form/test-utils/index.ts
```diff
@@ -0,0 +1,29 @@
...
+  draggable?: boolean;
+  accesskey?: boolean;
+  autocomplete?: boolean;
+  form?: string;
+}
+
+export const createAttributeMap = (element: HTMLElement): AttributesMap => {
+  const attributesMap = {};
+  const attributesNodeList: NamedNodeMap = element.attributes;
+  // eslint-disable-next-line unicorn/no-for-loop
```
andipants12, line 22: FYI I was getting two conflicting lint errors so I disabled this line. This lint rule required the use of for-in iterator instead of a loop, but after changing to use that iterator, another lint error came up instructing to _not_ use iterators, but instead use a loop.

packages/paste-core/components/form/src/select/Select.tsx
```diff
@@ -0,0 +1,99 @@
...
+        size={props.multiple ? size : 0}
+        autoFocus={!!props.autoFocus}
+        {...safelySpreadFormControlProps(props)}
+      >
+        {children}
+      </SelectElement>
+      {insertAfter && <Suffix>{insertAfter}</Suffix>}
+      <SelectIconWrapper>
+        <ChevronDownIcon decorative aria-hidden="true" />
+      </SelectIconWrapper>
```
richbachman, line 84: We'll need to adjust the right position of the `SelectIconWrapper` when the `insertAfter` prop is used. Or maybe wrap the `SelectElement` and `SelectIconWrapper` in another container so `insertAfter` doesn't affect the right positioning.

andipants12: Also, I definitely forgot to include a conditional render here for when the `multiple` prop is used. Will make both adjustments. Thank you!

packages/paste-core/components/form/__tests__/select/option.test.tsx
```diff
@@ -0,0 +1,78 @@
...
+const MockOption: React.FC<MockOptionProps> = ({suffix = 'test', ...props}): React.ReactElement => {
+  return (
+    <Option data-testid={`option-${suffix}`} value={'option-1'} {...props}>
+      Option 1
+    </Option>
+  );
+};
+
+describe('Form | Option', () => {
+  afterEach(cleanup);
```
richbachman, line 22: You can drop the `afterEach(cleanup)` on the tests. Jest handles that: https://testing-library.com/docs/react-testing-library/api#cleanup

packages/paste-core/components/form/src/select/Select.tsx
```diff
@@ -0,0 +1,109 @@
...
+      color: 'colorTextWeak',
+    },
+  })
+);
+
+const Select = React.forwardRef<HTMLSelectElement, SelectProps>(
+  ({disabled, hasError, insertBefore, insertAfter, children, size, multiple, ...props}, ref) => (
+    <FieldWrapper disabled={disabled} hasError={hasError}>
+      {insertBefore && <Prefix>{insertBefore}</Prefix>}
+      <Box display="flex" width="100%">
```
andipants12, line 78: @richbachman added this to the `<Select />` component, is this what you were thinking, or would you prefer it be another separate component?

packages/paste-core/components/form/src/select/Select.tsx
SiTaggart: are these the correct way round? should the suffix come after the chevron?

packages/paste-core/components/form/src/select/Select.tsx
SiTaggart: If this wrapper is relatively positioned, you shouldn't need to know whether or not the insertAfter is present. The absolute position is always based on this Box.

andipants12: @SiTaggart I'm using absolute positioning here because I noticed that if I used relative, the click event wasn't propagating correctly. In the commit I'm about to push I added normalizing spacing for the "right" value with absolute positioning, and gave the icon size 20 padding on the right. Is that an acceptable work around?

SiTaggart: Then you do a little 

```
top: 50%
transform: translateY(-50%)
```
on the icon wrapper and it's vertically centered

packages/paste-core/components/form/src/select/Select.tsx
```diff
@@ -0,0 +1,109 @@
...
+  value: string | string[];
+}
+
+export interface SelectIconWrapperProps {
+  ref?: string;
+  insertAfter?: React.ReactNode;
+  children: React.ReactNode;
+}
+
+const SelectIconWrapper = React.forwardRef<HTMLSelectElement, SelectIconWrapperProps>(
```
SiTaggart, line 27: You can skip forwarding the ref here, there's no need for it honestly

packages/paste-core/components/form/src/select/Select.tsx
```diff
@@ -0,0 +1,109 @@
...
+          ref={ref}
+          multiple={multiple}
+          size={multiple ? size : 0}
+          {...safelySpreadFormControlProps(props)}
+        >
+          {children}
+        </SelectElement>
+        {!multiple && (
+          <SelectIconWrapper insertAfter={insertAfter}>
+            <ChevronDownIcon decorative aria-hidden="true" />
```
SiTaggart: `decorative` should add `aria-hidden` for you

packages/paste-core/components/form/src/shared/Utils.tsx
```diff
@@ -1,9 +1,9 @@
-const PROPS_TO_BLOCK = ['className', 'style', 'size', 'height', 'width'];
+export const PROPS_TO_BLOCK: string[] = ['className', 'style', 'size', 'height', 'width'];
 
-export const safelySpreadFormControlProps = (props: {}): {} => {
+export const safelySpreadFormControlProps = (props: {}, propsToBlock: string[] = PROPS_TO_BLOCK): {} => {
```
SiTaggart, line 3: Going to end up suggesting what I suggested for the Button component this week. The props list is so small, lets just unset set them on the select and save the look up. e.g.

```
<SelectElement   
  {...props}
  className={undefined}
  style={undefined}       
  height={undefined}
  width={undefined}
  aria-invalid={hasError}
  ref={ref}
  multiple={multiple}
  size={multiple ? size : 0}
>

File: packages/paste-core/components/form/stories/select.stories.tsx [hunk 10]
```diff
@@ -0,0 +1,315 @@
...
+          <OptionGroup label="Group 2">
+            <Option value="option-3">Option 3</Option>
+            <Option value="option-4">Option 4</Option>
+          </OptionGroup>
+        </Select>
+        <FormHelpText>Info that helps a user with this field.</FormHelpText>
+      </>
+    );
+  })
+  .add('Select - Option Groups and Multiple', () => {
```
SiTaggart, line 259: Running this story local, I can't seem to select an option.

packages/paste-core/components/form/test-utils/index.ts
SiTaggart: If you didn't want to do that, you could do something like

```
Object.keys(attributesNodeList)
  .map((item) => attributesNodeList[item])
  .map(attr => attributesMap[attr.name] = attr.value)
```

packages/paste-core/components/form/src/select/Select.tsx
andipants12: Yeah I noticed that in the Chakra implementation, so then I can leave this as `position: absolute` and it's preferable to use the above styles?

packages/paste-core/components/form/src/shared/Utils.tsx
andipants12: Ok great, that makes sense. Will make this change.

packages/paste-core/components/form/src/select/Select.tsx
andipants12: I'm guessing the use of a computed `transform` here is more reliable than hard-coding the right spacing value?

packages/paste-core/components/form/stories/select.stories.tsx
andipants12: Yeah I just noticed that too, I broke something with the `onChange` handler for the story when I was adjusting the type interface. Will fix. Thank you for pointing this out!

packages/paste-core/components/form/src/select/Select.tsx
SiTaggart: No, the icon wrapper is absolute to the `Box`, relatively positioned wrapper. Then you can get rid of the top padding on the icon wrapper. 

Also curious why you are using right padding and right positioning together? Can it just be right positioning?

Also from what I can see the pointer events still passes through to the underlying select element in FF and chrome locally.

packages/paste-core/components/form/test-utils/index.ts
andipants12: Ah interesting, I like the work-around. I wonder how the TypeScript will like this... time to investigate!

packages/paste-core/components/form/src/select/Select.tsx
```diff
@@ -0,0 +1,109 @@
...
+      color: 'colorTextWeak',
+    },
+  })
+);
+
+const Select = React.forwardRef<HTMLSelectElement, SelectProps>(
+  ({disabled, hasError, insertBefore, insertAfter, children, size, multiple, ...props}, ref) => (
+    <FieldWrapper disabled={disabled} hasError={hasError}>
+      {insertBefore && <Prefix>{insertBefore}</Prefix>}
+      <Box display="flex" width="100%">
```
SiTaggart, line 78:
```suggestion
      <Box display="flex" width="100%" position="relative">
```

packages/paste-core/components/form/src/select/Select.tsx
```diff
@@ -0,0 +1,109 @@
...
+  ({children, insertAfter, ...props}, ref) => (
+    <Box
+      alignItems="center"
+      display="inline-flex"
+      paddingRight="space20"
+      paddingTop="space30"
+      position="absolute"
+      pointerEvents="none"
+      ref={ref}
+      right={insertAfter ? 'space140' : 'space40'}
```
SiTaggart, line 37:
```suggestion
      position="absolute"
      pointerEvents="none"
      ref={ref}
      right="space40"
      top="50%"
      transform="translateY(-50%)"
```

packages/paste-core/components/form/src/select/Option.tsx
```diff
@@ -0,0 +1,41 @@
...
+}
+
+const OPTION_PROPS_TO_BLOCK = [...PROPS_TO_BLOCK, 'selected'];
+
+const OptionElement = styled.option(
+  css({
+    background: 'transparent',
+    color: 'colorText',
+  })
+);
```
SiTaggart, line 19: Was running this locally and gave the options a little more space to breath with 

```
const OptionElement = styled.option(
  css({
    background: 'transparent',
    color: 'colorText',
    fontFamily: 'fontFamilyText',
    lineHeight: 'lineHeight50',
    paddingTop: 'space30',
    paddingBottom: 'space30',
    paddingLeft: 'space20',
  })
);
```
image

packages/paste-core/components/form/src/select/OptionGroup.tsx
```diff
@@ -0,0 +1,25 @@
...
+import * as PropTypes from 'prop-types';
+import {safelySpreadFormControlProps} from '../shared/Utils';
+
+export interface OptionGroupProps extends React.OptgroupHTMLAttributes<HTMLOptGroupElement> {
+  label: string;
+  children: React.ReactNode;
+}
+
+const OptionGroup = React.forwardRef<HTMLOptGroupElement, OptionGroupProps>(({label, children, ...props}, ref) => (
+  <optgroup ref={ref} label={label} {...safelySpreadFormControlProps(props)}>
```
SiTaggart, line 11: Also discovered you can give these _some_ styles, but limited 

```
const OptionGroupElement = styled.optgroup(
  css({
    background: 'transparent',
    color: 'colorText',
    fontFamily: 'fontFamilyText',
    fontWeight: 'fontWeightMedium',
    fontStyle: 'normal',
    margin: 'space20',
  })
);```

![image](https://user-images.githubusercontent.com/368249/80732992-f4497f00-8ac1-11ea-9525-72e30e537870.png)

File: packages/paste-core/components/form/src/shared/Utils.tsx (hunk 9 above)
Comment (andipants12) [reply packages/paste-core/components/form/src/shared/Utils.tsx:3]: @SiTaggart Just checked out `<Button />`, any benefit to adding [similar PropType validation](https://github.com/twilio-labs/paste/blob/master/packages/paste-core/components/button/src/index.tsx#L7)? Or is this too much oversight for this component?

Comment (richbachman) [reply packages/paste-core/components/form/src/shared/Utils.tsx:3]: We probably don't need to worry about doing the PropType validation here.

Comment (andipants12) [reply packages/paste-core/components/form/src/shared/Utils.tsx:3]: @richbachman not even for `this.props.children`?

File: packages/paste-core/components/form/src/shared/restricted-attributes.ts [hunk 15]
```diff
@@ -0,0 +1,16 @@
+export interface RestrictedPropsTypes {
+  className: undefined;
+  size: undefined;
+  style: undefined;
+  height: undefined;
+  width: undefined;
+  selected?: undefined;
+}
+
+export const restrictedProps: RestrictedPropsTypes = {
```
andipants12, line 10: @SiTaggart @richbachman added this here to reduce redundancy.

packages/paste-core/components/form/src/shared/Utils.tsx
richbachman: You could add `children: NonNullable<React.ReactNode>;` to the `SelectProps`. Check out [Radio](https://github.com/twilio-labs/paste/blob/master/packages/paste-core/components/form/src/Radio.tsx#L15) for reference.

packages/paste-core/components/form/src/select/Select.tsx
andipants12: Ah ok cool, I can remove this too.

packages/paste-core/components/form/src/shared/restricted-attributes.ts
SiTaggart: Nice 👍

packages/paste-core/components/form/test-utils/index.ts
```diff
@@ -0,0 +1,26 @@
...
+  draggable?: boolean;
+  accesskey?: boolean;
+  autocomplete?: boolean;
+  form?: string;
+}
+
+export const createAttributeMap = (element: HTMLElement): AttributesMap => {
+  const attributesNodeList: NamedNodeMap = element.attributes;
+
+  return Object.keys(attributesNodeList).reduce((attributesMap, key) => {
```
SiTaggart, line 22: Yeah, this is much better than my version 😉

packages/paste-core/components/form/__tests__/select/option.test.tsx
```diff
@@ -0,0 +1,76 @@
...
+    expect(attributeMap.title).toEqual('test-title');
+    expect(attributeMap.spellcheck).toEqual('true');
+    expect(attributeMap.hidden).toEqual('');
+    expect(attributeMap.draggable).toEqual('true');
+    expect(attributeMap.accesskey).toEqual('t e s t');
+    expect(attributeMap.disabled).toEqual('');
+    expect(attributeMap.label).toEqual('option-1-label');
+  });
+
+  it('should filter blacklisted props via safelySpreadFormControlProps', () => {
```
SiTaggart, line 47: Just need to update these test names as you're not using safelySpread anymore. Keep the test though

andipants12: Ohhhh yes, thank you for the reminder!

packages/paste-core/components/form/test-utils/index.ts
andipants12: 😸

packages/paste-core/components/form/src/shared/Utils.tsx
andipants12: Just added that, thank you!

PR #400, opened 2020-04-17
fix(tokens): Add shadow-border-input-* aliases, tokens to themes 

serifluous:
Add missing shadow-border-input-* aliases so they call console theme colors rather than default theme colors.


lerna.json
```diff
@@ -5,6 +5,7 @@
   "useWorkspaces": true,
   "command": {
     "publish": {
+      "ignoreChanges": ["**/stories/**", "**/__tests__/**"],
```
SiTaggart, line 8: preventing lerna from thinking changes to tests and stories warrant a publish of source code.

PR #405, opened 2020-04-23
fix(heading): set variant as required, remove default variant

richbachman:
- [x] Breaking Change(heading): Set `variant` as required for `Heading`
- [x] Removed default `variant` prop
- [x] Updated `Heading` prop table on website
- [x] Added missing variants when used on website


richbachman:
@SiTaggart our failed prop type console warning is thrown, and the default headingVariant switch statement kicks in:
```
default:
      return {
        marginBottom: marginBottom || 'space60',
        fontSize: 'fontSize70',
        lineHeight: 'lineHeight70',
      };
```

image

Probably still a breaking change since this could affect UI. Right?

SiTaggart:
Yeah, lets mark it as breaking, but the fact it doesn't crash and there are some styles on the page makes me feel less concerned.

Issue #368, opened 2020-04-01
<Button /> Allow more generic data attributes other than `data-test` 


ghostling:
We're already using `data-s` as selectors in our unit & e2e tests and would like to keep doing so. 

Noticed that the <Button /> component [only allows `data-test` attribute](https://github.com/twilio-labs/paste/blob/master/packages/paste-core/components/button/src/index.tsx#L97). This is oddly specific/restrictive.

SiTaggart:
Yeah, it is, you're right. The button was one of our first components and since then we've learnt a bunch. 

We generally allow for any applicable global html attribute now on alot of our components, so we should do the same here.

I'll add this to the backlog. Thanks for calling this out.

richbachman:
Hi @ghostling, we just merged a fix for `Button` that allows all standard button attributes, so `data-s` will work.

PR #406, opened 2020-04-24
fix(button): allow html button attributes

richbachman:
Fixes https://github.com/twilio-labs/paste/issues/368

- [x] Allowed all HTML button attributes
- [x] Added `data-foo` test
- [x] Added safely spread utility
- [x] Updated website button props table

### DIFF
--- a/packages/paste-core/components/button/src/index.tsx
+++ b/packages/paste-core/components/button/src/index.tsx
@@ -2,20 +2,27 @@ import * as React from 'react';
 import * as PropTypes from 'prop-types';
 import {Spinner} from '@twilio-paste/spinner';
 import {ButtonWrapper, ButtonChildren, SpinnerWrapper} from './styles';
-import {ButtonProps, ButtonStates} from './types';
+import {ButtonProps, ButtonStates, ButtonVariants, ButtonSizes, ButtonTabIndexes} from './types';
 
-const handlePropValidation = (props: ButtonProps): void => {
-  const {as: Tag, href, variant, size, tabIndex, fullWidth, children} = props;
+const handlePropValidation = (
+  children: React.ReactNode,
+  variant: ButtonVariants,
+  as?: string,
+  fullWidth?: boolean,
+  href?: string,
+  size?: ButtonSizes,
+  tabIndex?: ButtonTabIndexes
+): void => {
   const hasHref = href != null && href !== '';
   const hasTabIndex = tabIndex != null;
 
-  if (Tag !== 'a' && hasHref) {
+  if (as !== 'a' && hasHref) {
     throw new Error(`[Paste: Button] You cannot pass href into a button without the 'a' tag.  Use 'as="a"'.`);
   }
-  if (Tag === 'a' && !hasHref) {
+  if (as === 'a' && !hasHref) {
     throw new Error(`[Paste: Button] Missing href prop for link button.`);
   }
-  if (Tag === 'a' && variant === 'link') {
+  if (as === 'a' && variant === 'link') {
     throw new Error(`[Paste: Button] This should be a link. Use the [Paste: Anchor] component.`);
   }
   if (variant === 'reset' && size !== 'reset') {
@@ -32,7 +39,7 @@ const handlePropValidation = (props: ButtonProps): void => {
   }
 };
 
-const getButtonState = ({loading, disabled}: ButtonProps): ButtonStates => {
+const getButtonState = (disabled?: boolean, loading?: boolean): ButtonStates => {
   if (disabled) {
     return 'disabled';
   }
@@ -42,23 +49,34 @@ const getButtonState = ({loading, disabled}: ButtonProps): ButtonStates => {
   return 'default';
 };
 
-const Button: React.FC<ButtonProps> = props => {
-  const buttonState = getButtonState(props);
+const Button: React.FC<ButtonProps> = ({
+  as,
+  children,
+  disabled,
+  fullWidth,
+  href,
+  loading,
+  size,
+  tabIndex,
+  variant,
+  ...props
+}) => {
+  const buttonState = getButtonState(disabled, loading);
   const showLoading = buttonState === 'loading';
-  const disabled = buttonState !== 'default';
+  const showDisabled = buttonState !== 'default';
 
   // If size isn't passed, come up with a smart default:
   // - 'reset' for variant 'link'
   // - 'icon' if there's 1 child that's an icon
   // - 'default' otherwise
-  let defaultSize = props.size;
+  let defaultSize = size;
   if (defaultSize == null) {
     defaultSize = 'default';
 
-    if (props.variant === 'link' || props.variant === 'destructive_link') {
+    if (variant === 'link' || variant === 'destructive_link') {
       defaultSize = 'reset';
-    } else if (React.Children.count(props.children) === 1) {
-      React.Children.forEach(props.children, child => {
+    } else if (React.Children.count(children) === 1) {
+      React.Children.forEach(children, child => {
         if (React.isValidElement(child)) {
           // @ts-ignore
           if (typeof child.type.displayName === 'string' && child.type.displayName.includes('Icon')) {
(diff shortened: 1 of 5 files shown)


packages/paste-core/components/button/src/utils.tsx
```diff
@@ -0,0 +1,13 @@
+const PROPS_TO_BLOCK = ['className', 'style', 'height', 'width', 'loading'];
```
SiTaggart, line 1: As this is such a small list of props (of which I'm not sure loading should be included) I'm tempted to just set these properties to null on the button itself rather than filter them out.

richbachman: I like this idea better. I added `loading` here because when I added the safely spread function I was getting the following console error:

image

It may have something to do with the loading html attribute (https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes):
image

SiTaggart: This might just be something to do with the way the storybook Boolean knob is working

richbachman: Similar error when running tests:
image

SiTaggart: OK, so we don't actually use `loading` spread to the button element. You're going to have to rework the way the props come in and selective destructure some off of props before spreading.

You'll need to pull loading and disabled off of props for example, do to the way we handle them as states, not attributes on the button element. See getButtonStates(props) on line 46ish

packages/paste-core/components/button/src/types.ts
```diff
@@ -4,9 +4,12 @@ export type ButtonVariants = 'primary' | 'secondary' | 'destructive' | 'destruct
 export type ButtonStates = 'disabled' | 'loading' | 'default';
 export type ButtonTabIndexes = 0 | -1;
 
-export interface ButtonProps {
+export interface ButtonProps extends React.ButtonHTMLAttributes<HTMLButtonElement> {
   id?: never;
   className?: never;
+  style?: never;
+  width?: never | null;
+  height?: never | null;
```
SiTaggart, line 12: Remove all these `never` props on the interface it does nothing for us.

If anything its more confusing as it’ll try to auto complete the attribute, and then tell you once it’s done that you can’t use it.

That’ll at least hopefully solve some of the typing issues with classname and style. Setting them to null shouldn’t clash with the html types, unless it expects `undefined`

packages/paste-core/components/button/src/index.tsx
```diff
@@ -42,23 +42,43 @@ const getButtonState = ({loading, disabled}: ButtonProps): ButtonStates => {
...
+  onMouseEnter,
+  onMouseLeave,
+  onMouseUp,
+  size,
+  tabIndex,
+  type,
+  variant,
+  ...props
+}) => {
+  const buttonState = getButtonState({loading, disabled, variant, children});
```
SiTaggart, line 66: For these, lets not re-wrap them in an object. Let’s just pass comma separated arguments and modify the functions to not use `props` (or what ever it uses)

packages/paste-core/components/button/src/index.tsx
```diff
@@ -69,34 +89,40 @@ const Button: React.FC<ButtonProps> = props => {
     }
   }
 
-  handlePropValidation(props);
+  handlePropValidation({as, href, variant, size, tabIndex, fullWidth, children});
```
SiTaggart, line 92: Same here, comma separated list of arguments and modify the function

packages/paste-core/components/button/src/index.tsx
```diff
@@ -69,34 +89,40 @@ const Button: React.FC<ButtonProps> = props => {
...
-      onFocus={props.onFocus}
-      onBlur={props.onBlur}
+      aria-busy={buttonState === 'loading' ? 'true' : 'false'}
+      aria-controls={props['aria-controls']}
       aria-expanded={props['aria-expanded']}
       aria-haspopup={props['aria-haspopup']}
-      aria-controls={props['aria-controls']}
-      aria-busy={buttonState === 'loading' ? 'true' : 'false'}
-      data-test={props['data-test']}
+      as={as}
```
SiTaggart, line 100: For any of these that we actually don’t use in the component scope or logic and just passing down, lets just rely on the spreading of props instead of destructuring them off.

It’ll make this look a little tidier

packages/paste-core/components/button/src/types.ts
```diff
@@ -4,9 +4,12 @@ export type ButtonVariants = 'primary' | 'secondary' | 'destructive' | 'destruct
 export type ButtonStates = 'disabled' | 'loading' | 'default';
 export type ButtonTabIndexes = 0 | -1;
 
-export interface ButtonProps {
+export interface ButtonProps extends React.ButtonHTMLAttributes<HTMLButtonElement> {
```
SiTaggart, line 7: Because we’re now extending the button attributes into our interface, some of these declared props can probably just be removed as they would be present on the native. 

Unless we override anything with that prop, like make it required, we can just safely rely on the button element interface. We’re trying to make the button feel like a native HTML button after all. 

I’d go as far as to remove any opinions we have for things like `aria-haspopup` and just rely on the button element interface

packages/paste-core/components/button/src/types.ts
richbachman: Looks like they do want `undefined `. That seems to still work if I try to pass anything to either `className` or `style`. I removed `width` and `height` since they're not valid for button anyway.

packages/paste-core/components/button/src/index.tsx
richbachman: Restructured. Let me know if I have that setup correctly.

packages/paste-core/components/button/src/index.tsx
richbachman: Same. Restructured. Let me know if I have that setup correctly.

packages/paste-core/components/button/src/types.ts
richbachman: Cleaned up the props based on what we get from the native HTML button attributes. Let me know if you see anything I missed.

packages/paste-core/components/button/src/index.tsx
```diff
@@ -108,38 +116,20 @@ const Button: React.FC<ButtonProps> = props => {
 
 Button.defaultProps = {
   as: 'button',
-  type: 'button',
```
SiTaggart, line 111: We may want to keep the default type as button still

packages/paste-core/components/button/src/index.tsx
```diff
@@ -42,23 +49,34 @@ const getButtonState = ({loading, disabled}: ButtonProps): ButtonStates => {
...
+  loading,
+  size,
+  tabIndex,
+  variant,
+  ...props
+}) => {
+  const buttonState = getButtonState(disabled, loading);
   const showLoading = buttonState === 'loading';
-  const disabled = buttonState !== 'default';
+  const showDisabled = buttonState !== 'default';
```
SiTaggart, line 66: Keeping this as `disabled` would have been fine, it's less about showing disabled, but actually setting the disabled attribute

packages/paste-core/components/button/src/types.ts
```diff
@@ -1,42 +1,19 @@
...
-  variant: ButtonVariants;
-  size?: ButtonSizes;
-  fullWidth?: boolean;
-
   children: React.ReactNode;
-
-  disabled?: boolean;
+  fullWidth?: boolean;
+  href?: string;
+  id?: never;
```
SiTaggart, line 12: I think we should probably allow id's honestly.

packages/paste-website/src/pages/components/button/index.mdx
```diff
@@ -506,7 +508,6 @@ import {Button} from '@twilio-paste/button';
 | aria-expanded? | boolean                                  | A11y: For accordions                                                                | null      |
```
SiTaggart, line 508: Probably need to tidy up some of these now too

packages/paste-core/components/button/src/index.tsx
richbachman: Right. I renamed that because of the conflict with the destructured disabled prop, but since we're not passing that I'll change it back.

packages/paste-core/components/button/src/types.ts
richbachman: Pulled `id` out since we get it with the html attributes.

packages/paste-core/components/button/src/index.tsx
richbachman: Added back `type` prop.

PR #407, opened 2020-04-24
feat(website): add reset button example to button doc page

richbachman:
- [x] Added unset button example to button doc page

### DIFF
--- a/packages/paste-website/src/pages/components/button/index.mdx
+++ b/packages/paste-website/src/pages/components/button/index.mdx
@@ -285,6 +285,19 @@ their default size.
 </div>`}
 </LivePreview>
 
+#### Reset button styles
+
+The reset button is basically a primitive button. The button styles are
+reset, so styles won't interfere with any child elements. This allows you to
+create your own buttons using the Paste. This reset variant only resets
+styles. No other button functionality is reset.
+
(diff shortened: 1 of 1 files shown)


richbachman:
@serifluous would you mind weighing in on the "unset" vs "reset"? Thanks.

serifluous:
Can we make the heading "Unset button styles", "Reset button styles" or something similar?

I added the AC in the ticket to not call it "reset button" because that could be confusing/misleading.

SiTaggart:
I'm fine with "Reset button styles", it's just odd to have both props you have to set needing a value of `reset` but we don't call it that in the docs. But I might be thinking about it too literally.

If you think it's fine as unset, then just ignore me @serifluous @richbachman

serifluous:
Yeah, "Reset button styles" is fine imo, as long as the whole section is clear that we didn't create a button that resets stuff for consumers in their product.

I wasn't in the grooming where "unset" was chosen, so I don't have a strong opinion on using that word.

richbachman:
@SiTaggart @serifluous I just pushed up some text edits. Will you both take a look and let me know if you have any changes. Thanks.

richbachman:
@SiTaggart rebased with master and package sizes are now unchanged.

PR #409, opened 2020-04-27
fix(modal): reset margin on heading, add zIndex to overlay

richbachman:
- [x] Added margin reset to `ModalHeader`
- [x] Added `zIndex` to `ModalDialogOverlay`

### DIFF
--- a/packages/paste-core/components/modal/src/Modal.tsx
+++ b/packages/paste-core/components/modal/src/Modal.tsx
@@ -18,6 +18,10 @@ const ModalDialogOverlay = styled(ModalDialogPrimitiveOverlay)(
     alignItems: 'center',
     width: '100%',
     backgroundColor: 'colorBackgroundOverlay',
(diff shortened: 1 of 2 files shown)


packages/paste-core/components/modal/src/Modal.tsx
```diff
@@ -18,6 +18,7 @@ const ModalDialogOverlay = styled(ModalDialogPrimitiveOverlay)(
     alignItems: 'center',
     width: '100%',
     backgroundColor: 'colorBackgroundOverlay',
+    zIndex: '2000', // current zIndex tokens only go up to 90
```
SiTaggart, line 21: Can you add a comment just about this line to make it clear it’s there for console navigation reasons, so no one removes it accidentally.

Thanks

richbachman: Yep. Comment added.

PR #422, opened 2020-05-05
fix(stack): extend as prop from BoxProps on StackProps

richbachman:
- [x] Extend `as` prop from `BoxProps`
- [x] Add tests


packages/paste-core/layout/stack/__tests__/stack.test.tsx
```diff
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import {render as testRender} from '@testing-library/react';
+import {render as testRender, screen} from '@testing-library/react';
```
SiTaggart, line 2: Are you using `screen` as all?

SiTaggart: I don't think you need to re assign render as testRender here either

richbachman: Tried to use a `screen` method to test as a `header`, but it wasn't working as I thought. Just forgot to remove it.

packages/paste-core/layout/stack/__tests__/stack.test.tsx
```diff
@@ -99,4 +110,9 @@ describe('Stack', () => {
     const {asFragment} = testRender(<MockResponsiveStack />);
     expect(asFragment()).toMatchSnapshot();
   });
+
+  it('should render as a header', () => {
+    const {asFragment} = testRender(<MockHeaderStack />);
+    expect(asFragment()).toMatchSnapshot();
```
SiTaggart, line 116: I think for these I would like you to assert the element outside of a snapshot. And also provide some tests for arbitary html attributes

richbachman: I changed that test and added a couple more for valid attributes and one for the invalid className attribute. Let me know if I should add more.

packages/paste-core/layout/stack/__tests__/stack.test.tsx
```diff
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import {render} from '@testing-library/react';
+import {render, screen} from '@testing-library/react';
```
SiTaggart, line 2: Still importing screen

packages/paste-core/layout/stack/__tests__/stack.test.tsx
```diff
@@ -112,7 +120,22 @@ describe('Stack', () => {
   });
 
   it('should render as a header', () => {
-    const {asFragment} = render(<MockHeaderStack />);
-    expect(asFragment()).toMatchSnapshot();
+    const {getByTestId} = render(<MockHeaderStack />);
+    expect(getByTestId('header')).not.toBeNull();
```
SiTaggart, line 124: You need to test the element tag name or something here. This only tests that it exists by test id. If I change the as prop, the test will still pass

packages/paste-core/layout/stack/__tests__/stack.test.tsx
richbachman: Finally remove.

packages/paste-core/layout/stack/__tests__/stack.test.tsx
richbachman: Changed that to test for the actual element.

Issue #98, opened 2019-09-16
Sketch friendly values for tokens


seanmateer:
**Problem**
While working in Sketch, I find myself often referring to https://paste.twilio.design/tokens to confirm font-size or spacing sizes. Because these values are in rems, I have to either do mental math or inspect element to get a pixel size out to use in Sketch. 

**Suggestion**
There may be plugins that allow you to use rems in Sketch, but in the spirit of not requiring Sketch plugins, It would be nice to consider also including pixel values here, or perhaps a way to hover and get a pixel reference.

seanmateer:
Re: doing mental math above - I find that I expect that doubling a token value should double the pixel/rem value consistently. I.E. $space-10 (0.125) to $space-20 (0.25) doubles as I would expect, but $space-100(2.25) to $space-200 (4.75) does not. This may be less of a problem if there were pixel values to refer to, but causes some confusion for me.

serifluous:
Thanks, Sean! It's been added to our backlog.

SiTaggart:
Hi @seanmateer @Angelangg and @pguptaDS

Does this do an good enough job of solving this issue? https://paste-git-feat-websitefriendly-token-values.twilio-dsys.now.sh/tokens/

pguptaDS:
Yes, this is great! Thank you.

seanmateer:
Looks good, thanks for adding @SiTaggart!

PR #427, opened 2020-05-06
feat(website): friendly token values

SiTaggart:
- Replace the token filter input with the Paste Input
- add the pixel value to rem tokens on the token page

fixes https://github.com/twilio-labs/paste/issues/98


packages/paste-website/src/components/tokens-list/getTokenValue.ts
```diff
@@ -0,0 +1,15 @@
+const convertToPx = (value: string): string => {
+  return `${parseFloat(value.replace('rem', '')) * 16}px`;
```
richbachman, line 2: Never used `parseFloat`. Had to look that one up.

PR #433, opened 2020-05-07
fix(website): add clarifying text around installing core

SiTaggart:
Adding guidance around installing core and what that does. 

Also adding core install instructions on each component page.

Each component header now has a core install instruction

image

Engineering getting started page now gives a warning about core, and what's included in the core package

image


packages/paste-website/src/components/shortcodes/component-header/index.tsx
```diff
@@ -128,7 +124,8 @@ const ComponentHeader: React.FC<ComponentHeaderProps> = ({
         <Box marginBottom="space20">
           <PackageLabel>Install</PackageLabel>
           <PackageValue>
-            <PackageInstallSnippet>yarn add {packageName}</PackageInstallSnippet>
+            <PackageInstallSnippet>yarn add {packageName}</PackageInstallSnippet> &mdash; or &mdash;{' '}
+            <PackageInstallSnippet>yarn add @twilio-paste/core</PackageInstallSnippet>
```
richbachman, line 128: Should we swap these two? If we want most users to  use Core, maybe we should just put that first.

SiTaggart: Yeah potentially. Two thoughts

1. Does that take away from the fact these are individually versioned component? Does the version number then make sense?
2. I'm trying to avoid changing all the code examples to import from core, too.

richbachman: Right. That would take away from the individual component idea, and cause conflict with the code samples. Let's keep as is.

PR #440, opened 2020-05-12
chore: split the build up for faster CI

SiTaggart:
Make the builds slightly faster. Save roughly 3 minutes?


SiTaggart:
Spot on. Building the components is such a long running process, that even if you build first, cache and share between subsequent jobs, the upload and caching steps remove any gains concurrent remaining tasks have. Even over just sequentially running all jobs one after the other, that finishes before what we had before.

It’s also based on the fact that we were treating “build” as a check, in reality all checks require the build, so there’s not point in having it as a separate job.

In the future I’m considering dropping the type check job too, as it won’t build without correct types. Only thing that won’t cover is website and storybook types which has bitten us before.

richbachman:
That makes sense. Thanks for the explanation. Does the `pr-checks (10.16.x)` check just need to be disabled at this point?

SiTaggart:
Yup, I'll switch the required checks before merging

PR #446, opened 2020-05-14
feat: disclosure primitive and reakit library

TheSisb:
This PR adds the awesome Reakit library (https://reakit.io/docs/) as a wrapped package in our system, and creates the disclosure primitive component wrapper around their primitive export.


packages/paste-libraries/reakit/package.json
```diff
@@ -0,0 +1,38 @@
+{
+  "name": "@twilio-paste/reakit-library",
+  "version": "0.0.0",
+  "category": "library",
+  "status": "backlog",
+  "private": true,
```
richbachman, line 6: Remove the `"private" : true,` here.

packages/paste-core/primitives/disclosure/README.md
```diff
@@ -0,0 +1,104 @@
+# Disclosure Primitive
```
richbachman, line 1: We've fallen out of the habit of adding a README for the individual component packages. Looks like there are only a couple that have them right now. Thoughts on each package needing one.

packages/paste-core/primitives/disclosure/stories/index.stories.tsx
```diff
@@ -0,0 +1,89 @@
...
+    <>
+      <Heading as="h3" variant="heading30">
+        Basic Example{' '}
+        <sup>
+          <Anchor href="https://reakit.io/docs/disclosure/#usage">1</Anchor>
+        </sup>
+      </Heading>
+      <Separator orientation="horizontal" />
+
+      <DisclosurePrimitive {...disclosure}>Toggle</DisclosurePrimitive>
```
richbachman, line 19: Should we use `as` our Paste Button on these examples?

TheSisb: Hmm.. If I do that, what would be left to do for the component Disclosure? 🤔

richbachman: Ha. True. Is that the only difference?

packages/paste-core/core-bundle/package.json
```diff
@@ -35,6 +35,7 @@
     "@twilio-paste/button": "^1.1.45",
     "@twilio-paste/card": "^1.3.30",
     "@twilio-paste/combobox-primitive": "^0.1.0",
+    "@twilio-paste/disclosure-primitive": "^0.0.2",
```
SiTaggart, line 38: You need to add `@twilio-paste/reakit-library` as a peer dependency to core. Right now it’s bundling it into core, hence the size increase of 3kb or so. If it’s a peer you should see that size drop.

Eventually we’d create a bundle like core, for the libraries so we don’t need to do this for every library package.

packages/paste-core/primitives/disclosure/README.md
SiTaggart: I tended not to care too much as the website package page should basically tell you the same information. 

Feels dangerously redundant too as they get out of sync real quick.

What would be cool long term might be the package read me _is_ the website page, but that would require a little more work to dynamically generate some common stuff on the component pages.

packages/paste-core/primitives/disclosure/stories/index.stories.tsx
SiTaggart: I tend to agree with Rich here, but I think the “how do I customise this” example should live in the website docs page, and not on the stories

packages/paste-libraries/reakit/rollup.config.js
```diff
@@ -0,0 +1,38 @@
...
+    {
+      file: pkg.module,
+      format: 'esm',
+    },
+  ],
+  external: [...Object.keys(pkg.peerDependencies || {})],
+  plugins: [
+    resolve(),
+    commonjs({
+      namedExports: {
```
SiTaggart, line 24: Just double checking this isn’t copy pasta from dropdown. You may not need too declare any namedExports

TheSisb: It is 🤣 Will fix.

SiTaggart:
I realize I didn't hit send on this yesterday... 🙄 
@TheSisb Can you rewrite the history on this to one commit per package

PR #447, opened 2020-05-14
feat: add paste animation library

TheSisb:
Adds react-spring v9 as our animation library of choice, as the `@twilio-paste/animation` library.  **Beta**, no docs until we're completely satisfied with this choice.


packages/paste-libraries/animation/tsconfig.build.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": "../../tsconfig.json",
```
SiTaggart, line 2: I think this path just needs another `../` to get the type check passing.

PR #449, opened 2020-05-15
feat(combobox): add combobox package

richbachman:
- [x] Combobox package
- [x] export styled InputElement, add FormControlWrapper

### DIFF
--- a/packages/paste-core/components/combobox/src/Combobox.tsx
+++ b/packages/paste-core/components/combobox/src/Combobox.tsx
@@ -1,0 +1,149 @@
+import * as React from 'react';
+import * as PropTypes from 'prop-types';
+import styled from '@emotion/styled';
+import css from '@styled-system/css';
+import {uid, useUID} from 'react-uid';
+import {useComboboxPrimitive, UseComboboxPrimitiveProps} from '@twilio-paste/combobox-primitive';
+import {ChevronDownIcon} from '@twilio-paste/icons/esm/ChevronDownIcon';
+import {Box} from '@twilio-paste/box';
+import {FormControlWrapper, FormHelpText, FormLabel, InputElement, FormInputProps} from '@twilio-paste/form';
+import {ComboboxInputWrapper} from './ComboboxInputWrapper';
+import {ComboboxListbox} from './ComboboxListbox';
+import {ComboboxListboxOption} from './ComboboxListboxOption';
+
+const ComboboxIconWrapper: React.FC = ({children}) => (
+  <Box
+    alignItems="center"
+    display="inline-flex"
+    position="absolute"
+    pointerEvents="none"
+    right="space40"
+    top="50%"
+    transform="translateY(-50%)"
+    zIndex="zIndex10"
+  >
+    {children}
+  </Box>
+);
+
+// Fixes chevron overlapping really long text
+/* eslint-disable emotion/syntax-preference */
+const StyledInputAsSelect = styled(InputElement)(
+  css({
+    paddingRight: 'space120',
+  })
+);
+/* eslint-enable */
+
+export interface ComboboxProps extends Omit<FormInputProps, 'id' | 'type' | 'value'> {
+  autocomplete?: boolean;
+  helpText?: string | React.ReactNode;
+  items: UseComboboxPrimitiveProps<any>['items'];
+  labelText: string | NonNullable<React.ReactNode>;
+  onInputValueChange?: UseComboboxPrimitiveProps<any>['onInputValueChange'];
+  itemToString?: UseComboboxPrimitiveProps<any>['itemToString'];
+  optionTemplate?: (item: string | {}) => React.ReactNode;
+}
+
+const defaultItemToString = (item: string): string => item;
+
+const Combobox = React.forwardRef<HTMLInputElement, ComboboxProps>(
+  (
+    {
+      autocomplete,
+      disabled,
+      hasError,
+      helpText,
+      insertAfter,
+      insertBefore,
+      items,
+      itemToString,
+      labelText,
+      optionTemplate,
+      onInputValueChange,
+      required,
+      ...props
+    },
+    ref
+  ) => {
+    const {
+      getComboboxProps,
+      getInputProps,
+      getItemProps,
+      getLabelProps,
+      getMenuProps,
+      getToggleButtonProps,
+      highlightedIndex,
+      isOpen,
+    } = useComboboxPrimitive({
+      items,
+      onInputValueChange,
+      itemToString: itemToString || defaultItemToString,
+    });
+    const helpTextId = useUID();
+    return (
+      <Box position="relative">
+        <FormLabel disabled={disabled} required={required} {...getLabelProps()}>
+          {labelText}
+        </FormLabel>
+        <FormControlWrapper
+          disabled={disabled}
+          hasError={hasError}
+          insertBefore={insertBefore}
+          insertAfter={insertAfter}
+        >
+          <ComboboxInputWrapper {...getComboboxProps({role: 'combobox'})}>
+            <StyledInputAsSelect
+              {...getInputProps({disabled})}
+              {...getToggleButtonProps({tabIndex: 0})}
+              {...(!autocomplete ? {onChange: event => event.preventDefault()} : undefined)}
+              aria-describedby={helpTextId}
+              {...props}
+              type="text"
+              ref={ref}
+              paddingRight="space90"
+            />
+            <ComboboxIconWrapper>
+              <ChevronDownIcon size="sizeIcon30" decorative aria-hidden="true" />
+            </ComboboxIconWrapper>
+          </ComboboxInputWrapper>
+        </FormControlWrapper>
+        <ComboboxListbox {...getMenuProps()}>
+          {isOpen &&
+            items.map((item, index) => {
+              return (
+                <ComboboxListboxOption
+                  {...getItemProps({item, index})}
+                  highlighted={highlightedIndex === index}
+                  key={uid(item)}
+                >
+                  {optionTemplate ? optionTemplate(item) : item}
+                </ComboboxListboxOption>
+              );
+            })}
+        </ComboboxListbox>
+        {helpText && (
+          <FormHelpText id={helpTextId} variant={hasError ? 'error' : undefined}>
+            {helpText}
+          </FormHelpText>
+        )}
+      </Box>
+    );
+  }
+);
+
+Combobox.displayName = 'Combobox';
+
+if (process.env.NODE_ENV === 'development') {
+  Combobox.propTypes = {
+    autocomplete: PropTypes.bool,
+    items: PropTypes.arrayOf(PropTypes.any).isRequired,
+    helpText: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
+    labelText: PropTypes.oneOfType([PropTypes.string, PropTypes.node]).isRequired,
+    onInputValueChange: PropTypes.func,
+    itemToString: PropTypes.func,
+    optionTemplate: PropTypes.func,
+  };
+}
+
+export {Combobox};
--- a/packages/paste-core/components/combobox/src/ComboboxListbox.tsx
+++ b/packages/paste-core/components/combobox/src/ComboboxListbox.tsx
@@ -1,0 +1,45 @@
+import * as React from 'react';
+import * as PropTypes from 'prop-types';
+import {Box, safelySpreadBoxProps} from '@twilio-paste/box';
+
+export interface ComboboxListboxProps {
+  children: NonNullable<React.ReactNode>;
+}
+
+const ComboboxListbox = React.forwardRef<HTMLUListElement, ComboboxListboxProps>(({children, ...props}, ref) => {
+  // TODO:
+  // Right now the listbox is absolutely positioned. This will
+  // fail if a combobox is placed at the bottom of a screen
+  // so we need upgrade the positioning using popperjs in the future.
+  return (
+    <Box
+      {...safelySpreadBoxProps(props)}
+      as="ul"
+      backgroundColor="colorBackgroundBody"
+      borderBottomLeftRadius="borderRadius20"
+      borderBottomRightRadius="borderRadius20"
+      boxShadow="shadowCard"
+      listStyleType="none"
+      margin="space0"
+      overflow="hidden"
+      padding="space0"
+      position="absolute"
+      width="100%"
+      zIndex="zIndex10"
+      marginTop="space20"
+      ref={ref}
+    >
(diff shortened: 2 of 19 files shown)


packages/paste-core/components/combobox/src/index.tsx
```diff
@@ -0,0 +1,147 @@
...
+  const helpTextId = useUID();
+  return (
+    <>
+      <FormLabel disabled={disabled} htmlFor={id} required={required} {...getLabelProps()}>
+        {labelText}
+      </FormLabel>
+      <Box {...getComboboxProps({role: 'combobox'})} display="flex" position="relative">
+        <FormInput
+          {...getInputProps()}
+          {...getToggleButtonProps({tabIndex: 0})}
```
SiTaggart, line 79: For the non-autocomplete combobox, I shouldn't be able to type into the input at all. Only Select something from the list. if you add an `onInput` handler to the input, and prevent the default behaviour of that event if it's not autocomplete, that _should_ do the trick. Ask if you don't know what those words mean.

packages/paste-core/components/combobox/src/index.tsx
```diff
@@ -0,0 +1,147 @@
...
+      if (autocomplete && inputValue) {
+        setInputItems(items.filter(item => item.toLowerCase().startsWith(inputValue.toLowerCase())));
+      }
+      return inputValue;
+    },
+  });
+  const helpTextId = useUID();
+  return (
+    <>
+      <FormLabel disabled={disabled} htmlFor={id} required={required} {...getLabelProps()}>
```
SiTaggart, line 73: Lets pull label from Combobox, and rely on composition. If you want FormLabel you can get it from `form`. Reason being this is just starting to do too much. If I just want a simple search combobox, I want to provide a hidden label, and then we have to start extending all the props to handle those edge case. 

Lets keen this simple.

packages/paste-core/components/combobox/src/index.tsx
```diff
@@ -0,0 +1,147 @@
...
+              padding="space30"
+              paddingLeft="space50"
+              paddingRight="space50"
+              _hover={{textDecoration: 'underline'}}
+            >
+              <Text as="span">{item}</Text>
+            </Box>
+          ))}
+      </Box>
+      {helpText && (
```
SiTaggart, line 120: Same here, help text is just composition.

packages/paste-core/components/combobox/src/index.tsx
```diff
@@ -0,0 +1,147 @@
...
+        <FormInput
+          {...getInputProps()}
+          {...getToggleButtonProps({tabIndex: 0})}
+          disabled={disabled}
+          hasError={hasError}
+          id={id}
+          name={name}
+          placeholder={placeholder}
+          required={required}
+          type="text"
```
SiTaggart, line 86: What if I wanna do insertAfter and insertBefore? Like selects?

richbachman: `insertBefore` works, but `insertAfter` complicates things since that is part of the `FormInput` we have overlaps with the chevron:

image

Unless we create another prefix and suffix for combobox, or export those from the form package. What do you think?

packages/paste-core/components/combobox/src/index.tsx
richbachman: So in pulling the label, we lose the `getLabelProps` function in the hook, and get dinged with some accessibility errors.

SiTaggart: OK, that's fair. If you do want something to look slightly different that's what the primitive is for, for building your own.

So that leads to be the next part then. Can you create components for each of the pieces that we need to compose a new combobox with? Say Everything we use Box for.

So when I do need use the primitive, say I want to swap the input, I still might want the input wrapper, the listbox and it's options. So I could do 

```
import {FormHelpText, FormInput, FormLabel} from '@twilio-paste/form';
import {ComboboxInputWrapper, ComboboxListBox, ComboboxListboxOption} from '@twilio-paste/combobox;

<FormLabel disabled={disabled} htmlFor={id} required={required} {...getLabelProps()}>
  {labelText}
</FormLabel>
  <FormInput
    {...getInputProps()}
    {...getToggleButtonProps({tabIndex: 0})}
    disabled={disabled}
    hasError={hasError}
    id={id}
    name={name}
    placeholder={placeholder}
    required={required}
    type="text"
  />
  <ComboboxIconWrapper>
    <ChevronDownIcon size="sizeIcon30" decorative aria-hidden="true" />
  </ComboboxIconWrapper>
</ComboboxInputWrapper>
<ComboboxListbox
  {...getMenuProps()}
>
  {isOpen &&
    inputItems.map((item, index) => (
      <ComboboxListboxOption
        {...getItemProps({item, index})}
      >
        <Text as="span">{item}</Text>
      </ComboboxListboxOption>
    ))}
</ComboboxListbox>
{helpText && (
  <FormHelpText id={helpTextId} variant={hasError ? 'error' : undefined}>
    {helpText}
  </FormHelpText>
)}

```

packages/paste-core/components/combobox/src/index.tsx
SiTaggart: Checkout what @andipants12 did in select https://github.com/twilio-labs/paste/blob/master/packages/paste-core/components/form/src/select/Select.tsx

packages/paste-core/components/combobox/src/index.tsx
SiTaggart: If we're keeping this in the control with Label, that's fine. Just wiring it up with id and aria-describedby on the input

richbachman: Added back in. Added the `aria-describedby`.

packages/paste-core/components/combobox/src/index.tsx
richbachman: Worked this out using `selectedItem` from the primitive.

packages/paste-core/components/combobox/src/index.tsx
richbachman: Still need to add the actual export files, but I think I worked this out. Running into one issue with spreading the props on `ComboboxListboxOption`. If I use `safelySpreadBoxProps`, it doesn't seem to want to pass the `highlightedIndex`. It will if I just spread the props.

packages/paste-core/components/combobox/src/Combobox.tsx
```diff
@@ -0,0 +1,134 @@
...
+  autocomplete?: boolean;
+  helpText?: string | React.ReactNode;
+  items: UseComboboxPrimitiveProps<any>['items'];
+  labelText: string | NonNullable<React.ReactNode>;
+  onInputValueChange?: UseComboboxPrimitiveProps<any>['onInputValueChange'];
+  itemToString?: UseComboboxPrimitiveProps<any>['itemToString'];
+  optionTemplate?: (item: string | {}) => React.ReactNode;
+}
+
+/* eslint-disable react/no-array-index-key */
```
SiTaggart, line 37: Let’s use react-uid as the key for mapped options and remove this disable

packages/paste-core/components/form/src/shared/FormControlWrapper.tsx
```diff
@@ -0,0 +1,40 @@
...
+    {insertAfter && <Suffix>{insertAfter}</Suffix>}
+  </FieldWrapper>
+);
+
+FormControlWrapper.displayName = 'FormControlWrapper';
+
+if (process.env.NODE_ENV === 'development') {
+  FormControlWrapper.propTypes = {
+    disabled: PropTypes.bool,
+    hasError: PropTypes.bool,
```
SiTaggart, line 36: Missing some props

packages/paste-core/components/form/src/shared/FormControlWrapper.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+interface FormControlWrapperProps {
+  children: NonNullable<React.ReactNode>;
+  disabled?: boolean;
+  hasError?: boolean;
+  insertAfter?: React.ReactNode;
+  insertBefore?: React.ReactNode;
+}
+
+const FormControlWrapper: React.FC<FormControlWrapperProps> = ({
```
SiTaggart, line 16: Can you use this in Input, Select and Textarea as it’s designed to be a shared things for all?

packages/paste-core/components/form/src/shared/FormControlWrapper.tsx
```diff
@@ -0,0 +1,40 @@
...
+  children,
+  disabled,
+  hasError,
+  insertAfter,
+  insertBefore,
+  ...props
+}) => (
+  <FieldWrapper {...safelySpreadFormControlProps(props)} disabled={disabled} hasError={hasError}>
```
SiTaggart, line 24: I thought we got rid of the need for this specific form control spread 🤔

packages/paste-libraries/dropdown/src/index.tsx
```diff
@@ -1,3 +1,6 @@
-import Downshift, {useCombobox} from 'downshift';
+import Downshift from 'downshift';
 
-export {Downshift, useCombobox};
+export * from 'downshift';
```
SiTaggart, line 3: Better 👍

packages/paste-core/components/combobox/stories/index.stories.tsx
```diff
@@ -0,0 +1,158 @@
...
+              <InformationIcon decorative={false} title="Get more info" />
+            </Anchor>
+          }
+          labelText="Choose a component:"
+          helpText="This is the help text"
+        />
+      </>
+    );
+  })
+  .add('Combobox - Object', () => {
```
SiTaggart, line 133: These are some really nice examples showing the power and flexibility of combo box

packages/paste-core/components/combobox/src/Combobox.tsx
```diff
@@ -0,0 +1,134 @@
...
+/* eslint-enable */
+
+Combobox.displayName = 'Combobox';
+
+if (process.env.NODE_ENV === 'development') {
+  Combobox.propTypes = {
+    autocomplete: PropTypes.bool,
+    helpText: PropTypes.oneOfType([PropTypes.string, PropTypes.node]),
+    labelText: PropTypes.oneOfType([PropTypes.string, PropTypes.node]).isRequired,
+    optionTemplate: PropTypes.func,
```
SiTaggart, line 130: Missing a couple props down here

packages/paste-core/components/combobox/src/ComboboxListbox.tsx
```diff
@@ -0,0 +1,40 @@
+import * as React from 'react';
+import * as PropTypes from 'prop-types';
+import {Box, safelySpreadBoxProps} from '@twilio-paste/box';
+
+export interface ComboboxListboxProps {
+  children: NonNullable<React.ReactNode>;
+}
+
+const ComboboxListbox: React.FC<ComboboxListboxProps> = ({children, ...props}) => {
+  return (
```
SiTaggart, line 10: I wonder if it’s worth leaving a TODO here as a note around needing to upgrade listbox positioning to use popperjs in the future?

packages/paste-core/components/combobox/src/Combobox.tsx
```diff
@@ -0,0 +1,134 @@
...
+    right="space40"
+    top="50%"
+    transform="translateY(-50%)"
+    zIndex="zIndex10"
+  >
+    {children}
+  </Box>
+);
+
+export interface ComboboxProps extends Omit<FormInputProps, 'id' | 'type' | 'value'> {
```
SiTaggart, line 27: Nice 👍

packages/paste-core/components/combobox/src/Combobox.tsx
```diff
@@ -0,0 +1,134 @@
...
+          disabled={disabled}
+          hasError={hasError}
+          insertBefore={insertBefore}
+          insertAfter={insertAfter}
+        >
+          <ComboboxInputWrapper {...getComboboxProps({role: 'combobox'})}>
+            <InputElement
+              {...getInputProps({disabled})}
+              {...getToggleButtonProps({tabIndex: 0})}
+              {...(!autocomplete ? {value: selectedItem || ''} : undefined)}
```
SiTaggart, line 89: Wonder if you can use this syntax on line 71 to save us having to create a default item to string function.

`{...(itemToString ? : itemToString : undefined)}`

packages/paste-core/components/form/src/shared/FormControlWrapper.tsx
richbachman: Yes. I did some testing with Input, and there's just some slight restructuring that will need to happen with that as well as Select and Textarea. I think we could get rid of FieldWrapper after that refactor. Adding a Type prop for use with Input later.

packages/paste-core/components/form/src/shared/FormControlWrapper.tsx
richbachman: Oh I see. We should just use `{...restrictedProps}` like Select. We'll need to go back and refactor Input and Textarea.

packages/paste-core/components/combobox/src/Combobox.tsx
richbachman: Looks like that translates to: `itemToString || undefined`, so back to where we were before adding the default item to string function.

packages/paste-core/components/combobox/src/ComboboxListbox.tsx
richbachman: Agreed.

packages/paste-core/components/combobox/src/index.tsx
richbachman: I actually had to change this a bit for an instance when a user doesn't want autocomplete, and has a complete item object. `selectedItem` only handles strings, so it was returning `object Object`. Using `{...(!autocomplete ? {onChange: event => event.preventDefault()} : undefined)}` solves the issue.

packages/paste-core/components/form/src/shared/FormControlWrapper.tsx
richbachman: Added the missing ones.

packages/paste-core/components/combobox/src/Combobox.tsx
richbachman: Added the missing ones.

packages/paste-core/components/combobox/stories/index.stories.tsx
richbachman: I added one more story showing a combobox with a custom(unstyled) input just to make sure the exported components were working.

PR #450, opened 2020-05-18
feat: add animations to modal, globally respect 'prefers-reduced-motion'

TheSisb:
In the process of adding animations to our Modal component, I also wanted to make sure we handled the prefers-reduced-motion media query.  

This PR:
- [x] Unfortunately, react-spring v8 didn't have a mechanism to globally modify/disable animations so I bumped that library to v9 RC3.  
- [x] Add the animation-library dependency to other Paste packages.
- [x] Makes animations globally (CSS and JS using our lib) respect the [prefers-reduced-motion](https://css-tricks.com/introduction-reduced-motion-media-query/) system setting.
- [x] Adds an open/close animation to the Modal component
- [x] Disables JS animations by default on Applitools (for VRT)
- [x] Adds a minHeight to the Modal component to prevent the footer buttons from overflowing
- [x] Adds an exported React Hook "useReducedMotion" to the animation-library, to determine whether the user has that flag toggled on or off
- [x] Updates the ThemeProvider to use our more recent style of writing CSS, moves the `box-sizing` styles into the global styles block (so it doesn't dupe), and globally disables CSS animations for any user of ThemeProvider if prefers-reduced-motion is on.

---

Sidenote: a neat animation visualizer https://react-spring-visualizer.com/

### DIFF
--- a/packages/paste-core/components/modal/package.json
+++ b/packages/paste-core/components/modal/package.json
@@ -33,6 +33,7 @@
     "@styled-system/css": "^5.1.5",
     "@styled-system/theme-get": "^5.1.2",
     "@twilio-paste/absolute": "^2.0.38",
+    "@twilio-paste/animation-library": "^0.1.1",
     "@twilio-paste/box": "^2.4.6",
     "@twilio-paste/button": "^1.2.2",
     "@twilio-paste/design-tokens": "^5.2.0",
@@ -53,6 +54,7 @@
   },
   "devDependencies": {
     "@twilio-paste/absolute": "^2.0.38",
+    "@twilio-paste/animation-library": "^0.1.1",
     "@twilio-paste/box": "^2.4.6",
     "@twilio-paste/button": "^1.2.2",
     "@twilio-paste/design-tokens": "^5.2.0",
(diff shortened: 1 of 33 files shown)


packages/paste-libraries/animation/package.json
```diff
@@ -24,7 +24,7 @@
     "type-check": "tsc --noEmit"
   },
   "dependencies": {
-    "react-spring": "9.0.0-rc.3"
```
SiTaggart, line 27: Just noticed this is missing 
```
"publishConfig": {	  "publishConfig": {
    "access": "public"	    "access": "public"
  },```

File: packages/paste-core/components/modal/package.json [hunk 2]
```diff
@@ -53,6 +54,7 @@
   },
   "devDependencies": {
     "@twilio-paste/absolute": "^2.0.36",
+    "@twilio-paste/animation-library": "^0.0.0",
```
SiTaggart, line 57: I think this also needs to be added as a peer dep for the core bundle now

SiTaggart: Definitely needs to be a dep to the website if it's not already too

PR #453, opened 2020-05-20
feat(Modal docs): add guideline about using h4 in Modal Body

serifluous:
Add guideline to Composition Notes about using h4 in Modal Body

https://paste-git-featmodal-docs-use-h4-in-body.twilio-dsys.now.sh/components/modal/#composition-notes

### DIFF
--- a/packages/paste-website/src/pages/components/modal/index.mdx
+++ b/packages/paste-website/src/pages/components/modal/index.mdx
@@ -191,7 +191,7 @@ Include a Header, Body, and Footer in your modal.
 
 Modal Header text should be concise (2-4 words), start with a verb, and clearly describe the information or action the modal presents.
 
-Modal Body should be focused to 1 main section of content that's described by Modal Header text. If you do need to create multiple sections of content with Headings in Modal Body, use [Heading with the `heading40` variant](/components/heading#heading-as-h-2-with-heading-40-variant).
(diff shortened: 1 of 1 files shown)


packages/paste-website/src/pages/components/modal/index.mdx
```diff
@@ -187,9 +187,13 @@ To set a different initial focus target, set the `initialFocusRef` prop on the `
 
 ## Composition Notes
 
-Include a Header, Body, and Footer in your modal. Heading text should be concise (2-4 words), start with a verb, and clearly describe the information or action the modal presents.
+Include a Header, Body, and Footer in your modal.
 
-Footers should include 1–2 actions aligned to the right side of the modal. Place the primary action farthest on the right to indicate to users they're progressing through their task.
+Modal Header text should be concise (2-4 words), start with a verb, and clearly describe the information or action the modal presents.
+
+Modal Body should have 1 main section of content that's described by the text in the Modal Header. If you do need to create multiple sections of content with Headings in Modal Body, use a [Heading with the `heading40` variant](/components/heading#heading-as-h-2-with-heading-40-variant).
```
serifluous, line 194: @SiTaggart Should this be more specific? Like should they use an h1 with the heading40 variant? Not sure how document hierarchy works within the modal

SiTaggart: It's fine, my original intention was to have a modal a new sectioning content container, meaning you can technically restart the document hierarchy in the modal. We could then be specific as we'd be able to predict what levels headings you should start using.

Couldn't change the outer element though, so we're stuck with unpredictable outlines and trying to suggest one wouldn't end well. We'd never guess right

PR #457, opened 2020-05-22
fix(paragraph): can now unset marginBottom prop

TheSisb:
This PR updates the `Paragraph` component to allow passing `space0` for `marginBottom`. This is useful when using the component as the last child in a `Card` component, for example.

I also updated the tests to use the new library.


packages/paste-core/components/paragraph/__tests__/paragraph.test.tsx
```diff
@@ -1,58 +1,20 @@
...
-    render(
-      <Theme.Provider theme="console">
-        <Paragraph>This is a paragraph</Paragraph>
-      </Theme.Provider>,
-      container
-    );
-    const results = await axe(document.body);
+describe('Accessibility', () => {
+  it('Should have no accessibility violations', async () => {
+    const {container} = testRender(<Paragraph>Hello world!</Paragraph>);
```
TheSisb, line 16: I noticed that we were [doing quite a bit in other tests](https://github.com/twilio-labs/paste/blob/master/packages/paste-core/components/form/__tests__/checkbox.test.tsx#L194), but I _think_ this just works?

package.json
```diff
@@ -41,6 +41,7 @@
     "test:tools": "node --experimental-worker ./node_modules/.bin/jest --maxWorkers=2 ./tools/ -c ./tools/jest.config.js",
     "test:coverage": "yarn pre-test && node --experimental-worker ./node_modules/.bin/jest --maxWorkers=2 --coverage",
     "test:vrt": "eyes-storybook",
+    "test:one": "yarn pre-test && node --experimental-worker ./node_modules/.bin/jest --maxWorkers=2 ./packages/paste-core/components/paragraph",
```
SiTaggart, line 44: probably wanna remove?

packages/paste-core/components/paragraph/__tests__/paragraph.test.tsx
SiTaggart: Yup, I've switched [to this in other components](https://github.com/twilio-labs/paste/blob/master/packages/paste-core/components/menu/__tests__/index.spec.tsx#L111), still learning the ropes of testing library

PR #460, opened 2020-05-26
fix(select): bug fixes

SiTaggart:
- Fix right spacing
- Fix long text overflow
- Add data attribute to prevent console bootstrap seletize overrides

### DIFF
--- a/packages/paste-core/components/form/src/select/Select.tsx
+++ b/packages/paste-core/components/form/src/select/Select.tsx
@@ -25,7 +25,7 @@ const SelectIconWrapper: React.FC<{children: React.ReactNode}> = ({children}) =>
     display="inline-flex"
     position="absolute"
     pointerEvents="none"
-    right="space40"
+    right="space30"
     top="50%"
     transform="translateY(-50%)"
     zIndex="zIndex10"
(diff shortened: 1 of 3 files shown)


packages/paste-core/components/form/src/select/Select.tsx
```diff
@@ -49,7 +49,7 @@ const SelectElement = styled.select(
     fontWeight: 'fontWeightNormal',
     color: 'colorText',
     paddingTop: 'space30',
-    paddingRight: 'space40',
+    paddingRight: 'space100',
```
richbachman, line 52: Should this and the combobox styles match up here? I used `space120` and kept the `space40` right position above on combobox. Totally open to changing combobox. Just let me know and I'll create a ticket.

SiTaggart: Probably change Combobox. The intention was to make the gap between the chevron and the right border tighter. So then I matched (as close as I could) the whitespace between the left hand side of the chevron and the text.

You'd probably need a smaller token than 120 in that case too.

richbachman: Ticket created for this.

PR #461, opened 2020-05-26
feat(website): setup Algolia search

richbachman:
- [x] Algolia search feature
- [x] Added `slug` frontmatter to mdx pages
- [x] Combobox Primitive: export UseComboboxProps
- [x] Combobox: expose onSelectedItemChange and UseComboboxProp, set maxHeight, allow value


packages/paste-website/src/components/search/index.tsx
```diff
@@ -0,0 +1,83 @@
...
+  const [focus, setFocus] = React.useState(false);
+
+  const appId = process.env.GATSBY_ALGOLIA_APP_ID;
+  const searchKey = process.env.GATSBY_ALGOLIA_SEARCH_KEY;
+  const searchClient = React.useMemo(() => algoliasearch(appId as string, searchKey as string), [appId, searchKey]);
+
+  useOnClickOutside(ref, () => setFocus(false));
+
+  return (
+    <Box marginBottom="space120">
```
aayushpi, line 46:
```suggestion
    <Box marginBottom="space70">
```

serifluous:
Would be nice to disable autofill on the input

screenshot of autofill

serifluous:
I don't think we need the "# results for..." text when there are results in the list, esp. since it takes up a whole bunch of space at the top.

Keep it for when there are no results. Then, the text can be default body text size and unitalicized, without the separator underneath. It'll be helpful to add quotation marks around the search term, too.

Happy to wait for some alternate opinions tho.

richbachman:
@serifluous I've disabled the autofill suggestions on the input.

packages/paste-website/src/components/search/SearchInput.tsx
```diff
@@ -0,0 +1,28 @@
...
+import {FormInput} from '@twilio-paste/form';
+
+interface SearchFormProps {
+  currentRefinement: string;
+  refine: (value: string) => void;
+  onFocus?: (event: React.FocusEvent<HTMLInputElement>) => void;
+}
+
+const SearchForm: React.FC<SearchFormProps> = ({currentRefinement, refine, onFocus, ...props}) => (
+  <form noValidate action="" role="search" autoComplete="off">
```
SiTaggart, line 12: Just need to add an `onSubmit` event handler on this form to prevent the form submitting and reloading the page when I hit enter in the input

SiTaggart:
This could probably do with a little polish

image

packages/paste-website/src/components/search/SearchResults.tsx
```diff
@@ -0,0 +1,24 @@
...
+interface StateResultsProps {
+  searchResults?: [] | any;
+  searchState?: [] | any;
+}
+
+const StateResults: React.FC<StateResultsProps> = ({searchResults, searchState}) => {
+  const hasResults = searchResults && searchResults.nbHits !== 0;
+
+  return (
+    <>
```
TheSisb, line 14: Don't need these fragments

packages/paste-website/src/components/search/index.tsx
```diff
@@ -0,0 +1,84 @@
+import * as React from 'react';
+import {InstantSearch, Index, Configure} from 'react-instantsearch-dom';
+import algoliasearch from 'algoliasearch/lite';
+import {Box} from '@twilio-paste/box';
+import {SearchInput} from './SearchInput';
+import {SearchHits} from './SearchHits';
+import {SearchResults} from './SearchResults';
+import {SearchByAlgolia} from './SearchByAlgolia';
+
+const useOnClickOutside = (ref: any, handler: any): void => {
```
TheSisb, line 10: You may want to put this in it's own file, somewhere more re-useable

packages/paste-website/src/pages/components/modal/index.mdx
```diff
@@ -2,6 +2,7 @@
 title: Modal
 package: '@twilio-paste/modal'
 description: A modal displays content over the main page and blocks any interaction with the page.
+slug: /components/modal/
```
TheSisb, line 5: Is there no other way to generate this? From the graphql maybe? This could be error prone..

TheSisb: Ok

packages/paste-website/src/components/search/SearchInput.tsx
```diff
@@ -0,0 +1,28 @@
...
+  onFocus?: (event: React.FocusEvent<HTMLInputElement>) => void;
+}
+
+const SearchForm: React.FC<SearchFormProps> = ({currentRefinement, refine, onFocus, ...props}) => (
+  <form noValidate action="" role="search">
+    <FormInput
+      aria-label="Search"
+      id="search"
+      onChange={e => refine(e.target.value)}
+      placeholder="Search..."
```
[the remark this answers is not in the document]
richbachman: We've gone a new route with an accessible Combobox, so we're good here.

packages/paste-core/components/combobox/__tests__/__snapshots__/index.spec.tsx.snap
```diff
@@ -167,7 +167,8 @@ exports[`Combobox  Render should render 1`] = `
   box-shadow: shadowCard;
   list-style-type: none;
   margin: space0;
-  overflow: hidden;
+  max-height: size50;
```
TheSisb, line 170: Doesnt look like the token was converted to css here and the line above it :O

richbachman: So no changes here?

TheSisb: Ah, good idea!
LGTM

PR #463, opened 2020-05-27
docs(website): content update for home page

aayushpi:
Homepage content update to call out components, content, and layout more explicitly. Here's a screenshot of the intended effect:

image

**Contributing to Twilio**

> Can't really explain what about the anchors on the headings is throwing me off, but could you try putting anchors underneath each text block?
> 
> Might be throwing me off because all the clickable things are bunched together in the middle of the screen, or that the flow of the top section has static text above interactive things, but the bottom sections have interactive things above static text.

tried something different in the newest update


SiTaggart:
@aayushpi any chance we can add a third button after roadmap so we can keep a link to the sandbox on the homepage?

serifluous:
Can't really explain what about the anchors on the headings is throwing me off (maybe it's because all the clickable things are bunched together in the middle of the screen), but could you try putting anchors underneath each text block?

aayushpi:
> I like this, but I do have a concern that there is no longer a clear primary call to action. Maybe that's fine and we want this to be a "choose your own adventure" situation, but I could see folks getting a little bit overwhelmed about what to do next. I personally liked the "getting started" action as a clear primary CTA.

Would a "Get Started" heading above the unordered list help?

serifluous:
> > I like this, but I do have a concern that there is no longer a clear primary call to action. Maybe that's fine and we want this to be a "choose your own adventure" situation, but I could see folks getting a little bit overwhelmed about what to do next. I personally liked the "getting started" action as a clear primary CTA.
> 
> Would a "Get Started" heading above the unordered list help?

Or a "Getting started" link that's the same text size as the text block above the list, though that might make the two other bullet points look a little lonely.

The separator helps the content flow more now from top to bottom imo  👍

katieporter:
> > > I like this, but I do have a concern that there is no longer a clear primary call to action. Maybe that's fine and we want this to be a "choose your own adventure" situation, but I could see folks getting a little bit overwhelmed about what to do next. I personally liked the "getting started" action as a clear primary CTA.
> > 
> > 
> > Would a "Get Started" heading above the unordered list help?
> 
> Or a "Getting started" link that's the same text size as the text block above the list, though that might make the two other bullet points look a little lonely.
> 
> The separator helps the content flow more now from top to bottom imo 👍

Yeah I think a "Getting started" header could help. I also wonder if we could take the links out of the unordered list or just stack them? Doesn't feel totally necessary to have the bullet points but I don't feel super strongly about it. I think it looks good!

PR #466, opened 2020-05-27
docs(website): combobox primitive docs

richbachman:
- [x] Combobox primitive doc

### DIFF
--- a/packages/paste-website/src/pages/primitives/combobox-primitive/index.mdx
+++ b/packages/paste-website/src/pages/primitives/combobox-primitive/index.mdx
@@ -1,0 +1,293 @@
+---
+title: Combobox Primitive
+description: An unstyled and accessible basis upon which to build a combobox.
+slug: /primitives/combobox-primitive/
+---
+
+import {graphql} from 'gatsby';
+import {useUID} from 'react-uid';
+import {useComboboxPrimitive} from '@twilio-paste/combobox-primitive';
+import {Box} from '@twilio-paste/box';
+import {Button} from '@twilio-paste/button';
+import {ChevronDownIcon} from '@twilio-paste/icons/esm/ChevronDownIcon';
+import {FormLabel, FormInput} from '@twilio-paste/form';
+import {SidebarCategoryRoutes} from '../../../constants';
+import {Callout, CalloutTitle, CalloutText} from '../../../components/callout';
+import {autocompleteExample, defaultExample} from '../../../component-examples/ComboboxPrimitiveExamples';
+import Changelog from '@twilio-paste/combobox-primitive/CHANGELOG.md';
+
+export const pageQuery = graphql`
+  {
+    allPastePrimitive(filter: {name: {eq: "@twilio-paste/combobox-primitive"}}) {
+      edges {
+        node {
+          name
+          description
+          status
+          version
+        }
+      }
+    }
+    mdx(fields: {slug: {eq: "/combobox-primitive/"}}) {
+      headings {
+        depth
+        value
+      }
+    }
+  }
+`;
+
+<ComponentHeader
+  name="Combobox Primitive"
+  categoryRoute={SidebarCategoryRoutes.PRIMITIVES}
+  githubUrl="https://github.com/twilio-labs/paste/tree/master/packages/paste-core/primitives/combobox"
+  storybookUrl="https://twilio-labs.github.io/paste/?path=/story/primitives-combobox--dropdown-combobox"
+  data={props.data.allPastePrimitive.edges}
+/>
+
+---
+
+<contentwrapper>
+
+<TableOfContents headings={props.data.mdx.headings} />
+
+<content>
+
+## Guidelines
+
+### About the Combobox Primitive
+
+This package provides a foundation upon which developers can implement a [WAI-Aria compliant Combobox](https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html).
+It can be used to build functional and accessible Comboboxes with or without autocomplete/typeahead features.
+Our Combobox is built on top of this primitive.
+
+The purpose of providing these unstyled primitives is to cater for instances when the styled Combobox provided
+by Paste, doesn't meet the requirements needed to solve a unique or individual customer problem. At that point
+you are welcome to fallback to this functional primitive to roll your own styled Combobox whilst still providing
+a functional and accessible experience to your customers.
+
+This primitive should be used to compose all custom Comboboxes to ensure accessibility and upgrade paths.
+
+<Callout variant="warning">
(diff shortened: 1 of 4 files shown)


packages/paste-website/src/pages/primitives/combobox-primitive/index.mdx
```diff
@@ -0,0 +1,293 @@
...
+  <CalloutText>
+    We strongly suggest that all components built on top of this primitive get reviewed by the Design Systems team and
+    goes through the UX Review process to ensure an excellent experience for our customers.
+  </CalloutText>
+</Callout>
+
+## Usage Guide
+
+This package is a wrapper around the [`Downshift` package](https://downshift.netlify.com).
+Our wrapper currently only exposes the useSelect and useCombobox hooks, but renamed for Paste.
```
SiTaggart, line 81: > only exposes the useSelect and useCombobox hooks

How true is this still?

packages/paste-website/src/pages/primitives/combobox-primitive/index.mdx
```diff
@@ -0,0 +1,293 @@
...
+The API's for both hooks are otherwise identical. The reason we chose to just expose the hooks
```
SiTaggart, line 82: We expose the render prop and hook. We need to update this paragraph

packages/paste-website/src/pages/primitives/combobox-primitive/index.mdx
```diff
@@ -0,0 +1,293 @@
...
+
+<LivePreview
+  scope={{useComboboxPrimitive, Box, Button, ChevronDownIcon, FormLabel, FormInput, useUID}}
+  noInline
+  language="jsx"
+>
+  {autocompleteExample}
+</LivePreview>
+
+### useComboboxPrimitive Arguments
```
SiTaggart, line 145: need to document the render prop version too, I'm afriad.

packages/paste-website/src/pages/primitives/combobox-primitive/index.mdx
richbachman: Should we mention exposing the render prop here? Or add a separate paragraph about using the render prop if not able to use hooks, along with a link out to their docs?

packages/paste-website/src/pages/primitives/combobox-primitive/index.mdx
richbachman: Different now. I'll update that to only include the useCombobox hook.

packages/paste-website/src/pages/primitives/combobox-primitive/index.mdx
```diff
@@ -0,0 +1,293 @@
...
+
+The combobox primitive is an unstyled functional version of a Combobox. It can be used to build functional and
+accessible comboboxes with or without autocomplete/typeahead features. Our Combobox is built on top of this primitive.
+
+The purpose of providing these unstyled primitives is to cater for instances when the styled Combobox provided
+by Paste, doesn't meet the requirements needed to solve a unique or individual customer problem. At that point
+you are welcome to fallback to this functional primitive to roll your own styled combobox whilst still providing
+a functional and accessible experience to your customers.
+
+This primitive should be used to compose all custom comboboxes to ensure accessibility and upgrade paths.
```
TheSisb, line 68: The copy needs a quick find/replace to make all combobox capitalized

packages/paste-website/src/pages/primitives/combobox-primitive/index.mdx
TheSisb: I wonder if the future upgrade path would be easier if we don't mention the render props here at all. If there's a lot of demand we can add it in, but removing it might cause more friction down the line than it's worth.

packages/paste-website/src/pages/primitives/combobox-primitive/index.mdx
```diff
@@ -0,0 +1,293 @@
...
+
+#### Basic Combobox
+
+A basic Combobox is a direct replacement for the native HTML select element. It should function in roughly
+the same way, the difference being in the ability to style everything about it; the input trigger, option list
+and option content.
+
+Contrary to the Downshift documentation and example, the basic Combobox should have an HTML input element as
+the trigger, and not an HTML button. Use an input element and set the role of combobox in the `getToggleButtonProps`
+getter or directly on the element itself. You can see this demonstrated below:
```
TheSisb, line 117: Might this be something we do directly in the primitive before re-exposing it to consumers, to prevent misuse?

packages/paste-website/src/pages/primitives/combobox-primitive/index.mdx
```diff
@@ -0,0 +1,293 @@
...
+
+##### `toggleMenu: () => void`
+
+Toggle the menu open state.
+
+##### `selectItem: (item: Item) => void`
+
+Selects the given item.
+
+##### `setHighlightedIndex: (index: number) => void`
```
TheSisb, line 283: 🤔 I now kind of wish we adopted this props format everywhere...

SiTaggart: As in, not the table of props?

If so, +1

packages/paste-website/src/pages/primitives/combobox-primitive/index.mdx
SiTaggart: Fair, let’s just mention the hook and keep the render prop as an unadvertised back up plan

packages/paste-website/src/pages/primitives/combobox-primitive/index.mdx
SiTaggart: Edit: as per other comment, ignore me here.

packages/paste-website/src/pages/primitives/combobox-primitive/index.mdx
richbachman: 👍

packages/paste-website/src/pages/primitives/combobox-primitive/index.mdx
richbachman: Same. I like this format better. Easier for mobile as well.

packages/paste-website/src/pages/primitives/combobox-primitive/index.mdx
richbachman: Because the primitive is just the hook, I don't think we want to lock it down to a particular composition. Guidance might be our best path forward here.

packages/paste-website/src/pages/primitives/combobox-primitive/index.mdx
TheSisb: Yeah.  Let this be our path forward and we will eventually be consistent :D

PR #478, opened 2020-06-04
docs(website): add alpha combobox doc page

richbachman:
- [x] Combobox doc page
- [x] Expose more props for Combobox that may be useful to user
  - onHighlightedIndexChange
  - onIsOpenChange


TheSisb:
The autocomplete example crashes when I try typing a letter into the combobox
image

richbachman:
@TheSisb will you verify the crash is fixed?

SiTaggart:
Just spotted a bug with Combobox component. The label for and input id's don't match

image

Reckon you can do a quick patch in this PR for that?

richbachman:
@SiTaggart it looks like that might be an issue with Downshift.
image

I tried passing an `htmlFor` value directly to the label and in the `getLabelProps` function, but those get overwritten by Downshift.

SiTaggart:
So the problem is this line in the combobox, we should be able to work around it. 

https://github.com/twilio-labs/paste/blob/master/packages/paste-core/components/combobox/src/Combobox.tsx#L96

If you need both, input props and toggle button props, then it might be the _order_ of those that matter. toggle button id is overrideing the input id.

richbachman:
@SiTaggart just pushed up the fix for the label for attribute and a test to verify.

PR #480, opened 2020-06-04
feat(tabs): adding alpha tabs component

TheSisb:
WIP

Adding the Paste Tabs component (alpha).

todo: tests


richbachman:
I was thinking about the `TabsPanel` wrapper. Should we set that up that just accepts padding for the alpha. Purely stylistic, but it might be nice to be able to set a consistent spacing between the tabs and the tabs content. Thoughts?

packages/paste-core/components/tabs/src/Tab.tsx
```diff
@@ -0,0 +1,111 @@
...
+        borderBottomStyle: 'solid',
+        borderBottomColor: 'transparent',
+        _selected: {
+          borderBottomColor: 'colorBorderPrimary',
+        },
+      };
+  }
+};
+
+const getTabTextStyles = (props: TabProps): {} => {
```
richbachman, line 46: Should the return type here be `TextColor` or `TextColorOptions` from our style props?

TheSisb: I just changed this to also include _hover, so I'm going to leave it ambiguous for now.

packages/paste-core/components/tabs/package.json
```diff
@@ -0,0 +1,50 @@
...
+    "@emotion/styled": "^10.0.10",
+    "@twilio-paste/box": "^2.4.8",
+    "@twilio-paste/flex": "^0.3.27",
+    "@twilio-paste/style-props": "1.1.9",
+    "@twilio-paste/tabs-primitive": "^0.1.0",
+    "@twilio-paste/text": "^2.1.9",
+    "prop-types": "^15.7.2",
+    "react": "^16.8.6",
+    "react-dom": "^16.8.6",
+    "styled-system": "^5.1.2"
```
SiTaggart, line 41: Most recently for peer dependencies, after fixing a bunch of @yarnpkg/doctor warnings (now seems to be working again), I came across this article explaining the `Unmet transitive peer dependency` warning. https://dev.to/arcanis/implicit-transitive-peer-dependencies-ed0. The crux of it was, peer deps of peer deps should also be listed.

Otherwise you get a whole bunch of these looking errors
image.

In some packages that causes the compiled code to blow out with some peer deps that it needs that aren't listed. In other cases like this one, they seem to compile fine without it.

I'm been leaning on the side of caution and just listing them as it does no harm and clears the warnings about unmet peers

TheSisb: :O yarn doctor works again?! 
Will do! 👍

packages/paste-core/components/tabs/src/Tab.tsx
```diff
@@ -0,0 +1,118 @@
...
+  }
+  return {
+    color: 'colorTextWeak',
+    _hover: {
+      color: 'colorText',
+    },
+  };
+};
+
+export interface TabProps {
```
richbachman, line 66: If you import `TabPrimitiveProps` and extend it here, I think that will fix the TS error here.

richbachman: Ignore that. Still having errors.

packages/paste-core/components/tabs/__tests__/tabs.test.tsx
```diff
@@ -0,0 +1,49 @@
...
+    it('relevant html and aria attributes', () => {
+      render(<HorizontalTabsExample />);
+
+      const renderedTabList = screen.getByRole('tablist');
+      expect(renderedTabList.getAttribute('aria-orientation')).toEqual('horizontal');
+      expect(renderedTabList.getAttribute('aria-label')).toEqual('My tabs');
+
+      const renderedTabs = screen.queryAllByRole('tab');
+      renderedTabs.forEach(renderedTab => {
+        expect(renderedTab.getAttribute('aria-controls')).toBeDefined();
```
SiTaggart, line 27: Is it worth testing the values of these?

For this component to be accessible we know that `aria-controls` should match the id of the panel that it shows. We know `aria-selected` should only be true when the tab is selected. `tabindex` should only be `0` also when it is selected. We know the panel should be `aria-labelledby` the tab that controls it.

That way if we bump the package and it regresses we the accessibility, we'll know about it

packages/paste-core/components/tabs/src/Tabs.tsx
```diff
@@ -0,0 +1,33 @@
...
+import {useTabPrimitiveState, TabPrimitiveInitialState} from '@twilio-paste/tabs-primitive';
+import {TabsContext} from './TabsContext';
+
+export type TabsProps = TabPrimitiveInitialState;
+
+// Set orientation to horizontal because undefined enables all arrow key movement
+const Tabs: React.FC<TabsProps> = ({children, orientation = 'horizontal', ...initialState}) => {
+  const tab = useTabPrimitiveState({orientation, ...initialState});
+  const value = React.useMemo(() => tab, Object.values(tab));
+  const returnValue = <TabsContext.Provider value={value}>{children}</TabsContext.Provider>;
```
SiTaggart, line 13: I really like this approach. I'm going to suggest we refactor Menu, which just exposes the hook, to do the same 👍

packages/paste-core/components/tabs/src/TabPanels.tsx
```diff
@@ -0,0 +1,39 @@
+import * as React from 'react';
+import {Box, safelySpreadBoxProps} from '@twilio-paste/box';
+import {isPaddingTokenProp, PaddingProps} from '@twilio-paste/style-props';
+
+export interface TabPanelsProps extends PaddingProps {
+  children: React.ReactNode;
+}
+
+const TabPanels = React.forwardRef<HTMLDivElement, TabPanelsProps>(
+  ({children, padding, paddingTop, paddingBottom, paddingRight, paddingLeft, ...props}, ref) => {
```
SiTaggart, line 10: Curious why you choose to expose padding on the panels, instead of the panel?

For the Alpha would it be long term safer to not expose these just yet, _just_ incase we actually don't want folks doing this when we do a full design round?

packages/paste-core/components/tabs/src/TabList.tsx
```diff
@@ -0,0 +1,45 @@
...
+  <Box borderRightWidth="borderWidth20" borderRightColor="colorBorderLight" borderRightStyle="solid" maxWidth="size30">
+    {children}
+  </Box>
+);
+
+const TabList = React.forwardRef<HTMLDivElement, TabListProps>(({children, ...props}, ref) => {
+  const tab = React.useContext(TabsContext);
+  const TabListWrapper = tab.orientation === 'vertical' ? VerticalTabList : HorizontalTabList;
+  return (
+    <TabPrimitiveList {...(tab as any)} {...props} ref={ref}>
```
SiTaggart, line 30: You could set this ` as={TabListWrapper}` if you wanted to remove the extra DOM element, but that would mean safely spreading props on `HorizontalTabList` and `VerticalTabList`

packages/paste-core/components/tabs/src/TabPanels.tsx
TheSisb: I put it on the wrapper because I figured we should enforce that each panel has consistent padding, whatever that may be. But I agree, I'll disable it for now.

packages/paste-core/components/tabs/src/TabList.tsx
TheSisb: I considered that but for the sake of separating the responsibility thought I'd just make it a child for clarity. Do you feel strongly about it?

SiTaggart: Nah, not in the slightest! I don’t that function from presentation separation at all

PR #482, opened 2020-06-05
feat(non-modal-dialog-primitive): add non-modal dialog primitive package

richbachman:
- [x] Non-modal dialog primitive package


SiTaggart:
I think even though we’ll end up calling the component a Popover, I would like the accessible primitive to be named non-modal dialog.

Mainly because it can be used for more than just a popover

richbachman:
That plan sounds good. I'll get the files updated.

PR #488, opened 2020-06-06
fix(website): tabbing overflow bug and adding two landmark roles

TheSisb:
This PR fixes the issue where tabbing many times would cause the focus to jump to the bottom of a long page of content, and cause the browser to shift the body area under the alerts at the top.  I figured out that it was happening because of the `100vh` making the browser think that the content area was the full height, when indeed it was `100vh - alertHeight`.

I also re-introduced the landmark roles and tabindex on the main body area, as well as on the sidebar.


packages/paste-website/src/components/site-wrapper/SiteBody.tsx
```diff
@@ -60,30 +76,39 @@ const BLMAlert: React.FC = () => {
...
-      <StyledSiteBody activeTheme={activeTheme}>
+      <BlackLivesMatterAlert />
+      <PasteThemeAlert isPasteTheme={isPasteTheme} />
+      <StyledSiteBody isPasteTheme={isPasteTheme}>
         <Sidebar />
-        <Box height="100vh" overflow="auto">
+        <Box height="100%" overflow="auto">
           <SiteHeader />
-          <SiteMain id="site-main">
+          <SiteMain id="site-main" tabIndex={0} role="main">
```
SiTaggart, line 111: No need for role here, it is a `<main />` element that comes with an implicit `man` role.

packages/paste-website/src/components/site-wrapper/sidebar/SidebarNavigation.tsx
```diff
@@ -124,7 +124,7 @@ const SidebarNavigation: React.FC<SidebarNavigationProps> = () => {
   const [tokensOpen, setTokensOpen] = React.useState(getCurrentPathname().startsWith(SidebarCategoryRoutes.TOKENS));
 
   return (
-    <SiteNav>
+    <SiteNav tabIndex={0} role="navigation" aria-label="Main">
```
SiTaggart, line 127: Again, no need for the role here, it is a `<nav />` element that comes with an implicit `navigation` role.

The label is a nice addition.

I am cautious of the focusable nav element here though. Ordinarily non-interactive elements shouldn't be focusable, even if they scroll. I'm OK with the main as we'll use it as a jump point when we have a "skip to main content" skip link, but I don't think you need it on the nav. The panel will scroll as you tab through the nav items

packages/paste-website/src/components/site-wrapper/SiteBody.tsx
TheSisb: That isn't supported in IE11: https://www.drupal.org/project/drupal/issues/2862904#comment-12161565

PR #496, opened 2020-06-10
feat(popover): add popover package

richbachman:
- [x] Popover package
- [x] Customize PopoverArrow component from Reakit to remove SVG class names
- [x] Export adjustments to non-modal-dialog-primitive package

### DIFF
--- a/packages/paste-core/components/popover/src/Popover.tsx
+++ b/packages/paste-core/components/popover/src/Popover.tsx
@@ -1,0 +1,73 @@
+import * as React from 'react';
+import * as PropTypes from 'prop-types';
+import {Absolute} from '@twilio-paste/absolute';
+import {Box, BoxProps, safelySpreadBoxProps} from '@twilio-paste/box';
+import {Button} from '@twilio-paste/button';
+import {CloseIcon} from '@twilio-paste/icons/esm/CloseIcon';
+import {StyledBase} from '@twilio-paste/theme';
+import {NonModalDialogPrimitive} from '@twilio-paste/non-modal-dialog-primitive';
+import {PopoverArrow} from './PopoverArrow';
+import {PopoverContext} from './PopoverContext';
+
+export interface PopoverProps {
+  'aria-label': string;
+  children: React.ReactNode;
+}
+
+const StyledPopover = React.forwardRef<HTMLDivElement, BoxProps>(({style, ...props}, ref) => {
+  return (
+    <Box
+      {...safelySpreadBoxProps(props)}
+      backgroundColor="colorBackgroundBody"
+      borderStyle="solid"
+      borderWidth="borderWidth10"
+      borderColor="colorBorderLight"
+      borderRadius="borderRadius20"
+      boxShadow="shadowCard"
+      maxWidth="size30"
+      minWidth="size30"
+      zIndex="zIndex10"
+      _focus={{outline: 'none'}}
+      style={style}
+      ref={ref}
+    />
+  );
+});
+
+const Popover = React.forwardRef<HTMLDivElement, PopoverProps>(({children, ...props}, ref) => {
+  const popover = React.useContext(PopoverContext);
+  return (
+    <NonModalDialogPrimitive {...(popover as any)} {...props} as={StyledPopover} ref={ref}>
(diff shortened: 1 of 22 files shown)


packages/paste-core/components/popover/src/PopoverContainer.tsx
```diff
@@ -0,0 +1,18 @@
...
+import {
+  useNonModalDialogPrimitiveState,
+  NonModalDialogPrimitivePopoverInitialState,
+} from '@twilio-paste/non-modal-dialog-primitive';
+import {PopoverContext} from './PopoverContext';
+
+export type PopoverContainerProps = NonModalDialogPrimitivePopoverInitialState;
+
+const PopoverContainer: React.FC<PopoverContainerProps> = ({children, placement, gutter, ...initialState}) => {
+  const popover = useNonModalDialogPrimitiveState({placement, gutter, ...initialState});
```
SiTaggart, line 11: I think there is an option to make a popover a focus trap, I would like to make this mandatory. 

You really shouldn’t be able to tab out of a popover.

packages/paste-core/components/popover/src/Popover.tsx
```diff
@@ -0,0 +1,53 @@
...
+      style={style}
+      ref={ref}
+    />
+  );
+});
+
+const Popover = React.forwardRef<HTMLDivElement, PopoverProps>(({children, ...props}, ref) => {
+  const popover = React.useContext(PopoverContext);
+  return (
+    <NonModalDialogPrimitive {...(popover as any)} {...props} as={StyledPopover} ref={ref}>
```
SiTaggart, line 37: Is there any way you can think of, of positioning a close x button in the top right and making that always render?

I’d like to make it a rule that all non-modal dialogs should have a visual dismiss action for dictation users, and if you can come up with a way to make it look nice, not get in the way of other people’s content, that would be great

packages/paste-core/components/popover/src/PopoverContainer.tsx
richbachman: I'm setting the `modal` option so now focus is trapped.

packages/paste-core/components/popover/src/Popover.tsx
richbachman: I was able to work this out: 
image

packages/paste-libraries/reakit/src/Popover/PopoverArrow.tsx
```diff
@@ -0,0 +1,85 @@
...
+      () => (
+        <svg viewBox="0 0 30 30" style={{transform}}>
+          <path
+            fill={fill}
+            stroke={stroke}
+            d="M23.7,27.1L17,19.9C16.5,19.3,15.8,19,15,19s-1.6,0.3-2.1,0.9l-6.6,7.2C5.3,28.1,3.4,29,2,29h26
+        C26.7,29,24.6,28.1,23.7,27.1z"
+          />
+          <path
+            fill={fill}
```
TheSisb, line 56: 👍

packages/paste-core/components/popover/src/Popover.tsx
```diff
@@ -0,0 +1,85 @@
...
+
+// import Paste Theme Based Styles due to portal positioning.
+// reakit portal is a sibling to the main app, so you are now
+// no longer a child of the theme provider. We need to re-set
+// some of the base styles that we rely on inheriting from
+// such as font-family and line-height so that compositions
+// of paste components in the modal are styled correctly
+
+// eslint-disable-next-line emotion/syntax-preference
+const StyledBase = styled('div')(pasteBaseStyles);
```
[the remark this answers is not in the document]
richbachman: Good catch!

packages/paste-core/components/popover/src/Popover.tsx
```diff
@@ -16,16 +15,6 @@ export interface PopoverProps {
   tabIndex?: number;
 }
 
-// import Paste Theme Based Styles due to portal positioning.
```
TheSisb: I think this comment is still valuable around where StyledBase is being used.

PR #498, opened 2020-06-11
docs(website): adding Tabs alpha component documentation

TheSisb:
Adding the docs for the alpha Tabs component.


katieporter:
@TheSisb it looks like, in the anatomy section, the Tab Label box background is listed as $color-background-required. I think it's supposed to be $color-background-body, maybe?

TheSisb:
@katieporter Good catch! I forgot to remove that line. Currently we don't set a background color on tabs, so it inherits the background of whatever we place it on.

PR #501, opened 2020-06-11
chore: lighten gray-70 in the console theme

katieporter:
Lightening the color-gray-70 alias (which is used for color-text-weak) in the console theme ever so slightly, so that it meets a 3:1 contrast ratio with color-gray-100 (which is used for color-text).

**Contributing to Twilio**


SiTaggart:
For this I would look to not committing the yarn.lock file changes. Just the token value change.

TheSisb:
Visual regression testing caught a side-effect of this change on the neutral Alert component's icon
image

Just confirming, is this fine? I can approve it afterwards

katieporter:
> Visual regression testing caught a side-effect of this change on the neutral Alert component's icon
> image
> 
> Just confirming, is this fine? I can approve it afterwards

@TheSisb yes, that's fine! Thanks for checking.

PR #502, opened 2020-06-12
feat(styling-library): adding Paste's styling library

TheSisb:
WIP

Creating a library called `@twilio-paste/styling-library` which encapsulates both [emotion](https://emotion.sh/docs/introduction) and [styled-system](https://styled-system.com/). This change aims to simplify dependency management and abstract our styling system in such a way that we can change the underlying technology with minimal impact to end-users. We also control what parts of those libraries we expose; this PR only exposes `@styled-system/css` and omits the `css` export from `emotion`, so that we consistently use the batteries-packed token version.

The `Button` component was making use of emotion's `css` function, so in order to swap to the styled-system css function I had to refactor the Button styles code.  It's easier to read now.


.eslintrc.js
```diff
@@ -66,9 +69,6 @@ module.exports = {
     // This rule tells people to do something (import foo = require('foo')) which doesn't work
     // with babel compiled typescript.
     '@typescript-eslint/no-var-requires': 'off',
-    // Enforce template strings for styles over objects for consistent codebase
-    // We can switch this to object at anytime, but lowering to warn as we have a mixed condebase right now.
-    'emotion/syntax-preference': [1, 'string'],
```
TheSisb, line 71: This whole file is prettier except the removal of this line, which we simply ignore everywhere.

SiTaggart:
The only thing that is catching my eye right now is the package size increase with theme.

In theory it should actually be reducing, as you've removed `emotion-theming` as a dependency and moved it to a peer, so that should be removed from the bundle

image

image

packages/paste-libraries/styling/src/index.tsx
```diff
@@ -0,0 +1,14 @@
+// Base styling system (emotion)
+import styled from '@emotion/styled';
+import * as emotion from 'emotion';
+
+export {Global as StylingGlobals, SerializedStyles, ThemeContext, keyframes} from '@emotion/core';
+export {ThemeProvider} from 'emotion-theming';
```
SiTaggart, line 6: `withTheme` is missing from here and is causing the theme package size to increase and some of the tests to fail. Not all tests

PR #509, opened 2020-06-17
fix(website): fix website scrolling

richbachman:
- [x] Fix website scrolling in Safari and mobile Safari


TheSisb:
This change brings the tabbing bug back

richbachman:
@TheSisb I found if we changed the tabIndex to -1 on SiteMain, that fixes the tabbing bug in Chrome and Firefox, and still allows users to use page up and page down when focused in the main area.

The page up and page down doesn't work in Safari, but scrolling still works in both desktop and mobile.

PR #522, opened 2020-06-23
fix(modal): add '__console_patch' prop to fix bug in Console

TheSisb:
Fixes an issue specific to Twilio's Console application which causes the page to shift upwards by 80px when the modal opens. This happens because of styling set on the app which clashes with the scroll-lock functionality of reach's modal. This prop is the least worst solution to a problem currently blocking adoption on the Modal component.

### DIFF
--- a/packages/paste-core/components/modal/src/utils/consoleUtils.ts
+++ b/packages/paste-core/components/modal/src/utils/consoleUtils.ts
@@ -1,0 +1,45 @@
+/*
+ * HACK NOTICE:
+ * This is an eggregious hack but currently the best way for us
+ * to maintain accessibility in our modal and keep it functional
+ * in Twilio's Console application.
+ *
+ * For more info, see: https://github.com/reach/reach-ui/issues/610
+ *
+ * FIXME: when the Console app redesigns navigation
+ */
+
+// 80px because the issue is the `calc(100vh - 80px)`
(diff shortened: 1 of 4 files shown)


richbachman:
Should we add a storybook example just for vrt?

packages/paste-core/components/modal/src/utils/consoleUtils.ts
```diff
@@ -0,0 +1,45 @@
...
+// 80px because the issue is the `calc(100vh - 80px)`
+const MARGIN_AMOUNT_BASED_ON_CALC_OFFSET = '80px';
+
+export function addConsoleHeightPatch(): void {
+  if (document == null) {
+    return;
+  }
+
+  const CONTENT_WRAPPER = document.querySelector('#content') as HTMLElement;
+  const SIDEBAR_WRAPPER = document.querySelector('#sidebar-wrapper') as HTMLElement;
```
SiTaggart, line 21: Given that the reason this happens in Console is reach modal removed padding top from the `body`, was there a particular reason you aren't just adding the 80px padding-top back to the body, rather than adding margin to 2 other elements?

TheSisb: As discussed on Slack, this is for more specificity for future maintainers.

PR #530, opened 2020-06-25
chore: improve build:dev command

TheSisb:
Since local builds have been rather slow on my machine, I took a little time to research ways to improve that DX. This PR does a couple things:

1. It fixes the `build:dev` command by removing the `w` in the `rollup -cw` part of every package. The `w` makes rollup startup a watcher, which basically pauses the build on the first package. We've never really used it since our dev cycle involves storybook which operates on the src files anyways, so we won't miss this.

2. It adds the lerna `--since HEAD` flag (https://github.com/lerna/lerna/issues/2211#issuecomment-535943543) which makes lerna only build packages with changes. There are one gotcha with this flag that I'm aware of:
   1. You can get into a weird state if you exclusively run `yarn build:dev` where, some of your packages might have been built on one unmerged branch, you switch branches, and those packages are now still built and stale. Previously since we always rebuilt everything, we would always be in a clean state. Also not a big deal, but something to be mindful of - we should periodically rebuild all the things to get back into a "clean" state.

### DIFF
--- a/packages/paste-core/core-bundle/package.json
+++ b/packages/paste-core/core-bundle/package.json
@@ -22,7 +22,7 @@
     "build:dev": "yarn clean && yarn compile:dev",
     "clean": "rm -rf ./dist && rm -rf tsconfig.build.tsbuildinfo && rm -rf .rpt2_cache",
     "compile": "rollup -c --environment NODE_ENV:production",
-    "compile:dev": "rollup -cw --environment NODE_ENV:development",
+    "compile:dev": "rollup -c --environment NODE_ENV:development",
     "prepublishOnly": "yarn build",
     "type-check": "tsc --noEmit"
   },
--- a/package.json
+++ b/package.json
@@ -21,7 +21,7 @@
     "watch:icons": "yarn workspace @twilio-paste/icons watch",
     "prebuild": "node ./tools/build/pre-build.js",
     "build": "lerna run build --ignore @twilio-paste/website",
-    "build:dev": "yarn prebuild && yarn build:tokens && lerna run build:dev",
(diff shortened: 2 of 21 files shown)


packages/paste-core/core-bundle/package.json
```diff
@@ -19,10 +19,11 @@
   "scripts": {
     "prebuild": "node ./tools/generate.js",
     "build": "yarn clean && yarn compile",
-    "build:dev": "yarn clean && yarn compile:dev",
+    "build:dev": "yarn build",
+    "build:watch": "yarn clean && yarn compile:dev",
```
richbachman, line 23: Out of curiosity, why id core getting a watch script, and not any other package?

SiTaggart:
> --since is short for --since HEAD, which means that you need to commit something for it to start detecting changes. Not a big deal.

Tried this locally and it started running the full build. Is there a change this only works on master?

TheSisb:
@SiTaggart How did you run this locally? If you cherry-picked this PR's commit onto your branch it will run pretty much everything because every package.json was changed.

packages/paste-core/core-bundle/package.json
TheSisb: Good catch, I'm not sure how that happened.. that change was meant to be in the design-tokens package! Fixed.

SiTaggart:
> How did you run this locally? 

I created a new commit on this branch locally...

TheSisb:
@SiTaggart Oops, I thought `--since` was short for `--since HEAD` but it's actually since the latest tags. I ammended the command to `--since HEAD` and now it is working as intended.

SiTaggart:
I don't know what I'm doing wrong but even commiting 2 commits to two different packages on this branch, it's not picking up any changes to build 

image

TheSisb:
Discussed in person

PR #532, opened 2020-06-25
feat(tooltip): add tooltip component package

richbachman:
- [x] Tooltip package
- [x] Stories
- [x] Tests
- [x] Doc page

### DIFF
--- a/packages/paste-core/components/anchor/src/index.tsx
+++ b/packages/paste-core/components/anchor/src/index.tsx
@@ -53,11 +53,11 @@ const StyledAnchor = styled.a(
   )
 );
 
-const Anchor: React.FC<AnchorProps> = props => (
-  <StyledAnchor {...secureExternalLink(props.href)} {...props}>
+const Anchor = React.forwardRef<HTMLAnchorElement, AnchorProps>((props, ref) => (
+  <StyledAnchor {...secureExternalLink(props.href)} {...props} ref={ref}>
     {props.children}
   </StyledAnchor>
-);
+));
 Anchor.displayName = 'Anchor';
 
 if (process.env.NODE_ENV === 'development') {
--- a/packages/paste-core/components/tooltip/__test__/index.spec.tsx
+++ b/packages/paste-core/components/tooltip/__test__/index.spec.tsx
@@ -1,0 +1,37 @@
+import * as React from 'react';
+import {axe} from 'jest-axe';
+import {render, screen} from '@testing-library/react';
+import {Button} from '@twilio-paste/button';
+import {Theme} from '@twilio-paste/theme';
+import {Tooltip} from '../src';
+
+const TooltipMock: React.FC<{}> = () => {
+  return (
+    <Theme.Provider theme="console">
+      <Tooltip text="Welcome to Paste!" data-testid="tooltip-example">
+        <Button variant="primary">Open Tooltip</Button>
+      </Tooltip>
+    </Theme.Provider>
+  );
+};
+
+describe('Tooltip', () => {
+  describe('Render', () => {
+    render(<TooltipMock />);
+    it('should render a tooltip button with aria attributes', () => {
+      const renderedTooltipButton = screen.getByRole('button');
+      expect(renderedTooltipButton.getAttribute('aria-describedby')).toEqual('paste-tooltip-1');
+
+      const renderedTooltip = screen.getByTestId('tooltip-example');
+      expect(renderedTooltip.getAttribute('role')).toEqual('tooltip');
+    });
+  });
+
+  describe('Accessibility', () => {
+    it('Should have no accessibility violations', async () => {
+      const {container} = render(<TooltipMock />);
+      const results = await axe(container);
+      expect(results).toHaveNoViolations();
+    });
+  });
+});
--- a/packages/paste-core/components/tooltip/package.json
+++ b/packages/paste-core/components/tooltip/package.json
@@ -1,0 +1,60 @@
+{
+  "name": "@twilio-paste/tooltip",
+  "version": "0.0.0",
+  "category": "interaction",
+  "status": "alpha",
+  "description": "The Tooltip is a popup that displays clarifying text related to an element that has been focused or hovered on.",
+  "author": "Twilio Inc.",
+  "license": "MIT",
+  "main:dev": "src/index.tsx",
+  "main": "dist/index.js",
+  "module": "dist/index.es.js",
+  "types": "dist/index.d.ts",
+  "sideEffects": false,
+  "publishConfig": {
+    "access": "public"
+  },
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "build": "yarn clean && yarn compile",
+    "build:dev": "yarn clean && yarn compile:dev",
+    "build:props": "typedoc --tsconfig ./tsconfig.json --json ./dist/prop-types.json",
+    "clean": "rm -rf ./dist && rm -rf tsconfig.build.tsbuildinfo && rm -rf .rpt2_cache",
+    "compile": "rollup -c --environment NODE_ENV:production",
+    "compile:dev": "rollup -cw --environment NODE_ENV:development",
+    "prepublishOnly": "yarn build",
+    "type-check": "tsc --noEmit"
+  },
+  "peerDependencies": {
+    "@twilio-paste/box": "^2.5.4",
+    "@twilio-paste/design-tokens": "^5.2.2",
+    "@twilio-paste/icons": "^2.2.9",
+    "@twilio-paste/reakit-library": "^0.6.0",
+    "@twilio-paste/spinner": "^1.2.18",
+    "@twilio-paste/style-props": "^1.2.5",
+    "@twilio-paste/styling-library": "^0.1.0",
+    "@twilio-paste/text": "^2.1.15",
+    "@twilio-paste/theme": "^3.2.5",
+    "@twilio-paste/tooltip-primitive": "^0.1.0",
+    "@twilio-paste/types": "^3.0.9",
+    "prop-types": "^15.7.2",
+    "react": "^16.8.6",
+    "react-dom": "^16.8.6",
+    "react-uid": "^2.2.0"
+  },
+  "devDependencies": {
+    "@twilio-paste/box": "^2.5.4",
+    "@twilio-paste/design-tokens": "^5.2.2",
+    "@twilio-paste/icons": "^2.2.9",
+    "@twilio-paste/reakit-library": "^0.6.0",
+    "@twilio-paste/spinner": "^1.2.18",
+    "@twilio-paste/style-props": "^1.2.5",
+    "@twilio-paste/styling-library": "^0.1.0",
+    "@twilio-paste/text": "^2.1.15",
+    "@twilio-paste/theme": "^3.2.5",
+    "@twilio-paste/tooltip-primitive": "^0.1.0",
+    "@twilio-paste/types": "^3.0.9"
+  }
+}
--- a/packages/paste-core/components/tooltip/src/index.tsx
+++ b/packages/paste-core/components/tooltip/src/index.tsx
@@ -1,0 +1,70 @@
+import * as React from 'react';
+import * as PropTypes from 'prop-types';
+import {useUID} from 'react-uid';
+import {Box, BoxProps, safelySpreadBoxProps} from '@twilio-paste/box';
+import {StyledBase} from '@twilio-paste/theme';
+import {Text} from '@twilio-paste/text';
+import {
+  TooltipPrimitiveInitialState,
+  useTooltipPrimitiveState,
+  TooltipPrimitive,
+  TooltipPrimitiveReference,
+} from '@twilio-paste/tooltip-primitive';
+import {TooltipArrow} from './TooltipArrow';
+
+const StyledTooltip = React.forwardRef<HTMLDivElement, BoxProps>(({style, ...props}, ref) => {
+  return (
+    <Box
+      {...safelySpreadBoxProps(props)}
+      backgroundColor="colorBackgroundBody"
+      borderColor="colorBorderLight"
+      borderRadius="borderRadius20"
+      borderStyle="solid"
+      borderWidth="borderWidth10"
+      boxShadow="shadowCard"
+      maxWidth="size30"
+      padding="space40"
+      paddingBottom="space30"
+      paddingTop="space30"
+      zIndex="zIndex90"
+      _focus={{outline: 'none'}}
+      style={style}
+      ref={ref}
+    />
+  );
+});
+
+export interface TooltipProps extends TooltipPrimitiveInitialState {
+  children: NonNullable<React.ReactElement>;
+  text: string;
+}
+
+const Tooltip = React.forwardRef<HTMLDivElement, TooltipProps>(({baseId, children, text, ...props}, ref) => {
+  const tooltip = useTooltipPrimitiveState({baseId: `paste-tooltip-${useUID()}`, ...props});
+  return (
+    <>
+      {React.Children.only(
+        <TooltipPrimitiveReference {...tooltip} ref={ref}>
+          {referenceProps => React.cloneElement(children, referenceProps)}
+        </TooltipPrimitiveReference>
+      )}
+      <TooltipPrimitive {...tooltip} {...props} as={StyledTooltip}>
+        <TooltipArrow {...tooltip} />
+        {/* import Paste Theme Based Styles due to portal positioning. */}
+        <StyledBase>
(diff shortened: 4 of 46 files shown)


packages/paste-core/components/tooltip/src/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+  children: NonNullable<React.ReactElement>;
+  text: string;
+}
+
+const Tooltip = React.forwardRef<HTMLDivElement, TooltipProps>(({children, text, ...props}, ref) => {
+  const tooltip = useTooltipPrimitiveState({...props});
+  return (
+    <>
+      {React.Children.only(
+        <TooltipPrimitiveReference {...tooltip} ref={ref} {...children.props}>
```
TheSisb, line 46: Just curious, why are we passing `children.props` to the PrimitiveReference?

packages/paste-core/components/tooltip/stories/index.stories.tsx
```diff
@@ -0,0 +1,116 @@
...
+    </Tooltip>
+  );
+};
+
+const IconButtonExample: React.FC<{}> = () => {
+  const id = useUID();
+  return (
+    <Tooltip baseId={id} text="Welcome to Paste!">
+      <Button variant="secondary" size="icon">
+        <InformationIcon decorative={false} title="Open Tooltip" />
```
TheSisb, line 26: This is an interesting example... What happens if you hover the icon svg? Does the title appear behind the tooltip?

Would it be appropriate here to match the icon title to the tooltip text?

packages/paste-core/components/tooltip/__test__/index.spec.tsx
```diff
@@ -0,0 +1,45 @@
...
+      </Tooltip>
+    </Theme.Provider>
+  );
+};
+
+describe('Tooltip', () => {
+  describe('Render', () => {
+    it('should render', () => {
+      const {asFragment} = render(<TooltipMock />);
+      expect(asFragment()).toMatchSnapshot();
```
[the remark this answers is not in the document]
richbachman: I'm good with dropping this one. Its not really telling us much anyway.

packages/paste-core/components/tooltip/stories/index.stories.tsx
richbachman: It looks like the title appears above the tooltip, even with the highest zIndex. Maybe we add some guidance in the doc page about matching the title to the tooltip text. What do you think @TheSisb?

packages/paste-core/components/tooltip/src/index.tsx
richbachman: Actually that was left over from an example on the reakit site. I don't think we need to them now. The reference just needs to clone the child.

packages/paste-core/components/tooltip/stories/index.stories.tsx
TheSisb: I _think_ we can mark the icon as decorative and leverage the tooltip to inform users, but I haven't tested that theory.

Documentation would be great yeah 👍

packages/paste-core/components/tooltip/package.json
```diff
@@ -0,0 +1,62 @@
+{
+  "name": "@twilio-paste/tooltip",
+  "version": "0.0.0",
+  "category": "interaction",
+  "status": "alpha",
+  "description": "Twilio's Tooltip component.",
```
aayushpi, line 6:
```suggestion
  "description": "The Tooltip is a popup that displays clarifying text related to an element that has been focused or hovered on.",
```

aayushpi: In general, let's be more descriptive in these descriptions since they show up in search as well.

packages/paste-website/src/pages/components/tooltip/index.mdx
```diff
@@ -0,0 +1,222 @@
...
+
+## Guidelines
+
+### About Tooltip
+
+The Tooltip component is a popup that displays text related to an element when the element is focused via
+the keyboard or on mouse hover. Tooltip follows the
+[WAI-ARIA Tooltip Pattern](https://www.w3.org/TR/wai-aria-practices/#tooltip).
+
+### Accessibility
```
aayushpi, line 75: I'd like this section to be more about how to use tooltips in an accessible manner. Are the items below things developers need to do? Or are these essential for accessible tooltips that we have already done?

If it's the latter, lets get rid of it. If not, lets re-write these as instructional. For example: "Do not add actionable elements to your tooltip since a user may not be able to reach them".

packages/paste-website/src/pages/components/tooltip/index.mdx
```diff
@@ -0,0 +1,222 @@
...
+  }}
+>
+  {`<Tooltip baseId="tooltip" text="Black Lives Matter. We stand with the Black community.">
+  <Button variant="primary">✊ Action</Button>
+</Tooltip>`}
+</LivePreview>
+
+### Controlling placement
+
+The placement of the tooltip is configurable via the `placement` prop. The available placement options are
```
aayushpi, line 101: Is there a default? Is there any merit in adding "in case your placement leads to an invisible tooltip, it adjusts" (assuming it does)

packages/paste-website/src/pages/components/tooltip/index.mdx
```diff
@@ -0,0 +1,222 @@
...
+    Tooltip,
+    Button,
+  }}
+>
+  {`<Tooltip baseId="tooltip-right" text="Welcome to Paste!" placement="right-start">
+  <Button variant="primary">Open Tooltip</Button>
+</Tooltip>`}
+</LivePreview>
+
+### Custom element
```
aayushpi, line 115: Unclear what custom means here. Would it be better to say "Passing tooltips on typically unfocusable elements"? 

And if so, we should provide guidance saying that passing tooltips on items that users may not be able to guess as tooltips is probably not a good idea

packages/paste-website/src/pages/components/tooltip/index.mdx
richbachman: These are all things that we are doing already. We're only allowing text in the actual tooltip, so users won't be able to add actionable elements. Removing this section for now.

packages/paste-website/src/pages/components/tooltip/index.mdx
richbachman: Content added.

packages/paste-website/src/pages/components/tooltip/index.mdx
richbachman: I've changed this title, and added some more descriptive text, and a call out about the focusable element / abuse. Let me know if that clears things up.

packages/paste-core/components/tooltip/stories/index.stories.tsx
richbachman: @TheSisb I adjusted the doc and story some to inform users more about leveraging an icon.

SiTaggart: > I think we can mark the icon as decorative and leverage the tooltip to inform users

Unfortunately, _technically_ not. A tooltip should describe the thing you are pointing at, but if you remove the text content of the thing you’re describing, there’s nothing to describe. Making it worse is decorative icons are essentially removed from the DOM to screen reader users.

richbachman: @SiTaggart should the icon title match the tooltip text? Or be more like "open tooltip"?

TheSisb: @SiTaggart Isn't it safe in the fact that the icon is removed from the DOM? The only thing to point to is the button, which will have the correct tooltip information.

SiTaggart: So the nuance is the accessible name the icon is giving the button element. If you set the icon to be decorative we `aria-hidden="true"` the icon, which removes the accessible name from the button. Effectively it removes the button text entirely, so the tooltip _describes_ a labelless button.

Often the tooltip text should not be the same as the button text, if you are describing it. It might be a pencil icon for an edit button, the button text might be "Edit" and the tooltip might describe what that means. They are independent pieces of information. 

image

image

packages/paste-website/src/pages/components/tooltip/index.mdx
```diff
@@ -0,0 +1,223 @@
...
+### Focusable element
+
+A tooltip can be placed on any **focusable** element, such as a button or a React component with a `ref`. Since our
+Box primitive passes a `ref`, it can be used to wrap a non-focusable element such as an icon. If you do use an icon,
+make sure you set the `decorative` prop to false, and give the icon a title. This ensures the icon is accessible
+to screen readers.
+
+<Callout variant="warning">
+  <CalloutText>
+    🚨<strong>CAUTION!</strong>🚨 Make sure not to abuse this method. Tooltips should only be placed on elements that
```
SiTaggart, line 116: So, I'm a little confused by what you're trying to say here as the example does what we say not to do.

This could be fixed by wrapping the icon in the example with an anchor, but it must be a natively focusable html element.

packages/paste-core/components/tooltip/stories/index.stories.tsx
SiTaggart: Have found a bug though, so that's exciting. As you can see in the screen shots the Description is "Open Tooltip" which is incorrect, it should be "Welcome to Paste". 

Reason is duplicate ids. The icon id's are unique but independently unique to the tooltip ids generated by reakit. I think we just need to set a default baseId to prefix `paste-tooltip` or something to the tooltip ids

image

TheSisb: Makes sense and good catch! We should prefix the icon IDs to `paste-icon-#` too 🤔

richbachman: @SiTaggart I've added a default baseId prefixed with `paste-tooltip-`, and using `useUID` to generate the number.

packages/paste-website/src/pages/components/tooltip/index.mdx
richbachman: @SiTaggart I've the example up. Removed this callout and simplified the text above. I also added a callout below about icon titles.

packages/paste-core/components/tooltip/stories/index.stories.tsx
```diff
@@ -0,0 +1,108 @@
...
+    </Tooltip>
+  );
+};
+
+const CustomExample: React.FC<{}> = () => {
+  return (
+    <Box display="flex" alignItems="center">
+      <Text as="span">Tooltip should appear from the icon.</Text>
+      <Tooltip text="Welcome to Paste!">
+        <a href="https://paste.twilio.com" style={{display: 'inline-flex'}}>
```
aayushpi, line 34:
```suggestion
        <a href="https://paste.twilio.design" style={{display: 'inline-flex'}}>
```

packages/paste-core/components/tooltip/package.json
```diff
@@ -0,0 +1,62 @@
...
+    "build:props": "typedoc --tsconfig ./tsconfig.json --json ./dist/prop-types.json",
+    "clean": "rm -rf ./dist && rm -rf tsconfig.build.tsbuildinfo && rm -rf .rpt2_cache",
+    "compile": "rollup -c --environment NODE_ENV:production",
+    "compile:dev": "rollup -cw --environment NODE_ENV:development",
+    "prepublishOnly": "yarn build",
+    "type-check": "tsc --noEmit"
+  },
+  "peerDependencies": {
+    "@twilio-paste/box": "^2.5.4",
+    "@twilio-paste/button": "^1.2.11",
```
TheSisb, line 32: This package isn't required. Stories work differently, so we don't need to list story packages in the package.json

packages/paste-core/components/tooltip/stories/index.stories.tsx
```diff
@@ -0,0 +1,108 @@
...
+    <Tooltip text="Welcome to Paste!" placement="right-start">
+      <Button variant="primary">Open Tooltip</Button>
+    </Tooltip>
+  );
+};
+
+const LeftExample: React.FC<{}> = () => {
+  return (
+    <Absolute preset="right" top={12} right={12}>
+      <Tooltip text="Welcome to Paste!" placement="left-start">
```
TheSisb, line 71: Since developers can't really know if the element they're wrapping in a tooltip will be on the edge of a screen, I believe the placement prop is more about customizing aesthetic.  Therefore, I think the stories might work better if you have 1 story called "Tooltip placements" with 4 triggers side-by-side showing all 4 placements in the same story, and another called "Automatic edge placement" which has four triggers on all 4 corners of the frame with no placement set, to show how it automatically just works.

TheSisb: Also it would be cool to have a story to show automatic adjustments if the trigger is, say, on the bottom right of the frame and placement is set to "bottom" (which would ordinarily be bad in that situation)

packages/paste-core/components/tooltip/stories/index.stories.tsx
```diff
@@ -0,0 +1,108 @@
...
+        <Button variant="primary">Open Tooltip</Button>
+      </Tooltip>
+    </Absolute>
+  );
+};
+
+storiesOf('Components|Tooltip', module)
+  .addDecorator(withKnobs)
+  .add('Default', () => {
+    return <Example />;
```
TheSisb, line 81: Would be nice to pass a knob here to toggle the placement and text

packages/paste-website/src/pages/components/tooltip/index.mdx
```diff
@@ -0,0 +1,222 @@
...
+  {`<Tooltip text="Black Lives Matter. We stand with the Black community.">
+  <Button variant="primary">✊ Action</Button>
+</Tooltip>`}
+</LivePreview>
+
+### Controlling placement
+
+The placement of the tooltip is configurable via the `placement` prop. The available placement options are
+available in the [props description here](#tooltip). By default the tooltip is positioned automatically so
+you don't have to worry about it going off the page, or being covered by another element.
```
TheSisb, line 94: Might be worth leaving a note about placement if setting it forces the tooltip off-screen, what happens? Or rewording " By default the tooltip is positioned automatically" to something like "no matter the placement, the tooltip will always adjust to remain in the screen"?

packages/paste-core/components/tooltip/stories/index.stories.tsx
richbachman: I've made these story adjusments.

packages/paste-core/components/tooltip/stories/index.stories.tsx
richbachman: I passed a text knob. I tried the placement one, but it doesn't want to change when the knob is changed. Not unless I refresh and change it manually. Maybe because its a hook in the component?

packages/paste-website/src/pages/components/tooltip/index.mdx
richbachman: Added some more text.

packages/paste-core/components/tooltip/stories/index.stories.tsx
```diff
@@ -1,108 +1,115 @@
...
+    );
   })
-  .add('Tooltip Left', () => {
-    return <LeftExample />;
+  .add('Tooltip Placements', () => {
+    return (
+      <Box marginTop="space120" paddingTop="space120" paddingLeft="space60">
+        <Stack orientation="horizontal" spacing="space40">
+          <Tooltip text="Welcome to Paste!" placement="top">
+            <Button variant="primary">Open Tooltip</Button>
```
TheSisb, line 26: Ultra nit but can you give these button texts the placement too? Like `Open to the top/right/left`

richbachman: Done!

SiTaggart:
Lets not have examples on the site not using our components and using inline styles 

image

Do that with the rebase, commit tidy and conflicts to reduce the amount of builds running, because afterwards it's good to go

packages/paste-core/components/anchor/src/index.tsx
```diff
@@ -53,11 +53,11 @@ const StyledAnchor = styled.a(
   )
 );
 
-const Anchor: React.FC<AnchorProps> = props => (
-  <StyledAnchor {...secureExternalLink(props.href)} {...props}>
+const Anchor = React.forwardRef<HTMLAnchorElement, AnchorProps>(({...props}, ref) => (
```
TheSisb, line 56: `{...props}` can be simplified to `props`

PR #542, opened 2020-06-30
feat(avatar): create alpha avatar

SiTaggart:
Create the alpha avatar component. 

- mimic icon sizes
- initials
- image
- response sizes
- tests
- stories


packages/paste-core/components/avatar/__tests__/avatar.test.tsx
```diff
@@ -0,0 +1,88 @@
...
+        expect(getCorrespondingFontSizeFromSizeToken('sizeIcon70')).toEqual('fontSize50');
+        expect(getCorrespondingFontSizeFromSizeToken('sizeIcon80')).toEqual('fontSize60');
+        expect(getCorrespondingFontSizeFromSizeToken('sizeIcon90')).toEqual('fontSize60');
+        expect(getCorrespondingFontSizeFromSizeToken('sizeIcon100')).toEqual('fontSize80');
+        expect(getCorrespondingFontSizeFromSizeToken('sizeIcon110')).toEqual('fontSize90');
+      });
+    });
+
+    describe('getComputedTokenNames', () => {
+      it('should handle singe size values', () => {
```
richbachman, line 36:
```suggestion
      it('should handle single size values', () => {
```

packages/paste-core/components/avatar/src/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+          textAlign="center"
+        >
+          {children}
+        </Text>
+      </Box>
+    </AvatarContext.Provider>
+  );
+};
+
+Avatar.displayName = 'Avatar';
```
richbachman, line 47: Looks like we're missing PropTypes for Avatar here.

packages/paste-core/components/avatar/src/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+  size: IconSize;
+}
+export interface AvatarImageProps {
+  src: string;
+  alt: string;
+}
+
+const AvatarContext = React.createContext<AvatarProps>({size: 'sizeIcon10'});
+
+export const AvatarImg: React.FC<AvatarImageProps> = ({src, alt, ...props}) => {
```
richbachman, line 17: Do we need PropTypes and a displayName for AvatarImg?

packages/paste-core/components/avatar/__tests__/avatar.test.tsx
```diff
@@ -0,0 +1,88 @@
...
+  getCorrespondingLineHeightFromSizeToken,
+  getCorrespondingFontSizeFromSizeToken,
+  getComputedTokenNames,
+} from '../src/utils';
+
+describe('Avatar', () => {
+  describe('Utils', () => {
+    describe('getCorrespondingLineHeightFromSizeToken', () => {
+      it('should return a line height to match icon size', () => {
+        expect(getCorrespondingLineHeightFromSizeToken('sizeIcon20')).toEqual('lineHeight20');
```
TheSisb, line 15: Should we add every possible sizeIcon / lineHeight pairing here, like we do for fontSize below?

packages/paste-core/components/avatar/src/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+  const {size} = React.useContext(AvatarContext);
+  return <Box {...safelySpreadBoxProps(props)} as="img" alt={alt} maxWidth="100%" src={src} size={size} />;
```
TheSisb, line 19: I think we should add `objectFit: 'cover'` and tweak `maxWidth and size` to `width: 100%, height: 100%`. This will allow us to get rid of the context provider, as the avatar image would just be sized to the container automatically.

packages/paste-core/components/avatar/src/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+  return (
+    <AvatarContext.Provider value={{size}}>
+      <Box
+        as="div"
+        backgroundColor="colorBackgroundPrimaryLight"
+        borderRadius="borderRadiusCircle"
+        overflow="hidden"
+        size={size}
+      >
+        <Text
```
TheSisb, line 33: Currently this wraps the image component if it's passed as a child.  What if we check if `typeof children === 'string'` and wrap it in a Text then?

packages/paste-core/components/avatar/src/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+  return <Box {...safelySpreadBoxProps(props)} as="img" alt={alt} maxWidth="100%" src={src} size={size} />;
+};
+
+const Avatar: React.FC<AvatarProps> = ({children, size = 'sizeIcon70'}) => {
+  const computedTokenNames = getComputedTokenNames(size);
+  return (
+    <AvatarContext.Provider value={{size}}>
+      <Box
+        as="div"
+        backgroundColor="colorBackgroundPrimaryLight"
```
TheSisb, line 28: Should this be customizable or nah? Like have 4 variant colors for text avatars?

I haven't ran this either, but I wonder if the color would bleed under the Avatar image in some browsers too 🤔

packages/paste-core/components/avatar/src/index.tsx
```diff
@@ -0,0 +1,49 @@
...
+export const AvatarImg: React.FC<AvatarImageProps> = ({src, alt, ...props}) => {
+  const {size} = React.useContext(AvatarContext);
+  return <Box {...safelySpreadBoxProps(props)} as="img" alt={alt} maxWidth="100%" src={src} size={size} />;
+};
+
+const Avatar: React.FC<AvatarProps> = ({children, size = 'sizeIcon70'}) => {
+  const computedTokenNames = getComputedTokenNames(size);
+  return (
+    <AvatarContext.Provider value={{size}}>
+      <Box
```
TheSisb, line 26: Would be nice to pass this an aria-label optionally

packages/paste-core/components/avatar/stories/index.stories.tsx
```diff
@@ -0,0 +1,80 @@
...
+        <Avatar size="sizeIcon100">ST</Avatar>
+        <Avatar size="sizeIcon110">ST</Avatar>
+      </Stack>
+    );
+  })
+  .add('Image', () => {
+    return (
+      <Stack orientation="horizontal" spacing="space40">
+        <Avatar size="sizeIcon10">
+          <AvatarImg alt="avatar example" src="/avatars/avatar2.png" />
```
TheSisb, line 30: Looking at this usage code, I'm starting to wonder what value having a composable `AvatarImg` provides 🤔 , maybe we just these as props to simplify the whole thing, and make `children?: string`

packages/paste-core/components/avatar/src/index.tsx
SiTaggart: It's not supported in IE11 unfortunately

packages/paste-core/components/avatar/src/index.tsx
SiTaggart: For now, not. We'll likely need to work on a palette of colors for things like badges and these which aren't necessarily usage specific, but like visualization colors or something. Maybe the same palette collection as charts, maybe a sibling palette. Either way, until we have that this is "fine" for the alpha I suppose

packages/paste-core/components/avatar/src/index.tsx
SiTaggart: Yeah, I should be safelySpreadingProps here really. Good catch

packages/paste-core/components/avatar/src/index.tsx
SiTaggart: Yeah, we could do that. Do you think it's worth the conditional check?

packages/paste-core/components/avatar/stories/index.stories.tsx
SiTaggart: Just pass `alt` and `src` to avatar if you want an image?

packages/paste-core/components/avatar/__tests__/avatar.test.tsx
SiTaggart: Doesn't do any harm

packages/paste-core/components/avatar/stories/index.stories.tsx
SiTaggart: Going with this, tidies things up a bunch

packages/paste-core/components/avatar/src/index.tsx
SiTaggart: Now that we don't have a child image component, it makes a bunch of sense to conditionally render the text

packages/paste-core/components/avatar/src/index.tsx
SiTaggart: No longer needed as we've removed the need for the context

packages/paste-core/components/avatar/src/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+    <Box
+      {...safelySpreadBoxProps(props)}
+      as="div"
+      backgroundColor="colorBackgroundPrimaryLight"
+      borderRadius="borderRadiusCircle"
+      overflow="hidden"
+      size={size}
+    >
+      {src != null ? (
+        <Box {...safelySpreadBoxProps(props)} as="img" alt={name} maxWidth="100%" src={src} size={size} title={name} />
```
richbachman, line 28: Does having the `title` and `alt` tags on the image version announce to screen readers two times?

SiTaggart: It shouldn't do, there is an algorithm for calculating the accessible name of the element. Both alt and title contribute to the accessible name, but alt overrides title in priority. As long as they are the same it's not a problem

packages/paste-website/src/pages/components/avatar/index.mdx
```diff
@@ -0,0 +1,216 @@
...
+
+##### `size` IconSize[]
+
+Responsive size property that takes IconSize token names.
+
+##### `src` string
+
+Used to set the image `src` attribute when setting an image as the Avatar.
+
+<Box marginTop="space90">
```
richbachman, line 210: Reminder to add the `ChangelogRevealer` here.

packages/paste-core/components/avatar/src/index.tsx
richbachman: Got it. Thanks for the clarification.

packages/paste-core/components/avatar/src/utils.ts
```diff
@@ -0,0 +1,69 @@
...
+    .reduce((previous, currentInitial, index, initials) => {
+      if (index === 0 || index === initials.length - 1) {
+        return `${previous}${currentInitial}`;
+      }
+      return previous;
+    }, '');
+};
+
+export const getCorrespondingLineHeightFromSizeToken = (size: IconSizeOptions): LineHeightOptions =>
+  `lineHeight${size.replace('sizeIcon', '')}` as LineHeightOptions;
```
TheSisb, line 23: nit: clearer as 
`size.replace('sizeIcon', 'lineHeight')`

SiTaggart: 🤦

PR #545, opened 2020-07-01
chore(menu): updated font color, width, and zindex styles

TheSisb:
Updates:
- Changed default fontColor for menuitem to colorText from colorTextLink
- Updated minWidth to the smallest token (sketch had a smaller value :/)
- Updated zIndex to the 20 value. For any overlay type component I think 20 is a safer default value because developer code can occasionally use zIndex10.

### DIFF
--- a/packages/paste-core/components/menu/src/index.tsx
+++ b/packages/paste-core/components/menu/src/index.tsx
@@ -41,8 +41,8 @@ const StyledMenu = React.forwardRef<HTMLDivElement, BoxProps>(({style, ...props}
       borderRadius="borderRadius20"
       boxShadow="shadowCard"
       maxWidth="size30"
-      minWidth="size30"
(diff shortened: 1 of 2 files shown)


packages/paste-core/components/menu/src/index.tsx
```diff
@@ -41,8 +41,8 @@ const StyledMenu = React.forwardRef<HTMLDivElement, BoxProps>(({style, ...props}
       borderRadius="borderRadius20"
       boxShadow="shadowCard"
       maxWidth="size30"
-      minWidth="size30"
-      zIndex="zIndex10"
+      minWidth="size10"
+      zIndex="zIndex20"
```
richbachman, line 45: These should be made to popover as well. I'll make a ticket for that.

Issue #556, opened 2020-07-07
Twilio Bootstrap styles conflict on <legend /> element


ghostling:
**Current behavior**

Currently observing this when using `<RadioGroup />` component. Twilio Bootstrap adds a bottom border to the "To get started choose" text below: 
![image](https://user-images.githubusercontent.com/2553665/86823175-6057de80-c041-11ea-81ad-04c816f3250e.png)

**Expected behavior**
No border.

richbachman:
Hi @ghostling. We just published a fix for this. If you update the forms package to the newest version, you should be good to go. Let us know if you have any other issues.

ghostling:
Thanks @richbachman ! It works perfectly.

PR #557, opened 2020-07-07
fix(form): add border-bottom-width rule for bootstrap override

richbachman:
Resolves: https://github.com/twilio-labs/paste/issues/556

- [x] Added `BorderBottomWidth: borderWidth0`


packages/paste-core/components/form/src/FormLabel.tsx
```diff
@@ -35,6 +35,7 @@ const FormLabel: React.FC<FormLabelProps> = ({as, marginBottom, required, disabl
     <Box
       {...safelySpreadBoxProps(props)}
       as={as}
+      borderBottomWidth="borderWidth0"
```
SiTaggart, line 38: Is it possible to leave a comment here to warn our future selves why we're setting this?

PR #571, opened 2020-07-13
feat(design-tokens): add color-border-light and color-border-lighter

richbachman:
- [x] Added color-border-light and color-border-lighter
- [x] Updated use of color-border-light to color-border-lighter


packages/paste-design-tokens/tokens/global/border-color.yml
```diff
@@ -5,15 +5,18 @@ imports:
   - ../aliases/color.yml
   - ../aliases/color-palette.yml
 props:
+  color-border-dark:
```
SiTaggart, line 8: What is dark for? Do we use it? Was it in the ticket?

richbachman: Looks like we're using it in Checkbox. I just moved the order of the tokens around to go from dark -> light.

SiTaggart: OMG, yup. I see it now 🤦

PR #577, opened 2020-07-14
chore(popover): remove minWidth and change zIndex

richbachman:
- [x] Updated popover styles to match minWidth and zIndex of Tooltip
- [x] Updated anatomy table on Popover doc page
- [x] Updated anatomy table on Tooltip doc page


packages/paste-core/components/popover/src/Popover.tsx
```diff
@@ -25,8 +25,7 @@ const StyledPopover = React.forwardRef<HTMLDivElement, BoxProps>(({style, ...pro
       borderRadius="borderRadius20"
       boxShadow="shadowCard"
       maxWidth="size30"
-      minWidth="size30"
-      zIndex="zIndex10"
+      zIndex="zIndex90"
```
TheSisb, line 28: Let's make this 80 since tooltip is 90 and should appear over popovers

richbachman: Sounds good.

PR #591, opened 2020-07-17
feat(design-tokens): add negative space tokens

richbachman:
- [x] Add space-negative tokens
- [x] Restructure Grid functions to used space-negative tokens
- [x] Rename helpers file to utils to be more inline with other components
- [x] Rework tests to use react testing library
  - Removed a lot of snapshots since Chromatic and Applitools covers them


packages/paste-core/layout/grid/src/utils.tsx
```diff
@@ -12,34 +11,66 @@ import {
...
 };
 
-// Gets the gutter and returns the value to be used as padding for Columns
-export const getColumnGutters = (theme: ThemeShape, gutter?: Space): Padding => {
+// Gets the gutter and returns the value to be used as negative margin to Grid
+export const getOuterGutterPull = (gutter?: Space): Margin => {
   if (Array.isArray(gutter)) {
     return (gutter as SpaceOptions[]).map((value: SpaceOptions) => {
-      return `${theme.space[value]}` as SpaceOptions;
+      return getCorrespondingNegativeSpacingToken(value);
```
SiTaggart, line 65: This could be simplified as a string replacement `value.replace('space', 'spaceNegative')` removing the need for the function entirely.

richbachman: I've modified that, and updated the tests.

packages/paste-core/layout/grid/src/utils.tsx
```diff
@@ -13,33 +12,30 @@ import {
...
-
-// Gets the gutter and returns the value to be used as padding for Columns
-export const getColumnGutters = (theme: ThemeShape, gutter?: Space): Padding => {
-  if (Array.isArray(gutter)) {
-    return (gutter as SpaceOptions[]).map((value: SpaceOptions) => {
-      return `${theme.space[value]}` as SpaceOptions;
-    });
+  if (gutter && gutter !== 'space0') {
+    // Ignoring this ts error becuase we're checking for space0 already.
+    // @ts-ignore
```
SiTaggart, line 30: What's the error? Would like to avoid `@ts-ignore` if I can help

packages/paste-core/layout/grid/src/utils.tsx
```diff
@@ -13,33 +12,30 @@ import {
...
+    return gutter.replace('space', 'spaceNegative') as SpaceOptions;
   }
 
-  if (gutter) {
-    return `${theme.space[gutter as SpaceOptions]}` as SpaceOptions;
+  if (gutter === 'space0') {
```
TheSisb, line 34: nit: move this `if` clause above the previous one and make the above one just `if (gutter)`

Would that fix the TS issue? 🤔

richbachman: Looks like that doesn't help with the TS error.

packages/paste-core/layout/grid/src/utils.tsx
```diff
@@ -13,33 +12,31 @@ import {
...
+    return 'space0';
   }
 
   if (gutter) {
-    return `${theme.space[gutter as SpaceOptions]}` as SpaceOptions;
+    // Ignoring this ts error: Type '"space0"' has no call signatures.
+    // We're checking for space0 already above.
+    // @ts-ignore
+    return gutter.replace('space', 'spaceNegative') as SpaceOptions;
```
SiTaggart, line 36: For whatever reason typescript thinks gutter _might_ be a number

Screen Shot 2020-07-22 at 10 39 17

if you add `&& typeof gutter === 'string'` to the condition, the error seems to go away

richbachman: That is a really odd one, since we're setting the type to `Space`.

PR #592, opened 2020-07-17
chore: remove publish checks

richbachman:
- [x] Change package publish to only run if commit message doesn't have `chore(release): publish`
- [x] Changed Chromatic to only run if commit message doesn't have `chore(release): publish`

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -6,14 +6,59 @@ on:
       - main
 
 jobs:
-  deploy:
+  main_build_test:
     runs-on: ubuntu-latest
     strategy:
       matrix:
         node-version: [10.16.x]
-
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@@v2
+      - uses: actions/setup-node@v1
+      - name: Get yarn cache
+        id: yarn-cache
+        run: echo "::set-output name=dir::$(yarn cache dir)"
+      - uses: actions/cache@v1
+        with:
+          path: ${{ steps.yarn-cache.outputs.dir }}
+          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-yarn-
+      - name: restore lerna
+        uses: actions/cache@master
+        with:
+          path: |
+            node_modules
+            */*/node_modules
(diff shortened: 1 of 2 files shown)


SiTaggart:
The ticket may not have been fully clear, but the intention was not to stop the checks altogether, but only run them when it's not a publish commit.

image

All checks including chromatics need to run on the merge commit to make sure everything is good before we publish the packages and to update the latest baselines for chromatic and applitool comparisons. But the exact same checks are run again in the "chore(release): publish" commit. But nothing has changed and we're running them for no reason. 

So conditionally, check if the last/current commit message was "chore(release): publish" if it was don't run most of the checks.

richbachman:
@SiTaggart, got it. I'll work through that on Monday. I felt like this was a little too easy.

richbachman:
@SiTaggart I split the publish action into two jobs. One that builds the main branch and runs tests, and another that runs conditionally that publishes to npm and deploys storybook. Let me know if I'm on the right track.

.github/workflows/publish.yml
```diff
@@ -6,24 +6,30 @@ on:
       - main
 
 jobs:
-  deploy:
```
SiTaggart, line 9: Honestly, I reckon you could probably do exactly the same here as the chromatic job. Leave everything as is and run the entire job _only_ if it's not `chore(release): publish`.

Given they are all reliant on the build steps it just seems easier, that way we don't run applitools either.

.github/workflows/publish.yml
```diff
@@ -6,24 +6,30 @@ on:
...
+  main_build_test:
     runs-on: ubuntu-latest
     strategy:
       matrix:
         node-version: [10.16.x]
-
     steps:
-      - uses: actions/checkout@v1
+      - uses: actions/checkout@@v2
```
SiTaggart, line 15: you've got some double `@` for the checkout version, but as an aside I had a _bunch_ of issues with checkout v2, hence publishing uses v1 right now.

richbachman: Ah.. I saw v2 in pull_request workflow, so wondered if we just missed it here. Either way, reverting this change in favor of your previous suggestion.

.github/workflows/publish.yml
richbachman: That makes sense. I've made that change.

PR #598, opened 2020-07-22
feat(disclosure): add state prop

richbachman:
- [x] Add `state` prop
- [x] Add example story
- [x] Add tests
- [x] Update Disclosure doc with `state` example

### DIFF
--- a/packages/paste-core/components/disclosure/__tests__/disclosure.test.tsx
+++ b/packages/paste-core/components/disclosure/__tests__/disclosure.test.tsx
@@ -1,7 +1,15 @@
 import * as React from 'react';
 import {axe} from 'jest-axe';
 import {render, screen, fireEvent} from '@testing-library/react';
-import {Disclosure, DisclosureContent, DisclosureHeading, DisclosureHeadingProps, DisclosureProps} from '../src';
+import {
+  Disclosure,
+  DisclosureContent,
+  DisclosureHeading,
+  DisclosureHeadingProps,
+  DisclosureProps,
+  DisclosureStateReturn,
+  useDisclosureState,
+} from '../src';
 
 export const MockDisclosure: React.FC<{
   visible?: DisclosureProps['visible'];
@@ -15,6 +23,32 @@ export const MockDisclosure: React.FC<{
       </DisclosureHeading>
       <DisclosureContent data-testid="disclosure">Disclosure content</DisclosureContent>
     </Disclosure>
+  );
+};
+
+const useVisibleDisclosureState = (): DisclosureStateReturn => {
+  const disclosure = useDisclosureState();
+  const [visible, setVisible] = React.useState(true);
+  return {
+    ...disclosure,
+    visible,
+    toggle: () => {
+      setVisible(false);
+    },
+  };
+};
+
+const StateHookMock: React.FC = () => {
+  const disclosure = useVisibleDisclosureState();
+  return (
+    <>
+      <Disclosure variant="contained" state={disclosure}>
+        <DisclosureHeading as="h2" variant="heading20">
+          Clickable heading
+        </DisclosureHeading>
+        <DisclosureContent>Disclosure content</DisclosureContent>
+      </Disclosure>
+    </>
   );
 };
 
@@ -49,6 +83,13 @@ describe('Disclosure', () => {
     const renderedDisclosureButton = screen.getByRole('button');
     expect(renderedDisclosureButton.getAttribute('aria-disabled')).toEqual('true');
     expect(renderedDisclosureButton.getAttribute('tabindex')).toEqual('0');
+  });
+  it('should render a disclosure open and update attributes when clicked using a state hook', () => {
+    render(<StateHookMock />);
+    const renderedDisclosureButton = screen.getByRole('button');
+    expect(renderedDisclosureButton.getAttribute('aria-expanded')).toEqual('true');
+    fireEvent.click(renderedDisclosureButton);
+    expect(renderedDisclosureButton.getAttribute('aria-expanded')).toEqual('false');
   });
   describe('accessibility', () => {
     it('should have no accessibility violations', async () => {
--- a/packages/paste-core/components/disclosure/src/Disclosure.tsx
+++ b/packages/paste-core/components/disclosure/src/Disclosure.tsx
@@ -16,12 +16,18 @@ export interface DisclosureContextProps {
 
 export const DisclosureContext = React.createContext<DisclosureContextProps>({} as any);
 
+export interface DisclosureStateReturn extends DisclosurePrimitveStateReturn {
+  [key: string]: any;
+}
+
 export interface DisclosureProps extends DisclosurePrimitiveInitialState {
   children: NonNullable<React.ReactNode>;
+  state?: DisclosureStateReturn;
   variant?: Variants;
 }
-const Disclosure: React.FC<DisclosureProps> = ({children, variant = 'default', ...props}) => {
-  const disclosure = useDisclosurePrimitiveState({...props});
+
+const Disclosure: React.FC<DisclosureProps> = ({children, variant = 'default', state, ...props}) => {
+  const disclosure = state || useDisclosurePrimitiveState({...props});
   const disclosureContext = {
     disclosure,
     variant,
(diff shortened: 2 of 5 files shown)


packages/paste-website/src/pages/components/disclosure/index.mdx
```diff
@@ -241,6 +242,36 @@ You can control the initial state of the disclosure to be expanded or collapsed
 </Disclosure>`}
 </LivePreview>
 
+### Using state hooks
+
+<Callout variant="warning">
+  <CalloutTitle>🚨 ATTENTION!! 🚨</CalloutTitle>
```
hharnisc, line 248: Perfect! Glad this is being called out as _at your own risk_

packages/paste-core/components/disclosure/src/Disclosure.tsx
```diff
@@ -16,12 +16,20 @@ export interface DisclosureContextProps {
...
 export interface DisclosureProps extends DisclosurePrimitiveInitialState {
   children: NonNullable<React.ReactNode>;
+  stateHook?: DisclosureStateReturn;
   variant?: Variants;
 }
-const Disclosure: React.FC<DisclosureProps> = ({children, variant = 'default', ...props}) => {
-  const disclosure = useDisclosurePrimitiveState({...props});
+
+const Disclosure: React.FC<DisclosureProps> = ({children, variant = 'default', stateHook, ...props}) => {
+  const disclosure = stateHook || useDisclosurePrimitiveState({...props});
```
hharnisc, line 32: 👍 IMO this is a viable path forward

richbachman:
@hharnisc interesting idea. Something we might look into in the future. If this current route serves your needs at the moment, let's run with this idea.

packages/paste-core/components/disclosure/src/Disclosure.tsx
```diff
@@ -16,12 +16,20 @@ export interface DisclosureContextProps {
 
 export const DisclosureContext = React.createContext<DisclosureContextProps>({} as any);
 
+interface LooseObject {
```
SiTaggart, line 19: Nit, I'd be tempted to lose this interface and just declare the `[key: string]: any` as a property on the DisclosureStateReturn interface

packages/paste-website/src/pages/components/disclosure/index.mdx
```diff
@@ -241,6 +242,36 @@ You can control the initial state of the disclosure to be expanded or collapsed
...
+    your application.
+  </CalloutText>
+</Callout>
+
+The `state` prop allows you to hook into, and keep in sync the state of the Disclosure and the state of your application.
+
+This will allow you to set initial state and merge the returned state from the hook with the application state,
+and keep it in sync with user interactions.
+
+In the example below, we're using a `state` to control visibility, and setting a transition delay of one second.
```
SiTaggart, line 260: This isn't necessarily true, I mean it kind of is as we delay the showing of the disclosure but we're not transitioning anything. We're also not doing anything with the transition state that we're creating. 

For this and the story, as a reader I'm not seeing the direct effect of setting extra state. In Reakit they at least use it to show a loading message before the toggle happens so I can see the connection directly in the example.

I'd suggest reworking this example in the docs and stories to be a little more visible of the effect you can take with mixed in state.

packages/paste-core/components/disclosure/__tests__/disclosure.test.tsx
```diff
@@ -50,6 +81,17 @@ describe('Disclosure', () => {
...
   });
+  it('should render a disclosure open using a state hook', () => {
+    render(<StateHookMock />);
+    const renderedDisclosureButton = screen.getByRole('button');
+    expect(renderedDisclosureButton.getAttribute('aria-expanded')).toEqual('true');
+  });
+  it('should update attributes when clicked using a state hook', () => {
+    render(<StateHookMock />);
+    const renderedDisclosureButton = screen.getByRole('button');
+    fireEvent.click(renderedDisclosureButton);
```
TheSisb, line 92: Prefer merging this with the previous `it()` call, to avoid having to render the same mock twice.

packages/paste-core/components/disclosure/src/Disclosure.tsx
```diff
@@ -16,12 +16,20 @@ export interface DisclosureContextProps {
 
 export const DisclosureContext = React.createContext<DisclosureContextProps>({} as any);
```
TheSisb, line 17: You aughta be able to type this as DisclosurePrimitveState I think?

packages/paste-core/components/disclosure/__tests__/disclosure.test.tsx
```diff
@@ -18,6 +26,29 @@ export const MockDisclosure: React.FC<{
   );
 };
 
+const useVisibleDisclosureState = (): DisclosureStateReturn => {
+  const disclosure = useDisclosureState({visible: true});
+  return {
+    ...disclosure,
+    toggle: () => {
+      disclosure.toggle();
```
TheSisb, line 34: This kind of doesn't test anything...
```
toggle: () => {disclosure.toggle();}
```
is equivalent to
```
toggle: disclosure.toggle
```
which does nothing.  Might as well return `disclosure` directly :P



Something like this might be better:
```js
const useVisibleDisclosureState = (): DisclosureStateReturn => {
  const [visible, setVisible] = React.useState(true);
  const toggle = () => {
    setVisible(!visible)
  };

  const disclosure = useDisclosureState({visible});
  return {
    ...disclosure,
    visible,
    toggle
  }
}
```
The above clearly shows that we now control the visible state, and we pass the value down to Disclosure to render. (I didn't test this code btw, hope it works 🤞)

packages/paste-core/components/disclosure/__tests__/disclosure.test.tsx
```diff
@@ -18,6 +26,29 @@ export const MockDisclosure: React.FC<{
...
+    },
+  };
+};
+
+const StateHookMock: React.FC = () => {
+  return (
+    <>
+      <Disclosure variant="contained" state={useVisibleDisclosureState()}>
```
TheSisb, line 42: The only way React has to distinguish hooks is if a function starts with the word `use`, so let's prefer to keep these outside the `return` and placed at the top of the component closure. This makes it more clear that it isn't an ordinary function call but a React Hook behavior to future maintainers.

packages/paste-core/components/disclosure/src/Disclosure.tsx
richbachman: Not sure I'm following this one. Something like:
`export const DisclosureContext = React.createContext<DisclosureContextProps>({} as DisclosurePrimitveState);`?

packages/paste-core/components/disclosure/__tests__/disclosure.test.tsx
richbachman: I've restructured this function as suggested. Just had to change it slightly to get it working.

packages/paste-website/src/pages/components/disclosure/index.mdx
richbachman: I've updated the story and doc example. I also changed the example copy to use a more real world idea. Let me know if I should add more to that.

packages/paste-core/components/disclosure/src/Disclosure.tsx
TheSisb: Yeah, I don't know about setting `any` here. Maybe `Partial<DisclosurePrimitveState>` if `DisclosurePrimitveState` doesn't work?

packages/paste-core/components/disclosure/__tests__/disclosure.test.tsx
```diff
@@ -27,19 +27,22 @@ export const MockDisclosure: React.FC<{
...
+    visible,
     toggle: () => {
-      disclosure.toggle();
+      setVisible(false);
     },
   };
 };
 
 const StateHookMock: React.FC = () => {
+  const {...disclosure} = useVisibleDisclosureState();
```
TheSisb, line 42: I think this can just be `const disclosure =`

packages/paste-core/components/disclosure/src/Disclosure.tsx
richbachman: @TheSisb changing to that partial gets me this error: https://github.com/typescript-eslint/typescript-eslint/blob/v1.13.0/packages/eslint-plugin/docs/rules/no-object-literal-type-assertion.md

Looks like `any` might be the answer.

SiTaggart: This might have been one of those things where the typings from Reakit were a little jank too. Partial might do a decent enough job

PR #599, opened 2020-07-23
fix(radio): allow passing disable to individual radio in radioGroups

TheSisb:
Fixes https://github.com/twilio-labs/paste/discussions/535

- [x] Allow passing props directly to Radio, instead of solely retrieving them via RadioGroup. Defaults to direct props, fallback to context from RadioGroup.
- [x]  Radios now work in a controlled manner, and without RadioGroups.
- [x] Updated types to allow passing `border[dir]="none"` instead of never allowing `border` style prop.

### DIFF
--- a/packages/paste-core/components/form/__tests__/radio.test.tsx
+++ b/packages/paste-core/components/form/__tests__/radio.test.tsx
@@ -133,7 +133,7 @@ describe('Radio Group', () => {
         <Radio {...defaultProps}>foo</Radio>
       </RadioGroup>
     );
-    expect(getByRole('radio').name).toBe('bar');
+    expect(getByRole('radio').name).toBe(defaultProps.name);
   });
 
   it('renders a helpText message when helpText prop is present', () => {
--- a/packages/paste-core/components/form/src/Radio.tsx
+++ b/packages/paste-core/components/form/src/Radio.tsx
@@ -1,7 +1,6 @@
 import * as React from 'react';
 import * as PropTypes from 'prop-types';
 import {useUID} from 'react-uid';
-import {styled, css} from '@twilio-paste/styling-library';
 import {Box} from '@twilio-paste/box';
 import {
   BaseRadioCheckboxControl,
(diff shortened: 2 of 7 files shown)


packages/paste-core/components/form/__tests__/radio.test.tsx
```diff
@@ -133,7 +133,7 @@ describe('Radio Group', () => {
         <Radio {...defaultProps}>foo</Radio>
       </RadioGroup>
     );
-    expect(getByRole('radio').name).toBe('bar');
+    expect(getByRole('radio').name).toBe(defaultProps.name);
```
TheSisb, line 136: Not sure how this worked before...

packages/paste-core/components/form/src/Radio.tsx
```diff
@@ -1,59 +1,76 @@
...
-    padding: '0',
-    position: 'absolute',
-    textTransform: 'none',
-    whiteSpace: 'nowrap',
-    width: '1px',
-  })
+type HiddenRadioProps = Pick<RadioProps, 'checked' | 'value' | 'id' | 'disabled' | 'name' | 'onChange'> & {
+  ref?: any | undefined;
+};
+const HiddenRadio: React.FC<HiddenRadioProps> = props => (
```
SiTaggart, line 28: Now this is a functional component do you need to forward the ref to `Box` here too?

packages/paste-core/components/form/stories/radio.stories.tsx
```diff
@@ -94,6 +94,44 @@ storiesOf('Forms|Radio', module)
       </Radio>
     );
   })
+  .add('Radios - Controlled', () => {
+    const [value, setValue] = React.useState('1');
+    const handleChange = (event: React.ChangeEvent<HTMLInputElement>): void => setValue(event.target.value);
+    return (
+      <>
```
SiTaggart, line 101: Nit: we don't technically allow this in our docs, as a radio must be part of a radio group, but I get why it's pulled out to show controlled state works on it's own

TheSisb: Yeah I realize that this is a departure from the original design. I'm thinking we don't document this behavior _perse_ , but  it's helpful to showcase how it works in a controlled manner just in storybook for devs.

SiTaggart: Question; Are you able to understand how to use RadioGroup with Formik, given that RadioGroup is essentailly a requirement from our library for accessibility reasons?

Asked another way, by giving you an example that _doesn't_ use it, are we in danger of confusing people when we say "you must use the RadioGroup to wrap radios?

TheSisb: I haven't really dug into Formik in a long time, but iirc Formik works best with simple controlled components.  The fact that RadioGroup is controlling Radios would add complexity to implementation.

PR #602, opened 2020-07-23
feat(tabs): fitted tabs and update styles

richbachman:
- [x] Updated styles
- [x] Fitted tabs
- [x] Fitted tabs story
- [x] Added Fitted tabs to docs
- [x] Removed vertical tab examples from docs

### DIFF
--- a/packages/paste-core/components/tabs/src/Tab.tsx
+++ b/packages/paste-core/components/tabs/src/Tab.tsx
@@ -4,40 +4,75 @@ import {Box, safelySpreadBoxProps} from '@twilio-paste/box';
 import {Text} from '@twilio-paste/text';
 import {TabPrimitive} from '@twilio-paste/tabs-primitive';
 import {TabsContext} from './TabsContext';
+import {Orientation, Variants} from './types';
 
-type Orientation = 'horizontal' | 'vertical' | undefined;
-
-const getTabBoxStyles = (orientation: Orientation): {} => {
+// TODO:
+// Split vertical tabs into a separate component
+// because fitted tabs do nothing when orientation
+// is vertical, which feels broken. Vertical tabs
+// shouldn't get the fitted variant.
+const getTabBoxStyles = (orientation: Orientation, variant: Variants): {} => {
   switch (orientation) {
     case 'vertical':
       return {
         display: 'block',
-        right: '-2px',
-        paddingBottom: 'space40',
-        paddingTop: 'space40',
+        paddingBottom: 'space20',
+        paddingTop: 'space20',
         paddingLeft: 'space50',
         paddingRight: 'space50',
-        borderRightWidth: 'borderWidth20',
-        borderRightStyle: 'solid',
-        borderRightColor: 'transparent',
+        borderLeftWidth: 'borderWidth20',
+        borderLeftStyle: 'solid',
+        borderLeftColor: 'transparent',
+        marginBottom: 'space40',
+        _last: {
+          marginBottom: 'space0',
+        },
         _selected: {
-          borderRightColor: 'colorBorderPrimary',
+          borderLeftColor: 'colorBorderPrimary',
+        },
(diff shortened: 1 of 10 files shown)


packages/paste-core/components/tabs/src/Tab.tsx
```diff
@@ -6,39 +6,71 @@ import {TabPrimitive} from '@twilio-paste/tabs-primitive';
 import {TabsContext} from './TabsContext';
 
 type Orientation = 'horizontal' | 'vertical' | undefined;
+type Variant = 'flexed' | undefined;
```
SiTaggart, line 9: Thinking about this more and how it ties into the orientation. 

For a while I was thinking that orientation should be variants as they are significantly different visual appearances. But then I was wondering how that plays with `flexed`, do vertical tabs get flexed? I don't think they actually play well together.

I'm sure flexed only applies to horizontal.

I'm wondering if vertical tabs should be their own component, and not a variant. 🤔

richbachman: @SiTaggart as we discussed, I've added the TODO note about vertical tabs, and removed the vertical tabs from the doc page.

packages/paste-core/primitives/box/src/index.tsx
```diff
@@ -159,6 +161,7 @@ const extraConfig = system({
   willChange: true,
   textDecoration: true,
   textTransform: true,
+  textAlign: true,
```
SiTaggart, line 164: Does this need to be added to `safelySpreadBoxProps` (I realize there maybe more missing)

packages/paste-core/components/tabs/src/Tabs.tsx
```diff
@@ -4,12 +4,14 @@ import {Flex} from '@twilio-paste/flex';
 import {useTabPrimitiveState, TabPrimitiveInitialState} from '@twilio-paste/tabs-primitive';
 import {TabsContext} from './TabsContext';
 
-export type TabsProps = TabPrimitiveInitialState;
+export interface TabsProps extends TabPrimitiveInitialState {
+  variant?: 'fitted' | undefined;
```
SiTaggart, line 8: nit: probably would have been nicer to share these across the various components rather than repeat them 3 times

packages/paste-core/primitives/box/src/index.tsx
richbachman: @SiTaggart I added `TYPOGRAPHY_PROPS` to `safelySpreadBoxProps`. Does that cover it, or do I need to call out `textAlign` specifically?

packages/paste-core/components/tabs/src/Tabs.tsx
richbachman: I added a types file and moved `Orientation` and `Variants` types there.

packages/paste-core/primitives/box/src/index.tsx
SiTaggart: Dunno, you'd have to check whats in the typography props. if we list it in there then it's enough, if we don't then it needs to be a special prop we add on it's own to the object

serifluous:
Should both horizontal and vertical tabs support left/right as well as up/down arrow key support?

packages/paste-core/primitives/box/src/index.tsx
SiTaggart: OK, yeah it's listed in the props `https://github.com/twilio-labs/paste/blob/main/packages/paste-style-props/src/typography.ts#L8

SiTaggart:
> Should both horizontal and vertical tabs support left/right as well as up/down arrow key support?

if it's possible, this would be cool. But maybe in another ticker

richbachman:
@SiTaggart agreed. I'll create a ticket for our backlog.

PR #608, opened 2020-07-28
feat:(combobox):  option groups and controlled input

SiTaggart:
This PR introduces the ability to create options groups and use Combobox as a controlled input.

- Add the ability to create options groups within the Combobox Listbox.
- Add necessary props to use Combobox as a controlled component.
- Document options groups and controlled component

### DIFF
--- a/packages/paste-core/components/combobox/src/Combobox.tsx
+++ b/packages/paste-core/components/combobox/src/Combobox.tsx
@@ -1,8 +1,12 @@
 import * as React from 'react';
 import * as PropTypes from 'prop-types';
 import {css, styled} from '@twilio-paste/styling-library';
-import {uid, useUID} from 'react-uid';
-import {useComboboxPrimitive, UseComboboxPrimitiveProps} from '@twilio-paste/combobox-primitive';
+import {useUID, useUIDSeed} from 'react-uid';
+import {
+  useComboboxPrimitive,
+  UseComboboxPrimitiveProps,
+  UseComboboxPrimitiveState,
+} from '@twilio-paste/combobox-primitive';
 import {ChevronDownIcon} from '@twilio-paste/icons/esm/ChevronDownIcon';
 import {Box} from '@twilio-paste/box';
 import {
@@ -15,7 +19,11 @@ import {
 } from '@twilio-paste/form';
 import {ComboboxInputWrapper} from './ComboboxInputWrapper';
 import {ComboboxListbox} from './ComboboxListbox';
+import {ComboboxListboxGroup} from './ComboboxListboxGroup';
 import {ComboboxListboxOption} from './ComboboxListboxOption';
+
+// This module can only be referenced with ECMAScript imports/exports by turning on the 'esModuleInterop' flag and referencing its default export
+const groupBy = require('lodash.groupby');
 
 // Fixes chevron overlapping really long text
 /* eslint-disable emotion/syntax-preference */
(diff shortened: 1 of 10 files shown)


packages/paste-core/components/combobox/src/Combobox.tsx
```diff
@@ -15,8 +15,12 @@ import {
 } from '@twilio-paste/form';
 import {ComboboxInputWrapper} from './ComboboxInputWrapper';
 import {ComboboxListbox} from './ComboboxListbox';
+import {ComboBoxGroup} from './ComboboxGroup';
```
richbachman, line 18: Should this be named `ComboboxListboxGroup`?

packages/paste-core/components/combobox/src/Combobox.tsx
```diff
@@ -38,10 +43,11 @@ export interface ComboboxProps extends Omit<FormInputProps, 'id' | 'type'> {
   onIsOpenChange?: UseComboboxPrimitiveProps<any>['onIsOpenChange'];
   onSelectedItemChange?: UseComboboxPrimitiveProps<any>['onSelectedItemChange'];
   optionTemplate?: (item: string | {}) => React.ReactNode;
+  selectedItem?: UseComboboxPrimitiveProps<any>['selectedItem'];
```
TheSisb, line 46: Just curious why are these all typed to 'any' 🤔

packages/paste-core/components/combobox/src/Combobox.tsx
```diff
@@ -77,13 +87,61 @@ const Combobox = React.forwardRef<HTMLInputElement, ComboboxProps>(
     } = useComboboxPrimitive({
       initialSelectedItem,
       items,
-      itemToString: itemToString || defaultItemToString,
       onHighlightedIndexChange,
       onInputValueChange,
       onIsOpenChange,
       onSelectedItemChange,
+      ...(itemToString && {itemToString}),
```
TheSisb, line 94: Can't we just pass it directly? `itemToString,`? Or does it not like the key to be there with a null value? If it doesn't like null values, maybe you can set these to default values when you destructure them like `initialIsOpen = false,` and then it should just work maybe

packages/paste-core/components/combobox/src/Combobox.tsx
```diff
@@ -77,13 +87,61 @@ const Combobox = React.forwardRef<HTMLInputElement, ComboboxProps>(
...
+        </ComboboxListboxOption>
+      );
+    };
+
+    const renderItems = (): React.ReactNode[] =>
+      items.map((item, index) => {
+        return renderItem(item, index);
+      });
+
+    const renderGroupedItems = (): React.ReactNode[] | null => {
```
TheSisb, line 121: Would be great to move this and the previous functions out of the component. It means these wont reinitialze every re-render, but also it will force us to pass arguments to these functions, making them pure and generally more testable.

packages/paste-core/components/combobox/src/Combobox.tsx
```diff
@@ -111,20 +169,7 @@ const Combobox = React.forwardRef<HTMLInputElement, ComboboxProps>(
...
-                  {...getItemProps({item, index})}
-                  highlighted={highlightedIndex === index}
-                  key={uid(item)}
-                >
-                  {optionTemplate ? optionTemplate(item) : item}
-                </ComboboxListboxOption>
-              );
-            })}
-        </ComboboxListbox>
+        <ComboboxListbox {...getMenuProps()}>{isOpen && renderListBox()}</ComboboxListbox>
```
TheSisb, line 172: Since the listbox doesn't need to remain in the DOM to animate in/out, I wonder if we could move the `isOpen` outside?

```
{isOpen ? <ComboboxListbox .... > : null}
```

packages/paste-core/components/combobox/src/Combobox.tsx
```diff
@@ -77,13 +87,61 @@ const Combobox = React.forwardRef<HTMLInputElement, ComboboxProps>(
...
       items,
-      itemToString: itemToString || defaultItemToString,
       onHighlightedIndexChange,
       onInputValueChange,
       onIsOpenChange,
       onSelectedItemChange,
+      ...(itemToString && {itemToString}),
+      ...(initialIsOpen && {initialIsOpen}),
+      ...(inputValue && {inputValue}),
+      ...(selectedItem && {selectedItem}),
```
TheSisb, line 97: What happens if selectedItem and initialSelectedItem are different on mount? 🤔

packages/paste-core/components/combobox/src/ComboboxListboxGroup.tsx
```diff
@@ -0,0 +1,38 @@
...
+import {Text} from '@twilio-paste/text';
+
+export interface ComboboxListboxGroupProps {
+  children: NonNullable<React.ReactNode>;
+  groupName: string;
+}
+
+const ComboboxListboxGroup: React.FC<ComboboxListboxGroupProps> = ({children, groupName}) => {
+  return (
+    <Box as="div" role="group" aria-label={groupName} paddingTop="space20">
```
TheSisb, line 13: Is `role="group"` working as intended here? Just double checking if it was just a `Menu` issue

packages/paste-core/components/combobox/src/ComboboxListboxGroup.tsx
```diff
@@ -0,0 +1,38 @@
...
+      <Box
+        as="div"
```
TheSisb, line 15: nit: can omit, this is default

packages/paste-core/components/combobox/src/ComboboxListboxOption.tsx
```diff
@@ -6,20 +6,35 @@ import {Text} from '@twilio-paste/text';
 export interface ComboboxListboxOptionProps {
   children: NonNullable<React.ReactNode>;
   highlighted?: boolean;
+  variant: 'default' | 'groupOption';
 }
 
+const getVariantStyles = (variant: ComboboxListboxOptionProps['variant']): {} => {
```
TheSisb, line 12: I'm not sure which I prefer, but in the Button to Box refactor I started just making a map instead of using switches:

```
const VariantStyles = {
  default: {
    paddingLeft: 'space70',
    paddingRight: 'space70',
  },
  groupOption: {
    paddingLeft: 'space50',
    paddingRight: 'space50',
  }
};

// usage
{...VariantStyles[variant]}
```

Pros: not a function, so no function call. No switch statement verboseness.
Cons: no `default:` case fallback, but with typescript it's unlikely to fall into the default case

Thoughts?

packages/paste-core/components/combobox/src/Combobox.tsx
SiTaggart: I _think_ it looks like we can remove them. I was just following the pattern, and I think we copy and pasted from Typescript

packages/paste-core/components/combobox/src/Combobox.tsx
SiTaggart: Yeah it doesn't like them to be null, it seems to still ignore the internal implementation that downshift does hence we had to originally supply a `defaultItemToString` so it wouldn't break on us. 

For things like `itemToString`, `inputValue` and `selectedItem` we don't really know what the internal version is 
doing or what it's dependent on. I'm not confident `defaultItemToString` was doing the right thing. If I left `initialSelectedItem`, `selectedItem` and `inputValue` all pass through, and only set `initialSelectedItem` as having a value, it doesn't update the inputValue so it looks broken. If these are only passed when present it works as intended. 

I feel like doing this more closely replicate the default behaviour of downshift

packages/paste-core/components/combobox/src/Combobox.tsx
SiTaggart: I guess the same question could be asked of Downshift directly, it's just a pass through replicating their API. 🤔

packages/paste-core/components/combobox/src/Combobox.tsx
SiTaggart: Apparently not, but I have a way to at least make it more explicit.

packages/paste-core/components/combobox/src/Combobox.tsx
SiTaggart: Sure

packages/paste-core/components/combobox/src/ComboboxListboxGroup.tsx
SiTaggart: Works a treat 👍

packages/paste-core/components/combobox/src/ComboboxListboxOption.tsx
SiTaggart: Pretty neat and tidy, I like it. Typescript is a bit weird about it, but it's not impossible to type the values so `box` doesn't complain.

packages/paste-core/components/combobox/src/Combobox.tsx
```diff
@@ -26,21 +34,129 @@ const StyledInputAsSelect = styled(InputElement)(
...
   optionTemplate?: (item: string | {}) => React.ReactNode;
+  selectedItem?: UseComboboxPrimitiveProps<Item>['selectedItem'];
+  inputValue?: UseComboboxPrimitiveProps<Item>['inputValue'];
+  groupItemsBy?: string;
+}
+
+interface RenderItemProps {
+  item: Item;
+  index: number;
+  getItemProps: any;
```
SiTaggart, line 60: This is something that we don't get the types for from Downshift and we don't even use it so i'm happy with an any here

PR #609, opened 2020-07-28
feat(icons): add UDL icons and upgrade svg-to-react

TheSisb:
- [x] Updates SVG to React to fix occasional failures on the convert command
- [x] Update all icons for UDL


packages/paste-icons/tools/templates/reactIconTemplate.js
```diff
@@ -20,7 +20,7 @@ const ${componentName}: React.FC<${componentName}Props> = ({as, display, size, i
   return (
     <IconWrapper as={as} display={display} size={size} iconColor={iconColor}>
       <UID>
-        {uid => (
+        {titleId => (
```
TheSisb, line 23: New SVGR forces this variable name, so I had to swap `uid` to `titleId` in the upgrade.

PR #616, opened 2020-07-29
feat(box): add typography style props and others

TheSisb:
- [x] Add typography style props to Box
- [x] Allow passing `none, initial, inherit, unset` to `border`, `background`, and `textShadow`
- [x] Enable margin/padding[X/Y] props
- [x] Export a BoxStyleProps interface we can use to style plain JS objects of style props, to avoid needing to type each key individually.
- [x] Export a BoxElementProps, which is useful for typing against native HTML elements
- [x] Typings fixes to Disclosure and Menu components
- [x] Added missing react-uid peerDep to Disclosure
- [x] Updated Text typings to match Box typings enhancement


packages/paste-core/primitives/box/src/index.tsx
```diff
@@ -133,33 +132,31 @@ const extraConfig = system({
...
   resize: true,
-  transition: true,
   listStyleType: true,
   listStylePosition: true,
   listStyleImage: true,
   objectFit: true,
   objectPosition: true,
-  backgroundAttachment: {
-    property: 'backgroundAttachment',
-  },
```
SiTaggart, line 156: We might need to check this is supported by Styled-System

packages/paste-core/primitives/box/src/index.tsx
```diff
@@ -39,92 +39,91 @@ import {
...
+  _groupHover?: BoxStyleProps;
+  _notFirst?: BoxStyleProps;
+  _notLast?: BoxStyleProps;
+  _placeholder?: BoxStyleProps;
+}
+
+// Omits potential clashes from our style props with HTMLAttributes (i.e.: color)
+export interface BoxElementProps extends Omit<React.HTMLAttributes<HTMLElement>, keyof BoxStyleProps> {
+  as?: keyof JSX.IntrinsicElements;
   type?: string;
```
SiTaggart, line 112: We need to support the following 
```
href?: string;
rel?: string;
target?: string;```

PR #619, opened 2020-07-29
throwaway PR for line height test

serifluous:
- Updated line-height-30 alias to 1.25 rem (20px)
- Updated Paragraph and List line heights to lineHeight40

### DIFF
--- a/packages/paste-core/components/list/src/index.tsx
+++ b/packages/paste-core/components/list/src/index.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
-import { Text, safelySpreadTextProps } from '@twilio-paste/text';
-import { Space, isMarginTokenProp } from '@twilio-paste/style-props';
+import {Text, safelySpreadTextProps} from '@twilio-paste/text';
+import {Space, isMarginTokenProp} from '@twilio-paste/style-props';
 
 interface BaseListProps extends React.OlHTMLAttributes<HTMLElement> {
   className?: never;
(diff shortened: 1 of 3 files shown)


SiTaggart:
Another place this is needed is with changing the button and input heights. With needing to change the line height to 20px to achieve the desired new overall height, a default Icon placed in that button will push the height back up. Default icon is 24px high to match the line height of the default text.

serifluous:
> Another place this is needed is with changing the button and input heights. With needing to change the line height to 20px to achieve the desired new overall height, a default Icon placed in that button will push the height back up. Default icon is 24px high to match the line height of the default text.

Default icon size should be changed to 20px to match the default text line height.

katieporter:
This is perhaps a case where we should have used Paragraph and didn't, but the line-height in this disclosure example looks pretty tight.

image

I think one thing we'll need to anticipate if our default text and our paragraph text have different line heights is that folks may accidentally use the default text when they should use Paragraph, and there will be a visual difference if they do so.

katieporter:
Menu items could use a little extra spacing now, but I think UDL might take care of that?

image

katieporter:
This is minor since these are only used on the docs site, but the line-height for the body of our callouts looks pretty tight, especially for wordier ones.

image

katieporter:
I already shared this in a Slack thread sometime last week, but I'll add it here too so that everything is in one place. The radio and checkbox alignment will need to be adjusted.

image

katieporter:
> This is minor since these are only used on the docs site, but the line-height for the body of our callouts looks pretty tight, especially for wordier ones.
> 
> image

Related to this one, maybe we can give all `p` elements a 23px line-height? So, when folks use Text as `p`, it looks like Paragraph.

katieporter:
Labels and help text could use a little more spacing now.

image

SiTaggart:
@katieporter covered most of what I spotted. the only other thing I could see is that the margin between checkbox/radio label and helptext might need tightening up a little now.

image

PR #620, opened 2020-07-30
feat(tooltips): update tooltips design for UDL

SiTaggart:
Update the design of tooltips for UDL. Includes introducing 2 new tokens.

## Tokens
- add color-background-body-inverse
- add shadow
- update elevation tokens for default theme

## Tooltip
- update background
- update border
- update spacing
- update color

### DIFF
--- a/packages/paste-design-tokens/tokens/aliases/box-shadow.yml
+++ b/packages/paste-design-tokens/tokens/aliases/box-shadow.yml
@@ -4,6 +4,6 @@ imports:
   - ./spacing.yml
 aliases:
   shadow-elevation-0: "none"
-  shadow-elevation-10: "{!offset-0} {!offset-10}px {!offset-40}px {!offset-0} {!palette-gray-100-transparent-10}"
-  shadow-elevation-20: "{!offset-0} {!offset-20}px {!offset-60}px {!offset-0} {!palette-gray-100-transparent-20}"
-  shadow-elevation-30: "{!offset-0} {!offset-60}px {!offset-80}px {!offset-0} {!palette-gray-100-transparent-20}"
+  shadow-elevation-10: "{!offset-0} {!offset-10} {!offset-40} {!offset-0} {!palette-gray-100-transparent-10}"
+  shadow-elevation-20: "{!offset-0} {!offset-20} {!offset-60} {!offset-0} {!palette-gray-100-transparent-20}"
+  shadow-elevation-30: "{!offset-0} {!offset-60} {!offset-80} {!offset-0} {!palette-gray-100-transparent-20}"
--- a/packages/paste-design-tokens/tokens/themes/console/global/box-shadow.yml
+++ b/packages/paste-design-tokens/tokens/themes/console/global/box-shadow.yml
@@ -18,10 +18,10 @@ props:
     value: "{!shadow-elevation-10}"
     comment: Shadow for cards.
   shadow-focus:
-    value: "{!offset-0} {!offset-0} {!offset-0} {!offset-20}px {!palette-blue-60-transparent-80}"
+    value: "{!offset-0} {!offset-0} {!offset-0} {!offset-20} {!palette-blue-60-transparent-80}"
     comment: Shadow for focus ring on interactive elements.
   shadow-focus-inverse:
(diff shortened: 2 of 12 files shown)


SiTaggart:
TODO:

- [ ] This looks jank. Make the arrow smaller 
image

- [ ] Update the docs page anatomy section

packages/paste-design-tokens/tokens/aliases/box-shadow.yml
```diff
@@ -4,6 +4,6 @@ imports:
   - ./spacing.yml
 aliases:
   shadow-elevation-0: "none"
-  shadow-elevation-10: "{!offset-0} {!offset-10}px {!offset-20}px {!offset-0} {!palette-gray-100-transparent-30}"
-  shadow-elevation-20: "{!offset-0} {!offset-40}px {!offset-60}px {!offset-0} {!palette-gray-100-transparent-20}"
-  shadow-elevation-30: "{!offset-0} {!offset-50}px {!offset-80}px {!offset-20} {!palette-gray-100-transparent-20}"
+  shadow-elevation-10: "{!offset-0} {!offset-10}px {!offset-40}px {!offset-0} {!palette-gray-100-transparent-10}"
+  shadow-elevation-20: "{!offset-0} {!offset-20}px {!offset-60}px {!offset-0} {!palette-gray-100-transparent-20}"
+  shadow-elevation-30: "{!offset-0} {!offset-60}px {!offset-80}px {!offset-20} {!palette-gray-100-transparent-20}"
```
richbachman, line 9:
```suggestion
  shadow-elevation-30: "{!offset-0} {!offset-60}px {!offset-80}px {!offset-20}px {!palette-gray-100-transparent-20}"
```

TheSisb: 🤔 Should we just make the offset aliases strings that include the `px`?
image
More consistent with the other tokens and removes the need to comment on the offset-0

packages/paste-design-tokens/tokens/themes/console/global/box-shadow.yml
```diff
@@ -5,6 +5,9 @@ imports:
   - ../../../global/box-shadow.yml
   - ../aliases/box-shadow.yml
 props:
+  shadow:
```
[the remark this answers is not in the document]
SiTaggart: You would think, but unfortunately it's a lot more procedural in the way it computes the value of a token, especially if you're changing the alias value.

If you're are importing a token from another theme, the token is already created pointing to the value of the alias from that theme. If you simply import the new aliases, the token isn't regenerated or computed, so the value is the previous alias value still.

Importing is only effective if you are literally inheriting the token as is to the new theme, if you want to change the value of it, you need to create it on top of the imported one with the new alias value.

katieporter:
Looks like the arrow in the default theme has the wrong background color. Could we also make the arrow a tiny bit smaller?

image

SiTaggart:
> Looks like the arrow in the default theme has the wrong background and border color.
Looks like it's a switching problem, not necessarily a bug in the component. The theme switching isn't causing the inline component color to rerender. I don't foresee it being a problem. It's correct on initial theme render and dynamically changing the theme is just a demo edge case right now.

> Could we also make the arrow a tiny bit smaller?
I've tried to make it a little smaller but it starts to break when I go smaller and that's a bigger scope of work than this ticket I'm afriad.

katieporter:
> > Looks like the arrow in the default theme has the wrong background and border color.
> 
> Looks like it's a switching problem, not necessarily a bug in the component. The theme switching isn't causing the inline component color to rerender. I don't foresee it being a problem. It's correct on initial theme render and dynamically changing the theme is just a demo edge case right now.

Gotcha, that should be fine then. Do we want to make a bug ticket for it, or just leave it be?

> 
> > Could we also make the arrow a tiny bit smaller?
> 
> I've tried to make it a little smaller but it starts to break when I go smaller and that's a bigger scope of work than this ticket I'm afriad.

Cool, I can throw a ticket in the backlog to adjust that if that sounds good. Definitely not urgent.

SiTaggart:
> Do we want to make a bug ticket for it, or just leave it be?

Lets leave it for now, we need to replace it the svg for Box in the future, so this will go away when that happens

> Cool, I can throw a ticket in the backlog to adjust that if that sounds good. 

Sure go for it.

PR #625, opened 2020-07-31
chore(button): convert to use Box

richbachman:
Button now uses Box under the hood

### DIFF
--- a/packages/paste-core/components/button/src/DestructiveButton.tsx
+++ b/packages/paste-core/components/button/src/DestructiveButton.tsx
@@ -1,0 +1,74 @@
+import * as React from 'react';
+import {Box, BoxStyleProps, safelySpreadBoxProps} from '@twilio-paste/box';
+import {SizeStyles, BaseStyles} from './styles';
+import {DirectButtonProps, DirectButtonPropTypes} from './types';
+
+// This module can only be referenced with ECMAScript imports/exports by turning on the 'esModuleInterop' flag and referencing its default export
+const merge = require('lodash.merge');
+
+/*
+ * defensively resetting 'color' on pseudostyles from over zealous
+ * legacy global styles "a {...}" when button is set as an anchor
+ */
+const defaultStyles: BoxStyleProps = merge({}, BaseStyles.default, {
+  color: 'colorTextInverse',
+  backgroundColor: 'colorBackgroundDestructive',
+  borderColor: 'colorBorderDestructive',
+  _hover: {
+    color: 'colorTextInverse',
+    backgroundColor: 'colorBackgroundDestructiveDarker',
+    borderColor: 'colorBorderDestructiveDarker',
+  },
+  _focus: {
+    color: 'colorTextInverse',
+    borderColor: 'colorBorderDestructiveDarker',
+    boxShadow: 'shadowFocus',
+  },
+  _active: {
+    color: 'colorTextInverse',
+    backgroundColor: 'colorBackgroundDestructiveDarker',
+    borderColor: 'colorBorderDestructiveDarker',
+  },
+});
+
+const loadingStyles: BoxStyleProps = merge({}, BaseStyles.loading, {
+  color: 'colorTextInverse',
+  backgroundColor: 'colorBackgroundDestructiveDarker',
+  borderColor: 'colorBorderDestructiveDarker',
+});
+
+const disabledStyles: BoxStyleProps = merge({}, BaseStyles.disabled, {
+  color: 'colorTextInverse',
+  backgroundColor: 'colorBackgroundDestructiveLight',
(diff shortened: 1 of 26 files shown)


packages/paste-core/primitives/box/src/index.tsx
```diff
@@ -77,37 +78,41 @@ export interface BoxStyleProps
...
+  _even?: BoxBaseStyleProps;
+  _visited?: BoxBaseStyleProps;
+  _readOnly?: BoxBaseStyleProps;
+  _first?: BoxBaseStyleProps;
+  _last?: BoxBaseStyleProps;
+  _groupHover?: BoxBaseStyleProps;
+  _notFirst?: BoxBaseStyleProps;
+  _notLast?: BoxBaseStyleProps;
+  _placeholder?: BoxBaseStyleProps;
+  __moz_focus_inner?: BoxBaseStyleProps;
```
TheSisb, line 109: Added this as it was being used in CSS directly before.

packages/paste-core/components/button/src/DestructiveButton.tsx
```diff
@@ -0,0 +1,68 @@
+import * as React from 'react';
+// import * as PropTypes from 'prop-types';
```
richbachman, line 2: Do we need PropTypes on these individual components? If not, we can remove this comment.

packages/paste-core/components/button/src/LinkButton.tsx
```diff
@@ -0,0 +1,46 @@
+import * as React from 'react';
+// import * as PropTypes from 'prop-types';
```
richbachman, line 2: Same. Either remove or add PropTypes if needed.

packages/paste-core/components/button/src/PrimaryButton.tsx
```diff
@@ -0,0 +1,68 @@
+import * as React from 'react';
+// import * as PropTypes from 'prop-types';
```
richbachman, line 2: Same. Either remove or add PropTypes if needed.

packages/paste-core/components/button/src/ResetButton.tsx
```diff
@@ -0,0 +1,37 @@
+import * as React from 'react';
+// import * as PropTypes from 'prop-types';
```
richbachman, line 2: Same. Either remove or add PropTypes if needed.

packages/paste-core/components/button/src/SecondaryButton.tsx
```diff
@@ -0,0 +1,83 @@
+import * as React from 'react';
+// import * as PropTypes from 'prop-types';
```
richbachman, line 2: Same. Either remove or add PropTypes if needed.

packages/paste-core/layout/stack/src/index.tsx
```diff
@@ -19,7 +19,7 @@ type StackOrientationOptions = 'horizontal' | 'vertical';
 export type StackOrientation = ResponsiveValue<StackOrientationOptions>;
 
 interface StackStyleProps extends Pick<LayoutProps, 'display'>, Pick<FlexboxProps, 'alignItems' | 'flexWrap'> {}
-export interface StackProps extends Pick<BoxProps, 'as'>, React.HTMLAttributes<any> {
+export interface StackProps extends BoxProps {
```
SiTaggart, line 22: Extending `BoxProps` here isn't necessarily true as we `safelySpreadBoxProps`, removing style props passed down. You probably want `BoxElementProps` instead.

packages/paste-core/primitives/box/src/index.tsx
SiTaggart: You'll need to add this to here https://github.com/twilio-labs/paste/blob/main/packages/paste-core/primitives/box/src/PseudoPropStyles.ts

SiTaggart: In fact I think if you remove `: {[key: string]: string}` from `PseudoPropStyles` in https://github.com/twilio-labs/paste/blob/main/packages/paste-core/primitives/box/src/PseudoPropStyles.ts You might be able to dynamically create this interface like 

```
type PseudoSelectors = keyof typeof PseudoPropStyles;
type PseudoStylesProps = {
  [key in PseudoSelectors]: BoxStyleProps;
}```

so we don't have to keep these in sync

packages/paste-core/layout/stack/src/index.tsx
TheSisb: 🧠

packages/paste-core/primitives/box/src/index.tsx
TheSisb: Good idea, fixed!

PR #637, opened 2020-08-05
fix(icons): use display=block by default and rename 'iconColor' prop to 'color'

TheSisb:
Prior to this PR, icons were defaulting to `display="inline-flex"`.  This solved some issues, but created others. Through experience, we've come to realize that making icons display block by default would be most convenient.
- If you need to inline an icon, you can either wrap it in one of our Layout components (Stack, Flex, etc) and it will just work.
- If you need to inline it next to text, you can still pass a `display` prop to override the default block value (i.e.: `display="inline"`)

It's easier to modify the display prop when the icon is inline than when it is a nested child of a Layout component, which is why we're changing the default value (Developer Experience++).
Furthermore, icons are still `span` elements with the `display="block"` styled applied, for clearer semantics.

Also, since Box now accepts typography props and, namely `color`, I've gone ahead and made icons accept `color` for consistency too (previously `iconColor`).


packages/paste-icons/README.md
```diff
@@ -19,13 +19,13 @@ import AssetsIcon from '@twilio-paste/icons/react/AssetsIcon';
...
+| Prop       | Type                                   | Description                                                                                                                                           | Default                                                                  |
+| ---------- | -------------------------------------- | ----------------------------------------------------------------------------------------------------------------------------------------------------- | ------------------------------------------------------------------------ |
+| decorative | boolean                                | Whether or not the SVG is just visual flair or adds meaning to the page. Specifically for screenreaders to know whether to read out the title or not. | true                                                                     |
+| display?   | `ResponsiveValue<CSS.DisplayProperty>` |                                                                                                                                                       | inline-flex                                                              |
```
richbachman, line 25:
```suggestion
| display?   | `ResponsiveValue<CSS.DisplayProperty>` |                                                                                                                                                       | block                                                              |
```

packages/paste-website/src/pages/icon-system/index.mdx
```diff
@@ -199,7 +199,7 @@ import {PlusIcon} from '@twilio-paste/icons/esm/PlusIcon';
 | display?   | `ResponsiveValue<CSS.DisplayProperty>` |                                                                                                                                                       | `inline-flex`  |
```
richbachman, line 199:
```suggestion
| display?   | `ResponsiveValue<CSS.DisplayProperty>` |                                                                                                                                                       | `block`  |
```

PR #640, opened 2020-08-06
chore(button): update to UDL styles

TheSisb:
Updating some Button styles for UDL


serifluous:
Can you add the shadow border tokens to the console theme, too? The UDL colors are being passed to the console components.

serifluous:
Can you update these tokens in the default theme?:

color-background-primary-light: palette-blue-30
color-background-destructive-light: palette-red-30

SiTaggart:
I commented on one of the Chromatic stories where I saw a component variant _gain_ a border, but potentially that actually might be fixing it. It looks like it's a secondary button, `size="reset"` and that kind of makes sense that it should have a stroke, like the other sizes. 

Otherwise the code looks good

TheSisb:
Yeah I think that's a "fix" :). Moving forward the `reset` size option will be pulled out into a primitive, because I don't really see the value proposition of variant != reset + size == reset and we're missing a Button primitive generally.

PR #642, opened 2020-08-06
feat(form): add inverse variant

richbachman:
- [x] Inverse variant for input, textarea, select, combobox,
- [x] Inverse and Error inverse variant for help text
- [x] Inverse variant for label
- [x] Stories for vrt
- [x] Variant props on doc pages prop tables


serifluous:
Should've caught this in the other PR, but in the non-inverse state, can you change placeholder to colorTextWeaker when form fields are disabled?

packages/paste-core/components/form/src/shared/FieldWrapper.tsx
```diff
@@ -26,24 +35,90 @@ const FieldWrapper: React.FC<FieldWrapperProps> = ({disabled, hasError, readOnly
     boxShadow = null;
   }
 
+  // inverse boxShadow styles
+  let InverseBoxShadow = 'shadowBorderInverse' as BoxShadow;
```
SiTaggart, line 39: This is starting to get a bit gnarly, the `FieldWrapper` might benefit from a button like treatment where the variants are different components we conditionally render.

packages/paste-core/components/form/src/FormHelpText.tsx
```diff
@@ -3,38 +3,61 @@ import * as PropTypes from 'prop-types';
...
+    case FormHelpTextVariants.ERROR_INVERSE:
+      icon = <ErrorIcon color="colorTextErrorLight" decorative size="sizeIcon20" />;
+      break;
     default:
       break;
   }
 
+  let textColor = 'colorTextWeak' as TextColor;
+  if (disabled) {
+    textColor = 'colorTextWeaker';
```
SiTaggart, line 41: We can't disable help text I'm afraid, it always has to pass 4.5:1. It can be `weak`, but nothing lighter

katieporter: @SiTaggart does the label on a disabled field need to pass 4.5:1, too? Or just the help text?

SiTaggart: Both please

richbachman: @SiTaggart So we need drop the style of the disabled labels. Looks like our default labels are also not accessible when disabled then. I can go ahead and make that change in this PR if necessary.

richbachman: Going with `colorTextWeak` and `colorTextInverseWeak`

PR #648, opened 2020-08-11
feat(button): add inverse variant

TheSisb:
Adds a new variant called `inverse` for positioning buttons over dark backgrounds.


packages/paste-website/src/pages/components/button/index.mdx
```diff
@@ -193,6 +193,17 @@ possible, reversible action without much cost.
 </Button>`}
 </LivePreview>
 
+#### Inverse button
+
+The inverse button is best used over very strong colored backgrounds. In a lighter theme, this would mean over a dark background.
+Do reach for the other variants first and only use the inverse as a backup option.
+
+<LivePreview scope={{Button}} language="jsx">
+  {`<Button variant="inverse">
```
SiTaggart, line 202: Just need to wrap this is a Box with an inverse background so we can see the exaple.

PR #653, opened 2020-08-12
fix(button): reset variant should inherit type styles

SiTaggart:
With an update to the reset variant of button we accidentally regressed our roadmap page.

## Before button update: 

image

## After button update:

image

This PR applies a small fix to the reset variant that puts it back to original behavior, but also inherits weight and color. This places the reset variant back to its original intended purpose; use in compositions.

I also noticed we don't actually regression test a pretty common use case for buttons, icon next to text. Adding for regression purposes.

**Code includes:**

- inherit font size, weight and color on reset variants of button
- add story examples to cover those use cases so we catch future regressions
- add missing button examples text with icon for regression purposes

**N.B.** I get it that ideally we'd just replace the roadmap custom disclosures with the _real_, since released Disclosure component but I wanted to get us back to a before state before doing that.

### DIFF
--- a/packages/paste-core/components/button/src/ResetButton.tsx
+++ b/packages/paste-core/components/button/src/ResetButton.tsx
@@ -6,11 +6,15 @@ import {DirectButtonProps, DirectButtonPropTypes} from './types';
 // This module can only be referenced with ECMAScript imports/exports by turning on the 'esModuleInterop' flag and referencing its default export
 const merge = require('lodash.merge');
 
-const defaultStyles: BoxStyleProps = merge({}, BaseStyles.default);
+const defaultStyles: BoxStyleProps = merge({}, BaseStyles.default, {
+  fontSize: 'inherit',
+  fontWeight: 'inherit',
+  color: 'inherit',
+});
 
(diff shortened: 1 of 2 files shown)


packages/paste-core/components/button/src/ResetButton.tsx
```diff
@@ -6,11 +6,15 @@ import {DirectButtonProps, DirectButtonPropTypes} from './types';
 // This module can only be referenced with ECMAScript imports/exports by turning on the 'esModuleInterop' flag and referencing its default export
 const merge = require('lodash.merge');
 
-const defaultStyles: BoxStyleProps = merge({}, BaseStyles.default);
+const defaultStyles: BoxStyleProps = merge(BaseStyles.default, {
```
TheSisb, line 9: You shouldn't remove the first `{},` because `lodash.merge` mutates the first object passed. BaseStyles.default will be affected for the other variants.  If this currently works it might just be a fluke that reset is parsed last.

SiTaggart: Pushed

PR #656, opened 2020-08-13
feat: line height 20px

SiTaggart:
PR includes

- Adjust line-height scale slightly to gain line-height 20px and 24px for 30 and 40 respectively. Size * 1.4 rounded up to the nearest 4px. This allows a default 20px line-height and allows 24px for paragraphs and lists with a bump to line height 40
- scaled icons to base line-height of 20px
- adjusted bottom padding on alerts to accommodate line height
- adjusted spacing on radio and checkboxes to accommodate new line-height
- fixed indeterminate checkbox line color when disabled to match checked disabled


katieporter:
@SiTaggart does this misalignment have anything to do with the icon updates? 
image

katieporter:
Is it possible to make `Text as=p` the same line height as `Paragraph`?

SiTaggart:
> Is it possible to make Text as=p the same line height as Paragraph?
Not automatically, no. You have to set the line-height manually.

SiTaggart:
> does this misalignment have anything to do with the icon updates?

@katieporter no it's just broken cos we changed the default display of icons to be block and not inline-block. https://paste.twilio.design/components/card#card-with-centered-content.

I'll fix that up separately

katieporter:
> > Is it possible to make Text as=p the same line height as Paragraph?
> 
> Not automatically, no. You have to set the line-height manually.

Hrmmm okay...and that's because the new default line-height is 20px, yeah? That means Text inherits that line-height unless otherwise defined?

SiTaggart:
> > > Is it possible to make Text as=p the same line height as Paragraph?
> > 
> > 
> > Not automatically, no. You have to set the line-height manually.
> 
> Hrmmm okay...and that's because the new default line-height is 20px, yeah? That means Text inherits that line-height unless otherwise defined?

Correct. The default line height for Text is 20px, it doesn't care what HTML element it is at that point. Only when you use it in a composition are you then able to make it apply styles with your intended design in mind. I think it's fine honestly, if you're after a paragraph you should be using the paragraph component. if you're after incidental UI text, it should be the default line height.

PR #662, opened 2020-08-18
chore(website): set up env based search indexing

richbachman:
- [x] Change Algolia indexing so it only happens in the production deployment environment

### DIFF
--- a/packages/paste-website/gatsby-config.js
+++ b/packages/paste-website/gatsby-config.js
@@ -4,11 +4,11 @@
  * See: https://www.gatsbyjs.org/docs/gatsby-config/
  */
 
-// const queries = require('./src/utils/algolia');
+const queries = require('./src/utils/algolia');
 
 require('dotenv').config();
 
-module.exports = {
+const gatsbyConfig = {
   siteMetadata: {
     title: 'Paste',
     description: 'Paste: The Design System for building Twilio customer experiences.',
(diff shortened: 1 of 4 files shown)


packages/paste-website/gatsby-config.js
```diff
@@ -187,15 +189,24 @@ module.exports = {
...
+  options: {
+    appId: process.env.GATSBY_ALGOLIA_APP_ID,
+    apiKey: process.env.GATSBY_ALGOLIA_ADMIN_KEY,
+    indexName: process.env.GATSBY_ALGOLIA_INDEX_NAME,
+    queries,
+    chunkSize: 10000,
+  },
+};
+
+const activeEnv = process.env.DEPLOYMENT_ENV;
```
TheSisb, line 206: Where is DEPLOYMENT_ENV being set?

richbachman: In our Vercel environment variables

packages/paste-website/gatsby-config.js
```diff
@@ -4,11 +4,13 @@
...
 
-// const queries = require('./src/utils/algolia');
+const queries = require('./src/utils/algolia');
 
 require('dotenv').config();
 
-module.exports = {
+const gatsbyConfig = {};
+
+gatsbyConfig.data = {
```
TheSisb, line 13: Not sure why we need this `data` key. Can we do
```
const gatsbyConfig = {
  siteMetadata: ...
}

if (proces.env.DEPLOYMENT_ENV === 'production') {
  gatsbyConfig.plugins.push({
    resolve: `gatsby-plugin-algolia`,
    options: {
      appId: process.env.GATSBY_ALGOLIA_APP_ID,
      apiKey: process.env.GATSBY_ALGOLIA_ADMIN_KEY,
      indexName: process.env.GATSBY_ALGOLIA_INDEX_NAME,
      queries,
      chunkSize: 10000,
    },
  });
}

module.exports = gatsbyConfig;
```

richbachman: Looks like I was just over-complicating things. Your solution is simpler and works.

PR #666, opened 2020-08-19
fix(combobox): provide a means to use action methods from state hook

SiTaggart:
When using the Combobox there are sometimes needs where you need that little bit of extra control, for example clearing a selection based on some application logic. 

Currently we don't provide a good enough means to clear the selection and input value together as they are not explicitly tied to each other, and a combobox can get into a strange state once a selection has been made.

By exposing the internal state hook provided by Downshift, and allowing a consumer to pass the returned state props to the component as `state` we allow the consumer to use some of the action methods provided by downshift, such as `reset`.

This PR introduces the ability to provide Combobox a state override by using the state hook.

### DIFF
--- a/packages/paste-core/components/combobox/__tests__/index.spec.tsx
+++ b/packages/paste-core/components/combobox/__tests__/index.spec.tsx
@@ -1,12 +1,31 @@
 import * as React from 'react';
+import _ from 'lodash';
 import {axe} from 'jest-axe';
 import {uid} from 'react-uid';
-import {render, screen} from '@testing-library/react';
+import {render, screen, fireEvent} from '@testing-library/react';
 import {FormLabel, FormHelpText, InputElement} from '@twilio-paste/form';
+import {Button} from '@twilio-paste/button';
+import {CloseIcon} from '@twilio-paste/icons/esm/CloseIcon';
+import {Box} from '@twilio-paste/box';
 import {useCombobox, Combobox, ComboboxInputWrapper, ComboboxListbox, ComboboxListboxOption} from '../src';
 import {ComboboxProps} from '../src/types';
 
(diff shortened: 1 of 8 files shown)


packages/paste-core/components/combobox/__tests__/index.spec.tsx
```diff
@@ -107,6 +126,60 @@ const GroupedMockCombobox: React.FC<{groupLabelTemplate?: ComboboxProps['groupLa
   );
 };
 
+export const ControlledCombobox: React.FC = () => {
```
TheSisb, line 129: I noticed this is being exported; did you mean to re-use this with the story?

packages/paste-core/components/combobox/stories/index.stories.tsx
```diff
@@ -375,7 +377,7 @@ storiesOf('Components|Combobox', module)
               setValue(inputValue);
             }
           }}
-          itemToString={item => (item ? item.label : null)}
+          itemToString={(item: any) => (item ? item.label : null)}
```
TheSisb, line 380: I don't think these should be `any`. Maybe `{label: string}` instead? Like if it is a number this would break.

packages/paste-core/components/combobox/__tests__/index.spec.tsx
SiTaggart: nah, just part of a snippet I use to scaffold a component

packages/paste-core/components/combobox/stories/index.stories.tsx
SiTaggart: Yeah, because they're just stories I wasn't being too fussy with typing the returned data we feed the example. I was just trying to get rid of the red underline. But I can change them to strings

PR #676, opened 2020-08-21
chore: update applitools eyes

TheSisb:
wip


SiTaggart:
Let's be a little cautious here. We currently do 60k to 70k snapshots a month and that comes from the Twilio central funded budget.

Our limit with Applitools right now is 150k a month. If we add just one browser we'll be consuming 80-90% of the entire companies budget of snapshots on a monthly basis.

Based on the need, this is super easy to add, and it's great too see that most components render really well in IE11, but we might need a budget conversation to happen first if we want to proceed.

TheSisb:
Just playing around. Nothing to see here haha

PR #685, opened 2020-08-25
fix(button): correct type clashes 

SiTaggart:
- Integration Codesandbox CI to test packages before merge
- Fix typescript clashes in build external packages on the button


SiTaggart:
So apparently this isn't a thing. In the preview sandbox no ts errors. IN a local create react app no ts errors. only in the Sandboxes from the template. Ignoring for now...

PR #686, opened 2020-08-25
fix(website): fix topbar zIndex

richbachman:
Prevents Combobox, Menu, and Popover from overlapping the top bar area of the docs site.

**QUESTION:**
Does not fix the Tooltip overlap. We intentionally set tooltip to have our highest zIndex token. So tooltips will always overlap. Should we change the Tooltip zIndex?

**ANSWER:**
No, this feels very intentional and expected behaviour.


SiTaggart:
> Does not fix the Tooltip overlap. We intentionally set tooltip to have our highest zIndex token. So tooltips will always overlap. Should we change the Tooltip zIndex?

No, this feels very intentional and expected behaviour for me

image

PR #692, opened 2020-08-26
chore: add codesandbox ci integration

SiTaggart:
Allow testing the packages in codesandbox in PRs using a create react app example as the bootstrap for the sandbox.

### DIFF
--- a/package.json
+++ b/package.json
@@ -177,8 +177,5 @@
       "commit-msg": "commitlint -E HUSKY_GIT_PARAMS",
       "pre-push": "if git-branch-is main; then yarn pre-push; fi"
     }
-  },
-  "engines": {
-    "node": "12.x"
   }
 }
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -177,8 +177,5 @@
       "commit-msg": "commitlint -E HUSKY_GIT_PARAMS",
       "pre-push": "if git-branch-is main; then yarn pre-push; fi"
     }
-  },
-  "engines": {
-    "node": "12.x"
```
SiTaggart, line 182: Because Codesandbox uses node v10

SiTaggart: I could give it a range I suppose. Either `"node" : ">=10.x <=12.x"` or `"node" : ">=10.x"`

richbachman: The range is probably the safest path, right?

PR #696, opened 2020-08-26
chore(anchor): refactor to use Text

richbachman:
- [x] Refactored Anchor to use Text
- [x] Refactored Anchor test to use React Testing Library
- [x] Added boxShadow, outline, and ref to Text


packages/paste-core/components/anchor/__tests__/anchor.test.tsx
```diff
@@ -1,59 +1,48 @@
 import * as React from 'react';
-import renderer from 'react-test-renderer';
-import {render} from 'react-dom';
+import {render, screen} from '@testing-library/react';
```
SiTaggart, line 2: 👏

PR #703, opened 2020-08-31
feat(website): setup Algolia DocSearch

richbachman:
- [x] Setup Algolia DocSearch


packages/paste-website/src/components/site-wrapper/SiteHeader.tsx
```diff
@@ -13,6 +13,19 @@ import {SiteBodyContext} from './SiteBodyContext';
 export const SiteHeader: React.FC = () => {
   const {isPasteTheme} = React.useContext(SiteBodyContext);
   const theme = useTheme();
+  const [value, setValue] = React.useState('');
+
+  React.useEffect(() => {
+    if (typeof window !== `undefined` && window.location != null) {
+      // @ts-ignore
```
richbachman, line 20: Ignoring because `docsearch` comes from the external js. Error is:
`Property 'docsearch' does not exist on type 'Window & typeof globalThis`

serifluous:
Is there any way to remove the left column? It feels pretty redundant with the page heading and right column.

Is there a way to make the heading for the page results smaller, too?

packages/paste-website/package.json
```diff
@@ -65,10 +65,12 @@
     "gatsby": "^2.19.19",
     "gatsby-image": "^2.2.41",
     "gatsby-plugin-algolia": "^0.11.0",
```
SiTaggart, line 67: Should we remove this now?

packages/paste-website/src/components/site-wrapper/SiteHeader.tsx
SiTaggart: You can add the check for `window.docsearch` to the condition above, should remove the ts error

packages/paste-website/package.json
richbachman: Yup. Removed all the old search deps and files.

richbachman:
@serifluous 

> Is there any way to remove the left column? It feels pretty redundant with the heading for page results and the right column.

Yes. I've hidden those.

> Is there a way to make the heading for the page results smaller, too?

And made the headings smaller.

packages/paste-website/src/components/site-wrapper/SiteHeader.tsx
```diff
@@ -13,6 +13,18 @@ import {SiteBodyContext} from './SiteBodyContext';
 export const SiteHeader: React.FC = () => {
   const {isPasteTheme} = React.useContext(SiteBodyContext);
   const theme = useTheme();
+  const [value, setValue] = React.useState('');
+
+  React.useEffect(() => {
+    if (typeof window !== `undefined` && window.docsearch != null) {
+      window.docsearch({
```
SiTaggart, line 20: I'd be pretty tempted to put this in the `html.tsx` file as well, to be honest. Just as a `<script>` tag after `postBodyComponents`

SiTaggart (approved):

PR #708, opened 2020-09-01
feat(anchor): add inverse variant

richbachman:
- [x] Added inverse Anchor variant
- [x] Added variant to Anchor doc props table


packages/paste-core/components/anchor/src/index.tsx
```diff
@@ -1,17 +1,17 @@
 import * as React from 'react';
-import * as PropTypes from 'prop-types';
-import {Text, safelySpreadTextProps} from '@twilio-paste/text';
+import {safelySpreadTextProps} from '@twilio-paste/text';
+import {DefaultAnchor} from './DefaultAnchor';
+import {InverseAnchor} from './InverseAnchor';
+import {AnchorProps, AnchorPropTypes, AnchorVariants} from './types';
```
SiTaggart, line 5: I think because you've moved them, if anyone is doing

```
import {AnchorProps} from '@twilio-paste/anchor'
```

You may break them. Might need to export the types from here

`export * from './types'` at the bottom probably

packages/paste-core/components/anchor/src/index.tsx
```diff
@@ -26,43 +26,26 @@ export const secureExternalLink = (href: string): {} | undefined => {
...
-      textDecoration: 'none',
-    }}
-    {...secureExternalLink(props.href)}
-    {...safelySpreadTextProps(props)}
-  >
-    {props.children}
-  </Text>
-));
+const Anchor = React.forwardRef<HTMLAnchorElement, AnchorProps>((props, ref) => {
+  const AnchorComponent = getAnchorComponent(props.variant || 'default');
```
SiTaggart, line 30: I know we started doing variant styles as maps, but should we do the same for variants themselves too?

```jsx
const AnchorVariants = {
  'inverse': InverseAnchor,
  'default': DefaultAnchor,
}

const Anchor = React.forwardRef<HTMLAnchorElement, AnchorProps>((({variant='default', ...props}), ref) => {
  const AnchorComponent = AnchorVariants[variant];
...
```

packages/paste-website/src/components/paste-mdx-provider/index.tsx
```diff
@@ -1,7 +1,8 @@
 import * as React from 'react';
 import {MDXProvider} from '@mdx-js/react';
 import {styled} from '@twilio-paste/styling-library';
-import {Anchor, AnchorProps} from '@twilio-paste/anchor';
+import {Anchor} from '@twilio-paste/anchor';
+import {AnchorProps} from '@twilio-paste/anchor/dist/types';
```
SiTaggart, line 5: Yeah, so this. In relation to my above point

packages/paste-core/components/anchor/src/index.tsx
richbachman: @SiTaggart Does that help with performance, by getting rid fo the `getAnchorComponent` function?

SiTaggart: Same theory I guess, one less function and switch statement to run

richbachman: Looks like I may get into a dependency cycle issue because AnchorVariants relies on InverseAnchor and DefaultAnchor, which rely on AnchorProps (which AnchorVariants is a part of).

packages/paste-core/components/anchor/__tests__/anchor.test.tsx
```diff
@@ -9,6 +9,15 @@ describe('Anchor', () => {
     expect(asFragment()).toMatchSnapshot();
   });
 
+  it('it should render an inverse anchor', () => {
```
TheSisb, line 12: Let's omit snapshotting component renders from now on as a rule. Since we have VRT (applitools and chromatic) this now becomes noise. As long as we add a story, that now becomes the test.

packages/paste-core/components/anchor/src/index.tsx
richbachman: Just pushed changes. Maybe what I have now works.

SiTaggart (approved):

PR #710, opened 2020-09-01
chore(website): add ids for search indexing

richbachman:
Added a couple of IDs to content we need to remove from the DocSearch index.

### DIFF
--- a/packages/paste-website/src/components/shortcodes/ChangelogRevealer.tsx
+++ b/packages/paste-website/src/components/shortcodes/ChangelogRevealer.tsx
@@ -1,10 +1,11 @@
 import * as React from 'react';
+import {Box} from '@twilio-paste/box';
 import {Separator} from '@twilio-paste/separator';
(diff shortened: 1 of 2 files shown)


packages/paste-website/src/components/shortcodes/ChangelogRevealer.tsx
```diff
@@ -1,17 +1,18 @@
 import * as React from 'react';
+import {Box} from '@twilio-paste/box';
 import {Separator} from '@twilio-paste/separator';
 import {Disclosure, DisclosureContent, DisclosureHeading} from '@twilio-paste/disclosure';
 
 export const ChangelogRevealer: React.FC<{children: NonNullable<React.ReactNode>}> = ({children}) => {
   return (
-    <>
+    <Box id="changelog">
```
SiTaggart, line 8: I might be tempted to make this a little more specific, but otherwise, fine.

richbachman: Thats fair. I'll change to `component-changelog`.

PR #715, opened 2020-09-02
fix(website): redirect trailing slash to no slash

TheSisb:
per https://github.com/algolia/docsearch-configs/pull/2349#issuecomment-685426430

> When trailingSlash: undefined, visiting a path with or without a trailing slash will not redirect.
> For example, both /about and /about/ will serve the same content without redirecting.
> This is not recommended because it could lead to search engines indexing two different pages with duplicate content.

https://vercel.com/docs/configuration#project/trailingslash

### DIFF
--- a/now.json
+++ b/now.json
@@ -2,34 +2,14 @@
   "name": "paste",
   "scope": "twilio-dsys",
   "version": 2,
-  "builds": [{
-    "src": "now-build.sh",
-    "use": "@now/static-build",
-    "config": {
-      "distDir": "packages/paste-website/public"
-    }
-  }],
-  "routes": [{
-      "src": "/form-elements",
-      "status": 301,
-      "headers": {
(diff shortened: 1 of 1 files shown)


now.json
```diff
@@ -2,14 +2,18 @@
   "name": "paste",
   "scope": "twilio-dsys",
   "version": 2,
-  "builds": [{
-    "src": "now-build.sh",
-    "use": "@now/static-build",
-    "config": {
-      "distDir": "packages/paste-website/public"
+  "trailingSlash": false,
```
richbachman, line 5:
```suggestion
  "trailingSlash": true,
```

I think we want the trailing slash. It looks like setting it `true` will forward any non-slash urls to slash urls.

TheSisb: 👍

TheSisb:
Closing for now

PR #721, opened 2020-09-03
chore: make codesandbox actually build and use core on PRs

SiTaggart:
Try and test codesnadbox


richbachman (approved):
Looks like that covers everything.

PR #730, opened 2020-09-04
chore: updates to combobox and primitive docs

SiTaggart:
- fix(combobox): handling refs in downshift v6 is different
- docs(combobox-primitive): update to reflect v6 of downshift

### DIFF
--- a/packages/paste-core/components/combobox/src/Combobox.tsx
+++ b/packages/paste-core/components/combobox/src/Combobox.tsx
@@ -235,12 +235,11 @@ const Combobox = React.forwardRef<HTMLInputElement, ComboboxProps>(
           <ComboboxInputWrapper {...getComboboxProps({role: 'combobox'})}>
             <StyledInputAsSelect
               {...getToggleButtonProps({tabIndex: 0})}
-              {...getInputProps({disabled})}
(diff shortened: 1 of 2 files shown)


packages/paste-core/components/combobox/src/Combobox.tsx
```diff
@@ -235,33 +235,30 @@ const Combobox = React.forwardRef<HTMLInputElement, ComboboxProps>(
...
-              highlightedIndex,
-              optionTemplate,
-              groupItemsBy,
-              groupLabelTemplate,
-              optionUID,
-              groupUID,
-            })}
-          </ComboboxListbox>
-        )}
+        <ComboboxListbox hidden={!isOpen} {...getMenuProps()}>
```
richbachman, line 250: Is the hidden prop the same as the hidden html attribute here?

richbachman (approved):

packages/paste-core/components/combobox/src/Combobox.tsx
SiTaggart: Yes, exactly the same. V6 expects the listbox to be rendered and warns if it's not, which to be fair is correct. This is just a nicer way to do it and hide the shadow we added to the listbox.

PR #731, opened 2020-09-04
fix(icon): reset line-height on icon wrapper

SiTaggart:
When aligning icons next to text, the line height of the text is inherited by the icon wrapper.

This causes alignment issues even when using flex.

By setting the line height of the wrapper to 0 this issue is resolved.

Added a new line height token of 0 to achieve this with box.

Refactored icon wrapper to use box.

## before

image

## after 

image


richbachman (approved):
Nice fix for this one.

PR #733, opened 2020-09-04
fix(checkbox): replace check and dash with icons

SiTaggart:
Replace the css after and before elements for the check and indeterminate states with an icon


richbachman (approved):
So much cleaner.

PR #740, opened 2020-09-09
feat(icons): add Connected Devices and Close icons

katieporter:
Add Connected Devices product icons and updated Close icon.


richbachman:
@katieporter somehow the `titleid` messed up during the conversion which is causing the build failure. I pulled locally and re-ran the conversion, and that fixed the icons. I can just push those fixed icons up, or can hop on a zoom and help troubleshoot your local setup. Let me know.

katieporter:
@richbachman oh weird, thanks for investigating that. I might just try it again and see it if fixes it, but if it still doesn't, I'd love to hop on a zoom and take a look.

serifluous (approved):

PR #760, opened 2020-09-15
feat(toast): add toast package

richbachman:
- [x] Add Toast package
- [x] Add simple Toaster to control placement of toasts
- [x] Add color-border-success light token
- [x] Add Toast doc page
- [x] Add open Toast example to docs


packages/paste-core/components/toast/src/Toaster.tsx
```diff
@@ -0,0 +1,28 @@
...
+import {Box, safelySpreadBoxProps} from '@twilio-paste/box';
+import {Stack} from '@twilio-paste/stack';
+
+export interface ToasterProps {
+  children: NonNullable<React.ReactNode>;
+}
+
+const Toaster: React.FC<ToasterProps> = ({children, ...props}) => {
+  return (
+    <Box position="fixed" right="space40" top="space40" zIndex="zIndex90" {...safelySpreadBoxProps(props)}>
```
richbachman, line 12: @SiTaggart should we add a placement prop now that controls whether the toaster is placed? Or wait until we further flush out the toaster?

packages/paste-core/components/toast/src/types.ts
```diff
@@ -0,0 +1,42 @@
+import * as PropTypes from 'prop-types';
+import {ValueOf} from '@twilio-paste/types';
+
+export const ToastIconVariants = {
```
SiTaggart, line 4: This is redundant, you can just use the ToastVariant constant.

packages/paste-core/components/toast/src/types.ts
```diff
@@ -0,0 +1,42 @@
...
+  ERROR: 'error',
+  NEUTRAL: 'neutral',
+  SUCCESS: 'success',
+  WARNING: 'warning',
+} as const;
+
+export const ToastRoles = {
```
SiTaggart, line 11: For toasts lets make them all role="status". For alerts it's fine as the idea is that they are not stacked. For toasts it's different, as they are designed to be stacked. A role of alert is an assertive live region, meaning it will often jump the queue when being announced by a screen reader. If you have 2 toasts and the last one is assertive, it'll cause the first one not to be announced.

packages/paste-core/components/toast/src/types.ts
```diff
@@ -0,0 +1,42 @@
...
+  ERROR: 'alert',
+  NEUTRAL: 'status',
+  SUCCESS: 'status',
+  WARNING: 'alert',
+} as const;
+
+export const ToastVariants = {
```
SiTaggart, line 18: Can this live in a constants file. It's not really a type.

packages/paste-core/components/toast/src/Toaster.tsx
SiTaggart: So firstly lets not call this a toaster. This can just be the toast container and it only cares about positioning the toasts on the screen relative to the window.

On the placement prop, I'm not sure we're even allowing you to change it. It should be fixed for every page and every app so toasts always appear in the same position.

packages/paste-core/components/toast/src/types.ts
richbachman: Got it. Thanks for the explanation.

packages/paste-core/components/toast/src/types.ts
richbachman: Done.

packages/paste-core/components/toast/src/Toaster.tsx
richbachman: Renamed Toaster to ToastContainer.

packages/paste-core/components/toast/src/Toast.tsx
```diff
@@ -0,0 +1,68 @@
...
+    default:
+      return <InformationIcon color="colorTextNeutral" decorative={false} title="information: " size="sizeIcon20" />;
+  }
+};
+
+const Toast: React.FC<ToastProps> = ({children, hidden, onDismiss, variant = 'neutral', ...props}) => {
+  const ToastComponent = ToastComponentVariants[variant];
+
+  return (
+    <ToastComponent hidden={hidden} role="status" variant={variant} {...safelySpreadBoxProps(props)}>
```
richbachman, line 42: @SiTaggart am I on the right track here by providing a prop to hide an individual toast? Then showing it via button click. See here: https://github.com/twilio-labs/paste/pull/760/files#diff-cec5e6eba87fe35785110f3a7946cd6cR148

packages/paste-website/src/pages/components/toast/index.mdx
```diff
@@ -0,0 +1,362 @@
...
+- **Error.** Communicate an issue that's preventing a user from completing their current action and can be fixed within their current flow.
+
+In general, use Toasts for response to user action. To decide whether you should use Toasts or another type of messaging, jump to [Toasts vs. Alerts vs. Inline messaging](#toast-vs-alerts-vs-inline-messaging).
+
+### Accessibility
+
+For [toasts that automatically time out](#toast-with-automatic-time-out), consider users with different reading speeds, vision levels, and literacy levels. We recommend giving users about 0.5 seconds per word. For short toasts at 5–10 words, that's about 2–4 seconds until you time-out a toast. For toasts at our [max suggested length of 140 characters](#composition-notes), that's about 30 seconds for about 55 English words. In this case, consider not making the toast automatically time out since the longer content length could also suggest that there's information worth persisting for the customer.
```
serifluous, line 83:
```suggestion
For [toasts that automatically time out](#toast-with-automatic-time-out), consider users with different reading speeds, vision levels, and literacy levels (more details in the research insight below). We recommend giving users about 0.5 seconds per word. For short toasts at 5–10 words, that's about 2–4 seconds until you time-out a toast. For toasts at our [max suggested length of 140 characters](#composition-notes), that's about 30 seconds for about 55 English words. In this case, consider not making the toast automatically time out since the longer content length could also suggest that there's information worth persisting for the customer.
```

packages/paste-website/src/pages/components/toast/index.mdx
```diff
@@ -0,0 +1,362 @@
...
+  <CalloutText>
+    Additional research by{' '}
+    <Anchor href="https://www.ncbi.nlm.nih.gov/pmc/articles/PMC5557829/">
+      researchers at the University of Alabama at Birmingham
+    </Anchor>{' '}
+    and by{' '}
+    <Anchor href="https://www.microsoft.com/en-us/research/uploads/prod/2019/01/readerview_chi2019.pdf">
+      Microsoft Research
+    </Anchor>{' '}
+    suggest roughly similar or higher reading rates due to increased age and dyslexia in adults.
```
serifluous, line 104:
```suggestion
    suggest reading rates are similar or slightly faster with increased age or dyslexia in adults.
```

packages/paste-core/components/toast/src/Toast.tsx
richbachman: Removed the hidden prop since its a valid HTML attribute.

packages/paste-website/src/pages/components/toast/index.mdx
```diff
@@ -0,0 +1,384 @@
...
+
+- The user wouldn't need to remember or copy the information in order to continue with their current flow, or any future flow.
+- The toast doesn't have any interactive content, such as a link or button.
+
+We recommend giving users about 0.5 seconds per word before timing out a toast.
+
+### Showing and Hiding a Toast
+
+For a basic method of hiding and showing, you can use the `hidden` HTML attribute. Passing a true or false value to that attribute will show or hide the Toast respectively.
+We'll be providing a more comprehensive method for displaying Toasts in the near future.
```
SiTaggart, line 232: I'd be tempted to make this a callout at the start of this section to make it _really_ clear that we acknowledge this is just an interim solution.

packages/paste-core/components/toast/src/Toast.tsx
```diff
@@ -0,0 +1,68 @@
...
+    default:
+      return <InformationIcon color="colorTextNeutral" decorative={false} title="information: " size="sizeIcon20" />;
+  }
+};
+
+const Toast: React.FC<ToastProps> = ({children, onDismiss, variant = 'neutral', ...props}) => {
+  const ToastComponent = ToastComponentVariants[variant];
+
+  return (
+    <ToastComponent role="status" variant={variant} {...safelySpreadBoxProps(props)}>
```
SiTaggart, line 42: You don't need to safely spread here as you're doing it already at the Component level

packages/paste-core/components/toast/src/ToastContainer.tsx
```diff
@@ -0,0 +1,35 @@
+import * as React from 'react';
+import * as PropTypes from 'prop-types';
+import {Box, safelySpreadBoxProps} from '@twilio-paste/box';
+import {Portal} from '@twilio-paste/reakit-library';
+import {Stack} from '@twilio-paste/stack';
+import {StyledBase} from '@twilio-paste/theme';
+
+export interface ToastContainerProps {
```
SiTaggart, line 8: As you have a types files, lets' put all type declarations in there

packages/paste-core/components/toast/src/Toast.tsx
```diff
@@ -0,0 +1,68 @@
...
+
+Toast.displayName = 'Toast';
+
+if (process.env.NODE_ENV === 'development') {
+  Toast.propTypes = ToastPropTypes;
+}
+
+export {Toast};
+
+export * from './types';
```
SiTaggart, line 68: Let's export all types from the index.ts file

packages/paste-core/components/toast/src/types.ts
```diff
@@ -0,0 +1,20 @@
+import * as PropTypes from 'prop-types';
+import {ValueOf} from '@twilio-paste/types';
+import {ToastVariantObject} from './constants';
+
+export type ToastVariants = ValueOf<typeof ToastVariantObject>;
+
+export interface ToastProps {
+  children: NonNullable<React.ReactNode>;
+  onDismiss?: () => void;
+  role?: string;
```
SiTaggart, line 10: I don't think you mean to have a role as part of the interface here. You don't actually allow folks to set it

packages/paste-core/components/toast/stories/index.stories.tsx
```diff
@@ -0,0 +1,162 @@
...
+import {Button} from '@twilio-paste/button';
+import {Text} from '@twilio-paste/text';
+import {Truncate} from '@twilio-paste/truncate';
+import {Toast, ToastContainer} from '../src';
+
+storiesOf('Components|Toast', module)
+  .addDecorator(withKnobs)
+  .add('Neutral', () => {
+    return (
+      <ToastContainer>
```
SiTaggart, line 14: It looks liek you're going to have set a box around these with a min height for the snapshots to work. Because the container is fixed, the story doesn't actually have any height

image

packages/paste-core/components/toast/package.json
```diff
@@ -0,0 +1,62 @@
...
+    "build:dev": "yarn clean && yarn compile:dev",
+    "build:props": "typedoc --tsconfig ./tsconfig.json --json ./dist/prop-types.json",
+    "clean": "rm -rf ./dist && rm -rf tsconfig.build.tsbuildinfo && rm -rf .rpt2_cache",
+    "compile": "rollup -c --environment NODE_ENV:production",
+    "compile:dev": "rollup -c --environment NODE_ENV:development",
+    "prepublishOnly": "yarn build",
+    "type-check": "tsc --noEmit"
+  },
+  "peerDependencies": {
+    "@twilio-paste/absolute": "^2.0.70",
```
SiTaggart, line 31: I don't think you're using absolute, is it a peer dependency of another thing you're using?

packages/paste-website/src/pages/components/toast/index.mdx
richbachman: I think thats a good call. Done.

packages/paste-core/components/toast/src/types.ts
richbachman: Yep. Missed that when I removed the different roles.

packages/paste-core/components/toast/stories/index.stories.tsx
richbachman: Good idea on the minHeight. Looks like that works.

packages/paste-core/components/toast/package.json
richbachman: Right. I had added because I thought there was a fixed variant, then forgot to remove it.

packages/paste-website/src/pages/components/toast/index.mdx
```diff
@@ -0,0 +1,390 @@
...
+
+### Showing and Hiding a Toast
+
+<Callout>
+  <CalloutText>
+    For a basic method of hiding and showing, you can use the <inlineCode>hidden</inlineCode> HTML attribute. Passing a
+    <inlineCode>true</inlineCode> or <inlineCode>false</inlineCode> value to that attribute will show or hide the Toast
+    respectively. We'll be providing a more comprehensive method for displaying Toasts in the near future.
+  </CalloutText>
+</Callout>
```
SiTaggart, line 155:
```suggestion
<Callout>
  <CalloutText>
    We'll be providing a more comprehensive method for displaying Toasts in the near future. Please bare with us.
  </CalloutText>
</Callout>

For a basic method of hiding and showing, you can use the `hidden` HTML attribute. Passing a `true` or `false` value to that attribute will show or hide the Toast respectively.
```

SiTaggart (approved):
Code looks good, just a very minor suggestion about the show/hide callout

SiTaggart (changes requested):
There's some kind of missing peer dependency that's causing the package to be 25kb. It should be no more than 1kb. Lets' look at all the files and what whats imported, and then follow those package peer dependency tree and make sure that all deps of deps are listed as peers

packages/paste-website/src/pages/components/toast/index.mdx
```diff
@@ -0,0 +1,390 @@
...
+Add an automatic time-out to a toast if it doesn't communicate any information that would help a user continue a flow, and would be annoying to the user to have to explicitly acknowledge and close.
+
+In most cases, it's better to dismiss the toast when a user reloads or navigates away from a page, rather than timing it out. Add an automatic time-out to a success toasts if the only thing it does it confirm that the action was completed. If the toast has any additional information or interactive content, like a link or button, don't make the toast time out.
+
+Add an automatic time-out if all these conditions are true:
+
+- The user wouldn't need to remember or copy the information in order to continue with their current flow, or any future flow.
+- The toast doesn't have any interactive content, such as a link or button.
+
+We recommend giving users about 0.5 seconds per word before timing out a toast.
```
serifluous, line 251:
```suggestion
Add an automatic time-out if all these conditions are true:

- The user wouldn't need to remember or copy the information in order to continue with their current flow, or any future flow.
- The toast doesn't have any interactive content, such as a link or button.

We recommend giving users about 0.5 seconds per word before timing out a toast.
```

packages/paste-website/src/pages/components/toast/index.mdx
```diff
@@ -0,0 +1,390 @@
...
+
+## Composition Notes
+
+Keep toast text brief and scannable by placing only the highest priority information in it. Too much text in a toast can overwhelm a user. Aim for no more than 140 characters in a toast, and keep the text directly related to the user's current action. Success toasts generally run about 50 characters long, while all other alerts might run about 100 characters long. However, don't [sacrifice being direct and straightforward](/content/product-style-guide#be-direct-and-straightforward) for a shorter message.
+
+Be specific in your response to the user's action and if applicable, provide a resolution by using an [anchor](/components/anchor/). Be cautious about using a button in a toast since it may compete with other buttons on the screen and should be used only to trigger an action. If you must use a button, use a single [small secondary button](/components/button#small-button) in a toast.
```
serifluous, line 257:
```suggestion
Be specific in your response to the user's action and if applicable, provide a resolution by navigating a user to another page using an [anchor](/components/anchor/). Be cautious about using a button in a toast since it should be used only to trigger an action. If you must use a button, use a [small secondary button](/components/button#small-button) in a toast.
```

packages/paste-website/src/pages/components/toast/index.mdx
```diff
@@ -0,0 +1,390 @@
...
+
+<Callout variant="warning">
+  <CalloutTitle>Don't forget the ToastContainer</CalloutTitle>
+  <CalloutText>
+    In order to control placement for Toasts, Paste provides you with a ToastContainer. The ToastContainer only handles
+    the positioning of Toasts in the top right area of the application. Multiple Toasts can be added to a single
+    ToastContainer. This ensures the correct spacing is added to each Toast.
+  </CalloutText>
+</Callout>
```
serifluous, line 266:
```suggestion
<Callout>
  <CalloutTitle>Don't forget the ToastContainer!</CalloutTitle>
  <CalloutText>
    In order to control placement for Toasts, Paste provides you with a ToastContainer. The ToastContainer only handles
    the positioning of Toasts in the top right area of the application. Multiple Toasts can be added to a single
    ToastContainer. This ensures the correct spacing is added to each Toast.
  </CalloutText>
</Callout>
```

packages/paste-website/src/pages/components/toast/index.mdx
```diff
@@ -0,0 +1,390 @@
...
+  </Stack>
+</Box>
+
+### Showing and Hiding a Toast
+
+<Callout>
+  <CalloutText>
+    We'll be providing a more comprehensive method for displaying Toasts in the near future. Please bare with us.
+  </CalloutText>
+</Callout>
```
serifluous, line 153:
```suggestion
<Callout variant="warning">
  <CalloutText>
    We'll be providing a more comprehensive method for displaying Toasts in the near future. Please bear with us.
  </CalloutText>
</Callout>
```

packages/paste-website/src/pages/components/toast/index.mdx
```diff
@@ -0,0 +1,390 @@
...
+  <CalloutText>
+    In order to control placement for Toasts, Paste provides you with a ToastContainer. The ToastContainer only handles
+    the positioning of Toasts in the top right area of the application. Multiple Toasts can be added to a single
+    ToastContainer. This ensures the correct spacing is added to each Toast.
+  </CalloutText>
+</Callout>
+
+## When to use a Toast
+
+Help users avoid scenarios where more than 3 toasts could show at a time. If more than 3 could appear at a time, show only one toast that summarizes the information you were intending to show in multiple toasts.
```
serifluous, line 270:
```suggestion
Help users avoid scenarios where more than 3 toasts could show at a time. If more than 3 could appear at a time, show only one toast that summarizes the information you were intending to show in all of the toasts.
```

packages/paste-website/src/pages/components/toast/index.mdx
```diff
@@ -0,0 +1,390 @@
...
+
+### Toast vs. Alerts vs. Inline messaging
+
+To communicate feedback on a user action, use a toast or in-page validation (_e.g._, full-page confirmation, inline error). You may also want to pair a toast with in-page validation. For example, when a user submits a long form, you can show an error toast that lets users know they need to fix some form fields, while also showing inline errors on each incorrect field.
+
+To communicate information from the system that isn't triggered by a user action, use an [alert](/components/alert) or other in-page messaging.
+
+##### Use Toasts when...
+
+- You need to communicate a direct response to a user action. **Example:** "asset.jpg" successfully added.
```
serifluous, line 280:
```suggestion
- You need to communicate a direct response to a user action. Example: _"asset.jpg" successfully added._
```

packages/paste-website/src/pages/components/toast/index.mdx
```diff
@@ -0,0 +1,390 @@
...
+- You're providing only one link or button to help a user continue a flow.
+
+##### Use Alerts when...
+
+- You need to notify a user about a high-priority or time-sensitive change that isn't triggered by a user action.
+- You're providing only one link or button to help a user resolve an issue.
+
+##### Use in-page messaging e.g., inline errors, in-page call-outs when...
```
serifluous, line 288:
```suggestion
##### Use in-page messaging, e.g., inline errors, in-page call-outs when...
```

packages/paste-website/src/pages/components/toast/index.mdx
```diff
@@ -0,0 +1,390 @@
...
+
+- You need to communicate a direct and contextual response to a user action.
+- You need to provide links or buttons contextually to an in-page element.
```
serifluous, line 291:
```suggestion
- You need to provide links or buttons next to an in-page element.
```

packages/paste-website/src/pages/components/toast/index.mdx
```diff
@@ -0,0 +1,390 @@
...
+- You need to communicate dynamically changing content inside a single toast, including showing progress bars. These cases should use a non-modal dialog instead.
```
serifluous, line 292:
```suggestion
- You need to communicate dynamically changing content, like a progress bar. Use a [non-modal dialog](primitives/non-modal-dialog-primitive) instead.
```

serifluous (changes requested):

packages/paste-website/src/pages/components/toast/index.mdx
```diff
@@ -0,0 +1,390 @@
...
+    Toast,
+    ToastContainer,
+    Text,
+    Button,
+  }}
+  noInline
+  language="jsx"
+>
+  {showHideExample}
+</LivePreview>
```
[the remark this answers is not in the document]
richbachman: Let's backlog that, so we can get this package out this sprint.

packages/paste-website/src/pages/components/toast/index.mdx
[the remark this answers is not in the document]
richbachman: Yeah. It's a bit of manual process for downstream teams to do that now. They'll have to add some sort of time to the state of the app, and under-render the Toast when the timer runs out. Eventually the Toaster will handle this as well. Let's backlog the example for now so we can get this out the door.

SiTaggart (approved):

packages/paste-website/src/pages/components/toast/index.mdx
```diff
@@ -0,0 +1,386 @@
...
+  <Text as="span">
+    Your email address has been updated. You can view your profile <Anchor href="#">here</Anchor>.
+  </Text>
+</Toast>`}
+</LivePreview>
+
+<Callout>
+  <CalloutTitle>When to automatically time-out a success toast</CalloutTitle>
+  <CalloutText>
+    Add an automatic time-out to a success toast if the _only_ thing it does is confirm that the action was completed. If
```
serifluous, line 203: whoops, this is showing markdown syntax. Forgot this was in the callout

serifluous (approved):

PR #778, opened 2020-09-18
fix(website): adjust callout to use stack

katieporter:
Adjusts Callout to use Stack instead of a top margin on CalloutText so that Callouts without CalloutHeading don't have uneven spacing.


aayushpi:
@katieporter Got an example where I can see the difference? Most of the callouts have headings.

katieporter:
> @katieporter Got an example where I can see the difference? Most of the callouts have headings.

https://paste.twilio.design/getting-started/engineering/. There are a couple examples towards the end.

richbachman (approved):

PR #797, opened 2020-09-25
fix(button): center content within buttons

richbachman:
- [x] Center justify button content
- [x] Update snapshots

### DIFF
--- a/packages/paste-core/components/button/stories/index.stories.tsx
+++ b/packages/paste-core/components/button/stories/index.stories.tsx
@@ -19,31 +19,62 @@ const AllSizeOptions: React.FC<{variant: ButtonVariants}> = ({variant}) => {
     const children = size === 'icon' ? <PlusIcon title="Add item to cart" decorative={false} /> : variant;
 
     allButtons.push(
-      <Box key={`variant-${variant}-${size}`} marginBottom="space40" padding="space30">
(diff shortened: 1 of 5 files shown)


packages/paste-core/components/button/stories/index.stories.tsx
```diff
@@ -60,17 +60,174 @@ const AllSizeOptions: React.FC<{variant: ButtonVariants}> = ({variant}) => {
...
-  .add('Destructive Button', () => <AllSizeOptions variant="destructive" />)
-  .add('Destructive Secondary Button', () => <AllSizeOptions variant="destructive_secondary" />)
-  .add('Destructive Link Button', () => <AllSizeOptions variant="destructive_link" />)
-  .add('Link Button', () => <AllSizeOptions variant="link" />)
-  .add('Inverse Link Button', () => <AllSizeOptions variant="inverse_link" />)
+  .add('Primary Button', () => (
+    <>
+      <AllSizeOptions variant="primary" />
+      <Box padding="space30">
+        <Stack orientation="vertical" spacing="space40">
```
SiTaggart, line 67: I think you can just pop these into the `allButtons` array, pass the variant down and not have to touch the individual stories in this case

SiTaggart (approved):

PR #805, opened 2020-09-28
feat(table): add component package

richbachman:
- [x] Added Table component, stories, and tests
- [x] Added necessary props to Box
- [x] Added word-wrap to Text
- [x] Added Table to component Core


packages/paste-core/components/table/src/Tr.tsx
```diff
@@ -0,0 +1,27 @@
...
+    <Box
+      {...safelySpreadBoxProps(props)}
+      ref={ref}
+      as="tr"
+      borderStyle="solid"
+      borderColor="colorBorderLighter"
+      borderWidth="borderWidth0"
+      borderBottomWidth="borderWidth10"
+      verticalAlign={verticalAlignment}
+      _even={{backgroundColor: 'colorBackgroundRowStriped'}}
```
serifluous, line 16: We could potentially deprecate this token, and use colorBackground here instead.

@SiTaggart originally added this token, so will wait for him to chime in.

SiTaggart: It depends on if you ever feel like you'd like specific control over row striping (across all components, not just tables), independent from a generic background.

Would a dark theme potentially need it? Doubt it, but you never know.

packages/paste-core/components/table/src/types.ts
```diff
@@ -0,0 +1,74 @@
+import * as PropTypes from 'prop-types';
+import {WidthOptions, isWidthTokenProp} from '@twilio-paste/style-props';
+
+export type AlignmentOptions = 'left' | 'center' | 'right';
+export type TableVariants = 'auto' | 'fixed';
+export type VerticalAlignmentOptions = 'top' | 'middle' | 'bottom';
```
SiTaggart, line 6: You could potentially keep these types in sync with the prop-types below by using constants.

Then you can derive the types from the constant, and populate the oneOf values from the constant too. That way it's a single update in the future to keep them in sync with each other

packages/paste-style-props/src/types/typography.ts
```diff
@@ -55,4 +57,5 @@ export interface TypographyProps {
   textTransform?: TextTransform;
   whiteSpace?: WhiteSpace;
   wordBreak?: WordBreak;
+  wordWrap?: WordWrap;
```
SiTaggart, line 60: Is there a prop-type that also needs this? I can't remember if we do prop types for typography

packages/paste-core/components/table/src/Td.tsx
```diff
@@ -0,0 +1,38 @@
...
+      fontSize="fontSize30"
+      lineHeight="lineHeight30"
+      paddingBottom="space40"
+      paddingLeft="space40"
+      paddingRight="space40"
+      paddingTop="space40"
+      position="relative"
+      textAlign={alignment}
+      verticalAlign="inherit"
+      width={width}
```
SiTaggart, line 20: I'm tempted to only have width set on the column header. Setting width on a cell seems odd, and limiting to headers focuses the size setting API to one area of a table. Thoughts?

richbachman: Yeah. I think thats a good idea. Seems weird to set the width in multiple areas.

packages/paste-style-props/src/types/typography.ts
richbachman: Yep, but forgot to add it to the typography prop types, so good catch. I just added it.

packages/paste-core/components/table/src/Table.tsx
```diff
@@ -0,0 +1,26 @@
+import * as React from 'react';
+import {Box, safelySpreadBoxProps} from '@twilio-paste/box';
+import {TableProps, TablePropTypes} from './types';
+
+const Table = React.forwardRef<HTMLTableElement, TableProps>(({variant = 'auto', ...props}, ref) => {
```
TheSisb, line 5: Just wondering if, for simplicity, we should call the `variant` prop `tableLayout`.  Similarly for the other parts of the table: `alignment` -> `textAlign`, `verticalAlignment` -> `verticalAlign`.

Reason being that it's a familiar API for folks who know how HTML tables work and we aren't really doing anything special with these props.  What do you think?

richbachman: I like that better. More familiar API, and it also opens up `variant` if we ever need that in the future for say an inverse table.

packages/paste-core/components/table/src/types.ts
```diff
@@ -0,0 +1,76 @@
...
+
+export interface TrProps {
+  children: NonNullable<React.ReactNode>;
+  verticalAlign?: TableVerticalAlignmentOptions;
+}
+
+export const TrPropTypes = {
+  children: PropTypes.node.isRequired,
+  verticalAlign: PropTypes.oneOf(
+    Object.keys(TableVerticalAlignmentObject).map(verticalAlignment => TableVerticalAlignmentObject[verticalAlignment])
```
SiTaggart, line 44: can't you do `Object.values(TableVerticalAlignmentObject)` for these?

packages/paste-core/components/table/stories/index.stories.tsx
```diff
@@ -0,0 +1,364 @@
...
+import {MoreIcon} from '@twilio-paste/icons/esm/MoreIcon';
+import {Text} from '@twilio-paste/text';
+import {Truncate} from '@twilio-paste/truncate';
+import {Table, THead, TBody, TFoot, Tr, Td, Th} from '../src';
+
+const MoreMenu: React.FC<{}> = () => {
+  const menu = useMenuState();
+  return (
+    <Box display="flex" justifyContent="flex-end">
+      <MenuButton {...menu} tableLayout="reset" size="reset">
```
SiTaggart, line 17: Not sure this is supposed to be `tableLayout`. It's given you some funky blue icon buttons

image

richbachman: Damn find/replace all!

packages/paste-core/components/table/src/types.ts
richbachman: Much simpler.

SiTaggart (approved):

packages/paste-core/components/table/stories/index.stories.tsx
```diff
@@ -0,0 +1,364 @@
...
+              </Text>
+            </Td>
+            <Td>Incoming</Td>
+            <Td>(602) 609-6747</Td>
+          </Tr>
+        </TBody>
+      </Table>
+    );
+  })
+  .add('Fixed table', () => {
```
TheSisb, line 96: The layperson may read this as "not broken". We should be a little more specific that this is a `tableLayout: fixed` Table

packages/paste-core/components/table/stories/index.stories.tsx
```diff
@@ -0,0 +1,364 @@
...
+            <Td>&nbsp;</Td>
+            <Td>Cell</Td>
+            <Td textAlign="center">Content</Td>
+            <Td textAlign="right">3</Td>
+          </Tr>
+        </TFoot>
+      </Table>
+    );
+  })
+  .add('Row textAlign', () => {
```
TheSisb, line 255: Isn't this example still a column textAlign? 😕

packages/paste-core/components/table/stories/index.stories.tsx
richbachman: Changing to `TableLayout: Fixed table`.

packages/paste-core/components/table/stories/index.stories.tsx
richbachman: I changed this one to `Row verticalAlign` to be more clear.

TheSisb (approved):

PR #811, opened 2020-09-30
feat(website): add alpha Table doc page

richbachman:
- [x] Added alpha Table doc page
- [x] Swapped out existing doc site tables with Table


katieporter:
I love that you used Table for the anatomy! For that section, could the "modifiable?" all be right aligned? And could all the columns across every anatomy table use set widths, so that it all lines up from table to table? I feel like the middle column should be the widest, and the two outer columns can be narrower.

packages/paste-website/src/pages/components/table/index.mdx
```diff
@@ -0,0 +1,436 @@
...
+| border-bottom- color | color-border-lighter                     | No          |
+
+### Th
+
+| Property    | Default token                                                            | Modifiable? |
+| ----------- | ------------------------------------------------------------------------ | ----------- |
+| font-size   | font-size-30                                                             | No          |
+| line-height | line-height-30                                                           | No          |
+| font-weight | font-weight-bold                                                         | No          |
+| padding     | space-40, First - padding-left: space-50, Last - padding-right: space-50 | No          |
```
katieporter, line 342: I'm having trouble understanding what "first" and "last" are referring to here.

katieporter: Same for the Td table.

packages/paste-website/src/pages/components/table/index.mdx
```diff
@@ -0,0 +1,436 @@
...
+| Property     | Default token        | Modifiable? |
+| ------------ | -------------------- | ----------- |
+| border-width | border-width-10      | No          |
+| border-color | color-border-lighter | No          |
+
+### Tr
+
+| Property             | Default token                            | Modifiable? |
+| -------------------- | ---------------------------------------- | ----------- |
+| background-color     | Even rows - color-background-row-striped | No          |
```
katieporter, line 331: Could you do "Even rows:" instead of "Even rows -"? Feels a little too dash-y haha which makes it harder to read.

packages/paste-website/src/pages/components/table/index.mdx
```diff
@@ -0,0 +1,436 @@
...
+    </Tr>
+    <Tr verticalAlignment="top">
+      <Td>W.E.B. Du Bois</Td>
+      <Td>Herein lies the tragedy of the age: not that men are poor—all men know something of poverty; not that men are wicked—who is good? Not that men are ignorant—what is truth? Nay, but that men know so little of men.</Td>
+    </Tr>
+  </TBody>
+</Table>`}
+</LivePreview>
+
+## Anatomy
```
katieporter, line 303: Do you want to include THead in this section, too?

richbachman: The `THead` is just a box `as="thead"`, so theres nothing really to mention on it. We're not doing any other custom styling for it.

richbachman:
@katieporter 

> For the props section, could all the columns across every anatomy table use set widths, so that it all lines up from table to table?

Good catch. I looked at the old website table, and it had a fixed table layout so I changed all of the prop and anatomy tables to follow that pattern.

packages/paste-website/src/pages/components/table/index.mdx
richbachman: I've updated this to read a little better. Let me know if that works.

packages/paste-core/components/table/src/Tr.tsx
```diff
@@ -0,0 +1,27 @@
...
+    <Box
+      {...safelySpreadBoxProps(props)}
+      ref={ref}
+      as="tr"
+      borderStyle="solid"
+      borderColor="colorBorderLighter"
+      borderWidth="borderWidth0"
+      borderBottomWidth="borderWidth10"
+      verticalAlign={verticalAlignment}
+      _even={{backgroundColor: 'colorBackgroundRowStriped'}}
```
serifluous, line 16: We could potentially deprecate this token, and use colorBackground here instead.

@SiTaggart originally added this token, so will wait for him to chime in.

katieporter (approved):

packages/paste-website/src/pages/components/table/index.mdx
katieporter: Much more clear!

packages/paste-website/src/pages/components/table/index.mdx
```diff
@@ -0,0 +1,503 @@
...
+- **TBody:** The main body of your table.
+- **TFoot:** The footer of your table.
+- **Tr:** A row of cells in your table.
+- **Th:** A header cell which can be used as a column title.
+- **Td:** A cell of table content.
+
+### Accessibility
+
+Tables are used to organize data with a logical relationship in grids. Accessible tables need HTML markup that indicates
+header cells and data cells and defines their relationship. Assistive technologies use this information to provide context to users.
```
SiTaggart, line 91: Under here it might be nice to list out 2 must haves:

- Tables **must** have a `THead`, containing a Table Header `Th` for each column.
- Tables **must** have a `TBody` wrapping the table body of rows.

packages/paste-website/src/pages/components/table/index.mdx
```diff
@@ -0,0 +1,503 @@
...
+  </TBody>
+</Table>`}
+</LivePreview>
+
+### Fixed Column Width Table
+
+The Fixed Table variant sets equal column widths for the table.
+
+<LivePreview scope={{Table, THead, TBody, TFoot, Tr, Th, Td}} language="jsx">
+  {`<Table variant="fixed">
```
SiTaggart, line 154: Just a reminder this API changed

SiTaggart (approved):

PR #818, opened 2020-10-02
allow Anchor prop override

vnguyen94:
Note: I made this PR from the Github UI; will let the PR build and see how it turns out first

This PR allows the user to override provided props if needed. My original use-case was to add custom ~line-height~ flex styling to an Anchor, but saw that it was always being overriden by the provided props (which might be by design).

edit: I probably won't merge this in as I see that I've bypassed the commitlint hook 😅  but I hope it at least shows the intent of the change

**Contributing to Twilio**

### DIFF
--- a/packages/paste-core/components/anchor/src/DefaultAnchor.tsx
+++ b/packages/paste-core/components/anchor/src/DefaultAnchor.tsx
@@ -1,10 +1,9 @@
 import * as React from 'react';
-import {Box} from '@twilio-paste/box';
+import {Box, safelySpreadBoxProps} from '@twilio-paste/box';
 import {AnchorProps, AnchorPropTypes} from './types';
 
 const DefaultAnchor = React.forwardRef<HTMLAnchorElement, AnchorProps>((props, ref) => (
   <Box
-    {...props}
     as="a"
     color="colorTextLink"
     fontSize="inherit"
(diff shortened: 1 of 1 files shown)


SiTaggart:
@vnguyen94 can you open up a discussion topic on this please so we can understand what you're trying to do with the Anchor component?

The anchor doesn't set a display property, and sets inherit on line height, font size and font weight too. In effect all it really does it set the color and add some attributes for security (very over simplified). 

You should be able to set a display property, and if placed inside of text that is already styled, say a heading, it'll inherit the overall text characteristics, but just set the color and hover/focus styles for underline.

SiTaggart:
Hey @vnguyen94 following up here. Can you open a discussion to go over the issues you faced? I'm still of the opinion that Anchor is fine the way it is due to it inheriting typography styles and I'd like to close this if that's true. Thanks

vnguyen94:
Sorry we had a team hackathon last week so I wasn't able to get around to this in a timely fashion. I'll close this for now

vnguyen94:
@SiTaggart so the real issue was that i previously was wrapping my text in a Box which i think was working, until i then started wrapping my text in a Truncate. my current solution was to tack on custom styling to the Truncate: 

```jsx
  return (
    <Box maxWidth="size50" display="inline-flex">
      <Anchor href={payload.url}>
        <Truncate style={{ display: 'flex' }}>{payload.title}</Truncate>
      </Anchor>
    </Box>
  );
```

SiTaggart:
@vnguyen94 it's still hard to tell from the snippet, what's not working? What are the expected behaviors and actual behaviors in this situation? Codesandbox would be helpful too to demonstrate it.

PR #819, opened 2020-10-02
feat(website): full width and inverse header

richbachman:
- [x] Added full width inverse header
- [x] Adjusted initial theme to default theme
- [x] Reworked sidebar
- [x] Added `size-sidebar` token


katieporter:
Can the focus state on the toggle match the focus state of the search box?

katieporter:
Focus state of the nav items in the sidebar should match what's documented here: https://share.goabstract.com/98d036a3-34b9-476c-adf6-ac222095f445?mode=design&sha=ecda37d859190c31c5f97a01d613ff6b015c0b5c

katieporter:
Sidebar is scrolling even when it doesn't need to: http://g.recordit.co/azMNH2PsSO.gif

katieporter:
Can the "Contact us" menu appear to the right instead of on top?
image

More like so:
image

katieporter:
IMO we can get rid of the warning banner on the Default theme, though @aayushpi may need to make the final call on that one.

Also, I think we may want to leave the toggle as-is (with Current / Preview) until we launch the new homepage. That was my bad - I mocked up a future state in the design and forgot to communicate that.

katieporter:
Can you add 4px of margin between the chevron icon and the text in the sidebar?

katieporter (changes requested):
This looks amaaaazing! Left a few comments about tiny tweaks, but overall this is looking so good.

aayushpi:
OK to remove the alert.

richbachman:
@katieporter I've pushed changes to account for your comments. Let me know if you see anything else. Thanks!

katieporter:
Looks great! I think the only remaining thing is to adjust the theme toggle values back to Current / Preview like we have today. We'll want to wait to change the toggle until we launch the new homepage.

richbachman:
> I think the only remaining thing is to adjust the theme toggle values back to Current / Preview like we have today. We'll want to wait to change the toggle until we launch the new homepage.

@katieporter that should be fixed now.

katieporter (approved):

packages/paste-design-tokens/tokens/aliases/sizing.yml
```diff
@@ -35,3 +35,5 @@ aliases:
   size-100: "64rem" #1024px
   size-110: "70.5rem" #1128px
   size-120: "77rem" #1264px
+
+  size-sidebar-10: "13.75rem" #220px
```
SiTaggart, line 39: I don't mind the concept of a sharable sidebar width. Especially if we're in the business of making sure our navs are cohesive across products.

But are we really going to have more than one sidebar size? You might for the states, expanded and collapsed, but I'm not sure they they be on a 10 step scale. They're just one off use case tokens

packages/paste-website/src/components/ContactUsMenu.tsx
```diff
@@ -16,33 +18,45 @@ export const ContactUsMenu: React.FC<{}> = () => {
...
-          Report a bug
-        </MenuItem>
-      </Menu>
-    </>
+    <Box position="fixed" bottom={0} left={0}>
+      <Box backgroundColor="colorBackground" padding="space20">
+        <MenuButton {...menu} variant="reset" size="reset">
+          {/* 212px / 13.25rem minWidth because of the boxshadow focus state being cut off by the sidebar width and edge of browser.
+        sizeSidebar10 - 8px =  212px / 13.25rem. */}
+          <Box as="span" display="flex" alignItems="center" padding="space40" minWidth="13.25rem">
```
SiTaggart, line 26: rather than do all this, can't you just put it in a box with 8px padding?

packages/paste-website/src/components/ContactUsMenu.tsx
```diff
@@ -16,33 +18,45 @@ export const ContactUsMenu: React.FC<{}> = () => {
...
-        <MenuItem
-          {...menu}
-          href="https://github.com/twilio-labs/paste/issues/new?assignees=&labels=Type%3A+Bug&template=bug_report.md&title="
-          onClick={() => handleClick('Contact Menu', 'Report a bug')}
-        >
-          Report a bug
-        </MenuItem>
-      </Menu>
-    </>
+    <Box position="fixed" bottom={0} left={0}>
```
SiTaggart, line 21: When hovering over a nav item, this kind of bugs me

Screen Shot 2020-10-05 at 16 30 36

packages/paste-website/src/components/site-wrapper/sidebar/SidebarNavigation.styles.ts
```diff
@@ -17,86 +17,117 @@ interface SiteNavNestListProps {
...
 `;
 
 export const SiteNavItem = styled.li`
-  margin: ${themeGet('space.space10')} 0;
+  margin: ${themeGet('space.space20')};
   font-size: ${themeGet('fontSizes.fontSize30')};
   line-height: ${themeGet('lineHeights.lineHeight20')};
 `;
 
+export const SiteNavNestedItem = styled(SiteNavItem)`
```
SiTaggart, line 29: Any scope to convert a bunch of this to `Box`?

packages/paste-website/src/components/site-wrapper/SiteHeader.tsx
```diff
@@ -29,34 +42,64 @@ export const SiteHeader: React.FC = () => {
...
-      borderColor="colorBorderLight"
-      borderStyle="none none solid none"
-      borderBottomWidth="borderWidth10"
-      paddingLeft="space200"
-      paddingRight="space80"
-      paddingTop="space50"
-      paddingBottom="space50"
+      display="flex"
+      alignItems="center"
+      backgroundColor="colorBackgroundPrimaryDarkest"
```
SiTaggart, line 47: Shouldn't this be using an inverse background token?

SiTaggart: Also found a bug in inverse anchor. Hover state doesn't pass contrast 

image

packages/paste-website/src/components/site-wrapper/SiteHeader.tsx
```diff
@@ -29,34 +42,64 @@ export const SiteHeader: React.FC = () => {
...
                 value={value}
                 onChange={event => {
                   setValue(event.target.value);
                 }}
+                variant="inverse"
+                insertBefore={<SearchIcon decorative={false} title="search" />}
               />
             </Box>
-            <ContactUsMenu />
             <Anchor
```
SiTaggart, line 103: Can you make this anchor display flex or block? I was expecting to see the focus state around the entire thing. Right now it collapses

image

packages/paste-website/src/components/site-wrapper/SiteHeader.tsx
```diff
@@ -29,34 +42,64 @@ export const SiteHeader: React.FC = () => {
...
+      display="flex"
+      alignItems="center"
+      backgroundColor="colorBackgroundPrimaryDarkest"
       position="sticky"
-      top={`${getMainContentComputedOffset(isPasteTheme)}px`}
+      top={`${getMainContentComputedOffset()}px`}
       zIndex="zIndex80"
     >
-      <Box display="flex" justifyContent="space-between" alignItems="center" flexDirection="row">
+      <Box display="flex" alignItems="center" paddingRight="space70" paddingLeft="space70" minWidth="sizeSidebar10">
```
SiTaggart, line 52: Curious why you chose to copy paste this and delete the component, rather than update the component and import it into a different location

packages/paste-website/src/components/site-wrapper/sidebar/SidebarNavigation.tsx
```diff
@@ -96,6 +101,24 @@ const pageQuery = graphql`
   }
 `;
 
+const SideNavDisclosureButton: React.FC<SiteNavButtonProps> = props => {
```
SiTaggart, line 104: If you convert SiteNavButton to box you probably don't need this extra component.

packages/paste-website/src/components/site-wrapper/sidebar/SidebarNavigation.tsx
```diff
@@ -130,208 +153,238 @@ const SidebarNavigation: React.FC<SidebarNavigationProps> = () => {
...
-            </SiteNavItem>
-            <SiteNavItem>
-              <SiteNavAnchor to="/getting-started/about-paste">About Paste</SiteNavAnchor>
-            </SiteNavItem>
-            <SiteNavItem>
-              <SiteNavAnchor to="/getting-started/faqs">FAQs</SiteNavAnchor>
-            </SiteNavItem>
-          </SiteNavNestList>
+          </SideNavDisclosureButton>
+          <DisclosurePrimitiveContent visible={gettingStartedOpen}>
```
SiTaggart, line 163: I see, so yeah, we should use all the bits from the disclosure primitive. What is happening is you're losing all the accessibility goodness like `aria-controls` on the button, which points to the id of this div when using all the pieces together

packages/paste-website/src/utils/stickyUtils.ts
```diff
@@ -1,15 +1,13 @@
-import {PSA_ALERT_HEIGHT, PASTE_THEME_ALERT_HEIGHT, SITE_MASTHEAD_HEIGHT} from '../constants';
+import {PSA_ALERT_HEIGHT, SITE_MASTHEAD_HEIGHT} from '../constants';
 
-export const getMainContentComputedOffset = (isPasteTheme: boolean): number => {
-  if (isPasteTheme) {
-    return PSA_ALERT_HEIGHT + PASTE_THEME_ALERT_HEIGHT;
-  }
+export const getMainContentComputedOffset = (): number => {
```
SiTaggart, line 3: Are any of these even computed anymore? They all just return static numbers now, always. Could just be converted to constants

richbachman: You're right. I refactored these.

packages/paste-website/src/components/site-wrapper/sidebar/SidebarNavigation.tsx
richbachman: Ah. I see. I refactored this using the `DisclosurePrimitive`.

packages/paste-website/src/components/site-wrapper/sidebar/SidebarNavigation.tsx
richbachman: I've refactored to use `Box`.

packages/paste-website/src/components/site-wrapper/SiteHeader.tsx
richbachman: Oversight on my part. I copied the original just to get it in place, but then never refactored. Done now.

packages/paste-website/src/components/site-wrapper/SiteHeader.tsx
richbachman: I'll create a ticket it the backlog for this.

packages/paste-website/src/components/site-wrapper/sidebar/SidebarNavigation.styles.ts
richbachman: Yeah. I refactored it all to use `Box`.

packages/paste-design-tokens/tokens/aliases/sizing.yml
richbachman: Good point. Changed to not include the -10.

packages/paste-website/src/components/site-wrapper/SiteHeader.tsx
richbachman: Following this design for the background color: https://app.abstract.com/projects/39ed9aa7-cca8-4e4f-8f16-d8f9d36d251a/branches/master/commits/latest/files/6a521d86-90ac-4db8-83b6-9dbfd0c84e7f/layers/B33BCCE8-FABA-4C0D-A0CF-3BCC8986465A?collectionId=bc3992f2-7d22-4394-967d-a62ba262e26f&collectionLayerId=efe4c114-c1ec-47a1-a35d-8d2f2ad50c78

katieporter: Yeah, `colorBackgroundPrimaryDarkest` is the right token. It's not our inverse background color token, which is the darkest color on the gray scale. This is the darkest color on the blue / primary scale. We discussed naming of this token a lottttt and landed on this, since we don't intend for this to be the inverse background color in most cases.

packages/paste-website/src/components/ContactUsMenu.tsx
richbachman: I've moved the button up a bit so its no longer covered by that text.

packages/paste-website/src/components/ContactUsMenu.tsx
richbachman: Refactored so no longer need the width hack.

packages/paste-core/components/button/src/types.ts
```diff
@@ -29,6 +30,17 @@ export const DirectButtonPropTypes = {
   type: PropTypes.oneOf(['submit', 'button', 'reset']) as any,
   disabled: PropTypes.bool,
   buttonState: PropTypes.oneOf(['disabled', 'loading', 'default']).isRequired as any,
+  variant: PropTypes.oneOf([
```
richbachman, line 33: Created a ticket to refactor these as a map like we do with other components, i.e. Alert, Anchor, etc.

packages/paste-website/src/utils/stickyUtils.ts
```diff
@@ -1,15 +1,5 @@
...
-
-export const getStickyColumnComputedOffset = (isPasteTheme: boolean, topPad: number = 20): number => {
-  if (isPasteTheme) {
-    return topPad + PASTE_THEME_ALERT_HEIGHT + PSA_ALERT_HEIGHT + SITE_MASTHEAD_HEIGHT;
-  }
-  return topPad + PSA_ALERT_HEIGHT + SITE_MASTHEAD_HEIGHT;
-};
+const topPad = 20;
+export const getStickyColumnComputedOffset = topPad + PSA_ALERT_HEIGHT + SITE_MASTHEAD_HEIGHT;
+export const getStickySidebarComputedOffset = PSA_ALERT_HEIGHT + SITE_MASTHEAD_HEIGHT;
```
SiTaggart, line 5: Stick all these in `constants`. Change the name to something like `STICKY_PAGE_COLUMN_OFFSET` `STICKY_SIDEBAR_OFFSET`, get rid of `topPad`, just hard code it in the constant

SiTaggart (changes requested):

packages/paste-website/src/components/ContactUsMenu.tsx
```diff
@@ -16,9 +18,17 @@ export const ContactUsMenu: React.FC<{}> = () => {
   };
 
   return (
-    <>
-      <MenuButton {...menu} variant="link">
-        Contact us <ChevronDownIcon decorative />
+    <Box backgroundColor="colorBackground" marginTop="auto" padding="space20" paddingBottom="space70">
+      <MenuButton {...menu} variant="reset" size="reset" fullWidth>
+        <Box as="span" display="flex" alignItems="center" padding="space40" width="100%">
+          <ProductSupportIcon decorative={false} title="Description of icon" />
```
SiTaggart, line 24: > title="Description of icon"

Make this icon decorative. it adds nothing to a screen reader user

packages/paste-website/src/components/site-wrapper/SiteHeader.tsx
```diff
@@ -29,35 +30,46 @@ export const SiteHeader: React.FC = () => {
...
                 id="search-input"
-                type="search"
+                type="text"
                 placeholder="Search..."
                 value={value}
                 onChange={event => {
                   setValue(event.target.value);
                 }}
+                variant="inverse"
+                insertBefore={<SearchIcon decorative={false} title="search" />}
```
SiTaggart, line 68: Decorative icon here too.

packages/paste-website/src/components/ContactUsMenu.tsx
```diff
@@ -16,9 +18,17 @@ export const ContactUsMenu: React.FC<{}> = () => {
...
+          <Box as="span" marginLeft="space30" textAlign="left">
+            Contact us
+          </Box>
+          <Box as="span" marginLeft="auto">
+            <MoreIcon decorative={false} title="More options" />
```
SiTaggart, line 29: Decorative here too

packages/paste-website/src/components/icons/PasteIconInverse.tsx
```diff
@@ -0,0 +1,30 @@
...
+  color?: string;
+  decorative?: boolean;
+  display?: string;
+  size?: number;
+  title?: string;
+}
+
+const PasteIconInverse = React.memo(
+  ({title = 'Twilio Paste', className, color, decorative = true, display, size}: PasteIconInverseProps) => {
+    const titleId = useUID();
```
SiTaggart, line 15: Is there a way to give id a specific prefix, the id is not unique to the page due to a bug with Paste icons. 

We have 2 sets of useUID starting from 1.

TheSisb: `titleId = thing-${useUID()}` I believe

TheSisb (approved):

SiTaggart (approved):
Much better VoiceOver experience on the homepage now

PR #820, opened 2020-10-05
fix(modal): fix console-patch jank by removing animation on modal close

TheSisb:
The Modal component has some jittering in console because our hack to fix the paddingTop on the page body is getting removed before the modal close animation completes.

Currently react-spring v9 doesn't have a mechanism to tap into `onDestroyed` like v8 had (https://github.com/pmndrs/react-spring/issues/1114#issuecomment-675076537). To be fair, this is a pretty hacky fix on our part anyways, so this shouldn't be an issue that comes up frequently.

That said, the experience is suboptimal on Console so I've gone ahead and disabled the animation on Modal close if the patch is applied.  This removes the jank at not much visual cost, since the Modal still animates in.

### DIFF
--- a/packages/paste-core/components/modal/src/Modal.tsx
+++ b/packages/paste-core/components/modal/src/Modal.tsx
@@ -69,17 +69,22 @@ export interface ModalProps extends React.HTMLAttributes<HTMLDivElement> {
   __console_patch?: boolean;
 }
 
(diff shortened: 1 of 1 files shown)


richbachman (approved):

SiTaggart (approved):
Looks fine to me

PR #821, opened 2020-10-05
fix(theme): hide reach/dialog css warning

TheSisb:
Fixes: https://github.com/twilio-labs/paste/discussions/524


richbachman (approved):

packages/paste-theme/src/themeProvider.tsx
```diff
@@ -29,6 +29,12 @@ const pasteGlobalStyles = css({
       scrollBehavior: 'auto !important' as SystemStyleObject,
     },
   },
+  // To disable reach/dialog's style warning
+  // Issue: https://github.com/twilio-labs/paste/discussions/524
+  // Reach Docs Guidance: https://reach.tech/styling#skip-including-styles
+  ':root': {
+    ['--reach-dialog' as any]: '1',
+  },
```
SiTaggart, line 37: Should we do a global from inside the modal package?

TheSisb: I tried that initially but it wasn't working :(.  My hunch is that since Modal loads it's dependencies first (i.e.: reach/dialog) and then the CSS was being applied from emotion after, it wasn't being detected.  Possible I did it wrong too

SiTaggart (approved):

packages/paste-theme/src/themeProvider.tsx
SiTaggart: It's no biggy, just an idea I had, loading order would probably be a thing we'd be battling. If it works, it works. I'm happy with it

PR #825, opened 2020-10-05
feat: toaster 

SiTaggart:
Friday afternoon POC of a toaster.

Uses a hook to control state, pushing a new toast and popping an existing toast. Can be excluded to provide your own state management of toasts, just pass it an array of toasts to show.


serifluous:
Can the timeout be given in seconds instead of milliseconds? The recommendation in the accessibility section gives guidance in seconds.

serifluous (commented):
Can the timeout be given in seconds instead of milliseconds? The recommendation in the accessibility section gives guidance in seconds.

serifluous (changes requested):
Both the ToastContainer and Toaster should place toasts space-90 (32px) from the top of the page. It looks like it's ~12px right now.

serifluous:
When Toasts overflow the page, is there a way to scroll to whichever Toast is in focus (just through tabbing, not with mouse scroll)?

Our guidance for max # of toasts is 3 per page anyway, so it's not a thing people should be doing...but would be nice.

packages/paste-core/layout/stack/src/index.tsx
```diff
@@ -91,12 +91,13 @@ const Stack: React.FC<StackProps> = ({children, orientation, spacing, ...props})
...
   const stackStyles = React.useMemo(() => getStackStyles(orientation), [orientation]);
   const childMargins = React.useMemo(() => getStackChildMargins(orientation, spacing), [orientation, spacing]);
+  const keySeed = useUIDSeed();
 
   return (
     <Box {...safelySpreadBoxProps(props)} {...stackStyles}>
       {validChildren.map((child, index) => {
         return (
-          <Box {...(childrenCount !== index + 1 ? childMargins : null)} key={useUID()}>
+          <Box {...(childrenCount !== index + 1 ? childMargins : null)} key={keySeed(`${index}-stack`)}>
```
TheSisb, line 100: Is this change working as intended? From what I've gathered, this will return `val-${index}-stack` and not provide "uid" functionality :( https://github.com/thearnica/react-uid/blob/master/src/uid.ts#L23

This may work better as `keySeed(child)`. The API is confusing and I plan to improve this in our UID lib hopefully.

packages/paste-core/components/toast/src/types.ts
```diff
@@ -8,19 +8,47 @@ export interface ToastProps extends React.HTMLAttributes<HTMLDivElement> {
...
+}
+
+export type ToasterProps = Pick<UseToasterReturnedProps, 'toasts' | 'pop'>;
+export const ToasterPropTypes = {
+  toasts: PropTypes.arrayOf(
+    PropTypes.shape({
+      id: PropTypes.string,
+      message: PropTypes.node.isRequired,
+      dismissAfter: PropTypes.number,
+      timeOutId: PropTypes.number,
```
TheSisb, line 49: Looks like these two types aren't used here, only on "ToasterToast"

packages/paste-website/src/pages/components/toast/index.mdx
```diff
@@ -163,6 +193,12 @@ Add an automatic time-out if all these conditions are true:
 
 We recommend giving users about 0.5 seconds per word before timing out a toast.
 
+#### Basic
+
+For a basic method of hiding and showing, you can just render a `<Toast />` inside of a `<ToastContainer />`.
+
+When showing multiple toasts, place the most recent toast at the bottom of the stack.
```
TheSisb, line 200: I know what you mean but it reads weird haha.  At first I thought "at the top of the stack" and now I'm not sure sure. FIFO? New toasts go into the front of an array?

TheSisb (commented):
This is very solid work and I love the animations.  Might need a few minor tweaks but overall looks great. Left a few comments but mostly nits.

packages/paste-core/components/toast/src/ToastPortal.tsx
```diff
@@ -0,0 +1,32 @@
...
+export interface ToastPortalProps {
+  children: NonNullable<React.ReactNode>;
+}
+
+const ToastPortal: React.FC<ToastPortalProps> = ({children, ...props}) => {
+  return (
+    <Portal>
+      {/* import Paste Theme Based Styles due to portal positioning. */}
+      <StyledBase>
+        <Box {...safelySpreadBoxProps(props)} position="fixed" right="space40" top="space40" zIndex="zIndex90">
```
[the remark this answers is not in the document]
SiTaggart: The position to the window is shared across `ToastContainer` and `Toaster`. So both the `Toaster` and `ToastContainer` now use the `ToastPortal` for positioning. 

First approach was to try and use the `ToastContainer` for the `Toaster`, but as `ToastContainer` came first, I couldn't really think of a nice way to not introduce a breaking change to anyone already using it as a basic toast positioner because of the use of `Stack`. So all I did was abstract the common things; needs to be a portal, needs base styles and positioned to the window top right. I don't _love_ it, but hoping the `ToastContainer` is deprecated in the future.

packages/paste-core/components/toast/src/Toaster.tsx
```diff
@@ -0,0 +1,79 @@
...
+  trigger: HTMLElement | null;
+}
+
+const Toaster: React.FC<ToasterProps> = ({toasts, pop, ...props}) => {
+  const [refMap] = React.useState(() => new WeakMap());
+  const [returnTrigger, setReturnTrigger] = React.useState<ReturnTriggerState>({trigger: null});
+
+  const transitions = useTransition(toasts, {
+    from: {height: 0, opacity: 0, transform: 'translateX(100%)'},
+    enter: item => async next => {
```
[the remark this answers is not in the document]
SiTaggart: It's not particularly well documented, but if you don't feed it an async function, you can't actually get access to `item`, even though it's in the closure. They use this pattern a bunch in their examples too. :shrug:

packages/paste-core/components/toast/src/Toaster.tsx
```diff
@@ -0,0 +1,79 @@
...
+    leave: {height: 0, opacity: 0, transform: 'translateX(0px) scale(0.8)'},
+    config: {
+      mass: 0.8,
+      tension: 350,
+      friction: 28,
+    },
+  });
+
+  const clearReturnTrigger = React.useCallback(() => {
+    setReturnTrigger({trigger: null});
```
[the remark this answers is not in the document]
SiTaggart: Sure, there's some "stuff" going on in here for sure

packages/paste-core/components/toast/src/Toaster.tsx
```diff
@@ -0,0 +1,79 @@
...
+import {ToasterProps, ToasterPropTypes} from './types';
+
+export const AnimatedToast = animated(Box);
+
+interface ReturnTriggerState {
+  trigger: HTMLElement | null;
+}
+
+const Toaster: React.FC<ToasterProps> = ({toasts, pop, ...props}) => {
+  const [refMap] = React.useState(() => new WeakMap());
```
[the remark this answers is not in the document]
SiTaggart: My only concern here is I can't seem to remove anything from it when I'm popping toasts from the stack. `refMap.delete(item)` "works", air quotes, in that if I step through it and inspect the map before and after the item is deleted. But it doesn't actually see to persist or reflect in the map used in the state

packages/paste-core/components/toast/src/types.ts
SiTaggart: Well toasts is an array of `ToasterToast`. They are valid if you're _not_ using the `useToaster` hook and for what ever reason you are wanting to roll your own state. In which case you're dealing with setting this stuff

SiTaggart:
@serifluous 

> Can the timeout be given in seconds instead of milliseconds? The recommendation in the accessibility section gives guidance in seconds.

It is a developer API though and when dealing with time we deal with milliseconds.

> When Toasts overflow the page, is there a way to scroll to whichever Toast is in focus (just through tabbing, not with mouse scroll)?

Can you explain this in a different way? I'm not sure I understand what you're asking?

serifluous:
> It is a developer API though and when dealing with time we deal with milliseconds.

Would be worth translating the timing recommendations to milliseconds in the docs ("0.5 seconds (500 milliseconds)"), so it's more of a no-brainer when people play around with the Toaster in the docs.

serifluous:
InformationIcon can be swapped to NeutralIcon now that the icon's been added.

packages/paste-core/layout/stack/src/index.tsx
SiTaggart: It seems to work, but I prefer `keySeed(child)` 👍

SiTaggart: OK, back to using an index in the string of a seed. It will create a unique seed for each stack, and within each stack the children are unique based on index.

If I used `keySeed(child)` it causes an infinite loop due to too many state updates when the children are dynamic.

packages/paste-core/components/toast/stories/index.stories.tsx
```diff
@@ -154,18 +161,108 @@ storiesOf('Components|Toast', module)
...
+              />
+            </div>
+            <div>
+              <Label htmlFor={dismissAfterID}>Dismiss after</Label>
+              <Input
+                id={dismissAfterID}
+                value={timeout}
+                type="number"
+                onChange={e => setTimeout(e.target.value)}
+                placeholder="milliseconds"
```
serifluous, line 235: "Milliseconds" should be moved either to help text or into the label

packages/paste-website/src/pages/components/toast/index.mdx
```diff
@@ -144,38 +150,114 @@ Here are ways that you can compose neutral, warning, error, success, and dismiss
...
-We recommend giving users about 0.5 seconds per word before timing out a toast.
+We recommend giving users about 500 milliseconds per word before timing out a toast.
+
+#### Toaster component
+
+The `<Toaster />` component handles rendering the toasts it is passed via props. It handles how the Toast enters and leaves the UI.
+
+If you pass it the returned state of the `useToast` hook, state, pushing, popping and handling browser window time-outs is taken care of for you.
+
+If you choose to handle state entirely yourself, you can use `<Toaster />` on it's own, but you are fully responsible for handling state, timeouts and the dismiss functionality.
```
serifluous, line 246:
```suggestion
If you choose to handle state entirely yourself, you can use `<Toaster />` on its own, but you are fully responsible for handling state, time-outs, and the dismiss functionality.
```

packages/paste-website/src/pages/components/toast/index.mdx
```diff
@@ -144,38 +150,114 @@ Here are ways that you can compose neutral, warning, error, success, and dismiss
...
 - The toast doesn't have any interactive content, such as a link or button.
 
-We recommend giving users about 0.5 seconds per word before timing out a toast.
+We recommend giving users about 500 milliseconds per word before timing out a toast.
+
+#### Toaster component
+
+The `<Toaster />` component handles rendering the toasts it is passed via props. It handles how the Toast enters and leaves the UI.
+
+If you pass it the returned state of the `useToast` hook, state, pushing, popping and handling browser window time-outs is taken care of for you.
```
serifluous, line 244:
```suggestion
If you pass it the returned state of the `useToaster` hook, state, pushing, popping, and handling browser window time-outs is taken care of for you.
```

packages/paste-website/src/pages/components/toast/index.mdx
```diff
@@ -144,38 +150,114 @@ Here are ways that you can compose neutral, warning, error, success, and dismiss
...
 
 Add an automatic time-out if all these conditions are true:
 
 - The user wouldn't need to remember or copy the information in order to continue with their current flow, or any future flow.
 - The toast doesn't have any interactive content, such as a link or button.
 
-We recommend giving users about 0.5 seconds per word before timing out a toast.
+We recommend giving users about 500 milliseconds per word before timing out a toast.
+
+#### Toaster component
```
serifluous, line 240: In what scenario would someone want to use Toaster by itself? Do we need to include this section up here, or can we move it to the API section?

serifluous (dismissed):
"Don't forget the ToastContainer!" callout under Composition Notes can be removed now.

packages/paste-core/components/toast/src/Toaster.tsx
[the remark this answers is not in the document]
TheSisb: I don't quite follow. Are you saying that manually calling "delete()" works, but by deleting the item the `refMap` still logs it as part of the state? The UI seems to respond to the changes correctly, so it could just be a weird React thing? Have you tried forcing garbage collection to see if that cleared it: https://developers.google.com/web/tools/chrome-devtools/memory-problems#visualize_memory_leaks_with_timeline_recordings ?

TheSisb:
Taking a closer look at the animation, I wonder if it's too fast for a toast 🤔 I generally like snappy motion for UI elements that react or are caused by user interaction. In the case of a toast, it is trying to grab the user's attention via their periphery and so a slower animation feels less sudden or excited.  What do you think about the following config:
```
config: {
    mass: 1,
    tension: 120,
    friction: 18,
},
```


--- 

Just for my future self, I'm pasting some other scribbles of mine that I wrote while playing with the animation. No better place to store this lol.
```
  const leaveItemHeight = React.useRef();

  // The units have to match for the animations to calculate correctly
  const marginBottom0 = `${theme.space.space0}rem`;

  const transitions = useTransition(toasts, {
    from: {opacity: 0, height: 0, marginBottom: marginBottom0, transform: 'translateX(100%)'},
    enter: item => async next => {
      await next({
        height: refMap.get(item).offsetHeight,
        marginBottom: theme.space.space40,
        opacity: 1,
        transform: 'translateX(0%)',
      });
    },
    leave: {
      to: {opacity: 0.3, height: 0, marginBottom: marginBottom0, transform: 'translateX(0%)'},
      from: {
        opacity: 1,
        height: leaveItemHeight.current,
        marginBottom: theme.space.space40,
        transform: 'translateX(0%)',
        overflow: 'hidden',
      },
    },
    config: {
      mass: 0.8,
      tension: 350,
      friction: 28,
    },
  });

  // Call this instead of pop in the render
  const removeItem = (item: ToasterToast): void => {
    // Set the item height so we can animate it out
    leaveItemHeight.current = refMap.get(item).offsetHeight;
    pop(item.id);
  };
```

packages/paste-core/components/toast/src/Toaster.tsx
```diff
@@ -0,0 +1,103 @@
...
+  const [returnTarget, setReturnTarget] = React.useState<ReturnTargetState>({trigger: null});
+  const theme = useTheme();
+
+  const transitions = useTransition(toasts, {
+    from: {height: 0, marginBottom: 0, opacity: 0, transform: 'translateX(100%) scale(1)'},
+    enter: item => async next => {
+      await next({
+        height: refMap.get(item).offsetHeight,
+        marginBottom: theme.space.space40,
```
TheSisb, line 25: So the marginBottom isn't currently animating correctly because `theme.space.space40` returns a string value with `rem` units, and the initial state is a number (`0`).  To fix this, we should either change this to `40` or the other marginBottoms from `0` to `"0rem"`.  Looks smoother with the change.

Sidebar: we should probably replace the 0 tokens (i.e.: space0) to be a string with the correct unit: `0` -> `"0rem"` to make this type of work tokenizable.

TheSisb (approved):
Some animation nits but this looks great to me.

packages/paste-website/src/pages/components/toast/index.mdx
SiTaggart: It's a very specific use-case where you are controlling state for everything in the UI from the application in a redux store, for example. It might be best in the API section though, I think you might be right about that. I don't think many people will do it.

packages/paste-core/components/toast/src/Toaster.tsx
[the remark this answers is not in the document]
SiTaggart: hmmm, yeah it actually seems kinda fine on the memory front.

PR #832, opened 2020-10-07
feat(icons): add TwiML Bins product icon

kateschwarz:
Add TwiML Bins product icon

**Contributing to Twilio**

- [x ] I acknowledge that all my contributions will be made under the project's license.


serifluous:
Hey @kateschwarz. Can you send me the SVG for this icon so I can add it to our Sketch library? I might be able to compress it a bit more, too.

kateschwarz:
@serifluous I removed the original SVG and replaced it with the compressed version you provided. Thanks!

serifluous (approved):

TheSisb (approved):

packages/paste-icons/src/ProductTwimlBinsIcon.tsx
```diff
@@ -0,0 +1,44 @@
+/**
```
TheSisb, line 1: This file name has incorrect capitalization. Can you rename it to `ProductTwiMLBinsIcon.tsx` please?

Renaming files in git on osx is awkward, you might need this command: `git mv ProductTwimlBinsIconProps.tsx ProductTwiMLBinsIconProps.tsx` for git to recognize the new filename

TheSisb (changes requested):
This rename should fix breaking builds.

packages/paste-icons/src/ProductTwimlBinsIcon.tsx
kateschwarz: Super weird... it was the correct name in my text editor and when I would `ls` all of the files in that folder, but I ran the command you suggested above, and looks like it changed the name 👍

TheSisb: Yeah it's an issue with macs. You have to do the `git mv` command if you're renaming a file that is already checked in by just changing capitalization. Thanks!

TheSisb (approved):

PR #833, opened 2020-10-08
fix(combobox): hide chevron if set to autocomplete

richbachman:
- [x] Hides chevron and removes extra right padding if Combobox has autocomplete

### DIFF
--- a/packages/paste-core/components/combobox/src/Combobox.tsx
+++ b/packages/paste-core/components/combobox/src/Combobox.tsx
@@ -20,10 +20,11 @@ import {RenderGroupItemsProps, RenderItemProps, RenderItemsProps, RenderListBoxP
 const groupBy = require('lodash.groupby');
 
 // Fixes chevron overlapping really long text
+// Extra right padding is removed when autocomplete is true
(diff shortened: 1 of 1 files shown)


packages/paste-core/components/combobox/src/Combobox.tsx
```diff
@@ -221,6 +221,7 @@ const Combobox = React.forwardRef<HTMLInputElement, ComboboxProps>(
           {labelText}
         </Label>
         <InputBox
+          autocomplete={autocomplete}
```
SiTaggart, line 224: Why does `InputBox` need to know about autocomplete?

richbachman: It made sense to me to pass it like the other props, but maybe I just need the autocomplete type to be available. Thinking about it more now, it probably doesn't need to be passed since its not doing anything to `InputBox`.

TheSisb (approved):
What Simon said, then lgtm!

richbachman:
@SiTaggart I cleaned this one up a bit. Ready for your re-review.

SiTaggart (approved):
Much better

PR #836, opened 2020-10-09
test: setup up cypress website tests

richbachman:
- [x] Set up cypress
- [x] Add initial tests for sidebar, token list, and theme switcher


cypress/integration/e2e/theme-switcher/index.spec.ts
```diff
@@ -0,0 +1,17 @@
...
+  it('check the console theme radio', () => {
+    cy.get('input[name="sitetheme"]')
+      .check('console')
+      .should('be.checked');
+  });
+
+  it('check the default theme radio', () => {
+    cy.get('input[name="sitetheme"]')
+      .check('default')
+      .should('be.checked');
```
SiTaggart, line 15: I don't necessarily see how these cam both be checked

.github/workflows/cypress.yml
```diff
@@ -0,0 +1,64 @@
...
+          key: ${{ runner.os }}-${{ hashFiles('**/yarn.lock') }}
+      - name: Install dependencies
+        run: yarn install
+      - name: Print URL
+        run: echo Testing URL ${{ github.event.deployment_status.target_url }}
+      - name: Run cypress
+        run: yarn ci:website
+        env:
+          APPLITOOLS_API_KEY: ${{ secrets.APPLITOOLS_API_KEY }}
+          CYPRESS_BASE_URL: ${{ github.event.deployment_status.target_url }}
```
SiTaggart, line 38: Are these still running on the deployment or the website running in the ci box?

richbachman: They're running on the Vercel deployment. Changing that to run just on the website running on the CI box might speed up the tests since we wouldn't have to wait for the deployment.

cypress/integration/e2e/theme-switcher/index.spec.ts
richbachman: They're two separate tests. First one loads the site, and basically checks to see if the console theme is checked. The second one checks the default theme and verifies its checked.

.github/workflows/cypress.yml
SiTaggart: Yeah, that might have been a miscommunication. When asking for the way to run tests locally, I was expecting to just use the same script for ci too

cypress/integration/e2e/theme-switcher/index.spec.ts
SiTaggart: oh I see now

.github/workflows/cypress.yml
richbachman: In my mind it made sense to test the production version of the site, which led to testing the preview deployment. That said I could alter the test:website script to be something like:

```
start-server-and-test 'yarn build:website' 'yarn serve:website' http://localhost:9000 'yarn run cypress run --spec './cypress/integration/**/*.spec.ts''
```

That will also simplify reporting back to github, as I could run the action on `push` instead of `deployment_status`.

SiTaggart: Are you only reporting back to the checks api for the applitools report?

richbachman: I had to report it to the checks api, because the run wasn't really being tied to the PR. See here: https://share.getcloudapp.com/4guJ2ZBY

Since Vercel handled the deployment status, it kicked off the Cypress action and somehow lost the tie to the PR. I'm testing my above idea about changing the script locally, and that seems to be working better. I don't think I'd need to report back to the checks api since the action runs on `pull_request`.

SiTaggart: I see, so yeah, that makes sense then to me not to use the deployment, just for not having to do a custom report back to the checks API.

cypress/integration/applitools/index.spec.ts
```diff
@@ -0,0 +1,31 @@
...
+    cy.wait(2000);
+    cy.eyesOpen({
+      appName: 'paste_docs',
+      testName: 'Tokens',
+      browser: {width: 1440, height: 1020, name: 'chrome'},
+    });
+    cy.eyesCheckWindow('Tokens');
+    cy.eyesClose();
+  });
+});
```
SiTaggart, line 31: For now, drop the Applitools screenshots. I'm having some feels about Applitools right now.

richbachman: Got it. Agreed. Doesn't seem to work well with our set up.

SiTaggart (approved):

PR #842, opened 2020-10-12
Experiment using babel for builds

TheSisb:
Currently we build our packages using typescript via rollup, and we run babel on the output from TS. This PR explores using babel to transpile our TS via rollup, and using tsc to compile just the typings.


TheSisb:
Closing. Probably will use https://esbuild.github.io/

Issue #810, opened 2020-09-29
@twilio-paste/animation-library not transpiled to ES5


negomi:
**Describe the bug**
I'm trying to use Paste components in an older project which uses Webpack 3.

When building that project, I get an ES2015+ syntax error (Webpack 3 uses UglifyJS, which does not minify ES2015+ syntax).

I narrowed the issue down to `@twilio-paste/animation-library` which ships with untranspiled arrow functions.

Is this intentional, or should we expect Paste packages to be transpiled to ES5?

**To Reproduce**
1. Install `@twilio-paste/animation-library` in a project
2. Confirm `=>` is present in the dist files (`node_modules/@twilio-paste/animation-library/dist/index.js` and `node_modules/@twilio-paste/animation-library/dist/index.es.js`)

**Expected behavior**
Arrow functions not present in the library code.

TheSisb:
Hi @negomi! Good find! That totally slipped past my radar.  It looks like our build config file isn't finding the relevant babel config file and those arrow functions aren't being transpiled in our CJS build.  I'll make a ticket to fix this in our upcoming sprint.

TheSisb:
@negomi Please try the new release

negomi:
Just installed `0.2.2` and I'm still seeing `=>`/`let`/`const` etc. in the dist files 🤔

negomi:
Hey @TheSisb, can we reopen this?

SiTaggart:
The fix should have been included in `0.2.3`, and still looks fine in `0.2.4` https://unpkg.com/@twilio-paste/animation-library@0.2.4/dist/index.js

No lets or consts that I can see in there. Is it errorring on anything else?

negomi:
Ah it is fixed in `0.2.3` onwards, I was checking for the fix one version too early. Thanks 😄

PR #845, opened 2020-10-13
fix(animation-library): make sure babel is run correctly

TheSisb:
Fixes: https://github.com/twilio-labs/paste/issues/810.

Seems like the babel.config.js in our repo root isn't being picked up by our rollup-babel plugin. This fix only applies to the animation package. I plan to roll this out (and other improvements) elsewhere soon.

### DIFF
--- a/packages/paste-libraries/animation/babel.config.json
+++ b/packages/paste-libraries/animation/babel.config.json
@@ -1,0 +1,1 @@
+{"extends": "../../../babel.config.js"}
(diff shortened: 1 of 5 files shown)


packages/paste-libraries/animation/babel.config.json
```diff
@@ -0,0 +1 @@
+{"extends": "../../../babel.config.js"}
```
SiTaggart, line 1: Simple fix 👍

SiTaggart (approved):

richbachman (approved):

SiTaggart (approved):
Bundle size increase is unfortunate but progress over perfection. 👍

PR #846, opened 2020-10-13
feat(website): add useful resources to design guidelines

serifluous:


katieporter:
@serifluous it's showing like so in the preview:
image

I think you probably just need to delete any extra empty lines above and below the list of resources.

serifluous:
> @serifluous it's showing up like so in the preview:
> image
> I think you probably just need to delete any extra empty lines above and below the list of resources.

fixed 👍

katieporter (approved):

SiTaggart (approved):

PR #847, opened 2020-10-13
fix(core-bundle): make react-uid a dependency for easier usage

TheSisb:
Having to install `react-uid` manually when using only `@twilio-paste/core` is not an ideal DX. I hardcode it as a dependency for now as a fast-follow to the unbarrel PR, and in the future we will wrap the implementation with our own `@twilio-paste/uid` library package which will automatically address this issue.

### DIFF
--- a/examples/create-react-app/package.json
+++ b/examples/create-react-app/package.json
@@ -11,8 +11,7 @@
     "prop-types": "^15.7.2",
     "react": "^16.13.1",
     "react-dom": "^16.13.1",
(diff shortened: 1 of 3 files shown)


SiTaggart (approved):
If you remove the dep from the create react app example here https://github.com/twilio-labs/paste/blob/main/examples/create-react-app/package.json#L15

The codesandbox we generate in the PR will also reflect what we now need for a functioning app too

richbachman (approved):

PR #850, opened 2020-10-14
fix(website): correctly load the monospace font, fix accessibility issues

SiTaggart:
Following watching @katieporter video on switching to the default theme, when talking about fonts I realised we aren't even loading or setting the monospace font from the system. In prep for the announcement tomorrow, this PR: 

- Loads Fira Mono. 
- Sets the font on code blocks and inline code to Fira Mono.
- uses the Night owl theme for all code blocks as the GitHub one is super inaccessible
- fixes a tiny accessibility problem with anchored headings where the link is not distinguishable from the text surrounding it.

image


serifluous:
Vvvvvvvvv nice

The change makes it a little hard to tell the difference between editable and non-editable code blocks, though.

TheSisb (approved):

PR #858, opened 2020-10-15
fix(checkbox): use context for CheckboxGroup

richbachman:
- [x] Allow passing props directly to Checkbox, instead of retrieving them via CheckboxGroup. Defaults to direct props, fallback to context from CheckboxGroup.
- [x] Use Box for HiddenCheckbox


packages/paste-core/components/checkbox/src/CheckboxGroup.tsx
```diff
@@ -1,35 +1,63 @@
...
+  value,
   ...props
 }) => {
+  const onChangeHandler = React.useMemo(() => {
+    return (event: React.ChangeEvent<HTMLInputElement>): void => {
+      if (onChange != null) {
+        onChange(event.target.checked);
+      }
+    };
+  }, [onChange]);
```
vnguyen94, line 32: `useCallback` might be more valid here?

```tsx
const onChangeHandler = React.useCallback((event: React.ChangeEvent<HTMLInputElement>): void => {
  onChange?.(event.target.checked);
}, [onChange]);
```

packages/paste-core/components/checkbox/src/Checkbox.tsx
```diff
@@ -12,27 +11,48 @@ import {
...
+
+type HiddenCheckboxProps = Pick<
+  CheckboxProps,
+  'checked' | 'disabled' | 'id' | 'indeterminate' | 'name' | 'onChange' | 'value'
+> & {
+  ref?: any | undefined;
+};
+const HiddenCheckbox = React.forwardRef<HTMLInputElement, HiddenCheckboxProps>((props, ref) => (
+  <Box
+    {...props}
```
SiTaggart, line 35: Can you use safelySpreadBoxProps here? Just in case

PR #871, opened 2020-10-20
feat(uid-library): add package

TheSisb:
**TL;DR:** I added a new package called `@twilio-paste/uid-library` that wraps `react-uid`. This library is similar to several of our others, like reakit-library and animation-library, in that it provides the same advantages.

---

**Things this PR does**
- [x] Add the new package and only exports the React Hooks API for consistency and better SSR support. I don't rename any exports because I think the ones provided are great already. I export the vanilla `uid` export for unimportant ID generation outside of React, like is used by the Toaster PR.
- [x] Added stories that _shouldn't_ be VRT'd, but help play with the library.
- [x] Updates all package.jsons and usage of `react-uid` to now use the new library. In a couple cases, I had to migrate some code that was using APIs we don't export in this new package.
- [x] Updated the icon generation template and re-built all our icons with the new API
- [x] Wrapped Theme.Provider with UIDFork so folks get it out of the box. https://github.com/twilio-labs/paste/pull/871/commits/69a4a20f02edb0f10645c25cde0842dc05d05deb

Documentation in a follow-up PR.

---

**Other notes:**
- ~~I'm not sure if it's a great idea to add UIDFork in Theme.Provider, but I pushed the code up to show what it looks like and discuss it with the team more easily.~~ I scrapped this idea.
- The stories don't need to be VRT'd, but they help play around while learning the API so I kept them in and will be used as examples when building out the docs. I attempt to add flags to prevent those stories from going through VRT.


packages/paste-core/components/combobox/src/Combobox.tsx
```diff
@@ -29,99 +29,104 @@ const StyledInputAsSelect = styled(InputElement)<ComboboxProps>(props =>
...
+  if (groupItemsBy == null) {
+    return null;
   }
-  return null;
+
+  const groupedItems = groupBy(items, (item: {}) => item[groupItemsBy]);
+  const groupedItemKeys = Object.keys(groupedItems);
+
+  return (
+    <>
```
SiTaggart, line 83: Nit: don't necessarily need the fragment here either

SiTaggart (approved):
Very nice. Nice lil combo box refactor too 👍

packages/paste-core/components/combobox/src/Combobox.tsx
TheSisb: Because this component is optionally returning an array, we need to wrap it in a fragment. 

TS warning too:
image

PR #875, opened 2020-10-22
feat(codemods): add barreled-to-unbarreled codemod

TheSisb:
## What changed (non-technical)

We have added a system to allow downstream teams to run automatic code updates, called "codemods". Learn more about them here https://github.com/facebook/codemod#background

## What changed (technical)

This PR creates a new `paste-codemods` folder in our monorepo that publishes the `@twilio-paste/codemods` package.

The package is structured to contain multiple "transforms", which are jscodeshift functions we can write to transform code.  The first and only transform included is the `barreled-to-unbarreled` transform, which converts imports from paste-core to the new unbarreled style (`@twilio-paste/core/button`).

The package also includes a `bin` field in the package json which points to a CLI nodejs script to make running this package easy and ergonomic for folks. This is heavily inspired (read: copied) from the [react-codemods](https://github.com/reactjs/react-codemod) repo.

## How to test it

I've published the code as the `@next` version, you can run it locally to watch it in action with:
```shell
npx @twilio-paste/codemods
```

We currently only have one codemod, generously contributed by @GiantRobots

### DIFF
--- a/packages/paste-codemods/package.json
+++ b/packages/paste-codemods/package.json
@@ -1,0 +1,31 @@
+{
+  "name": "@twilio-paste/codemods",
+  "version": "0.0.1",
+  "status": "beta",
+  "description": "A collection of codemods for maintaining projects built with Paste.",
+  "author": "Twilio Inc.",
+  "license": "MIT",
+  "publishConfig": {
+    "access": "public"
+  },
+  "bin": "./bin/paste-codemod.js",
+  "scripts": {
+    "build": "node ./tools/create-package-mappings.js",
+    "test": "jest"
+  },
+  "dependencies": {
+    "chalk": "^2.4.2",
+    "execa": "^3.2.0",
+    "globby": "^10.0.1",
+    "inquirer": "^7.0.0",
+    "is-git-clean": "^1.1.0",
+    "jscodeshift": "^0.11.0",
+    "meow": "^5.0.0"
+  },
+  "devDependencies": {
+    "@babel/core": "^7.11.4",
+    "@babel/plugin-proposal-object-rest-spread": "^7.8.3",
+    "@babel/preset-env": "^7.8.4",
+    "@twilio-paste/core": "*"
+  }
+}
--- a/packages/paste-codemods/tools/create-package-mappings.js
+++ b/packages/paste-codemods/tools/create-package-mappings.js
@@ -1,0 +1,14 @@
+const path = require('path');
+const {writeToFile} = require('../../../tools/utils/writeToFile');
+const {generatePackageExportsMap} = require('./generatePackageExportsMap');
+
+(async () => {
+  const mapping = await generatePackageExportsMap();
+
+  // Write to disk
+  writeToFile(path.join(__dirname, 'mappings.json'), mapping, {
+    successMessage: 'Wrote core packages export mapping to mappings.json file!',
+    errorMessage: 'Could not generate mappings!',
+    formatJson: true,
+  });
+})();
--- a/packages/paste-codemods/transforms/__tests__/barreled-to-unbarreled.spec.tsx
+++ b/packages/paste-codemods/transforms/__tests__/barreled-to-unbarreled.spec.tsx
@@ -1,0 +1,3 @@
+const {defineTest} = require('jscodeshift/dist/testUtils');
+
+defineTest(__dirname, 'barreled-to-unbarreled', null, 'barreled-to-unbarreled', {parser: 'ts'});
--- a/packages/paste-website/src/pages/libraries/codemods/index.mdx
+++ b/packages/paste-website/src/pages/libraries/codemods/index.mdx
@@ -1,0 +1,112 @@
+---
+title: Codemods
+description: A collection of jscodeshift codemods to help update projects built with Paste.
+slug: /libraries/codemods/
+---
+
+import {graphql} from 'gatsby';
+import {SidebarCategoryRoutes} from '../../../constants';
+import {Callout, CalloutTitle, CalloutText} from '../../../components/callout';
+
+export const pageQuery = graphql`
+  {
+    allPasteLibraries(filter: {name: {eq: "@twilio-paste/codemods"}}) {
+      edges {
+        node {
+          name
+          description
+          status
+          version
+        }
+      }
+    }
+    mdx(fields: {slug: {eq: "/libraries/codemods/"}}) {
+      fileAbsolutePath
+      frontmatter {
+        slug
+        title
+      }
+      headings {
+        depth
+        value
+      }
+    }
+  }
+`;
+
+<ComponentHeader
+  name="Codemods"
+  categoryRoute={SidebarCategoryRoutes.LIBRARIES}
+  githubUrl="https://github.com/twilio-labs/paste/tree/main/packages/paste-codemods"
+  data={props.data.allPasteLibraries.edges}
+/>
+
+---
+
+<contentwrapper>
+<PageAside data={props.data.mdx} />
+<content>
+
+## About
+
+[Codemods](https://github.com/facebook/codemod#background) are a collection of scripts that
+update your codebase from old patterns to new patterns automatically. In this way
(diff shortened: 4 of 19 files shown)


packages/paste-codemods/tools/mappings.json
```diff
@@ -0,0 +1,154 @@
+{
```
richbachman, line 1: Is this mappings file generated?

richbachman: Oh yep. Probably by the create-package-mappings.js file.

richbachman:
How does this work with new packages we release? Do we need to re-build this library in order to update the package mappings?

packages/paste-codemods/tools/create-package-mappings.js
```diff
@@ -0,0 +1,42 @@
...
+  const mapping = {};
+
+  // Get all Paste packages
+  const allPastePackages = await getRepoPackages();
+
+  // For each package in Paste:
+  allPastePackages.forEach(({name, location}) => {
+    // We don't want to use the 'form' or 'typography' packages because they are deprecated
+    // Skip them for now
+    if (name === '@twilio-paste/typography' || name === '@twilio-paste/form') return;
```
GiantRobots, line 16: This would probably be better if you had an array of deprecated packages and checked if they were in that array instead of an if statement with or's for maintenance but such an easy change I probably wouldn't worry about it

packages/paste-codemods/tools/mappings.json
TheSisb: Yup!

packages/paste-codemods/tools/create-package-mappings.js
TheSisb: Totes. Done

TheSisb:
> How does this work with new packages we release? Do we need to re-build this library in order to update the package mappings?

The release step runs the build command, which creates a fresh mappings.json. That then gets bundled and published to NPM. We don't technically need to push mappings.json to github, but it's a small enough file and handy to lookup sometimes, so I commit it anyways. Adding new packages to core will end up updating mappings if you run the full project build. If that gets annoying, or if you think that might be annoying, I can add that file to the .gitignore too.

packages/paste-website/src/pages/libraries/codemods/index.mdx
```diff
@@ -0,0 +1,108 @@
+---
+title: Codemods
+description: A collection of jscodeshift codemods to help update projects built with Paste.
+slug: /libraries/codemods/
+---
+
+import {graphql} from 'gatsby';
+import {Box} from '@twilio-paste/box';
```
richbachman, line 8: I think you can drop Box here.

richbachman:
Last thing I can see here. Do we need tests for `generatePackageExportsMap` and `barreledToUnbarreled`?

TheSisb:
@richbachman Added tests now

richbachman (approved):
Just need to fix the lint errors, then I think this looks good.

packages/paste-codemods/package.json
```diff
@@ -0,0 +1,30 @@
...
+    "globby": "^10.0.1",
+    "inquirer": "^7.0.0",
+    "is-git-clean": "^1.1.0",
+    "jscodeshift": "^0.11.0",
+    "meow": "^5.0.0"
+  },
+  "devDependencies": {
+    "@babel/core": "^7.6.4",
+    "@babel/plugin-proposal-object-rest-spread": "^7.6.2",
+    "@babel/preset-env": "^7.6.3"
```
SiTaggart, line 28: Is there any reason these are different versions to the ones we hoist to root of the repo? In workspaces we can rely on the hoisted versions so potentially no need to declare them here if they don't need to be different to the ones we already have.

packages/paste-core/core-bundle/src/codemods.tsx
```diff
@@ -0,0 +1 @@
+export * from '@twilio-paste/codemods';
```
SiTaggart, line 1: Should codemods appear here as an export of Core?

packages/paste-codemods/tools/mappings.json
SiTaggart, line 1: Is there a requirement for this file to be checked in? Can it just behave like a build asset, similar to a dist js file? That way we don't need to keep updating it in PRs that introduce new packages?

packages/paste-codemods/transforms/__tests__/barreled-to-unbarreled.spec.tsx
```diff
@@ -0,0 +1,4 @@
+const {defineTest} = require('jscodeshift/dist/testUtils');
+const transform = require('../barreled-to-unbarreled');
```
SiTaggart, line 2: Do you need this?

packages/paste-website/src/pages/libraries/codemods/index.mdx
```diff
@@ -0,0 +1,107 @@
...
+
+---
+
+<contentwrapper>
+<PageAside data={props.data.mdx} />
+<content>
+
+## About
+
+This library provides codemods to alleviate the burden of teams making large sweeping changes related to Paste.
```
SiTaggart, line 52: What are codemods? Is there a page or something we can link someone to for an explainer or from where they came from?

packages/paste-website/src/pages/libraries/codemods/index.mdx
```diff
@@ -0,0 +1,107 @@
...
+When dealing with changes that affect many files in a project, we typically we have to
+make a trade-off between the benefit of the change and the risk of introducing errors and tedium.
```
SiTaggart, line 54: > we typically we

Small typo.

I also feel like this sentence is missing the punchline or leaves me hanging. What ties _our_ changes to many files, the trade-offs, introducing errors and tedium to codemods?

packages/paste-website/src/pages/libraries/codemods/index.mdx
```diff
@@ -0,0 +1,107 @@
...
+
+These codemods make the changes _for_ us. They remove the friction involved with planning and executing on the work.
```
SiTaggart, line 56: I'd angle this as "for you", the reader. You might need to explain a little more about what you mean with regards to planning and executing.

packages/paste-website/src/pages/libraries/codemods/index.mdx
GiantRobots: https://github.com/facebook/jscodeshift
https://github.com/reactjs/react-codemod

```
This library is a collection of scripts that update your codebase from old patterns to new patterns automagically. In this way we can effectively move API's and patterns to newer better versions without incurring tech debt across the entire Twilio organization.

Notable users of this approach are React, Ember and Vue. All of those frameworks provide assistance to downstream users when they introduce breaking changes to alleviate pain points and provide a path forward for large codebases.
```

That's my stab at wordage but I am not a great wordsmith

packages/paste-codemods/package.json
TheSisb: Copy pasta, fixing. Good catch!

packages/paste-core/core-bundle/src/codemods.tsx
TheSisb: I can't think of a reason to include them in core, I'll add it to the ignore list. Good point.

SiTaggart (approved):

packages/paste-codemods/package.json
```diff
@@ -0,0 +1,31 @@
...
+    "inquirer": "^7.0.0",
+    "is-git-clean": "^1.1.0",
+    "jscodeshift": "^0.11.0",
+    "meow": "^5.0.0"
+  },
+  "devDependencies": {
+    "@babel/core": "^7.11.4",
+    "@babel/plugin-proposal-object-rest-spread": "^7.8.3",
+    "@babel/preset-env": "^7.8.4",
+    "@twilio-paste/core": "*"
```
TheSisb, line 29: Added so that lerna builds this package last and so that the mappings get generated correctly

PR #886, opened 2020-10-26
fix(list): forward refs to component exports

TheSisb:
We weren't forwarding refs to these list components, so I wrapped them in forwardRef calls. See https://github.com/twilio-labs/paste/discussions/841#discussioncomment-98397


richbachman (approved):
lgtm!

PR #891, opened 2020-10-27
feat(grid): add ability to create equal height columns

SiTaggart:
Pass `equalColumnHeight` to grid to use Flexbox to stretch align content it's columns, and the columns content.

Has limitations that only a single child will work in a column, and `display: flex` collapses the child's width.

### DIFF
--- a/packages/paste-website/src/pages/layout/grid/index.mdx
+++ b/packages/paste-website/src/pages/layout/grid/index.mdx
@@ -164,6 +164,51 @@ Use `offset` to push a column to the right by a certain number of columns. For e
 </Grid>`}
 </LivePreview>
 
+### Equal column heights
+
+Set `equalColumnHeights` on a grid container and all columns in a single row will share the height of the tallest column.
+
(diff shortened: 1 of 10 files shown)


packages/paste-website/src/pages/layout/grid/index.mdx
```diff
@@ -164,6 +164,49 @@ Use `offset` to push a column to the right by a certain number of columns. For e
 </Grid>`}
 </LivePreview>
 
+### Equal column heights
+
+Set `equalColumnHeights` on a grid container to set the height all columns in a single row to be equal to the height of the tallest column.
```
TheSisb, line 169: Do we need to add a note on setting size on child elements?

TheSisb (approved):

richbachman (approved):

packages/paste-website/src/pages/layout/grid/index.mdx
SiTaggart: What about now?

PR #892, opened 2020-10-27
feat(breadcrumb): add alpha component package

richbachman:
- [x] Add alpha Breadcrumb package
- [x] Add Breadcrumb package to Core

### DIFF
--- a/packages/paste-core/components/breadcrumb/__tests__/breadcrumb.spec.tsx
+++ b/packages/paste-core/components/breadcrumb/__tests__/breadcrumb.spec.tsx
@@ -1,0 +1,68 @@
+import * as React from 'react';
+import {axe} from 'jest-axe';
+import {matchers} from 'jest-emotion';
+import {render, screen} from '@testing-library/react';
+import {Breadcrumb, BreadcrumbItem} from '../src';
+
+expect.extend(matchers);
+
+describe('Breadcrumb', () => {
+  const BreadcrumbExample: React.FC<{}> = () => {
+    return (
+      <Breadcrumb>
+        <BreadcrumbItem href="#">foo</BreadcrumbItem>
+        <BreadcrumbItem>bar</BreadcrumbItem>
+      </Breadcrumb>
+    );
+  };
+
+  it('it should render a nav with correct aria-label', () => {
+    const {getByRole} = render(<BreadcrumbExample />);
+    expect(getByRole('navigation').getAttribute('aria-label')).toBe('breadcrumb');
+  });
+
+  it('it should render a list with display: inline-flex', () => {
+    render(<BreadcrumbExample />);
+    const renderedList = screen.getByRole('list');
+    expect(renderedList).toHaveStyleRule('display', 'inline-flex');
+  });
+
+  it('it should render listitems', () => {
+    render(<BreadcrumbExample />);
+    const renderedListItem = screen.findByRole('listitem');
+    expect(renderedListItem).not.toBeNull();
+  });
+
+  it('it should render a breadcrumb separator', () => {
+    render(<BreadcrumbExample />);
+    const renderedSeparator = screen.getByText('/');
+    expect(renderedSeparator).toBeDefined();
+  });
+
+  it('it should render an anchor', () => {
+    render(
+      <Breadcrumb>
+        <BreadcrumbItem href="#">foo</BreadcrumbItem>
+      </Breadcrumb>
+    );
+    const renderedAnchor = screen.findByRole('link');
+    expect(renderedAnchor).not.toBeNull();
+  });
+
+  it('it should render text with correct aria-current', () => {
+    const {getByText} = render(
+      <Breadcrumb>
+        <BreadcrumbItem>foo</BreadcrumbItem>
+      </Breadcrumb>
+    );
+    expect(getByText('foo').getAttribute('aria-current')).toBe('page');
+  });
+
+  describe('Accessibility', () => {
+    it('Should have no accessibility violations', async () => {
+      const {container} = render(<BreadcrumbExample />);
+      const results = await axe(container);
+      expect(results).toHaveNoViolations();
+    });
+  });
+});
--- a/packages/paste-core/components/breadcrumb/src/index.tsx
+++ b/packages/paste-core/components/breadcrumb/src/index.tsx
@@ -1,0 +1,101 @@
+import * as React from 'react';
+import * as PropTypes from 'prop-types';
+import {Box, safelySpreadBoxProps} from '@twilio-paste/box';
+import {Anchor} from '@twilio-paste/anchor';
+import {Text} from '@twilio-paste/text';
+import {useUIDSeed} from '@twilio-paste/uid-library';
+
+const BreadcrumbSeparator: React.FC = () => (
+  <Text
+    as="span"
+    color="colorTextWeak"
+    fontSize="fontSize20"
+    lineHeight="lineHeight20"
+    paddingLeft="space20"
+    paddingRight="space20"
+    role="presentation"
+  >
+    /
+  </Text>
+);
+
+export interface BreadcrumbItemProps extends React.HTMLAttributes<HTMLLIElement> {
+  children: NonNullable<React.ReactNode>;
+  href?: string;
+  last?: boolean;
+}
+
+const BreadcrumbItem = React.forwardRef<HTMLAnchorElement, BreadcrumbItemProps>(
+  ({children, href, last, ...props}, ref) => {
+    return (
+      <Box
+        {...safelySpreadBoxProps(props)}
+        alignItems="center"
+        as="li"
+        color="colorText"
+        display="inline-flex"
+        fontSize="fontSize20"
+        lineHeight="lineHeight20"
+      >
+        {href ? (
+          <Anchor href={href} ref={ref}>
+            {children}
+          </Anchor>
+        ) : (
+          <Text aria-current="page" as="span" fontSize="fontSize20" lineHeight="lineHeight20" ref={ref}>
+            {children}
+          </Text>
+        )}
+        {!last && <BreadcrumbSeparator />}
+      </Box>
+    );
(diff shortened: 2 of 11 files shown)


packages/paste-core/components/breadcrumb/__tests__/breadcrumb.spec.tsx
```diff
@@ -0,0 +1,65 @@
+import * as React from 'react';
+import {axe} from 'jest-axe';
+import {matchers} from 'jest-emotion';
+import {render, screen} from '@testing-library/react';
+import {Breadcrumb, BreadcrumbItem} from '../src';
+
+expect.extend(matchers);
```
TheSisb, line 7: What's this for?

packages/paste-core/components/breadcrumb/src/index.tsx
```diff
@@ -0,0 +1,79 @@
+import * as React from 'react';
+import * as PropTypes from 'prop-types';
+import {Box, safelySpreadBoxProps} from '@twilio-paste/box';
+import {Text, safelySpreadTextProps} from '@twilio-paste/text';
+import {useUIDSeed} from '@twilio-paste/uid-library';
+
+const BreadcrumbSeparator: React.FC = () => (
+  <Text
+    as="span"
```
TheSisb, line 9: Let's give this `role="presentation"`

packages/paste-core/components/breadcrumb/stories/index.stories.tsx
```diff
@@ -0,0 +1,61 @@
...
+import {Truncate} from '@twilio-paste/truncate';
+import {Breadcrumb, BreadcrumbItem} from '../src';
+
+storiesOf('Components|Breadcrumb', module)
+  .addDecorator(withKnobs)
+  .add('Anchors', () => {
+    return (
+      <Breadcrumb>
+        <BreadcrumbItem>
+          <Anchor href="#">Phone Numbers</Anchor>
```
TheSisb, line 16: Question: what if BreadcrumbItem was a wrapper around Anchor if a href="" was passed, otherwise it has the standard behavior.  So this could be:
```
<BreadcrumbItem href="#">Phone Numbers</BreadcrumbItem>
<BreadcrumbItem>My Phone Number</BreadcrumbItem>
```
This would also be a more future-proof way of controlling the styles of links in breadcrumbs and, if href is absent, we could automatically apply the ` aria-current="page"` to make sure it's present. Essentially this would be to try and have less user error.

packages/paste-core/components/breadcrumb/stories/index.stories.tsx
```diff
@@ -0,0 +1,61 @@
...
+        </BreadcrumbItem>
+        <BreadcrumbItem>
+          <Anchor href="#">Active Numbers</Anchor>
+        </BreadcrumbItem>
+        <BreadcrumbItem>
+          <Text aria-current="page" as="span" fontSize="fontSize20" lineHeight="lineHeight20">
```
TheSisb, line 34: Do we need this `<Text>` element? Any way around it?

packages/paste-core/components/breadcrumb/stories/index.stories.tsx
```diff
@@ -0,0 +1,61 @@
...
+  })
+  .add('Anchors with truncation ', () => {
+    return (
+      <Breadcrumb>
+        <BreadcrumbItem>
+          <Anchor href="#">Phone Numbers</Anchor>
+        </BreadcrumbItem>
+        <BreadcrumbItem>
+          <Box as="span" display="inline-flex" maxWidth="size10">
+            <Truncate title="Some very long text to truncate">
```
TheSisb, line 49: An inline-flex Box with a Truncate child seems rather complex.  Can we hide this complexity behind a prop? Do we have any opinions on maxWidth?

packages/paste-core/components/breadcrumb/__tests__/breadcrumb.spec.tsx
richbachman: That extends `expect` with `toHaveStyleRule`, so we can test for a specific style. I ended up using it a lot with Table, since thats just a bunch of styled Boxes. I'm using it here to test the `inline-flex` style.

packages/paste-core/components/breadcrumb/stories/index.stories.tsx
richbachman: I had this same thought, but didn't wasn't sure if it was the right path to go down. Since you've mentioned it, and made me think about it more, this feels like a smart move. Less composing by the user, and would solve for the comment below about `Text`. I'll work this up.

packages/paste-core/components/breadcrumb/stories/index.stories.tsx
richbachman: Using the following API solves for that:
```
<Breadcrumb>
    <BreadcrumbItem href="#">Phone Numbers</BreadcrumbItem>
    <BreadcrumbItem href="#">Active Numbers</BreadcrumbItem>
    <BreadcrumbItem>Infrrd Germany hotline B</BreadcrumbItem>
</Breadcrumb>
```

packages/paste-core/components/breadcrumb/stories/index.stories.tsx
richbachman: Truncation is a little tricky here. I don't think we should have strong opinions about when text is truncated until we get feedback from users, especially in an alpha version.

The width may also need some research/testing on the design side. I'm sure there are some opinions there.

Maybe we just show guidance here for the alpha version. The new API does simplify it a little:
```
<BreadcrumbItem href="#">
    <Box as="span" display="inline-flex" maxWidth="size10">
        <Truncate title="Some very long text to truncate">Some very long text to truncate</Truncate>
    </Box>
</BreadcrumbItem>
```

TheSisb: Can we make `maxWidth` a prop on `BreadcrumbItem`, and if passed it does the truncation automagically?

packages/paste-core/components/breadcrumb/__tests__/breadcrumb.spec.tsx
SiTaggart: Might be worth looking if you can add this extend in the jest config so we have it everywhere

packages/paste-core/components/breadcrumb/__tests__/breadcrumb.spec.tsx
```diff
@@ -0,0 +1,84 @@
...
+  });
+
+  it('it should render a list with display: inline-flex', () => {
+    render(
+      <Breadcrumb>
+        <BreadcrumbItem href="#">foo</BreadcrumbItem>
+        <BreadcrumbItem>bar</BreadcrumbItem>
+      </Breadcrumb>
+    );
+    const renderedBreadcrumb = screen.getByRole('list');
```
SiTaggart, line 27: nit: all of these constants change their meaning. For example, this isn't really `renderedBreadcrumb`, it's the rendered list. It might be nice to have explicit names so the test code reads better.

packages/paste-core/components/breadcrumb/src/index.tsx
```diff
@@ -0,0 +1,100 @@
...
+        {...safelySpreadBoxProps(props)}
+        alignItems="center"
+        as="li"
+        color="colorText"
+        display="inline-flex"
+        fontSize="fontSize20"
+        lineHeight="lineHeight20"
+      >
+        {href ? (
+          <Anchor href={href} ref={ref}>
```
SiTaggart, line 41: Is it bad to forward a ref, but only render that component some of the time? If we're rendering Text, but still forwarding a ref, if I grab the ref from `BreadcrumbItem`...

packages/paste-core/components/breadcrumb/src/index.tsx
```diff
@@ -0,0 +1,100 @@
...
+      </Box>
+    );
+  }
+);
+
+BreadcrumbItem.displayName = 'BreadcrumbItem';
+
+if (process.env.NODE_ENV === 'development') {
+  BreadcrumbItem.propTypes = {
+    children: PropTypes.node.isRequired,
```
SiTaggart, line 59: `href`

packages/paste-core/components/breadcrumb/__tests__/breadcrumb.spec.tsx
richbachman: Yep. Totally makes sense.

packages/paste-core/components/breadcrumb/src/index.tsx
richbachman: Basically, if I don't pass the `ref` to `Text` as well, all of the properties that comes with `ref` from `BreadcrumbItem` won't get passed.

packages/paste-core/components/breadcrumb/__tests__/breadcrumb.spec.tsx
```diff
@@ -0,0 +1,84 @@
...
+import {axe} from 'jest-axe';
+import {matchers} from 'jest-emotion';
+import {render, screen} from '@testing-library/react';
+import {Breadcrumb, BreadcrumbItem} from '../src';
+
+expect.extend(matchers);
+
+describe('Breadcrumb', () => {
+  it('it should render a nav with correct aria-label', () => {
+    const {getByRole} = render(
```
TheSisb, line 11: nit: are you able to move this render into the describe() { block so that we can use it multiple times? Haven't tried it so not sure, but could speedup the test.

TheSisb (approved):

packages/paste-core/components/breadcrumb/__tests__/breadcrumb.spec.tsx
richbachman: Good suggestion. That sped the tests up from about 7s to 3.5s.

PR #897, opened 2020-10-28
chore(absolute): remove component package

richbachman:
- [x] Removed Absolute from components and core
- [x] Removed Absolute from docs site

BREAKING CHANGE: The Absolute component has been deprecated. For a replacement use the Box component with absolute positioning.


TheSisb (approved):

yarn.lock
```diff
@@ -4525,6 +4525,11 @@
     prettier "^1.19.1"
     tslib "^1.10.0"
 
+"@twilio-paste/absolute@^2.0.79":
```
SiTaggart, line 4528: This is odd, if you're removing it, and all it's usage, why is yarn adding it to the lockfile? Maybe needs another yarn install run?

SiTaggart (changes requested):
Everything is fine, I would just like to see the core-bundle commit be a breaking change. I know no one is using it, but we should honor our semver versioning

yarn.lock
richbachman: Yep. Just needed to run that again. Good catch.

SiTaggart (approved):

PR #909, opened 2020-11-04
test: adjust sidebar cypress tests

richbachman:
The sidebar test fails occasionally. Adding a timeout and changing the tests to see if the corresponding disclosure content is visible seems to work better.

I also added a test for the new patterns disclosure.

### DIFF
--- a/cypress/integration/e2e/sidebar-navigation/index.spec.ts
+++ b/cypress/integration/e2e/sidebar-navigation/index.spec.ts
@@ -1,6 +1,6 @@
 describe('Sidebar navigation', () => {
   beforeEach(() => {
-    cy.visit('http://localhost:9000/');
+    cy.visit('/');
   });
 
   it('opens the getting started disclosure', () => {
--- a/cypress/integration/e2e/sidebar-navigation/index.spec.ts
+++ b/cypress/integration/e2e/sidebar-navigation/index.spec.ts
@@ -4,50 +4,42 @@ describe('Sidebar navigation', () => {
   });
 
   it('opens the getting started disclosure', () => {
-    cy.get('[data-cy=getting-started-button]')
-      .click()
-      .should('have.attr', 'aria-expanded', 'true');
+    cy.get('[data-cy=getting-started-button]').click();
+    cy.get('[data-cy=getting-started-list]').should('not.have.attr', 'hidden');
   });
 
   it('opens the tokens disclosure', () => {
-    cy.get('[data-cy=design-tokens-button]')
-      .click()
-      .should('have.attr', 'aria-expanded', 'true');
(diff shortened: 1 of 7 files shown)


cypress/integration/e2e/theme-switcher/index.spec.ts
```diff
@@ -1,6 +1,6 @@
 describe('Theme switcher', () => {
   beforeEach(() => {
-    cy.visit('/');
+    cy.visit('/', {timeout: 10000});
```
SiTaggart, line 3: I don't think we should be waiting 10 seconds before each test. It's also only the sidebar test that is flakey, right? The others were fine without the timeout as far as I know. Do we need these anywhere else?

cypress/integration/e2e/sidebar-navigation/index.spec.ts
```diff
@@ -1,37 +1,45 @@
...
-    cy.get(layoutButton)
-      .click()
-      .should('have.attr', 'aria-expanded', 'true');
-    cy.get(iconsButton)
-      .click()
-      .should('have.attr', 'aria-expanded', 'true');
-    cy.get(contentButton)
-      .click()
-      .should('have.attr', 'aria-expanded', 'true');
+    cy.visit('/', {timeout: 10000});
```
SiTaggart, line 3: Do we need to revisit the homepage after expanding a section? Can we just do this in a single page visit? 

Waiting 10 seconds between each test seems unnecessarily excessive.

cypress/integration/e2e/sidebar-navigation/index.spec.ts
```diff
@@ -1,37 +1,45 @@
...
+  });
+
+  it('opens the getting started disclosure', () => {
+    cy.get('[data-cy=getting-started-button]').click();
+    cy.get('[data-cy=getting-started-list]').should('be.visible');
+  });
+
+  it('opens the tokens disclosure', () => {
+    cy.get('[data-cy=design-tokens-button]').click();
```
SiTaggart, line 12: Can we keep the accessibility tests too `.should('have.attr', 'aria-expanded', 'true');` for these please?

cypress/integration/e2e/theme-switcher/index.spec.ts
richbachman: I may have misunderstood the `timeout` argument here. I thought it was the time to wait for `cy.visit('/')` to resolve. Not an actual wait 10 seconds. 

I thought the localhost running in the action might be running slow, and not actually resolving in time for the test. Maybe thats not the case. I'll try and bulk these and remove the timeout.

TheSisb:
Test failed again D:

SiTaggart (approved):

PR #917, opened 2020-11-06
fix(combobox): bump listbox zIndex token

richbachman:
- [x] Bumped listbox zIndex value
- [x] Added story for VRT

### DIFF
--- a/packages/paste-core/components/combobox/stories/index.stories.tsx
+++ b/packages/paste-core/components/combobox/stories/index.stories.tsx
@@ -2,10 +2,13 @@ import * as React from 'react';
 import _ from 'lodash';
 import {storiesOf} from '@storybook/react';
 import {withKnobs} from '@storybook/addon-knobs';
+import {useUID} from '@twilio-paste/uid-library';
 import {Anchor} from '@twilio-paste/anchor';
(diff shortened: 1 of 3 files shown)


packages/paste-core/components/combobox/stories/index.stories.tsx
```diff
@@ -484,4 +487,46 @@ storiesOf('Components|Combobox', module)
         itemToString={(item: GroupedItem) => (item ? item.label : null)}
       />
     );
+  })
+  .add('Combobox - Listbox zIndex', () => {
+    const [selectValue, setSelectValue] = React.useState('');
+    const selectID = useUIDSeed();
```
TheSisb, line 493: nit: we should call this `seed` or `idSeed` for clarity, or you can use `useUID` which would return the `selectID` directly

richbachman: Good catch. No reason to use `useUIDSeed` here since I'm just generating one ID.

TheSisb:
I marked this "Ready For Review" so I could take a look at the story on Chromatic and wrap up the review.

TheSisb (approved):

PR #921, opened 2020-11-10
chore: extend the tooltip story timeouts

SiTaggart:
These two tooltip stories diff everytime in chromatic. The default doesn't even include a screenshot of the tooltip, and extending the timeout before the capture to hopefully iron out the positioning differences.

### DIFF
--- a/packages/paste-core/components/tooltip/stories/index.stories.tsx
+++ b/packages/paste-core/components/tooltip/stories/index.stories.tsx
@@ -39,12 +39,14 @@ storiesOf('Components|Tooltip', module)
     'Default',
     () => {
       return (
-        <Tooltip visible text={text('text', 'Welcome to Paste!')}>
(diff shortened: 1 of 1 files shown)


packages/paste-core/components/tooltip/stories/index.stories.tsx
```diff
@@ -39,12 +39,14 @@ storiesOf('Components|Tooltip', module)
     'Default',
     () => {
       return (
-        <Tooltip visible text={text('text', 'Welcome to Paste!')}>
-          <Button variant="primary">Open tooltip</Button>
-        </Tooltip>
+        <Box as="div" minHeight="400px">
```
TheSisb, line 42: 📟 `400px`

TheSisb (approved):
Worth a shot

richbachman (approved):

Issue #827, opened 2020-10-06
paste-core: latest version [2.16.61] is throwing error for Button and Anchor component


Kajal-Joshi:
I tried to update paste to it's latest version, but after updating it I started getting following error for components like Button and Anchor etc. Please let me know if anyone has faced this issue!

**Errors :
For Button Component**
error TS2739: Type '{ children: Element; "data-test": string; variant: "primary" | "secondary"; size: "icon"; onClick: () => void; key: string; }' is missing the following properties from type 'Pick<ButtonProps, "contextMenu" | "dir" | "form" | "slot" | "style" | "title" | "children" | "placeholder" | "name" | "id" | "disabled" | "loading" | "as" | "href" | ... 257 more ... | "css">': translate, css

Sample Implementation for which got above error: 

    <Button variant="secondary" onClick={() => generateAndDownloadFile()}>
                    Download HTML file  
    </Button>

**For Anchor Component**
error TS2739: Type '{ children: string; href: string; }' is missing the following properties from type 'Pick<AnchorProps, "contextMenu" | "dir" | "slot" | "style" | "title" | "children" | "width" | "minWidth" | "maxWidth" | "height" | "minHeight" | "maxHeight" | "size" | ... 278 more ... | "referrerPolicy">': translate, css
     
Sample Implementation for which got above error: 

       <Anchor href={href}>
          {anchorText}
        </Anchor>

![image](https://user-images.githubusercontent.com/72342162/95178981-a9086a80-07dd-11eb-9f49-f693af61804d.png)

SiTaggart:
hi @Kajal-Joshi, thanks for the report of the issue.

Can I ask some details about your environment? Could you list out what version of typescript and node you're running, also anything about the project, is it a Create React App project for example, something like that.

We've trying to replicate this locally and are struggling to get the same results.

Thanks

Kajal-Joshi:
@SiTaggart sure!

From initial commit of project doesn't looks like it has been created using Create React App! Custom settings I guess.
**Details that could be helpful :**
**Versions:**
"typescript": "^3.9.5",
"@typescript-eslint/eslint-plugin": "^3.6.1",
"@typescript-eslint/parser": "^3.6.1"


**Eslint configuration :** 
module.exports = {
    extends: [
        'airbnb-typescript',
        'airbnb/hooks',
        'plugin:@typescript-eslint/recommended',
        'plugin:jest/recommended',
        'prettier',
        'prettier/react',
[...]

PR #922, opened 2020-11-10
fix(types): set our type dependencies explicitly

SiTaggart:
Hoping to resolve issue https://github.com/twilio-labs/paste/issues/827 and our problems with CodeSandbox.

image

What seems to be happening is that we build our types against specific versions of `@type` packages. In certain environments though, the version of those same `@type` packages don't match, or are older. This then causes type issues. 

The clearest example is Anchor and Button, which both extend from the HTML anchor and button typings from `@types/react`. In `@types/react@16.9.x` they introduced new attributes that could be added to an anchor or button. Codesandbox is installing `@types/react@16.8.x` which doesn't include them. Hence the `translate`, `onAuxClick` and `onAuxClickCapture` is missing errors.

This should resolve that issue, as we are explicitly declaring what versions of typings you need to install. 

As it's technically a dev dependency, I didn't see much point in declaring it a peer dep.


TheSisb (approved):
Looks like it's working in the PR's codesandbox. Rubber stamp lgtm

richbachman (approved):

PR #947, opened 2020-11-17
chore(website): rework site-wrapper for responsiveness

TheSisb:
This work is getting rather large, so I'm going to break out any additional work on the MobileHeader into follow-up PRs.  This PR does the following:

- [x] Add React hook `useWindowSize` to get the browser dimensions in JS, so we can conditionally render components on mobile/desktop. This helps improve performance by not loading a whole bunch of stuff that is hidden in CSS. This hook includes telling us the currently active breakpoint index.
- [x] Adds a MobileHeader frame, with placeholder content that is still WIP for followup PR.  Needs Navigation improvements and Search built.
- [x] Adds 'transparent' as a valid option for `backgroundColor` style props. Needed this to make Navigation work both in Desktop and in Mobile.
- [x] Removes the core-bundle package from the website, since it will result in us possibly loading 2 versions of packages. We need the direct imports because we load in CHANGELOG.md and package.json files.
- [x] Fixes regression with doc pages by re-adding the `paddingX` on `siteMainInner`. Adding our negative space tokens to make the new homepage work without this hardcoded padding.
- [x] Removed the `minWidth` on our docs, so the homepage can be properly responsive.
- [x] Moved the graphQL query for navigation the an outer component (site-wrapper/index) and created a context provider so we can use the same result on Mobile and Desktop nav.
- [x] Created a fun lil' HamburgerToggle with animation.
- [x] Refactored (read: shuffled) some files around in the `site-wrapper` folder to make it more manageable.
- [x] Fixed scrolling on doc pages via the ToC; it scrolls with the heading in view now and does smooth scrolling. Added this while waiting for builds.


packages/paste-website/src/components/site-wrapper/sidebar/index.tsx
```diff
@@ -17,7 +17,7 @@ const Sidebar: React.FC = () => {
       position="sticky"
       top={`${SITE_MASTHEAD_HEIGHT}px`}
       width="sizeSidebar"
-      display="flex"
+      display={['none', 'none', 'flex']}
```
TheSisb, line 20: TODO rm and not render

packages/paste-website/src/components/ContactUsMenu.tsx
```diff
@@ -5,8 +5,28 @@ import {Menu, MenuButton, MenuItem, useMenuState} from '@twilio-paste/menu';
 import {MoreIcon} from '@twilio-paste/icons/esm/MoreIcon';
 import {ProductSupportIcon} from '@twilio-paste/icons/esm/ProductSupportIcon';
 
-export const ContactUsMenu: React.FC<{}> = () => {
-  const menu = useMenuState({placement: 'right-end'});
+interface ContactUsMenuProps {
+  placement?:
```
TheSisb, line 9: TODO: can we Pick<> the type from the Menu exports?

packages/paste-website/src/components/site-wrapper/SiteBody.tsx
```diff
@@ -14,22 +15,27 @@ const StyledSiteBody = styled.div`
...
       <SiteHeader />
       <StyledSiteBody>
-        <Sidebar />
-        <Box minWidth="size0">
-          <SiteMain role="main">
-            <ScrollAnchorIntoView />
-            <SiteMainInner>{children}</SiteMainInner>
-            <SiteFooter />
-          </SiteMain>
+        {isDesktop ? <Sidebar /> : null}
```
SiTaggart, line 34: I think it'd be better to just use the responsive display props here

packages/paste-website/src/components/site-wrapper/sidebar/SidebarNavigation.tsx
```diff
@@ -138,7 +61,13 @@ const SidebarNavigation: React.FC<SidebarNavigationProps> = () => {
   });
 
   return (
-    <Box as="nav" marginTop="space70" overflow="auto" role="navigation" aria-label="Main">
+    <Box
+      as="nav"
+      marginTop={breakpointIndex !== undefined && breakpointIndex <= 1 ? 'space70' : 'space0'}
```
SiTaggart, line 66: This would also be better as a responsive prop

packages/paste-website/src/components/site-wrapper/site-header/SiteHeaderMobile.tsx
```diff
@@ -0,0 +1,38 @@
...
+import {HamburgerToggle} from './HamburgerToggle';
+import {SiteHeaderLogo} from './SiteHeaderLogo';
+import {MobileNavigation} from './SiteHeaderMobileNavigation';
+
+interface SiteHeaderMobileProps {
+  searchValue: string;
+  onSearchChange: (newSearchValue: string) => void;
+}
+
+const SiteHeaderMobile: React.FC<SiteHeaderMobileProps> = (/* {searchValue, onSearchChange} */) => {
```
SiTaggart, line 12: Commented out arguments?

packages/paste-website/src/components/site-wrapper/site-header/HamburgerToggle.tsx
```diff
@@ -0,0 +1,71 @@
...
+    transition="all .25s ease-in-out"
+    transformOrigin="left center"
+    {...props}
+  />
+);
+
+const HamburgerToggle = React.forwardRef<HTMLButtonElement, HamburgerToggleProps>(({toggled, onToggle}, ref) => {
+  const backgroundColor = toggled ? 'colorBackgroundBodyInverse' : 'colorBackgroundBody';
+  return (
+    <Button ref={ref} variant="reset" size="reset" onClick={() => onToggle(!toggled)}>
```
SiTaggart, line 25: For the button in the masthead, should it not be using our inverse button? https://twilio-labs.github.io/paste/?path=/story/components-button--inverse-button

packages/paste-website/src/components/site-wrapper/site-header/index.tsx
```diff
@@ -0,0 +1,19 @@
...
+import {SiteHeaderMobile} from './SiteHeaderMobile';
+import {useWindowSize} from '../../../hooks/useWindowSize';
+
+export const SiteHeader: React.FC = () => {
+  const [value, setValue] = React.useState('');
+  const {breakpointIndex} = useWindowSize();
+
+  return (
+    <>
+      {breakpointIndex !== undefined && breakpointIndex <= 1 ? (
```
SiTaggart, line 12: Overall I think this approach is interesting but the flash of content is pretty bad experience wise.

I think wrapping these in Boxes and just using a responsive display prop would result is a better page experience

SiTaggart (commented):
The flash of content is pretty undesirable when using JS to detect screen size instead of the responsive props, unfortunately. 

image

The Theme switcher also needs to be on the internal pages. It's only the homepage where it's not shown.

There's a small regression in the side nav, where some spacing has been removed at the top 
image

and on a mobile, when you open the menu there is a small jump where the menu is pushed down after rendering.

Another weird quirk, is that if you start from the new homepage, it'll keep you at the initial zoom level, so all internal pages are mobile responsive. But if you go directly to a page it keeps the current behavior.

packages/paste-website/src/components/site-wrapper/site-header/SiteHeaderMobile.tsx
TheSisb: Yeah for this PR. They need to be passed through but I'm not using them yet. Mobile search in a follow up PR.

packages/paste-website/src/components/site-wrapper/site-header/HamburgerToggle.tsx
TheSisb: Designs currently look like this 
image

@corinne-nagel Want to swap to those?

packages/paste-website/src/components/site-wrapper/site-header/index.tsx
TheSisb: 100% agree. I've resolved the FoUC by rendering both the mobile and desktop components while the breakpointIndex is still undefined (i.e.: SSR and while rehydrating), and relying on CSS to show the right thing.  Once the page is rehydrated and all the JS starts to kick in, I go back to only rendering the elements that should be displayed. Best of both worlds!

packages/paste-website/src/components/site-wrapper/SiteBody.tsx
TheSisb: In this case I removed the `isDesktop` variable name and inlined the condition. Essentially I render it while `breakpointIndex` is still `undefined`, and use CSS within Sidebar to display it or not. Once the page is hydrated I conditionally render it.

packages/paste-website/src/components/site-wrapper/sidebar/SidebarNestedItem.tsx
```diff
@@ -3,7 +3,13 @@ import {Box} from '@twilio-paste/box';
 
 export const SidebarNestedItem: React.FC = props => {
   return (
-    <Box as="li" marginTop="space20" marginBottom="space20" fontSize="fontSize30" lineHeight="lineHeight20">
+    <Box
+      as="li"
+      marginTop="space20"
+      marginBottom="space20"
+      fontSize={['fontSize50', 'fontSize50', 'fontSize30']}
```
richbachman, line 10: Just want to verify the fontSize should get larger on tablet and mobile sizes.

packages/paste-website/src/components/site-wrapper/site-header/SiteHeaderDesktop.tsx
```diff
@@ -0,0 +1,55 @@
...
+import {useNavigationContext} from '../../../context/NavigationContext';
+
+interface SiteHeaderDesktopProps {
+  searchValue: string;
+  onSearchChange: (newSearchValue: string) => void;
+}
+
+const SiteHeaderDesktop: React.FC<SiteHeaderDesktopProps> = ({searchValue, onSearchChange}) => {
+  const {pathname} = useNavigationContext();
+  const isHomepage = pathname === '/' || pathname === '/new';
```
richbachman, line 17: I'm seeing the theme switcher on /new. https://share.getcloudapp.com/WnulR0dm
Might need to change this to:

```suggestion
  const isHomepage = pathname === '/' || pathname === '/new/';
```

TheSisb: Oh right, because both `/new` and `/new/` are valid URLs. I need to check for both. Good catch!

packages/paste-website/src/components/site-wrapper/sidebar/SidebarNestedItem.tsx
TheSisb: According to the designs, yup

SiTaggart:
I'm still seeing the responsive layout on all pages, is that intentional. They look super broken

image

image

packages/paste-website/src/components/site-wrapper/site-header/index.tsx
```diff
@@ -0,0 +1,25 @@
...
+import {SiteHeaderDesktop} from './SiteHeaderDesktop';
+import {SiteHeaderMobile} from './SiteHeaderMobile';
+import {useWindowSize} from '../../../hooks/useWindowSize';
+
+export const SiteHeader: React.FC = () => {
+  const [value, setValue] = React.useState('');
+  const {breakpointIndex} = useWindowSize();
+
+  // While SSR, render both and let CSS handle it (fixes FoUC)
+  if (breakpointIndex === undefined) {
```
SiTaggart, line 11: I'm unsure what value this technique brings right now. For the SSR payload you're having to render both anyway. From that point, I don't see what the benefit of adding JavaScript to take over something that CSS is already handling with the responsive display props.

I think we should just use what the system provides

TheSisb: Initially I was only using the responsive style props to change the display css property as you mentioned. As I began building the responsive UI, I thought it would be cleaner in both the code and the UX to share the navigation state rather than have a unique copy for desktop and mobile.  

When I started sharing the state and working on the mobile header's navigation animation, I wasn't getting smooth 60fps animations. In my efforts to improve the performance of the animation, I was investigating what React was doing that could cause lag.  Turns out, even though the content was hidden in CSS, it was still being (re)rendered in the DOM.  To cancel out all that work, I figured it would be best not to render it at all.  That's when you called out the FoUC.  I then came up with this approach which would render both desktop and mobile in the static SSR copy (which is fine), and then conditionally render whatever actually needs rendering.

This isn't really a necessary change, but it helps the UX by improving the performance a tad. I don't think I'd use this everywhere, and you rightfully called out a place where I used it instead of a responsive prop. In the case of this large DOM tree of a site-header, I thought it best to keep it.

packages/paste-website/src/components/ContactUsMenu.tsx
SiTaggart: From menu primitive yes, from menu no. We're not reexporting the types from the primitive in the menu. I'll create a ticket for it.

packages/paste-website/src/components/ContactUsMenu.tsx
```diff
@@ -18,13 +40,19 @@ export const ContactUsMenu: React.FC<{}> = () => {
...
   return (
-    <Box backgroundColor="colorBackground" marginTop="auto" padding="space20" paddingBottom="space70">
+    <Box marginTop="auto" padding="space20" paddingBottom="space70">
       <MenuButton {...menu} variant="reset" size="reset" fullWidth>
         <Box as="span" display="flex" alignItems="center" padding="space40" width="100%">
-          <ProductSupportIcon decorative />
-          <Box as="span" marginLeft="space30" textAlign="left">
-            Contact us
-          </Box>
+          <MediaObject verticalAlign="center">
```
SiTaggart, line 46: Look at that lovely media-object

packages/paste-website/src/components/site-wrapper/ScrollAnchorIntoView.tsx
```diff
@@ -1,6 +1,8 @@
 import * as React from 'react';
 import {getCurrentPathHash} from '../../utils/RouteUtils';
 
+const Y_OFFSET = -90;
```
SiTaggart, line 4: We do have this partly in a constant for the site assuming this is an offset to accomodate for the masthead https://github.com/twilio-labs/paste/blob/main/packages/paste-website/src/constants.ts#L5

Obviously doesn't include padding to make it look nice

packages/paste-website/src/pages/index.tsx
```diff
@@ -21,13 +21,15 @@ import {P} from '../components/Typography';
...
-const IndexPage: React.FC<{}> = (): React.ReactElement => {
+const IndexPage: React.FC<{location: {pathname: string}}> = ({location}): React.ReactElement => {
   return (
-    <SiteWrapper>
+    <SiteWrapper pathname={location.pathname}>
       <Helmet>
         <title>{SiteMetaDefaults.TITLE}</title>
         <link rel="canonical" href="https://paste.twilio.design" />
         <meta name="description" content={SiteMetaDefaults.DESCRIPTION} />
+        {/* needed for proper responsive but only on homepage */}
```
SiTaggart, line 31: if you're using this everywhere, you can remove the comment

SiTaggart (approved):

richbachman (approved):

PR #965, opened 2020-11-20
chore(website): implement mobile search

TheSisb:
Found out docsearch.js was on NPM so we no longer need to pollute the window global.  Also allows us to reinit when swapping from desktop to mobile.
Added some scss to make the popout work on mobile. It's kinda jank CSS, but it works!


packages/paste-website/src/components/site-wrapper/site-header/SiteHeaderMobileSearch.tsx
```diff
@@ -0,0 +1,100 @@
...
+  // such as font-family and line-height so that compositions
+  // of paste components in the modal are styled correctly
+  pasteBaseStyles
+);
+
+const StyledModalDialogContent = styled(ModalDialogPrimitiveContent)(
+  css({
+    position: 'fixed',
+    top: 10,
+    left: 10,
```
richbachman, line 39: Dropping this `left` style should center the modal in the content area.

richbachman:
For some reason the search results get out of alignment on mobile sizes. https://share.getcloudapp.com/ApuGOXqB

TheSisb:
> For some reason the search results get out of alignment on mobile sizes. https://share.getcloudapp.com/ApuGOXqB

I'm gonna call that within an acceptable range of degradation, considering we don't control any of that CSS and possibly a version bump to docsearch.js may fix it in the future.

PR #941, opened 2020-11-16
fix(radio-group): Remove additional spacing after the last element of the radio group.

roman-kulpa:
Remove additional spacing after the last element of the radio group

**Contributing to Twilio**

> All third-party contributors acknowledge that any contributions they provide will be made under the same open-source license that the open-source project is provided under.

- [x] I acknowledge that all my contributions will be made under the project's license.

PR #974, opened 2020-11-25
fix(inline-control-group): remove bottom spacing from last child control

SiTaggart:
Taking over from https://github.com/twilio-labs/paste/pull/941

Using stack to create the spacing and general tidy up. Also adding stories to verify no regressions.

### DIFF
--- a/.changeset/nervous-moles-sleep.md
+++ b/.changeset/nervous-moles-sleep.md
@@ -1,0 +1,6 @@
+---
+'@twilio-paste/radio-group': patch
+'@twilio-paste/core': patch
+---
+
+[RadioGroup] Added optional `value` string prop which this component needs and is no longer on the extended `InlineControlGroup` interface.
--- a/.changeset/stupid-zebras-wave.md
+++ b/.changeset/stupid-zebras-wave.md
@@ -1,0 +1,6 @@
+---
+'@twilio-paste/checkbox': patch
+'@twilio-paste/core': patch
+---
+
+[Checkbox] Remove unused `value` prop.
--- a/.changeset/two-humans-press.md
+++ b/.changeset/two-humans-press.md
@@ -1,0 +1,6 @@
+---
+'@twilio-paste/inline-control-group': patch
+'@twilio-paste/core': patch
+---
+
+[InlineControlGroup] Adjusted the vertical spacing of `children`. This change improve the handling of RadioGroup and CheckboxGroup children.
--- a/packages/paste-core/components/form/stories/form.stories.tsx
+++ b/packages/paste-core/components/form/stories/form.stories.tsx
@@ -1,7 +1,150 @@
 import * as React from 'react';
-import {useUIDSeed} from '@twilio-paste/uid-library';
+import {useUID, useUIDSeed} from '@twilio-paste/uid-library';
+import {Text} from '@twilio-paste/text';
 import {Box} from '@twilio-paste/box';
-import {Checkbox, FormLabel, FormInput, Select, Option, Radio, FormTextArea} from '../src';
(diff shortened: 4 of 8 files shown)


TheSisb (approved):

SiTaggart:
This is not as easy as first thought. The reason for the bottom margin is for when a group is orientated in a horizontal row and the number of items exceeds the row width space.

image

We can't remove the bottom margin in horizontal layouts entirely as the second row would remove the desired space between rows.

We can't _only_ remove the bottom margin on the last child, because as you can see in the screen shot, there might be more than one dangling options which will introduce the gap along the bottom of the groups.

I'll think about this some more. Maybe, because a group must have a legend all the spacing is top only?

TheSisb:
I kind of like a rule of marginTop only on our components, with the option to unset the margin. Mixing the two leads to scenarios where the margins collapse sometimes. We also can't predict what comes after an element, but we may have an opinion on what should come before.

packages/paste-core/components/form/stories/form.stories.tsx
```diff
@@ -65,8 +240,180 @@ export const All = (): React.ReactNode => {
       <Box marginBottom="space70">
         <FormLabel htmlFor={inputIDSeed('errorinput')}>Text Input</FormLabel>
         <FormInput hasError id={inputIDSeed('errorinput')} type="text" value="" />
+        <Grid>
+          <Column>
+            <CheckboxGroup
```
TheSisb, line 245: Seems like a lot of copy paste in this story. Can we component-ize this (like you did in inline-control-group) so the difference is easy to see?

.changeset/two-humans-press.md
```diff
@@ -0,0 +1,8 @@
+---
+'@twilio-paste/inline-control-group': patch
+'@twilio-paste/core': patch
+---
+
+Small visual adjustment to the the vertical spacing of children when passed to Inline Control Group.
+
+This also cascades down to Radio Group Checkbox Group spacing.
```
TheSisb, line 8:
```suggestion
[InlineControlGroup] Adjusted the vertical spacing of `children`.  This change improve the handling of RadioGroup and CheckboxGroup children.
```

.changeset/stupid-zebras-wave.md
```diff
@@ -0,0 +1,6 @@
+---
+'@twilio-paste/checkbox': patch
+'@twilio-paste/core': patch
+---
+
+Remove unused value prop from CheckboxGroup as it's now only required in RadioGroup
```
TheSisb, line 6:
```suggestion
[Checkbox] Remove unused `value` prop.
```

.changeset/nervous-moles-sleep.md
```diff
@@ -0,0 +1,6 @@
+---
+'@twilio-paste/radio-group': patch
+'@twilio-paste/core': patch
+---
+
+Value is added as a prop used specifically by RadioGroup as it was inheriting its prop interface from InlineContorlGroup which doens't use value at all.
```
TheSisb, line 6: I started reading the changes by only reading the changesets, to see if I could understand the changes. I had a hard time understanding the changesets so I'm providing some suggested improvements.

```suggestion
[RadioGroup] Added optional `value` string prop which this component needs and is no longer on the extended `InlineControlGroup` interface.
```

TheSisb (commented):
Really nice changes and improvements. Mostly nits.

packages/paste-core/components/form/stories/form.stories.tsx
SiTaggart: Probably good to give it a try

PR #979, opened 2020-11-30
chore(website): render footer angle on home page only

richbachman:


packages/paste-website/src/constants.ts
```diff
@@ -3,7 +3,8 @@ export const TWILIO_BLUE = '#0D122B';
 export const PSA_ALERT_HEIGHT = 40;
 export const PASTE_THEME_ALERT_HEIGHT = 54;
 export const SITE_MASTHEAD_HEIGHT = 76;
-export const SITE_BREAKPOINTS = ['768px', '1024px', '1220px'];
+// Added 1680 and above to account for really large monitors
```
TheSisb, line 6: This comment is lying about the value added 🙈

TheSisb (approved):

packages/paste-website/src/constants.ts
richbachman: Forgot to update that one. I'll do that now.

PR #988, opened 2020-12-02
fix: new homepage icon list responsive

SiTaggart:
Make the icon list page a little nicer on responsive screens

### DIFF
--- a/packages/paste-website/src/components/icons-list/IconsList.tsx
+++ b/packages/paste-website/src/components/icons-list/IconsList.tsx
@@ -54,12 +54,6 @@ const getFirstIcon = (iconsList: GroupedList): IconObject | null => {
   return null;
 };
 
-// Our Grid Column doesn't allow position for sticky, but our box doesn't allow percentage widths
-// Styled Box is fine I guess, but we might want to expand Box to take any width
-const IconColumn = styled(Box)({
-  width: '41.66666666666667%',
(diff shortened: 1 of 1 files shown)


TheSisb (approved):
I'm happy with these changes, thank you. nit: it would be nice if we could automatically scroll to the box way at the bottom when tapping on an icon

packages/paste-website/src/components/icons-list/IconsList.tsx
```diff
@@ -189,9 +185,14 @@ const IconsList: React.FC<IconsListProps> = () => {
             </Card>
           )}
         </Column>
-        <IconColumn paddingX="space60" position="sticky" top={STICKY_COLUMN_OFFSET}>
+        <Box
+          paddingX="space60"
+          position="sticky"
```
richbachman, line 190: It looks like this position sticky isn't working anymore, but its not working on the production site either: https://paste.twilio.design/icons

We can create another ticket to figure out that issue.

richbachman (approved):

packages/paste-website/src/components/icons-list/IconsList.tsx
SiTaggart: Yeah, that confused me too. I can't remember if we removed it or it broke at some point.

Issue #972, opened 2020-11-25
Input inline error icon size is changing


Dooffy:
**Describe the bug**
Input inline error icon size is relative and depends on error message length.

**To Reproduce**
1. Create new input element with multi-line error message.

**Expected behavior**
Icon size should not change.

**Screenshots**
<img width="803" alt="Screenshot 2020-11-25 at 11 43 07" src="https://user-images.githubusercontent.com/1443584/100224848-47c76280-2f15-11eb-9b98-ef17b564c6c8.png">

SiTaggart:
Thanks for the bug report @Dooffy. I've created a ticket in our backlog to address as soon as possible.

richbachman:
Hi @Dooffy. We just merged a fix in for this. You'll need to update your Paste dependencies once the build is finished. Let us know if you have any other issues.

PR #997, opened 2020-12-03
fix(help-text): stop icon from shrinking with long error text

richbachman:
- [x] Fixes icon shrinking when error text is really long
- [x] Added VRT stories to input

Fixes: https://github.com/twilio-labs/paste/issues/972


packages/paste-core/components/input/stories/input.stories.tsx
```diff
@@ -171,6 +171,35 @@ storiesOf('Components|Input', module)
       </>
     );
   })
+  .add('Input - Really Long Error', () => {
```
richbachman, line 174: Should we also add stories to Textarea, Select, and Combobox?

TheSisb: I think this is enough

packages/paste-core/components/help-text/__tests__/__snapshots__/helptext.test.tsx.snap
```diff
@@ -54,15 +54,15 @@ exports[`HelpText render it should render 1`] = `
 `;
 
 exports[`HelpText render it should render with an error icon 1`] = `
-.emotion-3 {
+.emotion-4 {
```
TheSisb, line 57: Can we remove the snapshot test? VRT covers it now.

TheSisb (approved):

PR #1000, opened 2020-12-04
fix: add fowardRefs to all exported components

richbachman:


packages/paste-core/components/combobox/src/Combobox.tsx
```diff
@@ -29,104 +29,108 @@ const StyledInputAsSelect = styled(InputElement)<ComboboxProps>(props =>
...
+        {items.map((item, index) => (
+          <Item
+            // eslint-disable-next-line react/no-array-index-key
+            key={index}
+            item={item}
+            index={index}
+            getItemProps={getItemProps}
+            highlightedIndex={highlightedIndex}
+            optionTemplate={optionTemplate}
+            ref={ref}
```
TheSisb, line 62: In this case, you're passing the same ref to several elements (each item in the map). I don't think we should forwardRef here

packages/paste-core/components/combobox/src/Combobox.tsx
```diff
@@ -29,104 +29,108 @@ const StyledInputAsSelect = styled(InputElement)<ComboboxProps>(props =>
...
+          if (groupedItemKey === 'undefined') {
+            return groupedItems[groupedItemKey].map((item: ItemType, index: number) => (
+              <Item
+                item={item}
+                index={index}
+                key={UIDSeed(`ungrouped-${index}`)}
+                getItemProps={getItemProps}
+                highlightedIndex={highlightedIndex}
+                optionTemplate={optionTemplate}
+                ref={ref}
```
TheSisb, line 95: We shouldn't pass the same ref in a map. This file may need more consideration.

packages/paste-core/components/disclosure/src/Disclosure.tsx
```diff
@@ -26,23 +26,27 @@ export interface DisclosureProps extends DisclosurePrimitiveInitialState {
...
+      return (
+        <DisclosureContext.Provider value={disclosureContext}>
+          <Card padding="space0" ref={ref}>
+            {children}
+          </Card>
+        </DisclosureContext.Provider>
+      );
+    }
+
+    return <DisclosureContext.Provider value={disclosureContext}>{children}</DisclosureContext.Provider>;
```
TheSisb, line 47: Hmm, is it fine that ref isn't passed for variant !== contained?

packages/paste-core/components/checkbox/src/CheckboxDisclaimer.tsx
```diff
@@ -10,22 +10,24 @@ export interface CheckboxDisclaimerProps
...
-        </Checkbox>
-      </Box>
-      {errorText && (
-        <Box marginTop="space30" marginLeft="space50">
-          <HelpText variant="error">{errorText}</HelpText>
+const CheckboxDisclaimer = React.forwardRef<HTMLDivElement, CheckboxDisclaimerProps>(
+  ({children, errorText, ...props}, ref) => {
+    return (
+      <>
+        <Box backgroundColor="colorBackground" borderRadius="borderRadius20" padding="space50" ref={ref}>
```
TheSisb, line 17: Unsure myself here, asking for other opinions: should we pass ref directly to the Checkbox in this case?

packages/paste-core/components/modal/src/Modal.tsx
```diff
@@ -86,56 +86,62 @@ const getAnimationStates = (isConsole: boolean): any => ({
...
+    return (
+      <ModalContext.Provider value={{onDismiss}}>
+        {transitions(
+          (styles, item) =>
+            item && (
+              <ModalDialogOverlay
+                onDismiss={onDismiss}
+                allowPinchZoom={allowPinchZoom}
+                initialFocusRef={initialFocusRef}
+                ref={ref}
```
TheSisb, line 126: We should move this ref to the ModalDialogContent

packages/paste-core/components/screen-reader-only/package.json
```diff
@@ -26,12 +26,24 @@
     "compile:dev": "rollup -c --environment NODE_ENV:development"
   },
   "peerDependencies": {
+    "@twilio-paste/box": "^2.11.13",
```
TheSisb, line 29: How was this working :O

packages/paste-core/components/screen-reader-only/src/index.tsx
```diff
@@ -1,22 +1,40 @@
-import {styled} from '@twilio-paste/styling-library';
+import * as React from 'react';
+import * as PropTypes from 'prop-types';
+import {safelySpreadBoxProps, Box, BoxProps} from '@twilio-paste/box';
 
-export interface ScreenReaderOnlyProps {
-  // as prop isn't typed correctly from styled package https://github.com/emotion-js/emotion/issues/1137
-  as?: keyof JSX.IntrinsicElements;
+export interface ScreenReaderOnlyProps extends Pick<BoxProps, 'as'> {
```
TheSisb, line 5: 👍

packages/paste-core/components/toast/src/ToastContainer.tsx
```diff
@@ -3,15 +3,15 @@ import {Stack} from '@twilio-paste/stack';
 import {ToastPortal} from './ToastPortal';
 import {ToastContainerProps, ToastContainerPropTypes} from './types';
 
-const ToastContainer: React.FC<ToastContainerProps> = ({children, ...props}) => {
+const ToastContainer = React.forwardRef<HTMLDivElement, ToastContainerProps>(({children, ...props}, ref) => {
   return (
-    <ToastPortal {...props}>
+    <ToastPortal {...props} ref={ref}>
```
TheSisb, line 8: I would move this ref on to the Stack to be more precise on what the user expects the ref to point to

packages/paste-core/components/truncate/src/index.tsx
```diff
@@ -1,12 +1,33 @@
...
+import {safelySpreadBoxProps, Box, BoxProps} from '@twilio-paste/box';
+
+export interface TruncateProps extends Pick<BoxProps, 'as'> {
+  children: NonNullable<React.ReactNode>;
+}
+
+const Truncate = React.forwardRef<HTMLSpanElement, TruncateProps>(({children, ...props}, ref) => {
+  return (
+    <Box
+      {...safelySpreadBoxProps(props)}
```
TheSisb, line 12: Do we care to safelySpread here? Truncate isn't really a designed component so it may be fine to just spread all Box props.. What do you think?

packages/paste-core/components/screen-reader-only/src/index.tsx
```diff
@@ -1,22 +1,40 @@
...
-  clip: rect(0 0 0 0);
-  text-transform: none;
-  white-space: nowrap;
-`;
-
+const ScreenReaderOnly = React.forwardRef<HTMLElement, ScreenReaderOnlyProps>(
+  ({as = 'span', children, ...props}, ref) => {
+    return (
+      <Box
+        {...safelySpreadBoxProps(props)}
```
TheSisb, line 13: Same question as on Truncate. Do we need to safelySpread here?

packages/paste-core/components/truncate/src/index.tsx
```diff
@@ -1,12 +1,33 @@
...
+import * as PropTypes from 'prop-types';
+import {safelySpreadBoxProps, Box, BoxProps} from '@twilio-paste/box';
+
+export interface TruncateProps extends Pick<BoxProps, 'as'> {
+  children: NonNullable<React.ReactNode>;
+}
+
+const Truncate = React.forwardRef<HTMLSpanElement, TruncateProps>(({children, ...props}, ref) => {
+  return (
+    <Box
```
TheSisb, line 11: It's great we can now use Box for these 👍 Good refactor

packages/paste-core/components/disclosure/src/Disclosure.tsx
richbachman: Good catch. I just missed that.

packages/paste-core/components/combobox/src/Combobox.tsx
richbachman: Right. This ref should only be passed to the `ComboboxListboxGroup`.

packages/paste-core/components/checkbox/src/CheckboxDisclaimer.tsx
richbachman: That actually makes more sense to me. The rest of the the disclaimer, i.e. Box is just a visual element.

packages/paste-core/components/truncate/src/index.tsx
richbachman: Yeah I don't think we need it here either. We probably don't need to be concerned with what gets passed here since there aren't any design implications.

packages/paste-core/components/screen-reader-only/src/index.tsx
richbachman: Same as Truncate.

TheSisb (approved):

PR #1001, opened 2020-12-07
chore: checkpoint storybook v6 [wip]

TheSisb:
Upgrading storybook to v6


.storybook/main.js
```diff
@@ -0,0 +1,4 @@
+module.exports = {
+  stories: ['../packages/**/stories/**/*.storiess.@(ts|tsx)'],
```
TheSisb, line 2: Only stories with two `s` will get parsed (for now, while migrating).

.storybook/preview.js
```diff
@@ -0,0 +1,14 @@
+import React from 'react';
+import {Theme} from '@twilio-paste/theme';
+
+export const parameters = {
+  actions: {argTypesRegex: '^on[A-Z].*'},
+};
+
+export const decorators = [
+  Story => (
+    <Theme.Provider theme="default">
```
TheSisb, line 10: We lose the theme switcher for now because we need to migrate the addon.

package.json
```diff
@@ -76,16 +77,15 @@
...
-    "@storybook/addon-actions": "^5.2.3",
-    "@storybook/addon-info": "^5.2.3",
-    "@storybook/addon-knobs": "^5.2.3",
-    "@storybook/addon-links": "^5.2.3",
-    "@storybook/addon-storyshots": "^5.2.3",
-    "@storybook/addon-viewport": "^5.3.14",
-    "@storybook/addons": "^5.2.3",
-    "@storybook/core": "^5.3.14",
-    "@storybook/react": "^5.2.3",
+    "@storybook/addon-a11y": "^6.1.10",
```
TheSisb, line 80: TODO: clean up these packages based on what we end up keeping

packages/paste-core/components/help-text/stories/HelpText.storiess.tsx
```diff
@@ -0,0 +1,32 @@
...
+import {Box} from '@twilio-paste/box';
+import {HelpText, HelpTextVariants} from '../src';
+
+/* eslint-disable-next-line import/no-default-export */
+export default {
+  title: 'Components/HelpText',
+  component: HelpText,
+};
+
+const Template: React.FC<{variant: HelpTextVariants}> = ({variant}) => (
```
TheSisb, line 11: TODO: Not sold on this API, need to think some more on how to make it so that we can use these stories in docs and elsewhere

GiantRobots: You can use the named Exported components in other places. We re-use them for tests so storybook has some additional testing for us.

package.json
```diff
@@ -76,16 +77,15 @@
...
+    "@storybook/addon-actions": "^6.1.10",
+    "@storybook/addon-links": "^6.1.10",
+    "@storybook/addon-essentials": "^6.1.10",
+    "@storybook/addon-storyshots": "^6.1.10",
+    "@storybook/addon-viewport": "^6.1.10",
+    "@storybook/addons": "^6.1.10",
+    "@storybook/core": "^6.1.10",
+    "@storybook/react": "^6.1.10",
```
GiantRobots, line 88: Just throwing this here because I couldn't comment where the linter was hollering at the knobs imports.

https://storybook.js.org/blog/storybook-controls/

Controls replaces knobs the api is nicer and pretty auto-magic which is really helpful.

.storybook/preview.js
GiantRobots: I wrote a theme switcher for the SendGrid storybook, it was pretty easy to create.

```typescript
import React from "react";
import { Theme } from "@twilio-paste/core/theme";

export const globalTypes = {
  theme: {
    name: "Paste Theme",
    description: "Global theme for components",
    defaultValue: "console", // console looks more like sendgrid than "sendgrid"
    toolbar: {
      // All available icons
      // https://github.com/storybookjs/storybook/blob/master/lib/components/src/icon/icons.tsx
      icon: "paintbrush",
      // array of plain string values or MenuItem shape (see below)
      items: ["sendgrid", "console", "flex"],
    },
  },
};

export const parameters = {
  actions: { argTypesRegex: "^on.*" },
  rootStyles: {
    height: "100vh",
    padding: "60px",
    backgroundColor: "white",
  },
};

export const decorators = [
  (Story, context) => {
    const theme = context.globals.theme;

    return (
      <Theme.Provider theme={theme}>
        <div style={context.parameters.rootStyles}>
          <Story />
        </div>
      </Theme.Provider>
    );
  },
];

```

PR #1002, opened 2020-12-07
feat: Paste token <--> package mappings

richbachman:
This PR adds a NodeJS script into the tools folder to detect which tokens are used in paste-core. The script is executed with `node tools/build/token-usage-detector.js`. The output is saved into the `tools/.cache` folder.
The current script captures:
1. The packages each token is used within.
2. Which tokens aren't being used at all (from within /paste-core/)
3. The minimal set of packages that, combined, use all the currently-in-use Paste tokens.

This information is useful because it helps us build a better preview pane on the Theme Designer tool and has the possibility to help us with token maintenance.

Open questions:
- When should we run this code?
- Where should we save the output?


tools/.cache/token-usage.json
```diff
@@ -0,0 +1,1206 @@
+{
```
SiTaggart, line 1: Is there a possibility we can get away with not committing the cache?

tools/build/token-usage-detector.js
```diff
@@ -0,0 +1,153 @@
+const path = require('path');
```
SiTaggart, line 1: In a super ideal world this would be typescript and have tests. I'd be ok with just tests for now.

tools/.cache/token-usage.json
TheSisb: Hmm... I'll remove the cache for now and we can build it when we need it in the future.

richbachman:
I wonder if this should live in/with reporting and run when that runs.

tools/.cache/token-usage.json
SiTaggart: It can be generated on the fly when ever this runs, I just think we can gitignore the cache. There's no need to commit the output either really

SiTaggart:
> I wonder if this should live in/with reporting and run when that runs.

Long term I see reporting as an ingestion point, it shouldn't have to run there, but we can fire data to it, if we want to track it.

tools/build/token-usage-detector.js
TheSisb: Take a look again. Full TS and tests added.

TheSisb:
> I wonder if this should live in/with reporting and run when that runs.

Since this has to scan through src files, it can't run outside of the monorepo as easily. I mean, I guess it can clone the monorepo and scan it, but it makes more sense inside the monorepo if we plan to use it for theme-designer and docs in the future.

SiTaggart (approved):
😍

PR #1016, opened 2020-12-09
feat: customization provider token designer

SiTaggart:
- add a generic theme type
- add a method to generate a theme from a set of known tokens
- add the customization provider to the preview and provide it a custom theme


richbachman (approved):

TheSisb (approved):
Awesome!

PR #1017, opened 2020-12-09
chore: refactor preview, fix console errors

richbachman:
- Refactored preview area and split into separate components
- Fixed the console errors associated with the preview components


packages/paste-token-designer/components/preview/components/AlertPreview.js
```diff
@@ -0,0 +1,29 @@
+import {Alert} from '@twilio-paste/core/alert';
```
TheSisb, line 1: nit: the folder structure of `components/preview/components` is a little awkward.  Would it be fine to move all of these up a level?

TheSisb (approved):

TheSisb (approved):

PR #1019, opened 2020-12-09
chore: remove playground page and plugin new preview

TheSisb:
Removed the old "playground" page and adds the new preview pane to the editor


packages/paste-token-designer/components/layout/editor-wrapper/index.js
```diff
@@ -21,7 +21,7 @@ const EditorWrapper = props => {
             </Button>
           </Box>
         </Box>
-        <PreviewPane />
+        <Preview />
```
TheSisb, line 24: So we can see the new new

richbachman (approved):

PR #1024, opened 2020-12-10
feat: remove the color key from theme

SiTaggart:
In the start we wanted explicit keys on the theme for each bucket of color types. `backgroundColor` => `backgroundColors`, `color` => `textColors` etc so that the theme was very focused.

When we introduced pseudo selectors to box and text, we had to place each type of color on a generic `colors` bucket as that's what the styled-system css function was hardcoded to look for.

This led to duplication.

When change a backgroundColorSuccess on a theme, you'd need to update it on `backgroundColors` and `colors`.

To get around this we fork the css function from styled-system, and manually replace the keys it looks for on the theme object to suit our own needs.

In doing this we also 

1. remove the colors key from all themes
2. modify the typings of the generic theme
3. Modify box to explicitly look for our color buckets on theme, rather than rely on the built in mappings to the colors key
4. add a bunch of tests to make this easier in the future if we ever need to do this again.


packages/paste-libraries/styling/src/css-function/css.ts
```diff
@@ -0,0 +1,224 @@
...
+
+const multiples = {
+  marginX: ['marginLeft', 'marginRight'],
+  marginY: ['marginTop', 'marginBottom'],
+  paddingX: ['paddingLeft', 'paddingRight'],
+  paddingY: ['paddingTop', 'paddingBottom'],
+  size: ['width', 'height'],
+};
+
+const scales = {
```
SiTaggart, line 50: None of these now look at the colors key

packages/paste-core/primitives/box/src/index.tsx
```diff
@@ -127,6 +127,22 @@ const extraConfig = system({
...
+    scale: 'borderColors',
+  },
+  borderRightColor: {
+    property: 'borderRightColor',
+    scale: 'borderColors',
+  },
+  borderTopColor: {
+    property: 'borderTopColor',
+    scale: 'borderColors',
+  },
```
SiTaggart, line 145: These were working from the colors key and we kinda didn't notice

packages/paste-libraries/styling/src/css-function/__tests__/css.test.ts
```diff
@@ -0,0 +1,480 @@
+import css from '../css';
```
SiTaggart, line 1: Mainly copy pasta from styled system, just a few tests update to meet our theme key requirements for colors

packages/paste-libraries/styling/src/css-function/css.ts
```diff
@@ -0,0 +1,224 @@
+// Very lightly TS Version of https://github.com/styled-system/styled-system/blob/master/packages/css/src/index.js
```
SiTaggart, line 1: Mainly copy pasta from styled-system. Simplified the get method a little, renames some arguments in it

packages/paste-theme/__tests__/themes.ts
```diff
@@ -0,0 +1,17 @@
...
+describe('Console theme', () => {
+  it('should match the snapshot', () => {
+    expect(ConsoleTheme).toMatchSnapshot();
+  });
+});
+describe('Sendgrid theme', () => {
+  it('should match the snapshot', () => {
+    expect(SendGridTheme).toMatchSnapshot();
+  });
+});
```
SiTaggart, line 17: These will obviously change as we add tokens or change token values, but I think it's worth it to have the fallback of knowning what the generated theme shape is, when / if we come to refactor the theming

packages/paste-libraries/styling/src/css-function/css.ts
```diff
@@ -0,0 +1,224 @@
...
+  width: 'sizes',
+  minWidth: 'sizes',
+  maxWidth: 'sizes',
+  height: 'sizes',
+  minHeight: 'sizes',
+  maxHeight: 'sizes',
+  flexBasis: 'sizes',
+  size: 'sizes',
+  // svg
+  fill: 'fillColors',
```
TheSisb, line 120: Should we make this `backgroundColors` or leave it as is?

packages/paste-libraries/styling/src/css-function/css.ts
```diff
@@ -0,0 +1,224 @@
+// Very lightly TS Version of https://github.com/styled-system/styled-system/blob/master/packages/css/src/index.js
+// not typed correctly in the slightest
```
TheSisb, line 2: Can you outline the specific lines you've changed from the src (logic wise) with comments to make future maintenance easier please
And can you also leave a note on which version of the css function you copied from Styled-System, preferably with a link to the commit/file in github

packages/paste-theme/__tests__/themes.ts
TheSisb: What if we did 
`expect(Object.keys(SendGridTheme).sort()).toMatchSnapshot()`?

packages/paste-core/primitives/box/src/index.tsx
richbachman: Good catch.

packages/paste-theme/__tests__/themes.ts
SiTaggart: Yeah, I buy that. We're only testing the buckets are right, don't care about the tokens, good shout

packages/paste-libraries/styling/src/css-function/css.ts
SiTaggart: I went with specific fill and stroke categories to be able to facilitate the possibility that illustrations have their own palette of colors separate to UI controls and to just keep the very specific nature of bucketing

TheSisb (approved):

PR #1045, opened 2020-12-15
feat: load default fonts from cdn and use tabular figures

SiTaggart:
To use Inters tabular figures we need to load the font from the author and not Google web fonts.

We might as well load the mono spaced font too, fira mono.

We now host the font files on our Twilio CDN.

Best practice is to use the stylesheet where the fonts are linked from and hosted on the CDN. If you do this you can preconnect and download the fonts faster. 

Fallback is that the theme provider loads includes CSS to load the fonts from the CDN too, if you don't use the stylesheet.

- This won't double download the assets, the browser already knows you've got the fonts, so even if the provider adds the CSS to request them again, nothing happens
- The added size to the provider is pretty small
- Best of both worlds. Super speedy downloads if you follow our advice, but you still get the fonts if you don't, just slower.


packages/paste-theme/src/styles/fonts.ts
```diff
@@ -0,0 +1,74 @@
...
+/* --------------------------------------------------------------------------
+[EXPERIMENTAL] Multi-axis, single variable font.
+
+Slant axis is not yet widely supported (as of February 2019) and thus this
+multi-axis single variable font is opt-in rather than the default.
+
+When using this, you will probably need to set font-variation-settings
+explicitly, e.g.
+
+  * { font-variation-settings: "slnt" 0deg }
```
TheSisb, line 32: I noticed we aren't setting this. Should we opt out of the experimental font for now? I'm not sure that we will need this.

TheSisb (approved):
Looks great

richbachman (approved):

serifluous:
Just dropping these here as a record (no need to change). The tabular figures look unnoticeable inline in most cases except numbers with 1s and 7s (thinner numbers).

Screen Shot 2020-12-17 at 10 11 29 AM
Screen Shot 2020-12-17 at 10 10 48 AM
Screen Shot 2020-12-17 at 10 09 35 AM

serifluous (approved):

packages/paste-theme/src/styles/fonts.ts
SiTaggart: I am changing the italic slant slightly using the experimental font here https://github.com/twilio-labs/paste/pull/1045/files#diff-ec0ccc84914af419df413ea5920db554fc99334c8b9d9af6ff59a9250aa0acccR40 on line 40

PR #1078, opened 2021-01-07
chore(table): borderless and striped, adjust styles, promote to beta

richbachman:
- [x] Added borderless variant
- [x] Added striped prop
- [x] Style adjustments
- [x] Added additional stories
- [x] Added additional tests

### DIFF
--- a/packages/paste-core/components/table/__tests__/index.spec.tsx
+++ b/packages/paste-core/components/table/__tests__/index.spec.tsx
@@ -18,7 +18,7 @@ describe('Table', () => {
         </TBody>
       </Table>
     );
-    const renderedTable = screen.getByTestId('mockDefaultTable').parentElement;
+    const renderedTable = screen.getByTestId('mockDefaultTable');
     expect(renderedTable).toHaveStyleRule('border-width', 'borderWidth10');
   });
   it('should render a borderless table', (): void => {
@@ -32,7 +32,7 @@ describe('Table', () => {
         </TBody>
       </Table>
     );
-    const renderedTable = screen.getByTestId('mockBorderlessTable').parentElement;
+    const renderedTable = screen.getByTestId('mockBorderlessTable');
     expect(renderedTable).toHaveStyleRule('border-width', 'borderWidth0');
   });
   it('should render a striped table', (): void => {
--- a/packages/paste-core/components/table/__tests__/index.spec.tsx
+++ b/packages/paste-core/components/table/__tests__/index.spec.tsx
@@ -7,6 +7,54 @@ import {Table, THead, TBody, TFoot, Td, Tr, Th} from '../src';
 expect.extend(matchers);
 
 describe('Table', () => {
+  it('should render a default table', (): void => {
+    render(
+      <Table data-testid="mockDefaultTable">
+        <TBody>
+          <Tr>
+            <Td>Column 1</Td>
+            <Td>Column 2</Td>
+          </Tr>
+        </TBody>
+      </Table>
+    );
+    const renderedTable = screen.getByTestId('mockDefaultTable').parentElement;
+    expect(renderedTable).toHaveStyleRule('border-width', 'borderWidth10');
+  });
+  it('should render a borderless table', (): void => {
+    render(
+      <Table variant="borderless" data-testid="mockBorderlessTable">
+        <TBody>
+          <Tr>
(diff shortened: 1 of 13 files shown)


packages/paste-core/components/table/__tests__/index.spec.tsx
```diff
@@ -7,6 +7,54 @@ import {Table, THead, TBody, TFoot, Td, Tr, Th} from '../src';
...
+          <Tr>
+            <Td>Column 3</Td>
+            <Td>Column 4</Td>
+          </Tr>
+        </TBody>
+      </Table>
+    );
+    const renderedTable = screen.getByTestId('mockStripedTable').querySelector('tr:nth-of-type(even)');
+    expect(renderedTable).toHaveStyleRule('background-color', 'colorBackgroundRowStriped', {
+      target: ':nth-of-type(even)',
```
TheSisb, line 55: neat

packages/paste-core/components/table/src/Tr.tsx
```diff
@@ -1,10 +1,13 @@
 import * as React from 'react';
 import {Box, safelySpreadBoxProps} from '@twilio-paste/box';
+import {TableContext} from './TableContext';
 import {TrProps, TrPropTypes} from './types';
 
 const Tr = React.forwardRef<HTMLTableRowElement, TrProps>(({verticalAlign = 'middle', ...props}, ref) => {
+  const table = React.useContext(TableContext);
```
TheSisb, line 7: nit: not a fan of the name `table` for this const. Maybe just `{striped}`?

packages/paste-core/components/table/src/types.ts
```diff
@@ -43,29 +53,29 @@ export const TrPropTypes = {
...
-  children: PropTypes.node.isRequired,
+  children: PropTypes.node,
   textAlign: PropTypes.oneOf(Object.values(TableAlignmentObject)),
   width: isWidthTokenProp,
 };
 
-export interface TdProps {
-  children: NonNullable<React.ReactNode>;
+export interface TdProps extends React.TdHTMLAttributes<HTMLTableCellElement> {
+  children?: React.ReactNode;
```
TheSisb, line 69: ReactNode already handles undefined, so you can omit the `?`

packages/paste-core/components/table/src/types.ts
```diff
@@ -43,29 +53,29 @@ export const TrPropTypes = {
   verticalAlign: PropTypes.oneOf(Object.values(TableVerticalAlignmentObject)),
 };
 
-export interface ThProps {
-  children: NonNullable<React.ReactNode>;
+export interface ThProps extends React.ThHTMLAttributes<HTMLTableHeaderCellElement> {
+  children?: React.ReactNode;
```
TheSisb, line 57: `?` this one too

TheSisb (approved):
Some minor nits; this looks solid

packages/paste-core/components/table/src/types.ts
```diff
@@ -1,39 +1,49 @@
...
 export type TableAlignmentOptions = ValueOf<typeof TableAlignmentObject>;
 export type TableLayoutOptions = ValueOf<typeof TableLayoutObject>;
+export type TableVariantOptions = ValueOf<typeof TableVariantObject>;
 export type TableVerticalAlignmentOptions = ValueOf<typeof TableVerticalAlignmentObject>;
 
+export interface TableContextProps {
+  striped: boolean;
+}
+
 export interface TableProps {
```
SiTaggart, line 10: Not extending from the table HTML element here?

packages/paste-core/components/table/src/types.ts
richbachman: Interesting. If I remove the `?`, TS still thinks it's required.
https://share.getcloudapp.com/Wnuln1xA

richbachman: I could just remove `children` from the interface.

packages/paste-core/components/table/src/types.ts
richbachman: Thanks for catching that.

packages/paste-core/components/table/src/types.ts
richbachman: Leaving as is for now.

PR #1084, opened 2021-01-11
feat: ESBuild for builds

TheSisb:
This PR switches off rollup and rollup-plugin-typescript2 for builds in favor for [ESbuild](https://esbuild.github.io/).
Locally, full builds have went **from 300s+ to around 70s!**

- [x] Add a build config to each package
- [x] Update tsconfig to be more streamlined and optimized
- [x] Fix all missing `import type` in our codebase
- [x] Clean up every package's `package.json` file. They all now have a `build:js` command that runs in ~6s to build only the JS.
- [x] Remove the no longer needed `tsconfig.build.json` file. Also removes the `prebuild` script that had to set package references.


SiTaggart (approved):

richbachman (approved):
Great work on this!

PR #1097, opened 2021-01-14
chore: add manypkg and fix issues

TheSisb:
Adds [manypkg](https://github.com/Thinkmill/manypkg) commands to Paste and runs the check command on `prebuild`. Very useful to make sure our package.jsons are well maintained.


richbachman (approved):
Builds all look good, so I think its good to go.

PR #1101, opened 2021-01-20
chore: storybook v6 deps and type fixes

SiTaggart:
- Upgrade all storybook deps
- re-added a tsconfig as Storybook default typescript set up uses it
  - We also lost some of the type checks we were using before like strict mode
- Fix type issues with test files and stories as it's now part of the storybook checks
- Fix some component types based on the fact that we weren't type-checking our own usage of our components, so their types were kinda jank and difficult to work with.
  - Most notably the Combobox


package.json
```diff
@@ -36,7 +36,7 @@
     "release:stable:npm-failed": "yarn lerna publish from-package -y --loglevel verbose -m 'chore(release): publish' --conventional-commits",
     "release:storybook": "yarn build:storybook && gh-pages -d docs",
     "version": "node ./tools/build/version.js",
-    "clean": "node ./tools/build/clean-repo.js && lerna clean --yes && lerna run clean && yarn && yarn bootstrap",
+    "clean": "node ./tools/build/clean-repo.js && lerna clean --yes && lerna run clean && rm -rf node_modules/ && yarn && yarn bootstrap",
```
SiTaggart, line 39: Had all kinds of issues with babel and its plugins, based on the order in which they are installed and giving variable build behaviors locally and in CI. Being able to clean _everything_ was very useful

package.json
```diff
@@ -174,7 +178,8 @@
   },
   "resolutions": {
     "**/websocket-extensions": "^0.1.4",
-    "**/graphql-playground-middleware-express": "^1.7.18"
+    "**/graphql-playground-middleware-express": "^1.7.18",
+    "@emotion/core": "10.0.28"
```
SiTaggart, line 182: Storybook depends on an older version, and was being used in the component build flow causing issues.

packages/paste-core/components/combobox/src/Combobox.tsx
```diff
@@ -17,7 +17,7 @@ import {ComboboxListbox} from './ComboboxListbox';
 import {ComboboxListboxGroup} from './ComboboxListboxGroup';
 import {ComboboxListboxOption} from './ComboboxListboxOption';
 import {getIndexedItems, getGroupedItems} from './helpers';
-import type {GroupItemsProps, Item as ItemType, ItemProps, ItemsProps, ListBoxProps, ComboboxProps} from './types';
+import type {GroupItemsProps, ItemProps, ItemsProps, ListBoxProps, ComboboxProps} from './types';
```
SiTaggart, line 20: non-used type based on the new per component tsconfigs not having some settings we once relied on like strict

packages/paste-core/components/combobox/src/types.ts
```diff
@@ -7,25 +7,29 @@ import type {InputVariants, InputProps} from '@twilio-paste/input';
...
+  optionTemplate?: OptionTemplateFn<any>;
   groupLabelTemplate?: (groupName: string) => React.ReactNode;
-  selectedItem?: UseComboboxPrimitiveProps<Item>['selectedItem'];
-  inputValue?: UseComboboxPrimitiveProps<Item>['inputValue'];
+  selectedItem?: UseComboboxPrimitiveProps<any>['selectedItem'];
+  inputValue?: UseComboboxPrimitiveProps<any>['inputValue'];
   groupItemsBy?: string;
   variant?: InputVariants;
-  state?: Partial<UseComboboxPrimitiveReturnValue<Item>>;
+  state?: Partial<UseComboboxPrimitiveReturnValue<any>>;
```
SiTaggart, line 32: All of these generics causes consumers to type their supplied items to the `Item` type and it was impossible to get around. By loosening all these you still get type inference, but it allows your items to be any shape you wish.

packages/paste-core/components/disclosure/stories/index.stories.tsx
```diff
@@ -38,7 +39,10 @@ export const ExampleDisclosures: React.FC<{
   );
 };
 
-const useDelayedDisclosureState = ({delay, ...initialState}): DisclosureStateReturn => {
+interface UseDelayedDisclosureStateProps extends DisclosureInitialState {
```
SiTaggart, line 42: Typed completely wrong

tsconfig.json
```diff
@@ -1,10 +1,36 @@
...
-  }
+    "resolveJsonModule": true,
+    "skipLibCheck": true,
+    "sourceMap": true,
+    "suppressImplicitAnyIndexErrors": true,
+    "target": "es6",
+    "typeRoots": ["./@types", "node_modules/@types"],
+    "strict": true,
+  },
+  "exclude": ["node_modules", "docs", "dist", "__testfixtures__"]
```
SiTaggart, line 35: So this is needed by storybook when compiling typescript and type checking. It's a copy of the component based tsconfigs, but with some of the extra settings that were missed from our previous set up

package.json
TheSisb: How do you feel about moving this to a `clean:full` command instead?

package.json
TheSisb: We'll need to make sure to update this if we update styling-library.

tsconfig.json
TheSisb: Can you move these to the storybook folder? the package tsconfigs extend this one, so I think it would be best to keep this root level tsconfig as generic as possible.

TheSisb (commented):
This is a colossal refactor, nicely done 💪

package.json
SiTaggart: I'm in

packages/paste-core/components/combobox/__tests__/index.spec.tsx
```diff
@@ -192,7 +195,7 @@ describe('Combobox', () => {
       render(<ComboboxMock />);
       const renderedLabel = document.querySelector('label');
       const renderedTextbox = screen.getByRole('textbox');
-      expect(renderedLabel.getAttribute('for')).toEqual(renderedTextbox.getAttribute('id'));
+      expect(renderedLabel!.getAttribute('for')).toEqual(renderedTextbox.getAttribute('id'));
```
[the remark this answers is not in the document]
SiTaggart: It's for null checks. `renderedTextbox` _might_ be null according to the typing of `getByRole`. So rather than do a null check to can use the `!` to basically say "I know this is here". It's called a non-null assertion https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#non-null-assertion-operator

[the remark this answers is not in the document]
SiTaggart: I'm being lazy in this instance, because I know there is a textbox on the page for this test

packages/paste-core/components/card/__test__/card.test.tsx
```diff
@@ -21,6 +22,7 @@ describe('Card', () => {
     const tree = renderer
       .create(
         <Theme.Provider theme="console">
+          {/* @ts-ignore intentional margin fail */}
```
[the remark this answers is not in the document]
SiTaggart: I always forget this exists

tsconfig.json
SiTaggart: I can, but my only worry is that the config in the packages now misses the following, not set by default, settings that we used to have for all packages:

```
"esModuleInterop": true,
"forceConsistentCasingInFileNames": true,
"newLine": "lf",
"noUnusedLocals": true,
"noUnusedParameters": true,
"resolveJsonModule": true,
"strict": true,
```

Some of which are pretty important and actually caught some things, like unused type imports for combobox that eslint isn't picking up.

TheSisb (approved):

PR #1112, opened 2021-01-27
chore: changesets

SiTaggart:
Adds changesets to the repo and release process.

The PR bot works, which is nice. It's hard to test if the release works until it's in main and we have a change set to deploy. 🤞 

I have basically crawled GitHub at big repos using changesets. Took a lot of inspiration from their configs. Chakras was pretty handy so hopefully its right first time.

Takes some updates to caches and action versions and applies them to all workflows so they are all running the same configuration. We can look to see how we organise those in the Actions cache ticket, there might be improvements to make, there might not.


package.json
```diff
@@ -30,32 +30,31 @@
     "build:theme-designer": "yarn workspace @twilio-paste/theme-designer build",
     "build:props": "lerna run build:props --ignore @twilio-paste/website --concurrency 2",
     "pre-push": "concurrently \"yarn:lint\" \"yarn:test\" \"yarn:prettier\" \"yarn:type-check\"",
-    "release": "yarn lerna publish -m 'chore(release): publish' --conventional-commits --canary --preid beta --dist-tag next",
-    "release:stable": "yarn lerna publish -m 'chore(release): publish' --conventional-commits",
-    "release:stable:ci": "yarn lerna publish -y --loglevel verbose -m 'chore(release): publish' --conventional-commits",
-    "release:stable:npm-failed": "yarn lerna publish from-package -y --loglevel verbose -m 'chore(release): publish' --conventional-commits",
+    "prerelease": "yarn build && yarn lint && yarn test",
+    "release": "yarn changeset publish",
```
TheSisb, line 34: Can remove the `changeset` script and replace this with `"release": "changeset publish"`

package.json
```diff
@@ -30,32 +30,31 @@
     "build:theme-designer": "yarn workspace @twilio-paste/theme-designer build",
     "build:props": "lerna run build:props --ignore @twilio-paste/website --concurrency 2",
     "pre-push": "concurrently \"yarn:lint\" \"yarn:test\" \"yarn:prettier\" \"yarn:type-check\"",
-    "release": "yarn lerna publish -m 'chore(release): publish' --conventional-commits --canary --preid beta --dist-tag next",
```
TheSisb, line 33: Does the new release script handle publishing the next tag now?

TheSisb (commented):
Changes look reasonable. I might be missing something, but does every package get its own changeset file/folder? What's the workflow like for developers now?

package.json
richbachman: I think we need the `changeset` script so we can generate a changeset for each PR, but I might be missing something there.

richbachman:
We should submit a PR to the changeset repo so we can get on the list of cool projects: https://github.com/atlassian/changesets#cool-projects-already-using-changesets-for-versioning-and-changelogs

richbachman (approved):

package.json
TheSisb: Good point.

SiTaggart:
> Changes look reasonable. I might be missing something, but does every package get its own changeset file/folder? What's the workflow like for developers now?

Nope, all changesets for all packages get placed in the root changeset directory until the release happens when they are deleted. example: https://github.com/chakra-ui/chakra-ui/tree/develop/.changeset

Workflow is just as the readme says; run `yarn changeset` to create a changeset, it has prompts to follow. Commit it with the PR. The GitHub action does the rest.

package.json
SiTaggart: Changesets can handle prereleases but it's a "whole thing". https://github.com/atlassian/changesets/blob/master/docs/prereleases.md

Much more involved and manual.

TheSisb: Can we use the old pre-release script for next and the new script for full releases? It was a really nice feature, would be sad to lose it.

SiTaggart: I guess it should work... it only looks at git tags and npm so the theory is sound. I'll add it back in until we can test it with something

SiTaggart: OK back in under a `release:next` script @TheSisb

PR #1117, opened 2021-01-27
feat(combobox): virtualize ungrouped combobox items

TheSisb:
Virtualizes combobox items that aren't using groupings and a custom optionTemplate.
Addresses: https://github.com/twilio-labs/paste/discussions/841


packages/paste-core/components/combobox/src/styles/ComboboxListboxOption.tsx
```diff
@@ -22,7 +25,17 @@ const VariantStyles: {[key in ComboboxListboxOptionProps['variant']]: BoxStylePr
...
 const ComboboxListboxOption = React.forwardRef<HTMLLIElement, ComboboxListboxOptionProps>(
-  ({children, highlighted, variant = 'default', ...props}, ref) => {
+  ({children, highlighted, variant = 'default', virtualItem, ...props}, ref) => {
+    const virtualItemProps = virtualItem
+      ? {
+          position: 'absolute' as PositionOptions,
+          top: 0,
+          left: 0,
+          width: '100%',
+          height: virtualItem.size,
```
richbachman, line 35: One thing I was having trouble with was this height. Were you able to get it to calculate correctly for items that run into two lines?

TheSisb: Couldnt get it working so only enabling virtualization iff no optionTemplate and no groupItemsBy props

packages/paste-core/components/combobox/src/ComboboxItems.tsx
```diff
@@ -0,0 +1,98 @@
...
+        <li role="presentation" key="total-size" style={{height: rowVirtualizer!.totalSize}} />
+        {rowVirtualizer!.virtualItems.map((virtualItem: VirtualItem) => {
+          const item = items[virtualItem.index];
+          return (
+            <ComboboxListboxOption
+              // @ts-ignore
+              ref={(element) => virtualItem.measureRef(element)}
+              {...getItemProps({item, index: virtualItem.index})}
+              highlighted={highlightedIndex === virtualItem.index}
+              key={UIDSeed(`item-${virtualItem.index}`)}
```
richbachman, line 49: Super nit: should we make `virtualItem.index` a `const` since we're re-using it here?

packages/paste-core/components/combobox/src/ComboboxItems.tsx
```diff
@@ -0,0 +1,98 @@
...
+  // for indexing within groups.
+  const indexedItems = getIndexedItems(items);
+  const groupedItems = getGroupedItems(indexedItems, groupItemsBy);
+  const groupedItemKeys = Object.keys(groupedItems);
+
+  return (
+    <>
+      {groupedItemKeys.map((groupedItemKey) => {
+        const isUncategorized = groupedItemKey === 'undefined';
+        const groupKey = isUncategorized ? 'Uncategorized' : groupedItemKey;
```
richbachman, line 73: I think at one point we talked about prefixing `Uncategorized` just in case a user has that as a group name. Do you think that is still necessary?

packages/paste-core/components/combobox/src/styles/ComboboxListbox.tsx
```diff
@@ -23,12 +23,12 @@ const ComboboxListbox = React.forwardRef<HTMLUListElement, ComboboxListboxProps>
       boxShadow="shadow"
       listStyleType="none"
       margin="space0"
-      maxHeight="size50"
       overflowY="auto"
       padding="space0"
       paddingY="space30"
       position="absolute"
       width="100%"
+      maxHeight="260px"
```
richbachman, line 31: Curious why `260px`?

packages/paste-core/components/combobox/src/styles/ComboboxListboxOption.tsx
```diff
@@ -22,7 +25,16 @@ const VariantStyles: {[key in ComboboxListboxOptionProps['variant']]: BoxStylePr
 };
 
 const ComboboxListboxOption = React.forwardRef<HTMLLIElement, ComboboxListboxOptionProps>(
-  ({children, highlighted, variant = 'default', ...props}, ref) => {
+  ({children, highlighted, variant = 'default', virtualItem, ...props}, ref) => {
+    const virtualItemProps = virtualItem
```
richbachman, line 29: Should this be `virtualItemStyles`?

packages/paste-core/components/combobox/src/styles/ComboboxListbox.tsx
TheSisb: So for infinite scroll to work correctly, we need to make the popout have the height of `numItems * itemHeight`, so that the scrollbar reflects the total set and not the 10 or so `li`s that are actually rendered. Doing so makes the height huge on long lists, so the actual popout is set to a max-height and we provide an inner `li` that is presentational with the total height to fake it.

That said, the maxHeight here was hardcoded as something that looked goodish, but I swapped it to a token now (size30) which is similar in PX value.

richbachman (approved):
Nice work getting this over the line!

packages/paste-core/components/combobox/__tests__/index.spec.tsx
```diff
@@ -289,7 +292,14 @@ describe('Combobox', () => {
   describe('Accessibility', () => {
     it('Should have no accessibility violations', async () => {
       const {container} = render(<ComboboxMock />);
-      const results = await axe(container);
+      const results = await axe(container, {
+        rules: {
+          // Incorrectly thinks our markup is bad for a11y.
+          // False negative
```
SiTaggart, line 298: nit: If you wanted to dig out that axe-core github issue this relates to it might be useful. But not really important

packages/paste-core/components/combobox/src/helpers.tsx
```diff
@@ -1,19 +1,23 @@
...
 };
 
 // Creates a grouped object of items with the original flat array index
-export const getGroupedItems = (items: ItemType[], groupItemsBy: string | undefined): any => {
-  // Should never happen
-  if (groupItemsBy == null) {
-    return null;
-  }
-
+export const getGroupedItems = (items: ItemType[], groupItemsBy: string): any => {
```
SiTaggart, line 21: Do we have tests for these helpers?

SiTaggart (approved):

packages/paste-core/components/combobox/src/helpers.tsx
TheSisb: Yup!

PR #1143, opened 2021-02-01
fix: pull request codesandboxes need their own tsconfig

SiTaggart:
When we use the NextJs template for the PR codesandbox build, the tsconfig was trying to inherit from the monorepo root tsconfig which didn't exist. So it broke.

This provides it's own isolated tsconfig for Codesandbox purposes

### DIFF
--- a/packages/paste-nextjs-template/tsconfig.json
+++ b/packages/paste-nextjs-template/tsconfig.json
@@ -1,10 +1,15 @@
 {
-  "extends": "../../tsconfig.json",
   "compilerOptions": {
     "allowJs": true,
+    "declaration": true,
+    "declarationMap": true,
     "emitDeclarationOnly":  false,
+    "emitDecoratorMetadata": true,
     "esModuleInterop": true,
+    "experimentalDecorators": true,
+    "forceConsistentCasingInFileNames": true,
     "incremental": false,
+    "isolatedModules": true,
     "jsx": "preserve",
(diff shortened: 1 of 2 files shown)


richbachman (approved):

packages/paste-nextjs-template/tsconfig.json
```diff
@@ -1,19 +1,39 @@
...
+    "moduleResolution": "node",
+    "newLine": "lf",
     "noEmit": true,
-    "resolveJsonModule": true
+    "noEmitOnError": false,
+    "noImplicitAny": true,
+    "noResolve": false,
+    "noUnusedLocals": true,
+    "noUnusedParameters": true,
+    "paths": {
```
TheSisb, line 28: I don't think this does anything here, since the path is relative to the root.

packages/paste-nextjs-template/tsconfig.json
```diff
@@ -1,19 +1,39 @@
...
+      "*": ["./@types/*"]
+    },
+    "resolveJsonModule": true,
+    "skipLibCheck": true,
+    "sourceMap": true,
+    "strict": true,
+    "target": "es6",
+    "typeRoots": ["./@types", "node_modules/@types"]
```
TheSisb, line 36: Same here.

TheSisb (approved):

PR #1146, opened 2021-02-02
chore: delay some flakey chromatic snapshots

SiTaggart:
Try and get Chromatic to pass all the time when you don't actually change anything in a component.


TheSisb:
Doesn't look like it worked. :(

I like this PR.

SiTaggart:
Opened a support ticket with Chromatic so we can use this to work through it

SiTaggart:
@TheSisb So I ran these 3 times in a row and everything seems to be passing each time. 

We obviously would see some diffs between main, as the timing of the snapshots is also different on main as the branch. But once the initial diff was accepted on this branch, if they were flakey I would expect to see if fail on re-runs, which it's not.

🤞 Fingers crossed

TheSisb (approved):
:|

richbachman (approved):

PR #1179, opened 2021-02-09
chore: prevent behavior where peerDeps patching causes major bump

TheSisb:
Changesets by default cause a major version bump to packages with peerDeps that minor bump. This prevents that behavior. See: https://github.com/atlassian/changesets/issues/524

**Contributing to Twilio**


SiTaggart (approved):
I don't necessarily know if this will change the peer does being a major bump or not, but worth a go

richbachman (approved):

PR #1232, opened 2021-02-26
fix(checkbox): add required dot to when checkbox is required

richbachman:
- [x] Added the required dot when the checkbox `required` prop is set to `true`
- [x] Added stories for VRT
- [x] Added tests
- [x] Reworked Label RequiredDot to use icon sizing and media object


TheSisb (approved):
Code lgtm

PR #1236, opened 2021-03-02
fix(design-tokens): change SG theme to inherit Default except font

richbachman:
- [x] Inherited everything except `font-family-10` from Default theme for SendGrid theme
- [x] Updated test in `paste-theme` package

Verified with design team that the code font family should be the same as the default theme.


TheSisb (approved):
💯

GiantRobots:
We don't use the SGTheme in SendGrid anymore so this isn't really required.

SiTaggart:
> We don't use the SGTheme in SendGrid anymore so this isn't really required.

This is more a risk mitigation backup plan, just in case someone is super attached to Colfax or a particular screen needs it. It's not really intended for use, and wont be advertised as such

GiantRobots:
Sounds good just wanted y'all to know we migrated off of it!

PR #1251, opened 2021-03-04
fix(input): allow input to work with hook form library

richbachman:
Changes to form components that allow components to work with hook form libraries like: https://github.com/vnguyen94/react-hook-form-paste


richbachman:
@vnguyen94 these are the initial code changes I found. Let me know if there's anything you that I'm missing. Thanks!

vnguyen94 (approved):
🚀

TheSisb (approved):

PR #1287, opened 2021-03-15
chore(toast): add console z-index patch prop

TheSisb:
Adds a prop to relevant Toast components called `__console_patch` which overrides the z-index token to set one that is higher than the masthead in the console product. This patch is temporary, and will be removed in the future.  It is only intended to be used in the Console product, which is why the prop name is described with a two underscore prefix. It should not be used outside of the Console product and doesn't affect our other customers.

**Contributing to Twilio**


.prettierignore
```diff
@@ -5,6 +5,7 @@ dist/
 public/
 __IconList.tsx
 rollup.icon-list.js
+build.icon-list.js
```
gloriliale, line 8: Does including build.icon-list.js here stop it from minify-ing back and forth? I was having an issue with that, so if this solves that issue, I'm very excited haha.

packages/paste-core/components/toast/src/types.ts
```diff
@@ -29,17 +29,25 @@ export const ToastPropTypes = {
 export interface ToastPortalProps extends React.HTMLAttributes<HTMLDivElement> {
   children: NonNullable<React.ReactNode>;
   ref?: any;
+  // FIXME: Overrides token zIndex to fix bug with Console product.
+  __console_patch?: boolean;
```
gloriliale, line 33: I hadn't thought of how this prop would work, and seeing it be a boolean, especially for a temp fix, makes so much sense.

packages/paste-website/src/pages/components/toast/index.mdx
```diff
@@ -487,9 +488,10 @@ If you choose to handle state entirely yourself, you can use `<Toaster />` on it
 
 #### ToastContainer Props
 
-| Prop     | Type        | Description | Default |
-| -------- | ----------- | ----------- | ------- |
-| children | `ReactNode` |             | null    |
+| Prop               | Type        | Description                                     | Default |
+| ------------------ | ----------- | ----------------------------------------------- | ------- |
+| children           | `ReactNode` |                                                 | null    |
+| \_\_console_patch? | `boolean`   | A temporary z-index fix for the Console product | null    |
```
gloriliale, line 494: Yay for documenting our changes! 
Would __console_patch be a prop of Toaster or ToasterContainer (or both?). What's the difference between these two?

gloriliale (approved):

packages/paste-website/src/pages/components/toast/index.mdx
SiTaggart: Both is good. Toaster is the "rite fancy" version which comes with some accessibility and animation and other jazz. 

ToastContainer is the no frills, doesn't do anything apart from shove it in the top corner, predecessor

packages/paste-core/components/toast/src/types.ts
```diff
@@ -29,17 +29,25 @@ export const ToastPropTypes = {
...
 }
 export const ToastPortalPropTypes = {
   children: PropTypes.node.isRequired,
+  // FIXME: Overrides token zIndex to fix bug with Console product.
+  __console_patch: PropTypes.bool,
```
SiTaggart, line 38: Do you need it on `ToastPortal`?

.prettierignore
TheSisb: Yup! That's the intent.

packages/paste-core/components/toast/src/types.ts
TheSisb: We do because the Toaster and ToastContainer ultimately pass this prop down to ToastPortal, where the ZIndex is applied.
image

richbachman (approved):
Nice work!

PR #1307, opened 2021-03-23
fix(button): improve Button handling of link behavior

TheSisb:
- [x] The Button component now displays an animated arrow icon by default for Buttons with props `as="a" href="string"` where `children` is a string. It doesn't add an icon if `children` is not a string type.
  - [x] Only adds icon to  variant `primary` `secondary`. Not to `reset`
- [x] Added similar functionality to Anchor component to handle external links correctly. 
- [x] Added more validation errors and tests for these throws.
  - [x] Throws if link button is disabled or loading
  - [x] Throws if link button isnt of variant `primary` `secondary` or `reset`
  - [x] Throws to inform to use anchor component if `as="a"` and `variant="link"` or `variant="inverse_link`   
- [x] Added stories

### DIFF
--- a/packages/paste-core/components/button/src/index.tsx
+++ b/packages/paste-core/components/button/src/index.tsx
@@ -2,6 +2,9 @@ import * as React from 'react';
 import {Box} from '@twilio-paste/box';
 import {Stack} from '@twilio-paste/stack';
 import {Spinner} from '@twilio-paste/spinner';
+import {secureExternalLink} from '@twilio-paste/anchor';
+import {useSpring, animated} from '@twilio-paste/animation-library';
+import {ArrowForwardIcon} from '@twilio-paste/icons/esm/ArrowForwardIcon';
 import type {
   ButtonProps,
   ButtonSizes,
@@ -21,6 +24,8 @@ import {InverseButton} from './InverseButton';
 import {InverseLinkButton} from './InverseLinkButton';
 import {ResetButton} from './ResetButton';
 
+const AnimatedBox = animated(Box);
+
 // If size isn't passed, come up with a smart default:
 // - 'reset' for variant 'link'
 // - 'icon' if there's 1 child that's an icon
@@ -29,12 +34,12 @@ const getButtonSize = (variant: ButtonVariants, children: React.ReactNode, size?
   let smartSize: ButtonSizes = 'default';
   if (size != null) {
     smartSize = size;
-  } else if (variant === 'link' || variant === 'destructive_link') {
+  } else if (variant === 'link' || variant === 'destructive_link' || variant === 'reset') {
     smartSize = 'reset';
   } else if (React.Children.count(children) === 1) {
     React.Children.forEach(children, (child) => {
       if (React.isValidElement(child)) {
-        // @ts-ignore
+        // @ts-expect-error we know displayName will exist in React
         if (typeof child.type.displayName === 'string' && child.type.displayName.includes('Icon')) {
           smartSize = 'icon';
         }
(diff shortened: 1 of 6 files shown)


packages/paste-core/components/button/src/index.tsx
```diff
@@ -29,12 +34,12 @@ const getButtonSize = (variant: ButtonVariants, children: React.ReactNode, size?
...
   if (size != null) {
     smartSize = size;
-  } else if (variant === 'link' || variant === 'destructive_link') {
+  } else if (variant === 'link' || variant === 'destructive_link' || variant === 'reset') {
     smartSize = 'reset';
   } else if (React.Children.count(children) === 1) {
     React.Children.forEach(children, (child) => {
       if (React.isValidElement(child)) {
-        // @ts-ignore
+        // @ts-expect-error we know displayName will exist in React
```
richbachman, line 42: Nice use of a newer TS feature! If anyone wants to read more about it, heres some info: https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-9.html#-ts-expect-error-comments

packages/paste-core/components/button/src/index.tsx
```diff
@@ -147,20 +177,55 @@ const getButtonComponent = (variant: ButtonVariants): React.FunctionComponent<Di
...
+          rest.onMouseEnter(event);
+        }
+        setHovered(true);
+      }}
+      onMouseLeave={(event) => {
+        if (typeof rest.onMouseLeave === 'function') {
+          rest.onMouseLeave(event);
+        }
+        setHovered(false);
+      }}
```
richbachman, line 228: Performance question. Is this way any more performant than the `useHover` hook thats used in Disclosure? https://github.com/twilio-labs/paste/blob/main/packages/paste-core/components/disclosure/src/utils.ts#L43

That might help answer the question of if we need to pull that out into a hooks library.

richbachman (approved):
Just one open question from me. Good refactor on the prop validation. That will help teach users how we want them to handle buttons.

packages/paste-core/components/button/src/index.tsx
```diff
@@ -54,25 +59,50 @@ const getButtonState = (disabled?: boolean, loading?: boolean): ButtonStates =>
...
+    }
+    if (variant === 'link' || variant === 'inverse_link') {
+      throw new Error(`[Paste: Button] Using Button component as an Anchor. Use the Paste Anchor component instead.`);
+    }
+    if (variant !== 'primary' && variant !== 'secondary' && variant !== 'reset') {
+      throw new Error(`[Paste: Button] <Button as="a"> only works with the following variants: primary or secondary.`);
+    }
+    if (disabled || loading) {
+      throw new Error(`[Paste: Button] <Button as="a"> cannot be disabled or loading.`);
+    }
```
SiTaggart, line 92: We should remember to call this out in the breaking changes as throwing new errors for potentially existing implementations might trip people up who aren't testing thoroughly

packages/paste-core/components/button/src/index.tsx
SiTaggart: Nice, good call out

SiTaggart (commented):
Looking great outside of a decision on potentially using a shared hover hook, and a change set

packages/paste-core/components/button/src/index.tsx
TheSisb: I haven't measured the performance, but my hunch is that they're reasonably equivalent. 
Some differences:
- My implementation used mouseEnter rather than mouseOver, which has a different behavior (https://www.w3schools.com/jquery/tryit.asp?filename=tryjquery_event_mouseenter_mouseover#:~:text=mouseenter%20and%20mouseover.-,The%20mouseover%20event%20triggers%20when%20the%20mouse%20pointer%20enters%20the,moved%20over%20the%20div%20element.)
- the `useHover` hook works with `ref` which may be cleaner for where you need the information about mouseOver on nested components defined in other files or files you don't control. For example, if the `mouseOver` event is not passed to a component that you are consuming, this hook uses ref to add the event via JS rather than react's event system (https://reactjs.org/docs/events.html)

I think both approaches are fine for now.

PR #1310, opened 2021-03-24
feat(design-tokens): token contrast check

SiTaggart:
Create a way to ensure that, for any given theme, the contrast ratio between a pair of design tokens always meets our requirements.

Not every color token should be used on every background token. We as a system dictate that usage. So there is no point in testing all colors on all backgrounds, as the result would just be noisy with a lot of intentional failures.

The PR first uses the Theo meta data on a token to create a system for annotating the color contrast requirements. We can categorize these into 2 categories; text color contrast, ui control color contrast. For any given token we can stipulate whether it has a text color contrast ratio requirement on a certain background or a ui control contrast requirement.

1. This token should be used for text color, on a specific background and should have a contrast ratio of 4.5:1 against the background color.
2. This token should be used as a color of a UI control or graphic, on a specific background and should have a contrast ration of 3:1 against the background color.

We create two categories `text_contrast_pairing` and `uicontrol_contrast_pairing`, under which we list the other color tokens this token should be paired with to create our spec. Example:

```
color-text-inverse-weak:
  value: "{!palette-gray-40}"
  comment: Weak inverse text color for dark backgrounds. Must pass AA color contrast with color-background-body-inverse.
  text_contrast_pairing:
    - color-background-body-inverse

color-text-brand-highlight:
  value: "{!amaranth}"
  comment: Twilio brand red, accessible on large text only.
  uicontrol_contrast_pairing:
    - color-background-body
```

This annotation creates the rules that:

1. `color-text-inverse-weak` when used on `color-background-body-inverse`, must have a color contrast ratio greater than or equal to 4.5:1
2. `color-text-brand-highlight` when used on `color-background-body`, must have a color contrast ration greater than or equal to 3:1

## Part II Theme utils

Create utilities and helpers that enable you to determine if a Paste theme is accessible based on the meta data provided by the token annotations.

Bundle them up in a hook you can use in a site to automatically determine whether the current theme is accessible.

## Part III Unit test

Create a unit test using Jest to assert that all the required color contrast rules for the default theme always pass

## Part IV UI

Create a super simple UI the lists all the token pairings and their contrast ratios, and updates in real time when updating color tokens to show that changes are not breaking the color contrast requirements.


packages/paste-design-tokens/tokens/global/border-color.yml
```diff
@@ -8,24 +8,67 @@ props:
   color-border-dark:
     value: "{!palette-gray-60}"
     comment: Dark border color
+    uicontrol_contrast_pairing:
```
TheSisb, line 11: How did you come up with these pairings and how should we review them? A little confused on how to provide helpful feedback here.

packages/paste-theme-utils/__tests__/themeContrast.spec.ts
```diff
@@ -0,0 +1,23 @@
...
+} from '../src/utils/colorContrastPairingUtils';
+
+const defaultThemeTextColorContrastRatings = getContrastRatingsOfTokensWithTextContrastRequirements(DefaultTheme);
+const defaultThemeUiControlColorContrastRatings = getContrastRatingsOfTokensWithUIControlContrastRequirements(
+  DefaultTheme
+);
+
+describe('Default Theme', () => {
+  describe('Text color contrast ratio for token pairs', () => {
+    test.each(defaultThemeTextColorContrastRatings)('ratio check for %p', (rating) => {
```
TheSisb, line 14: neat

packages/paste-theme-utils/package.json
```diff
@@ -0,0 +1,46 @@
+{
+  "name": "@twilio-paste/theme-utils",
+  "version": "0.0.0",
+  "description": "Paste theme utilities for checking color contrast of Paste themes.",
```
TheSisb, line 4: A few broad questions:

1) Any reason why we didn't put this into either the `design-tokens` or `theme` package?
2) Any plans to add more functionality to this? Right now it's focused on colors as per the description, which makes me wonder if this should be named something like `color-library`
3) Lastly, should this be a "library" package?

packages/paste-theme-utils/src/utils/colorContrastPairingUtils.ts
```diff
@@ -0,0 +1,205 @@
+import ColorCombos from 'color-combos';
+import type {ColorCombo} from 'color-combos';
+import * as DefaultRawTokenJSON from '@twilio-paste/design-tokens/dist/tokens.raw.json';
+import type {DesignToken, DesignTokensJSON} from '@twilio-paste/design-tokens/types';
+import type {GenericTokenShape, AllGenericTokens} from '@twilio-paste/theme';
```
TheSisb, line 5: Seeing this import from both design-tokens and theme reinforces wanting to put it in the design-tokens package, personally.

packages/paste-theme-utils/src/utils/colorContrastPairingUtils.ts
```diff
@@ -0,0 +1,205 @@
...
+ * by name only
+ *
+ * @param {Partial<GenericTokenShape>} tokens all the design tokens and their values in the current theme
+ * @return {*}  {AllGenericTokens} flatted object of design tokens
+ */
+export const flattenCategorizedTokens = (tokens: Partial<GenericTokenShape>): AllGenericTokens => {
+  const tokenCategories = Object.keys(tokens) as [keyof GenericTokenShape];
+  // build a new object based on raw design tokens, by taking the tokens from each category and just spreading them on a simple object
+  return tokenCategories.reduce((flatTheme, category) => {
+    return tokens[category] ? {...flatTheme, ...tokens[category]} : flatTheme;
```
TheSisb, line 50: Code may be simplified with something like
```js
let flatTheme = {};
Object.values(tokens).forEach(value => {
  flatTheme = {...flatTheme, ...value};
});
return flatTheme
```

packages/paste-theme/src/generateThemeFromTokens.ts
```diff
@@ -16,22 +16,38 @@ import {
...
-  textColors: Partial<{[key in keyof typeof TextColors]: any}>,
-  zIndices: Partial<{[key in keyof typeof ZIndices]: any}>
-): GenericThemeShape => {
+  sizings: {[key in keyof typeof Sizings]: any};
+  spacings: Partial<{[key in keyof typeof Spacings]: any}>;
+  textColors: Partial<{[key in keyof typeof TextColors]: any}>;
+  zIndices: Partial<{[key in keyof typeof ZIndices]: any}>;
+}
+
+export const generateThemeFromTokens = ({
```
TheSisb, line 36: Such a welcome change xD

packages/paste-token-contrast-checker/jestTransformer.js
```diff
@@ -0,0 +1,10 @@
+import babel from '@babel/core';
+
+module.exports = {
```
TheSisb, line 3: Why do we need this here?

packages/paste-design-tokens/tokens/global/border-color.yml
SiTaggart: Combination of kinda knowing what the system should do, whats used in the components currently, and best guess.

The one I have commented out are one that I think _should_ work but are currently failing too.

It's not definitive, but we can iterate on it. _Especially_ when we start thinking more about different background cases, as we really only cater for "on background body", they'll evolve

packages/paste-token-contrast-checker/jestTransformer.js
SiTaggart: Default create-react-app bootstrapping. Can probably delete some of these files actually

packages/paste-theme-utils/package.json
SiTaggart: OK, so yeah, I've been really struggling to find a home for these. It all started off in Design Tokens. Then I wanted it to be sharable with theme designer, and therefore based on `theme`, so it moved to Theme and the hook got created.

That added 12kb to theme for everyone, for no good reason. Why should the theme provider package suddenly ad 12k to your app because we want to test some themes in another internal app.

So then I pulled out of theme into it's own package, and I still kinda don't like it.

The fact that the utils require `GenericTokenShape` and `AllGenericTokens` from theme is problematic, as theme is a dependent of design tokens, so they can't live in Design tokens right now due to the circular dep. You could argue that those two should actually live in Design tokens though. 

I'm leaning towards utils living in Design Tokens, and the hook living in Theme as it's a hook to rate the current theme from theme provider.

packages/paste-token-contrast-checker/README.md
```diff
@@ -0,0 +1,55 @@
+# Design Token Contrast Checker
+
+Tool to run when you are checking the color contrast ration of a design tokens for a given theme, in real time.
```
richbachman, line 3:
```suggestion
Tool to run when you are checking the color contrast ratio of a design tokens for a given theme, in real time.
```

packages/paste-token-contrast-checker/src/components/Form.tsx
```diff
@@ -0,0 +1,96 @@
...
+          onChange={(newValue) => {
+            setRadioValue(newValue);
+          }}
+        >
+          <Radio
+            id={inputID('radio1')}
+            value="1"
+            helpText={
+              <Text as="span" color="currentColor">
+                This is some legend text with a <Anchor href="http://paste.twilio.com">link</Anchor>.
```
richbachman, line 67: Should this read: This is some help text...?

packages/paste-theme-utils/package.json
richbachman: > I'm leaning towards utils living in Design Tokens, and the hook living in Theme as it's a hook to rate the current theme from theme provider.

That feels like a good solution.

packages/paste-design-tokens/tokens/global/border-color.yml
richbachman: Just making sure I understand this correctly. This is basically saying that this border token should have a passing contrast ratio score when used on certain background color tokens. Right?

For the ones that _should_ work, do you have any insight on why they're not working?

If my understanding is correct, this feels like a good first pass. I'm sure these will continue to evolve and need to be iterated on.

SiTaggart: You are spot on. `color-border-dark` has ui control contrast ratio requirements against the following colors.

> For the ones that should work, do you have any insight on why they're not working?

We picked the wrong colors? 🤷🏼

packages/paste-design-tokens/contrast-rating/utils.ts
```diff
@@ -0,0 +1,204 @@
...
+import * as DefaultRawTokenJSON from '../dist/tokens.raw.json';
+import type {DesignToken, DesignTokensJSON, TokenPairContrastRating} from '../types';
+
+function toCamelCase(text: string): string {
+  return text.replace(/-\w/g, clearAndUpper);
+}
+
+function clearAndUpper(text: string): string {
+  return text.replace(/-/, '').toUpperCase();
+}
```
richbachman, line 13: Lodash has a camelCase function we could use here: https://lodash.com/docs/4.17.15#camelCase

SiTaggart: Nice catch, replaced

packages/paste-design-tokens/types/index.ds.ts
```diff
@@ -1,2 +1,2 @@
 declare module 'gulp-theo';
-declare module 'gulp-restart';
+export {};
```
TheSisb, line 2: What does this do? 🤔

packages/paste-design-tokens/build.js
```diff
@@ -0,0 +1,34 @@
+const esbuild = require('esbuild');
+
+// Resolve paths to dist files in compiled assets
+const DesignTokenDistResolver = {
+  name: 'DesignTokenDistResolver',
+  setup(build) {
+    // get any paths to dist and replace then as they become local
```
TheSisb, line 7: Can you expand on this comment please? I'm still unclear as to what this is trying to fix.

packages/paste-design-tokens/package.json
```diff
@@ -9,14 +9,23 @@
...
   "scripts": {
-    "build": "yarn clean && gulp",
+    "build": "yarn clean && gulp && yarn build:js && tsc",
+    "build:js": "NODE_ENV=production node build.js",
     "build:dev": "yarn build",
-    "build:watch": "yarn clean && gulp dev",
+    "build:watch": "gulp dev",
     "clean": "rm -rf ./dist",
     "tokens": "gulp default",
     "test": "jest"
```
TheSisb, line 23: Let's add the `"tsc": "tsc"` command here so that when we run `yarn type-check` in tests, it also runs the typechecker in this project: https://github.com/twilio-labs/paste/blob/main/package.json#L54

packages/paste-design-tokens/build.js
SiTaggart: So the javascript utilities this is trying to compile and place in the `dist/` folder to publish, are dependent on the compiled design token assets, which also live in the `dist/` folder. But before being compiled, the utils live outside of `dist/`.

Because the pre compiled Design Tokens are in yaml, we can't import those directly with a relative path, so we have to import the json they produce via the `dist/` compiled tokens.

Alls well and good locally, as the typescript compiles and doesn't complain when trying to run and build it as everything is local to the utils source, including the path to `./dist/what-ever.json`.

Problem comes when you distribute the utils with the design tokens in the `dist/` folder. The compiled utils now live _next to_ the design token JSON is needs, and trying to import them from `./dist/what-ever.json` won't work. All this does is remove the `dist/` and make the compile javascript have a relative path the json it needs that is now a sibling.

packages/paste-design-tokens/package.json
SiTaggart: Fair

packages/paste-design-tokens/types/index.ds.ts
SiTaggart: It's based on a TSconfig rule we have for all the other packages. I think for this package it's safe to turn off isolatedModules, so I can get rid of the empty export and stop typescript complaining.

packages/paste-design-tokens/build.js
TheSisb: Thanks

TheSisb (approved):

PR #1345, opened 2021-04-07
fix(avatar): make size optional and have default size

gloriliale:
- Fixed the size prop to make it optional
- Created a default variable for the size's default value to use in Avatar and AvatarContents components.


nkrantz (approved):
Looks great!

TheSisb (approved):

richbachman (approved):

PR #1346, opened 2021-04-07
fix: project lint and type-check commands now work locally

TheSisb:
I attempted to run `yarn lint` locally, but that caused a heap overflow error.  This thread https://github.com/typescript-eslint/typescript-eslint/issues/1551 pointed me towards improving the `.eslintignore` and `tsconfig.json` files, specifically around excludes, to reduce the amount of processed files and solve the error. I then took the liberty of cleaning up the `.prettierignore` file as well. Now all three files are fairly consistent on ignore paths.

Next up, running `yarn type-check` was only checking the website and the paste nextjs-template folders. I realized that we had changed the `type-check` command in each project folder to `tsc` during the esbuild tooling migration, so I amended the relevant package.json files to make the `yarn type-check` command work everywhere again.

Lastly, I removed rollup related dependencies from the root package.json file which are no longer in use.

Edit: the new chromatic version was crashing on how we imported images (https://github.com/twilio-labs/paste/pull/1346/checks?check_run_id=2290782702) so I modified the Avatar story to simplify the image prop.


nkrantz (approved):
Looks really good! 2 questions:

- What were we using rollup for and why is it no longer needed?
- Will the `.eslintignore` and `tsconfig` changes speed up those github checks? (not the vercel one which seems to be the biggest time suck)

richbachman:
Changes look good. Any luck with the Avatar storybook errors?

TheSisb:
> Looks really good! 2 questions:
> 
> * What were we using rollup for and why is it no longer needed?
> * Will the `.eslintignore` and `tsconfig` changes speed up those github checks? (not the vercel one which seems to be the biggest time suck)

Rollup (https://rollupjs.org/guide/en/) was previously used for our builds. We swapped over to ESbuild (https://esbuild.github.io/) earlier this year, however some references remained by accident. I removed the unneeded references here.

I wasn't able to capture a significant performance boost in my runs, so I think it's less about speed and more about memory consumption.

TheSisb:
> Changes look good. Any luck with the Avatar storybook errors?

I think so, see https://github.com/twilio-labs/paste/pull/1346/commits/885da2070ad04a60ae725982d04b82b83060373f

PR #1353, opened 2021-04-08
fix(docs): add prominence to font section of engineering guide

richbachman:
Moved font section up on the engineering guidelines page. Also added a callout about using the CDN being the most optimized way.

### DIFF
--- a/packages/paste-website/src/pages/getting-started/engineering/index.mdx
+++ b/packages/paste-website/src/pages/getting-started/engineering/index.mdx
@@ -172,6 +172,45 @@ const App = () => (
 );
 ```
 
+### Fonts
+
+#### Default theme
+
+Fonts for the default theme are available via the Twilio CDN and published from an [internal git repository](https://code.hq.twilio.com/DSYS/paste-fonts).
+
+The **best and most performant way** to load the fonts is to include the following snippet in the `<head />` of your web page.
+
+```
+<link rel="preconnect" href="https://assets.twilio.com" crossorigin />
+<link rel="stylesheet" href="https://assets.twilio.com/public_assets/paste-fonts/main-1.2.0/fonts.css">
+```
+
+<Callout>
+  <CalloutText>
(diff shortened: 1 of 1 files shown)


packages/paste-website/src/pages/getting-started/engineering/index.mdx
```diff
@@ -172,6 +172,45 @@ const App = () => (
...
+
+<Callout>
+  <CalloutText>
+    Loading the fonts directly from the Twilio CDN will provide the fastest download time, therefore giving you the most
+    optimized page load experience.
+  </CalloutText>
+</Callout>
+
+Alternatively, Paste will automatically load the fonts via JavaScript, so long as you wrap your application with the `Theme.Provider`
+and select the `default` theme. Though, you should note that this is not the most performant way.
```
gloriliale, line 196: This is much better, more prominent placement, hooray! I think it'll really up the chances that people follow these guidelines.

To me, the last sentence should be more specific, like "Note: this will result in slower download times than including the fonts in the `<head />`."

Also, we never explicitly say "we recommend" putting the fonts in the head. Is this on purpose, or is it something worth sneaking into a sentence in the callout text, for example?

gloriliale (approved):

packages/paste-website/src/pages/getting-started/engineering/index.mdx
richbachman: Both good things to call out. I like adding the "we recommend" text to the callout. Thanks for the suggestions!

packages/paste-website/src/pages/getting-started/engineering/index.mdx
```diff
@@ -172,6 +172,45 @@ const App = () => (
 );
 ```
 
+### Fonts
```
TheSisb, line 175: Let's rename this header to something like `How to load the correct fonts`

TheSisb (approved):

PR #1363, opened 2021-04-14
fix(button): conditionally use external link icon for anchor buttons

TheSisb:
Also updates the docs

### DIFF
--- a/packages/paste-core/components/button/stories/index.stories.tsx
+++ b/packages/paste-core/components/button/stories/index.stories.tsx
@@ -133,12 +133,12 @@ export const ButtonAsAnchor = (): React.ReactNode => {
     <>
       <Box padding="space30">
         <Button as="a" href="https://twilio.com" variant="primary">
-          Automatically adds link icon
+          Automatically adds external link icon
         </Button>
       </Box>
       <Box padding="space30">
-        <Button as="a" href="https://twilio.com" variant="secondary">
-          Automatically adds link icon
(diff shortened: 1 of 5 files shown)


packages/paste-core/components/button/stories/index.stories.tsx
```diff
@@ -133,12 +133,12 @@ export const ButtonAsAnchor = (): React.ReactNode => {
...
       <Box padding="space30">
         <Button as="a" href="https://twilio.com" variant="primary">
-          Automatically adds link icon
+          Automatically adds external link icon
         </Button>
       </Box>
       <Box padding="space30">
-        <Button as="a" href="https://twilio.com" variant="secondary">
-          Automatically adds link icon
+        <Button as="a" href="/" variant="secondary">
```
nkrantz, line 140: Would `href="/"` classify as an external link now, since the button component is checking for any `href` prop that is a string?

nkrantz (approved):
Looks good, nice fix

richbachman (approved):

packages/paste-core/components/button/stories/index.stories.tsx
richbachman: @nkrantz there's some other automagic stuff going on here. Button is using the `isExternalUrl` function from Anchor, which runs the string through regex to see if it starts with http or https. You can see that here: https://github.com/twilio-labs/paste/blob/main/packages/paste-core/components/anchor/src/index.tsx#L13-L16

PR #1367, opened 2021-04-16
chore(avatar): allow for only icon OR src prop

nkrantz:
- Added union type to AvatarProps to only allow for either `src` OR `icon` prop
- Added check for presence of both `src` AND `icon` to catch javascript users
- Fixed Avatar.propTypes and <AvatarContents /> to satisfy typescript
- Updated docs to reflect changes

**Contributing to Twilio**

### DIFF
--- a/.changeset/fifty-roses-brake.md
+++ b/.changeset/fifty-roses-brake.md
@@ -1,0 +1,7 @@
+---
+'@twilio-paste/avatar': patch
+'@twilio-paste/website': patch
+'@twilio-paste/core': patch
+---
+
+Restricted Avatar propTypes so that users cannot add both `src` and `icon` props to Avatar, only one or the other.
--- a/.changeset/fifty-roses-brake.md
+++ b/.changeset/fifty-roses-brake.md
@@ -1,6 +1,5 @@
 ---
 '@twilio-paste/avatar': patch
-'@twilio-paste/website': patch
 '@twilio-paste/core': patch
 ---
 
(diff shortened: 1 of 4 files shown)


.changeset/fifty-roses-brake.md
```diff
@@ -0,0 +1,7 @@
+---
+'@twilio-paste/avatar': patch
+'@twilio-paste/website': patch
```
TheSisb, line 3: Hmm we typically don't patch the website. Do you think we should start?

packages/paste-website/src/pages/components/avatar/index.mdx
```diff
@@ -138,6 +143,11 @@ Provide the Avatar with a source path via the `src` prop to render the Avatar as
 
 Provide the Avatar with an `icon` prop to display an icon. You must import the icon before passing it to the `icon` prop. You **must** still provide a `name` prop. The icon must be a <a href="https://paste.twilio.design/icons/">Paste Icon</a>.
 
+<Callout>
```
TheSisb, line 146: I think with the new typings we can probably remove these callouts now. Previously we added a note because the behavior was allowed and possibly unexpected, but now the code will throw an error informing the developer.

SiTaggart: That's a fair observation. Is it worth keeping for those who don't use TS and ignore propTypes?

I'm easy, either way

.changeset/fifty-roses-brake.md
nkrantz: Ah good catch - I don't think patching the website is necessary. Just an oversight.

packages/paste-website/src/pages/components/avatar/index.mdx
nkrantz: My instinct is to leave them there since I can't think of any downsides to having repeated information - what do you think? It seems like it could be more beneficial for an engineer to notice the rule in the docs and avoid having to encounter the error at all, rather than debugging the error.

richbachman (approved):

PR #1368, opened 2021-04-20
feat(pagination): add pagination package

richbachman:
- [x] Compositional components
- [x] Tests
- [x] Stories
- [x] Added inherit to `border-color` options
- [x] Added `font-variant-numeric` prop to `Box`

This doesn't include the `DirectionalPagination` or `NumberPagination` components. Let's get these compositional components out in the wild and see if those two components are needed. If so, we'll build them at that time.

### DIFF
--- a/packages/paste-core/components/pagination/__tests__/index.spec.tsx
+++ b/packages/paste-core/components/pagination/__tests__/index.spec.tsx
@@ -1,0 +1,174 @@
+import * as React from 'react';
+import {render, fireEvent} from '@testing-library/react';
+// @ts-ignore typescript doesn't like js imports
+import axe from '../../../../../.jest/axe-helper';
+import {PaginationArrow, PaginationNumber} from '../src';
+import {ArrowsPageLabelExample, NumbersExample, NumbersPageLabelExample} from '../stories/index.stories';
+
+describe('Pagination', () => {
+  it('should render a list of pagination numbers with a page label', () => {
+    const {getByText} = render(<NumbersPageLabelExample />);
+    expect(getByText('Page 5 of 10')).toBeDefined();
+  });
+
+  it('should render pagination arrow with correct previous aria-label', () => {
+    const {getByTitle} = render(<PaginationArrow label="Go to previous page" variant="back" />);
+    expect(getByTitle('Go to previous page')).toBeDefined();
+  });
+
+  it('should render pagination arrow with correct next aria-label', () => {
+    const {getByTitle} = render(<PaginationArrow variant="forward" label="Go to next page" />);
+    expect(getByTitle('Go to next page')).toBeDefined();
+  });
+
+  it('should render pagination arrow with a hidden visible label', () => {
+    const {getByText} = render(<PaginationArrow label="Go to previous page" variant="back" visibleLabel="Previous" />);
+    expect(getByText('Previous')).toBeDefined();
+    expect(getByText('Previous').getAttribute('aria-hidden')).toBeTruthy();
+  });
+
+  it('should render pagination arrow as a button', () => {
+    const {getByRole} = render(<PaginationArrow label="Go to previous" variant="back" />);
+    expect(getByRole('button')).toBeDefined();
+  });
+
+  it('should render pagination arrow as an anchor', () => {
+    const {getByRole} = render(<PaginationArrow as="a" href="#" label="Go to previous" variant="back" />);
+    expect(getByRole('link')).toBeDefined();
+    expect(getByRole('link').getAttribute('href')).toEqual('#');
+  });
+
+  it('should render arrow pagination with correct aria-label', () => {
+    const {getByRole} = render(<ArrowsPageLabelExample />);
+    expect(getByRole('navigation').getAttribute('aria-label')).toBe('pagination navigation');
+  });
+
+  it('should render pagination number as a button', () => {
+    const {getByRole} = render(<PaginationNumber label="Go to page 10">10</PaginationNumber>);
+    expect(getByRole('button')).toBeDefined();
+  });
+
+  it('should render pagination number as an anchor', () => {
+    const {getByRole} = render(
+      <PaginationNumber label="Go to page 10" as="a" href="#">
+        10
+      </PaginationNumber>
+    );
+    expect(getByRole('link')).toBeDefined();
+    expect(getByRole('link').getAttribute('href')).toEqual('#');
+  });
+
+  it('should render pagination number with correct aria-label', () => {
+    const {getByRole} = render(<PaginationNumber label="Go to page 10">10</PaginationNumber>);
+    expect(getByRole('button').getAttribute('aria-label')).toBe('Go to page 10');
+  });
+
+  it('should render pagination number with a hidden visible child', () => {
+    const {getByText} = render(<PaginationNumber label="Go to page 10">10</PaginationNumber>);
+    expect(getByText('10').getAttribute('aria-hidden')).toBeTruthy();
+  });
+
+  it('should render pagination number with correct aria-current', () => {
+    const {getByRole} = render(
+      <PaginationNumber label="Go to page 10" isCurrent>
+        10
+      </PaginationNumber>
+    );
+    expect(getByRole('button').getAttribute('aria-current')).toBeTruthy();
+  });
+
+  describe('pagination arrow event handlers', () => {
+    it('Should call the appropriate event handlers', () => {
+      const onClickMock: jest.Mock = jest.fn();
+      const onMouseDownMock: jest.Mock = jest.fn();
+      const onMouseUpMock: jest.Mock = jest.fn();
+      const onMouseEnterMock: jest.Mock = jest.fn();
+      const onMouseLeaveMock: jest.Mock = jest.fn();
+      const onFocusMock: jest.Mock = jest.fn();
+      const onBlurMock: jest.Mock = jest.fn();
+
+      const {getByTestId} = render(
+        <PaginationArrow
+          variant="forward"
+          label="Go to next"
+          data-testid="arrow-button"
+          onClick={onClickMock}
+          onMouseDown={onMouseDownMock}
+          onMouseUp={onMouseUpMock}
+          onMouseEnter={onMouseEnterMock}
+          onMouseLeave={onMouseLeaveMock}
+          onFocus={onFocusMock}
+          onBlur={onBlurMock}
+        >
+          Previous
+        </PaginationArrow>
+      );
+
+      fireEvent.click(getByTestId('arrow-button'));
+      expect(onClickMock).toHaveBeenCalledTimes(1);
+      fireEvent.mouseDown(getByTestId('arrow-button'));
+      expect(onMouseDownMock).toHaveBeenCalledTimes(1);
+      fireEvent.mouseUp(getByTestId('arrow-button'));
+      expect(onMouseUpMock).toHaveBeenCalledTimes(1);
+      fireEvent.mouseEnter(getByTestId('arrow-button'));
+      expect(onMouseEnterMock).toHaveBeenCalledTimes(1);
+      fireEvent.mouseLeave(getByTestId('arrow-button'));
+      expect(onMouseLeaveMock).toHaveBeenCalledTimes(1);
+      fireEvent.focus(getByTestId('arrow-button'));
+      expect(onFocusMock).toHaveBeenCalledTimes(1);
+      fireEvent.blur(getByTestId('arrow-button'));
+      expect(onBlurMock).toHaveBeenCalledTimes(1);
+    });
+  });
+
+  describe('pagination number event handlers', () => {
+    it('Should call the appropriate event handlers', () => {
+      const onClickMock: jest.Mock = jest.fn();
+      const onMouseDownMock: jest.Mock = jest.fn();
+      const onMouseUpMock: jest.Mock = jest.fn();
+      const onMouseEnterMock: jest.Mock = jest.fn();
+      const onMouseLeaveMock: jest.Mock = jest.fn();
+      const onFocusMock: jest.Mock = jest.fn();
+      const onBlurMock: jest.Mock = jest.fn();
+
+      const {getByTestId} = render(
+        <PaginationNumber
+          label="Go to page 1"
+          data-testid="number-button"
+          onClick={onClickMock}
+          onMouseDown={onMouseDownMock}
+          onMouseUp={onMouseUpMock}
+          onMouseEnter={onMouseEnterMock}
+          onMouseLeave={onMouseLeaveMock}
+          onFocus={onFocusMock}
+          onBlur={onBlurMock}
+        >
+          1
+        </PaginationNumber>
+      );
+
+      fireEvent.click(getByTestId('number-button'));
+      expect(onClickMock).toHaveBeenCalledTimes(1);
+      fireEvent.mouseDown(getByTestId('number-button'));
+      expect(onMouseDownMock).toHaveBeenCalledTimes(1);
+      fireEvent.mouseUp(getByTestId('number-button'));
+      expect(onMouseUpMock).toHaveBeenCalledTimes(1);
+      fireEvent.mouseEnter(getByTestId('number-button'));
+      expect(onMouseEnterMock).toHaveBeenCalledTimes(1);
+      fireEvent.mouseLeave(getByTestId('number-button'));
+      expect(onMouseLeaveMock).toHaveBeenCalledTimes(1);
+      fireEvent.focus(getByTestId('number-button'));
+      expect(onFocusMock).toHaveBeenCalledTimes(1);
+      fireEvent.blur(getByTestId('number-button'));
+      expect(onBlurMock).toHaveBeenCalledTimes(1);
+    });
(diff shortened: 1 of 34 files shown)


packages/paste-core/components/pagination/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,247 @@
...
+          </PaginationNumbers>
+          <PaginationArrow direction="right">Next</PaginationArrow>
+        </PaginationItems>
+      </Pagination>
+    );
+  };
+
+  it('should render a list of pagination items', () => {
+    const {asFragment} = render(<MockPagingationItems />);
+    expect(asFragment()).toMatchSnapshot();
```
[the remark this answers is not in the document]
richbachman: Yep. We are, but a screenshot made sense here because I'm testing to see the output of `PaginationItems`: the cloned flexed children.

packages/paste-core/components/pagination/src/PaginationArrow.tsx
```diff
@@ -0,0 +1,107 @@
...
+      ...props
+    },
+    ref
+  ) => {
+    handleLinkValidation({as, href});
+    return (
+      <Box
+        {...safelySpreadBoxProps(props)}
+        ref={ref}
+        as={as}
```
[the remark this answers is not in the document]
richbachman: Dig a little digging, and see the usefulness being mentioned more in the SEO realm. Mixed reviews on whether its actually worth adding. Maybe we just include some documentation guidance about that?

packages/paste-core/components/pagination/__tests__/index.spec.tsx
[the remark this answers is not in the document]
SiTaggart: > because I'm testing to see the output of PaginationItems

What exactly are you testing?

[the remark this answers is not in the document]
richbachman: @SiTaggart  I'm verifying the children are cloned, and styles are added. I guess that could also be done with VRT though, i.e. if the children are not inline something is broken.

packages/paste-core/components/pagination/src/PaginationArrow.tsx
```diff
@@ -0,0 +1,107 @@
...
+            borderWidth="borderWidth10"
+            display="flex"
+            height="sizeSquare80"
+            justifyContent="center"
+            width="sizeSquare80"
+          >
+            <ArrowForwardIcon decorative={false} display="inline-flex" title={nextArrowLabel} />
+          </Box>
+        ) : null}
+      </Box>
```
SiTaggart, line 96: This nested JSX is a bit on the wild side. I'd be tempted to pull them out into individual components. For example the boxes wrapping the arrow icons are identical.

I think I'd prefer to see the contents of the PaginationArrow JSX be a composition of styled components that just take children.

packages/paste-core/components/pagination/src/PaginationItems.tsx
```diff
@@ -0,0 +1,43 @@
...
+  return (
+    <Box {...safelySpreadBoxProps(props)} ref={ref} as="ul" display="inline-flex" margin="space0" padding="space0">
+      {validChildren.map((child, index) => {
+        return (
+          <Box
+            key={keySeed(`pagination-items-${index}`)}
+            as="li"
+            listStyleType="none"
+            margin="space0"
+            marginRight={childrenCount !== index + 1 ? 'space90' : 'space0'}
```
SiTaggart, line 26: is this some kind of first child, last child conditional styles?

packages/paste-core/components/pagination/__tests__/index.spec.tsx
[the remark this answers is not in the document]
SiTaggart: Hmmm I'm not seeing where you clone anything, what am I missing?

packages/paste-core/components/pagination/src/PaginationLabel.tsx
```diff
@@ -0,0 +1,28 @@
...
+      textAlign="center"
+    >
+      {children}
+    </Text>
+  );
+});
+
+PaginationLabel.displayName = 'PaginationLabel';
+
+if (process.env.NODE_ENV === 'development') {
```
SiTaggart, line 24: We totally don't need these checks. We don't strip them out on compile and PropTypes already doesn't run them in production mode.

packages/paste-core/components/pagination/src/PaginationNumbers.tsx
```diff
@@ -0,0 +1,49 @@
...
+              <Box
+                key={keySeed(`pagination-numbers-${index}`)}
+                as="li"
+                listStyleType="none"
+                margin="space0"
+                marginRight={childrenCount !== index + 1 ? 'space30' : 'space0'}
+                padding="space0"
+              >
+                {child}
+              </Box>
```
SiTaggart, line 33: Looks like the ul and li styling of PaginationNumbers and PaginationItems is shared. Maybe abstract them rather than dupe across files?

packages/paste-core/components/pagination/src/PaginationArrow.tsx
```diff
@@ -0,0 +1,107 @@
...
+
+const PaginationArrow = React.forwardRef<HTMLButtonElement, PaginationArrowProps>(
+  (
+    {
+      as = 'button',
+      children,
+      direction = 'left',
+      href,
+      nextArrowLabel = 'Go to next page',
+      previousArrowLabel = 'Go to previous page',
```
SiTaggart, line 18: I could technically set `nextArrowLabel` **and** `previousArrowLabel` and I probably shouldn't be able to

packages/paste-core/components/pagination/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,247 @@
...
+            </PaginationNumber>
+            <PaginationNumber aria-label="Go to page 6">6</PaginationNumber>
+            <PaginationEllipsis aria-label="Collapsed next pages" />
+            <PaginationNumber aria-label="Go to page 10">10</PaginationNumber>
+          </PaginationNumbers>
+          <PaginationArrow direction="right">Next</PaginationArrow>
+        </PaginationItems>
+      </Pagination>
+    );
+  };
```
SiTaggart, line 68: Do you reckon it's possible to use some of the stories as Mock components, instead of creating these?

packages/paste-core/components/pagination/package.json
```diff
@@ -0,0 +1,55 @@
+{
+  "name": "@twilio-paste/pagination",
+  "version": "0.0.0",
+  "category": "navigation",
+  "status": "beta",
```
SiTaggart, line 5: Let's go straight to production.

SiTaggart (commented):
There are some alignment issues in Safari on this one I'm afraid 
image

packages/paste-core/components/pagination/src/PaginationArrow.tsx
SiTaggart: On further thought we might need to think a little harder on the API of this one.

Right now if I set 

```
<PaginationArrow nextArrowLabel="Go to next page">Next</PaginationArrow>
```

The accessible text is "Go to nextpageNext". So not only do we have duplication for the consumer but also for the customer.

The fact I can also do

```
<PaginationArrow nextArrowLabel="Go to next page" previusArrowLabel="Go to previous page">Next</PaginationArrow>
```

Feels weird.

First and foremost we need to think of the customer experience to assistive tech users. How can we tidy this API, and make clear button text for the customer?

packages/paste-core/components/pagination/src/PaginationItems.tsx
richbachman: Yeah. Same thing we're doing with Stack. I'm mapping through the children, wrapping them in an `li` with a `marginRight`, then setting the `marginRight` of the last `li` to `space0`.

packages/paste-core/components/pagination/__tests__/index.spec.tsx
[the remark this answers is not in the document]
richbachman: Sorry. I used "clone" incorrectly here. I'm actually mapping through the children, then adding a `marginRight` to all except the last one.

packages/paste-core/components/pagination/src/PaginationArrow.tsx
richbachman: I've refactored PaginationArrow. Let me know what you think.

packages/paste-core/components/pagination/src/PaginationArrow.tsx
richbachman: Included a fix for this in the PaginationArrow refactor.

richbachman:
> There are some alignment issues in Safari on this one I'm afraid
> image

@SiTaggart looks like Safari adds a 2px border to buttons. Similar on our Button component. I can go ahead and fix it there as well.

packages/paste-core/components/pagination/__tests__/index.spec.tsx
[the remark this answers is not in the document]
richbachman: Removing these since VRT will handle any issues with the margins changing.

packages/paste-core/components/pagination/__tests__/index.spec.tsx
richbachman: Yup. I refactored the stories so they can be used in tests and elsewhere.

richbachman:
> looks like Safari adds a 2px margin to buttons. Similar on our Button component. I can go ahead and fix it there as well.

Actually looks like Safari is adding the 2px margin to all inputs, so I'll create a ticket to fix those.

packages/paste-core/components/pagination/src/PaginationItems.tsx
SiTaggart: Does `_firstChild` `_lastChild` not work? Did we implement it? Are there too many wrappers?

packages/paste-core/components/pagination/src/PaginationItems.tsx
```diff
@@ -0,0 +1,42 @@
...
+import type {PaginationItemsProps} from './types';
+import {PaginationItemsPropTypes} from './proptypes';
+
+const PaginationItems = React.forwardRef<HTMLUListElement, PaginationItemsProps>(({children, ...props}, ref) => {
+  const [validChildren] = React.useMemo(
+    () => [
+      React.Children.toArray(children).filter((child) => React.isValidElement(child) || typeof child === 'string'),
+    ],
+    [children]
+  );
```
SiTaggart, line 14: How much of this do you thing we _really_ need?

SiTaggart (approved):
Outside of my only remaining question on whether or not we need to validate the children of PaginationItems and numbers (which is not a blocker), I think this is great!! 🥇

packages/paste-core/components/pagination/src/Pagination.tsx
```diff
@@ -0,0 +1,18 @@
...
+  return (
+    <Box {...safelySpreadBoxProps(props)} ref={ref} aria-label={label} as="nav" role="navigation">
+      {children}
+    </Box>
+  );
+});
+
+Pagination.displayName = 'Pagination';
+
+Pagination.propTypes = PaginationPropTypes;
```
TheSisb, line 16: Just a lil curious why you opted to import these rather than define them inline?

packages/paste-core/components/pagination/src/PaginationArrow/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+  ({as = 'button', href, variant = 'back', ...props}, ref) => {
+    handleLinkValidation({as, href});
+    const PaginationArrowComponent = PaginationArrowVariants[variant];
+    return <PaginationArrowComponent {...props} ref={ref} as={as} href={href} />;
+  }
+);
+
+PaginationArrow.displayName = 'PaginationArrow';
+
+if (process.env.NODE_ENV === 'development') {
```
TheSisb, line 23: Forgot to remove this one

packages/paste-core/components/pagination/src/types.ts
```diff
@@ -0,0 +1,46 @@
...
+}
+
+export interface PaginationButtonProps extends React.HTMLAttributes<'button'> {
+  as?: AsTags;
+  href?: string;
+}
+
+export interface PaginationArrowProps extends PaginationButtonProps {
+  label: string;
+  variant?: Variants;
```
TheSisb, line 24: I think variant should be a required prop for Arrow, it's weird omitting it and makes code a little confusing as I wasn't sure what an arrow without `back/forward` meant without digging in the code.

packages/paste-core/components/pagination/src/PaginationArrow/PaginationArrowSVG.tsx
```diff
@@ -0,0 +1,24 @@
+import * as React from 'react';
+import {Box} from '@twilio-paste/box';
+
+const PaginationArrowSVG: React.FC = ({children}) => {
```
TheSisb, line 4: I don't _love_ this component name. This isn't quite an SVG, but a wrapper around the SVG Icon.

packages/paste-core/components/pagination/src/PaginationItems.tsx
richbachman: We could cut the ` || typeof child === 'string'` part because we'd only want consumers to use `PaginationArrow`, `PaginationNumbers`, or `PaginationNumber` here.

packages/paste-core/components/pagination/src/PaginationArrow/PaginationArrowSVG.tsx
TheSisb: This could also be rewritten as
```js
export const PaginationArrowIconWrapper = styled.div({
  css({
    alignItems: 'center',
    ...
   })
})
```

packages/paste-core/components/pagination/src/PaginationArrow/index.tsx
```diff
@@ -0,0 +1,27 @@
...
+const PaginationArrowVariants = {
+  back: PaginationBackArrow,
+  forward: PaginationForwardArrow,
+};
+
+const PaginationArrow = React.forwardRef<HTMLButtonElement, PaginationArrowProps>(
+  ({as = 'button', href, variant = 'back', ...props}, ref) => {
+    handleLinkValidation({as, href});
+    const PaginationArrowComponent = PaginationArrowVariants[variant];
+    return <PaginationArrowComponent {...props} ref={ref} as={as} href={href} />;
```
TheSisb, line 17: It's probably a little cleaner to just do the conditional in here without componentizing arrow back/forward, like:

```js
const VisibleLabel = ({visibleLabel, ...props}) => {
  if (!visibleLabel) return null;
  return (
    <Text aria-hidden="true" as="span" color="inherit" {...props}>
      {visibleLabel}
    </Text>
  );
}

....



<PaginationArrowButton {...props} ref={ref} as={as} visibleLabel={visibleLabel}>
    {variant === "forward"  ? <VisibleLabel marginRight="space30" label={visibleLabel} /> : null}
    <Box
      alignItems="center"
      borderColor="inherit"
      borderRadius="borderRadiusCircle"
      borderStyle="solid"
      borderWidth="borderWidth10"
      display="flex"
      height="sizeSquare80"
      justifyContent="center"
      width="sizeSquare80"
    >
       {variant === "forward" ? <ArrowForwardIcon decorative={false} display="block" title={label} /> : ...}
    </Box>
     {variant === "back"  ? <VisibleLabel marginLeft="space30" label={visibleLabel} /> : null}
</PaginationArrowButton>
```

packages/paste-core/components/pagination/src/types.ts
richbachman: Good point. Thats also inline with our Button component.

packages/paste-core/components/pagination/src/PaginationArrow/PaginationArrowSVG.tsx
SiTaggart: We should try to avoid styled elements where possible going forward. `Box` all the way

richbachman: I like the name change idea. Good call.

packages/paste-core/components/pagination/src/Pagination.tsx
richbachman: Since I'm sharing the `PaginationArrowButtonPropTypes` between the different `PaginationArrow` variants, I felt it was best to put all of the PropTypes in one place.

packages/paste-core/components/pagination/src/PaginationArrow/index.tsx
richbachman: I restructured this slightly to make it more clear which variant is being rendered. I kept the original basic idea here because the conditionals throw me off a bit. I like having the two separate variant components. For me, it makes it easier to see whats going on the files. Also, if anything needs to change I know exactly where to make that change.

TheSisb (approved):
LGTM

Issue #1305, opened 2021-03-23
Input type "disabled"  blends input value into background color in Safari


margusha:
Thanks for helping make Paste better! Note: For urgent requests, reach out at **#help-design-system**

**Describe the bug**
Paste documentation outlines the usage of input with disabled state. In the examples it shows visible value as it is defined in placeholder. When, instead value attribute is used for showing disabled input content, its color blends into background color in safari.

**To Reproduce**
1. create an new paste project
2. Use input component with disabled=true and specify value attribute with random content
3. Check in Safari - the input is invisible, but selectable

**Expected behavior**
Input value is visible, e.g. uses some other color

**Screenshots**
Disabled Input component using value attribute to show input content
![Screen Shot 2021-03-23 at 10 09 27 AM](https://user-images.githubusercontent.com/1866629/112113853-06dc1380-8bc0-11eb-98df-ab5d1046f0b9.png)

disabled input component when selection is highlighted:
![Screen Shot 2021-03-23 at 10 10 12 AM](https://user-images.githubusercontent.com/1866629/112114017-3ab73900-8bc0-11eb-97cb-7534ca9611d4.png)

**Desktop (please complete the following information):**
 - OSX
 - Safari
 - All versions

**Additional context**
Alternatively if the placeholder is desired approach, should hidden input be used instead to store the value? Quite often you would still expect to get the input value event hough its disabled.

richbachman:
Hi @margusha. Thanks for reaching out about this issue. We'll look into it, and get back to you as soon as possible.

richbachman:
Hi @margusha I've looked into this one further, and I think I found a solution. I've created a ticket in our backlog so we can get it fixed. We'll update this issue once the fix is deployed.

TheSisb:
Hi @margusha,

Unable to repro on Safari version 14.0.3 (15610.4.3.1.7, 15610) while looking at this example on our doc site: https://paste.twilio.design/components/input#disabled-input

Can you provide your Safari version please?

richbachman:
Hi @margusha,

We have a PR up that fixes this issue: https://github.com/twilio-labs/paste/pull/1408

PR #1408, opened 2021-04-30
fix(input): fixed the color of disabled input values

richbachman:
Fixes: https://github.com/twilio-labs/paste/issues/1305

- [x] Added `-webkit-text-fill-color` to css scales and enabled it to use `textColor` tokens
- [x] Fixed disabled `Input` value color
- [x] Fixed disabled `TextArea` value color
- [x] Adjusted disabled `Combobox` story to have an initial value for VRT

### DIFF
--- a/.changeset/itchy-countries-attend.md
+++ b/.changeset/itchy-countries-attend.md
@@ -1,0 +1,7 @@
+---
+'@twilio-paste/input': patch
+'@twilio-paste/textarea': patch
+'@twilio-paste/core': patch
+---
+
+Fix disabled value color in Safari.
--- a/packages/paste-core/components/textarea/src/TextArea.tsx
+++ b/packages/paste-core/components/textarea/src/TextArea.tsx
@@ -60,6 +60,9 @@ const TextAreaElement = styled(TextareaAutosize)<TextAreaProps>((props) =>
     '&:disabled': {
       color: props.variant === 'inverse' ? 'colorTextInverseWeaker' : 'colorTextWeaker',
       cursor: 'not-allowed',
+      // Fixes value color in Safari
+      '-webkit-text-fill-color': props.variant === 'inverse' ? 'colorTextInverseWeaker' : 'colorTextWeaker',
+      '-webkit-opacity': '1',
     },
   })
 );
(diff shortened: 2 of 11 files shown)


gloriliale (approved):
This looks great 🙌🏽! Do we need to include a changeset for combobox as well?

richbachman:
> This looks great 🙌🏽! Do we need to include a changeset for combobox as well?

@gloriliale I don't think so since we're just adding a story there. Nothing in the package actually changes.

.changeset/itchy-countries-attend.md
```diff
@@ -0,0 +1,6 @@
+---
+'@twilio-paste/textarea': patch
+'@twilio-paste/core': patch
```
TheSisb, line 3: Can add `'@twilio-paste/input': patch` here since it's the same change I think

packages/paste-core/components/textarea/src/TextArea.tsx
```diff
@@ -60,6 +60,9 @@ const TextAreaElement = styled(TextareaAutosize)<TextAreaProps>((props) =>
     '&:disabled': {
       color: props.variant === 'inverse' ? 'colorTextInverseWeaker' : 'colorTextWeaker',
       cursor: 'not-allowed',
+      // Fixes value color in Safari
+      '-webkit-text-fill-color': props.variant === 'inverse' ? 'colorTextInverseWeaker' : 'colorTextWeaker',
```
TheSisb, line 64: Just curious, why does this fix the issue? According to [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-text-fill-color), if this isn't provided the `color` will be picked up anyways: 
> If this property is not set, the value of the color property is used.

TheSisb (approved):
Just not sure how this works, but the fix seems reasonable devoid of alternatives.

packages/paste-core/components/textarea/src/TextArea.tsx
richbachman: @TheSisb unsure as well. I read that in the MDN page also, so using `color` should just work. I'm wondering if it's a bug in Safari, or they just decided to ignore the spec.

PR #1426, opened 2021-05-07
chore(website): filter out deprecated design tokens from list

SiTaggart:
Filter out the deprecated design tokens from the list page.

- I tried doing this by filtering the GraphQL so I wouldn't have to do this, but for whatever reason I just can't get it to filter the date

In terms of trying to add autocomplete annotations, we can deprecate methods, arguments or props, but not specific values of those things. I even tried with a conditional  type using JSDoc syntax but it just wouldn't infer the annotation based on value.

### DIFF
--- a/packages/paste-style-props/src/proptypes/background.ts
+++ b/packages/paste-style-props/src/proptypes/background.ts
@@ -1,7 +1,16 @@
 import {DefaultTheme} from '@twilio-paste/theme';
 import {propValidator} from './utils/propValidator';
+import {DEPRECATE_TOKENS} from '../constants';
 
 // Tokens
 const BackgroundColorOptions = ['none', 'transparent', ...Object.keys(DefaultTheme.backgroundColors)];
 
 export const isBackgroundColorTokenProp = propValidator(BackgroundColorOptions);
+export const isDeprecatedBackgroundColorTokenProp = (props: Record<string, unknown>): Error | null => {
+  if (props.backgroundColor != null && DEPRECATE_TOKENS.indexOf(props.backgroundColor) > 0) {
+    console.error(
+      `Paste: "${props.backgroundColor}" is a deprecated design token. It will be removed in the 2021.11.16 release of Paste`
+    );
+  }
+  return null;
+};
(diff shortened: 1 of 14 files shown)


packages/paste-style-props/src/proptypes/background.ts
```diff
@@ -1,7 +1,16 @@
...
+import {DEPRECATE_TOKENS} from '../constants';
 
 // Tokens
 const BackgroundColorOptions = ['none', 'transparent', ...Object.keys(DefaultTheme.backgroundColors)];
 
 export const isBackgroundColorTokenProp = propValidator(BackgroundColorOptions);
+export const isDeprecatedBackgroundColorTokenProp = (props: Record<string, unknown>): Error | null => {
+  if (props.backgroundColor != null && DEPRECATE_TOKENS.indexOf(props.backgroundColor) > 0) {
+    console.error(
+      `Paste Box: ${props.backgroundColor} is a deprecated design token. It will be removed in the 2021.11.16 release of Paste`
```
TheSisb, line 12: Not quite accurate on the `Paste Box:` prefix. I think the rest of the string is good enough, but also add quotes around the token `"${props.backgroundColor}"`

packages/paste-style-props/src/proptypes/border.ts
```diff
@@ -9,3 +10,11 @@ const BorderColorOptions = ['transparent', ...Object.keys(DefaultTheme.borderCol
 export const isBorderRadiusTokenProp = propValidator(BorderRadiusOptions);
 export const isBorderWidthTokenProp = propValidator(BorderWidthOptions);
 export const isBorderColorTokenProp = propValidator(BorderColorOptions);
+export const isDeprecatedBorderColorTokenProp = (props: Record<string, unknown>): Error | null => {
+  if (props.borderColor != null && DEPRECATE_TOKENS.indexOf(props.borderColor) > 0) {
+    console.error(
+      `Paste Box: ${props.borderColor} is a deprecated design token. It will be removed in the 2021.11.16 release of Paste`
```
TheSisb, line 16: Same comment and the other occurences

TheSisb (approved):
V nice!

PR #1431, opened 2021-05-10
fix(input): add margin space0 to fix Safari bug

richbachman:
- [x] Added `margin: space0` on input, select, and button. It wasn't needed for textarea. 

I played with adding `all: initial` or `all: unset`, but that also cancelled out things like `width`, even if I set it along with the `all` shorthand. It made more sense to just add the margin reset.


TheSisb (approved):
Code looks good, didn't check on Safari but I trust that you did :D

SiTaggart:
> > Code looks fine, I'll wait for Storybook in Chromatic to test in Safari
> 
> @SiTaggart here's the storybook preview: https://5e53448165911c0022e68c74-dkgjcugivq.chromatic.com/

Nice, looks great

PR #1438, opened 2021-05-11
feat: add create-package script and file templates

richbachman:
Added a create:package script that creates new package starter files based on handlebar templates using plop.js. See it in action: https://www.loom.com/share/7e0b7a52b7744090925b50c99d8a372f


TheSisb (approved):
Nice!

PR #1441, opened 2021-05-11
fix(site-input-page): resolve breaking error on site input page

zahnster:
* Resolves missing component reference that was breaking rendering of the Input page


richbachman:
Looks like the Select page is doing the same thing: https://deploy-preview-1441--paste-docs.netlify.app/components/select/

PR #1448, opened 2021-05-13
fix(checkbox): checkmark should always have a color

SiTaggart:
When checked state is changed via interaction the checked prop isn't
changed, so we can't rely on it to set the color. If you do rely only on
the checked prop, you can't see the check mark when you interact with
the control.

### DIFF
--- a/packages/paste-core/components/checkbox/src/Checkbox.tsx
+++ b/packages/paste-core/components/checkbox/src/Checkbox.tsx
@@ -2,7 +2,7 @@ import * as React from 'react';
 import * as PropTypes from 'prop-types';
 import {useUID} from '@twilio-paste/uid-library';
 import {Box, safelySpreadBoxProps} from '@twilio-paste/box';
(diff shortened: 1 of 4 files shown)


richbachman (approved):

packages/paste-core/components/checkbox/src/Checkbox.tsx
```diff
@@ -55,14 +55,7 @@ const CheckboxIcon: React.FC<{
   disabled: boolean | undefined;
   checked: boolean | undefined;
 }> = ({checked, disabled, indeterminate}) => {
-  let color = 'transparent' as TextColorOptions;
```
TheSisb, line 58: Hmm I wonder why this ever was an option?

TheSisb (approved):

packages/paste-core/components/checkbox/src/Checkbox.tsx
richbachman: It was my bad from the initial dark mode PR that carried over here.

PR #1451, opened 2021-05-14
feat(design-tokens): availability tokens

SiTaggart:


TheSisb (approved):
Do we need to update the other themes in any way? I don't think so, but double checking.

richbachman (approved):

packages/paste-design-tokens/tokens/global/text-color.yml
```diff
@@ -93,3 +93,17 @@ props:
...
+    comment: Color used to represent an entity or person as "available".
+  color-text-busy:
+    value: "{!palette-orange-60}"
+    comment: Color used to represent an entity or person as "busy".
+  color-text-unavailable:
+    value: "{!palette-red-60}"
+    comment: Color used to represent an entity or person as "unavailable".
+  color-text-offline:
+    value: "{!palette-gray-40}"
+    comment: Color used to represent an entity or person as "offline".
```
serifluous, line 109: What background colors is this supposed to be used on? Contrast is only 2.11 on background-body

SiTaggart: They are likely to be icon only colors, so I guess 3:1 would be beneficial. I don't know where they use them though, I think it must just be the background color to a status dot for them. 

Which if paired with text is technically fine it doesn't have to be 3:1.

Maaaaybe I'll just do backgrounds?

SiTaggart: I'm going to pull the text colors for as I don't think they are needed and don't even pass icon color contrast ratios

SiTaggart:
> Do we need to update the other themes in any way? I don't think so, but double checking.

Technically not. They get added to each theme in the snapshots by inheritance and design wise they aren't supposed to change between light and dark modes.

SG is just Default so thats fine. Console is basically deprecated at this point. There _might be_ value in adding them to dark  just to preserve their colors if they get changed for default.

PR #1452, opened 2021-05-14
feat(modal-dialog-primitive): replace reach dialog with reakit dialog

richbachman:
- [x] Modal Dialog Primitive now uses Reakit Dialog
- [x] Updated Modal Dialog Primitive doc page with new api
- [x] Bumped Reakit version in Reakit-Library
- [x] Updated snapshots for Menu and Menu Primitive

### DIFF
--- a/.changeset/tame-emus-hear.md
+++ b/.changeset/tame-emus-hear.md
@@ -1,0 +1,6 @@
+---
+'@twilio-paste/reakit-library': patch
+'@twilio-paste/core': patch
+---
+
+Bumped to the latest Reakit version.
--- a/packages/paste-core/primitives/modal-dialog/src/index.tsx
+++ b/packages/paste-core/primitives/modal-dialog/src/index.tsx
@@ -1,22 +1,20 @@
-/**
- * If you’re wondering why we wrapped that package into our own, we reasoned that it
- * would be best for our consumers’ DX, with reasons such as:
- *
- * - If we want to migrate the underlying nuts and bolts in the future, Twilio products
- *   that depend on this primitive would need to replace all occurrences of
- *   `import … from ‘@reach/dialog’` to `import … from ‘@some-new/package’`.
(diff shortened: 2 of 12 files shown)


packages/paste-core/primitives/modal-dialog/src/index.tsx
```diff
@@ -1,22 +1,20 @@
...
- */
+/* Keeping the Reach export until we transition Modal to use Reakit */
 import {
   DialogContent as ModalDialogPrimitiveContent,
   DialogOverlay as ModalDialogPrimitiveOverlay,
 } from '@reach/dialog';
 
 export {ModalDialogPrimitiveContent, ModalDialogPrimitiveOverlay};
+
+export type {
```
TheSisb, line 9: Waitaminute... we aren't replacing reach dialog in this PR, we're just adding Reakit dialog. I don't think this is a major change?  I do wonder if we should do the full swap in one go though, as having two dialog libraries is kind of odd maybe?

.changeset/tame-emus-hear.md
```diff
@@ -0,0 +1,6 @@
+---
+'@twilio-paste/reakit-library': patch
+'@twilio-paste/core': patch
+---
+
+Bumped to the latest Reakit version.
```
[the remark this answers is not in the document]
richbachman: Oh... Maybe. I thought changesets would do that for us.

packages/paste-core/primitives/modal-dialog/src/index.tsx
richbachman: For this PR, I didn't want to rebuild Modal yet, so I'm keeping the Reach export there until we rebuild Modal in a separate PR. As part of that one, I'd remove Reach export and dep.

richbachman:
After further investigation, changing to Reakit will causing too large of a breaking change to the current modal implementation.

PR #1454, opened 2021-05-15
fix(disclosure): add a background color to disclosure content

SiTaggart:
To address this discussion topic https://github.com/twilio-labs/paste/discussions/1320

Add a background color to the disclosure content so that you can see it on a background that is not the body color.

Also add story layout for VRT to at least display each component of a background that is not the body.


richbachman (approved):
Code and stories lgtm!

PR #1464, opened 2021-05-18
chore(modal-dialog-primitive): upgrade reach dialog

richbachman:
- [x] Upgraded Reach/Dialog which fixes the test times when a focusable element is in a dialog.
- [x] Upgraded React-Spring to the most recent stable version to fix a TS bug with opacity: https://github.com/pmndrs/react-spring/issues/653.
- [x] Added a test to Modal which renders a modal with focusable elements

### DIFF
--- a/packages/paste-core/components/modal/__tests__/index.spec.tsx
+++ b/packages/paste-core/components/modal/__tests__/index.spec.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
 import {useUID} from '@twilio-paste/uid-library';
 import {render} from 'react-dom';
-import {render as testRender, fireEvent, cleanup} from '@testing-library/react';
+import {render as testRender, fireEvent} from '@testing-library/react';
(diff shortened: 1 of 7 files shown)


packages/paste-core/components/modal/__tests__/index.spec.tsx
```diff
@@ -1,7 +1,7 @@
 import * as React from 'react';
 import {useUID} from '@twilio-paste/uid-library';
 import {render} from 'react-dom';
-import {render as testRender, fireEvent, cleanup} from '@testing-library/react';
```
gloriliale, line 4: It looks like we don't have to use cleanup() anymore, could you explain why that is?

richbachman: For sure. The basic answer is that Jest handles this for us. Here's more detail: https://testing-library.com/docs/react-testing-library/api/#cleanup

SiTaggart (approved):

PR #1466, opened 2021-05-18
fix(flex): correctly set layout style props

SiTaggart:
Based on [this discussion](https://github.com/twilio-labs/paste/discussions/1276), it became apparent that the inability to set layout props on the Flex component is actually a bug.

We extended the types but only half implemented the props. This PR addresses that and adds tests


packages/paste-core/layout/flex/src/index.tsx
```diff
@@ -60,24 +81,47 @@ const Flex = React.forwardRef<HTMLDivElement, FlexProps>(
...
+      );
+    }
+    if (
+      (paddingX && (padding || paddingBottom || paddingLeft || paddingRight || paddingTop)) ||
+      (paddingY && (padding || paddingBottom || paddingLeft || paddingRight || paddingTop))
+    ) {
+      console.error(
+        '[Paste Flex]: you cannot set a top, right, bottom or left padding when using the paddingX or paddingY attributes.'
+      );
+    }
```
richbachman, line 103: Nice! 🔥

richbachman (approved):

packages/paste-core/layout/flex/src/types.tsx
```diff
@@ -18,7 +18,7 @@ export type Basis = ResponsiveValue<BasisOptions>;
 export type WrapOptions = boolean;
 export type Wrap = ResponsiveValue<WrapOptions>;
 
-export interface FlexProps extends LayoutProps, MarginProps, PaddingProps {
+export interface FlexProps extends Omit<LayoutProps, 'verticalAlign'>, MarginProps, PaddingProps {
```
gloriliale, line 21: What is it that Omit is doing here? I don't think I've ever seen it used before.

packages/paste-core/layout/flex/src/index.tsx
```diff
@@ -60,24 +81,47 @@ const Flex = React.forwardRef<HTMLDivElement, FlexProps>(
...
+
+    const margins =
+      marginX || marginY ? {marginX, marginY} : {margin, marginBottom, marginLeft, marginRight, marginTop};
```
gloriliale, line 106: This was written in a way that was so easy to make sense of! Hooray!

packages/paste-core/layout/flex/src/types.tsx
SiTaggart: Good question! Because the Flex component provides it's own `vertical` and `vAlignContent` props, it doesn't make sense to allow you to use the Box style prop `verticalAlign` as they would potentially clash.

So I want all the layout style props, but I don't want `verticalAlign`.

packages/paste-core/layout/flex/src/index.tsx
```diff
@@ -60,24 +81,47 @@ const Flex = React.forwardRef<HTMLDivElement, FlexProps>(
       () => getFlexStyles({basis, hAlignContent, grow, shrink, vertical, vAlignContent, wrap}),
       [basis, hAlignContent, grow, shrink, vertical, vAlignContent, wrap]
     );
+
+    if (size && (width || height)) {
+      console.error('[Paste Flex]: you cannot set height or width with the size attribute.');
```
gloriliale, line 86: I really liked the wording of "you cannot set a  ________ when using the ______ attribute(s)" that was used in the other error messages, and I think it makes sense to use it here as well. 
```suggestion
      console.error('[Paste Flex]: you cannot set a height or width when using the size attribute.');
```

packages/paste-core/layout/flex/src/index.tsx
gloriliale: ^truth

packages/paste-core/layout/flex/__tests__/flex.test.tsx
```diff
@@ -198,261 +199,280 @@ describe('Flex Unit Tests', () => {
...
+  });
+
+  it('should set individual padding style props', (): void => {
+    render(
+      <Flex
+        data-testid="renderedFlexElement"
+        paddingTop="space20"
+        paddingRight="space20"
+        paddingBottom="space20"
+        paddingLeft="space20"
```
gloriliale, line 434: Looking at these tests is really useful for thinking of the ones I'm writing right now (seeing the modularity especially) 🤓
Would it make sense to have these attributes be all different space options to make absolutely certain that the styles match up exactly, or is that overkill?

gloriliale (approved):

packages/paste-core/layout/flex/src/types.tsx
gloriliale: Ah okay, so bc of that clash, we're saying that: from LayoutProps, omit the 'verticalAlign' one. Got it!

packages/paste-core/layout/flex/__tests__/flex.test.tsx
SiTaggart: Maybe? It certainly doesn't hurt, so I'll tweak them

packages/paste-core/layout/flex/src/index.tsx
SiTaggart: Agree, but I'll push it up separately

PR #1471, opened 2021-05-19
fix(website): align content in DisclosureHeading example

richbachman:
Fixes: https://github.com/twilio-labs/paste/discussions/1469

### DIFF
--- a/packages/paste-website/src/pages/components/disclosure/index.mdx
+++ b/packages/paste-website/src/pages/components/disclosure/index.mdx
@@ -8,6 +8,7 @@ slug: /components/disclosure/
 import {graphql} from 'gatsby';
 import {Anchor} from '@twilio-paste/anchor';
 import {Box} from '@twilio-paste/box';
+import {Flex} from '@twilio-paste/flex';
 import {Text} from '@twilio-paste/text';
 import {Stack} from '@twilio-paste/stack';
 import {MediaObject, MediaFigure, MediaBody} from '@twilio-paste/media-object';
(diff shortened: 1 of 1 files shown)


gloriliale (approved):
🎉 Nice! Is there a way to see this since we don't build for every branch anymore in Netlify? Would the best way be to ask you for a screenshot of what it looks like now in your branch (or pull this down to my machine to look at it)?
Also, I'm not super familiar w/ Flex, and was curious on why this was the solution for the misalignment!

richbachman:
> 🎉 Nice! Is there a way to see this since we don't build for every branch anymore in Netlify? Would the best way be to ask you for a screenshot of what it looks like now in your branch (or pull this down to my machine to look at it)?
> Also, I'm not super familiar w/ Flex, and was curious on why this was the solution for the misalignment!

@gloriliale We do still get deploys in every branch with Netlify. They're just hidden down in the checks: https://share.getcloudapp.com/d5u1mwJJ. Here's the deploy preview: https://deploy-preview-1471--paste-docs.netlify.app/components/disclosure#composition-notes

As for the Flex implementation there are multiple ways to center align this. The problem occurred because the DisclosureHeading has a different line-height that the base Text component. So one fix would be to set the line-height of Text to match.

Flex allows us to use the `vAlignContent` prop which returns `align-items: center / flex-start / flex-end`. That saves a consumer from trying to guess what line-height to use. Also, using a layout component for alignment is better here in case we ever change the line-height of DisclosureHeading. Since we're relying on Flex to handle the alignment, if the line-height changes, everything will still be aligned.

Issue #1472, opened 2021-05-19
Combobox does not value "required" attribute


ycao2020:
After adding that attribute on Combobox and then checking on UI, the input field rendered does not have required. Is this intended? Trying to do a form validation, and found the required is missing.

katieporter:
Hey @ycao2020 - thanks for raising this. We are looking into it and will get back to you shortly!

richbachman:
Hi @ycao2020 We just merged the fix. The package deploy should go out in the next 24 hours.

ycao2020:
@richbachman how do I get this change? do I need to upgrade my local Paste package or? my local has the same version as latest, 9.3.3.

richbachman:
@ycao2020 you should see the change in core version 9.3.3. You may need to run a fresh yarn or npm install. Let me know if you're still having trouble after doing that.

SiTaggart:
It'll go out in 9.3.4, we haven't merged the versioning PR yet https://github.com/twilio-labs/paste/pull/1480

PR #1478, opened 2021-05-20
fix(combobox): pass required prop to textbox

richbachman:
Fixes: https://github.com/twilio-labs/paste/issues/1472

- [x] Passes the required prop to the textbox inside combobox.
- [x] Test to verify required is being passed.


TheSisb (approved):
LGTM!

PR #1484, opened 2021-05-24
feat(alert-dialog): create package

gloriliale:
Created a new package for Alert Dialog component. Includes package w/:

- component code
- dependencies (for Modal)
- stories
- tests

### DIFF
--- a/.changeset/smooth-flies-own.md
+++ b/.changeset/smooth-flies-own.md
@@ -1,0 +1,6 @@
+---
+'@twilio-paste/modal': minor
+'@twilio-paste/core': minor
+---
+
+Created style file for stylings of the Modal so they can be used in other components, like Alert Dialog.
--- a/packages/paste-core/components/alert-dialog/__tests__/index.spec.tsx
+++ b/packages/paste-core/components/alert-dialog/__tests__/index.spec.tsx
@@ -1,0 +1,60 @@
+import * as React from 'react';
+import {matchers} from 'jest-emotion';
+import {render, screen} from '@testing-library/react';
+// @ts-ignore typescript doesn't like js imports
+import axe from '../../../../../.jest/axe-helper';
+import {AlertDialogWithTwoActions, DestructiveAlertDialog} from '../stories/index.stories';
+
+expect.extend(matchers);
+
+describe('Alert Dialog', () => {
+  it('Should render an alert dialog box', () => {
+    render(<AlertDialogWithTwoActions />);
+    expect(screen.getByText('Submit application')).toBeDefined();
+  });
+
+  it('Should have two, labeled buttons when a secondary label and action is given', () => {
+    render(<AlertDialogWithTwoActions />);
+    const buttons = screen.getAllByRole('button');
+    expect(buttons).toHaveLength(2);
+    expect(buttons[0]).toHaveTextContent('Cancel');
+    expect(buttons[1]).toHaveTextContent('Submit');
+  });
+
+  it('Should have a destructive button style when the destructive prop is included', () => {
+    render(<DestructiveAlertDialog />);
+    expect(screen.getByText('Delete')).toHaveStyle('background-color: color-background-destructive');
+  });
+
+  it('Should have a heading the same as the heading prop', () => {
+    render(<AlertDialogWithTwoActions />);
+    expect(screen.getByRole('heading')).toHaveTextContent('Submit application');
+  });
+
+  it('Should have the correct aria attributes', () => {
+    render(<AlertDialogWithTwoActions />);
+    expect(screen.getByRole('alertdialog')).toBeTruthy();
+    expect(screen.getByRole('alertdialog').getAttribute('aria-modal')).toEqual('true');
+    expect(screen.getByRole('alertdialog').getAttribute('aria-labelledby')).toEqual(
+      screen.getByRole('heading').getAttribute('id')
+    );
+    expect(screen.getByRole('alertdialog').getAttribute('aria-describedby')).toEqual(
+      screen
+        .getByText('Are you sure you want to submit this application? No information can be changed after submitting.')
+        .getAttribute('id')
+    );
+  });
+
+  it('Should have the initial focus land on the first focusable item', () => {
+    render(<AlertDialogWithTwoActions />);
+    expect(document.activeElement).toEqual(screen.getAllByRole('button')[0]);
+  });
+});
+
+describe('Accessibility', () => {
+  it('Should have no accessibility violations for two button alert dialogs', async () => {
+    const {container} = render(<AlertDialogWithTwoActions />);
+    const results = await axe(container);
+    expect(results).toHaveNoViolations();
+  });
+});
--- a/packages/paste-core/components/alert-dialog/package.json
+++ b/packages/paste-core/components/alert-dialog/package.json
@@ -1,0 +1,53 @@
+{
+  "name": "@twilio-paste/alert-dialog",
+  "version": "0.0.0",
+  "category": "overlay",
+  "status": "production",
+  "description": "An alert dialog displays content over the main page and blocks any interaction with the page.",
+  "author": "Twilio Inc.",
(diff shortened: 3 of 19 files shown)


.changeset/smooth-flies-own.md
```diff
@@ -0,0 +1,6 @@
+---
+'@twilio-paste/modal': patch
+'@twilio-paste/core': patch
+---
+
+Created style file for stylings of the Modal
```
nkrantz, line 6: It might help to add _why_ you pulled the Modal style out into its own file, for someone reading the Modal changelog who doesn't know about Alert Dialog

packages/paste-core/components/alert-dialog/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,66 @@
...
+    render(<AlertDialogExample />);
+    const buttons = screen.getAllByRole('button');
+    expect(buttons).toHaveLength(2);
+    expect(buttons[0]).toHaveTextContent('Cancel');
+    expect(buttons[1]).toHaveTextContent('Submit');
+  });
+
+  it('Should have a destructive button style when the destructive prop is included', () => {
+    render(<AlertDialogDestructive />);
+    expect(screen.getByText('Delete')).toHaveStyle('background-color: color-background-destructive');
```
nkrantz, line 32: Will this test fail if Alert Dialog is used with dark mode eventually?

packages/paste-core/components/alert-dialog/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+  it('Should have the initial focus land on the first focusable item', () => {
+    render(<AlertDialogExample />);
+    expect(document.activeElement).toEqual(screen.getAllByRole('button')[0]);
+  });
+});
+
+describe('Accessibility', () => {
+  it('Should have no accessibility violations', async () => {
+    const {container} = render(<AlertDialogExample />);
```
nkrantz, line 62: This is probably sufficient, but you might want to consider adding accessibility tests for the other storybook examples that you're testing too. When I was writing my tests, I found that one of my stories was failing because the child elements were lacking unique IDs.

packages/paste-core/components/alert-dialog/stories/index.stories.tsx
```diff
@@ -0,0 +1,68 @@
+import * as React from 'react';
+import {AlertDialog} from '../src';
+
+// Create FC's of all the stories you want to see/test
+// Should be 3: 2 choices regs, 2 choices destructive 1 choice notification
```
nkrantz, line 5: Could probs remove these

packages/paste-core/components/alert-dialog/stories/index.stories.tsx
```diff
@@ -0,0 +1,68 @@
...
+  return <AlertDialogExample />;
+};
+
+export const AlertDialogWithOneAction = (): React.ReactNode => {
+  return <AlertDialogForNotifying />;
+};
+
+export const DestructiveAlertDialog = (): React.ReactNode => {
+  return <AlertDialogDestructive />;
+};
```
nkrantz, line 68: Because you're not adding any metadata to the ReactNode examples here, you could just use the functional components as your stories and remove the `excludeStories` line. Storybook will use the FCs as the stories and you can export those ones for the tests as well!

nkrantz (commented):
It looks so great! Really like how you pulled out Modal styles to make it more usable. Your tests are super straightforward and easy to read. Is there anywhere you think it might be useful to add in-line comments for ease of reading?

packages/paste-core/components/alert-dialog/__tests__/index.spec.tsx
gloriliale: That's a good question- since what we're testing here is the token name not the value itself, this should pass even in dark mode (thanks to @richbachman for clarifying!)

packages/paste-core/components/alert-dialog/__tests__/index.spec.tsx
gloriliale: ah yes yes, I'll add one for the one-button version of alert dialog (the destructive version is a styling change, so I think it would be redundant to include)

packages/paste-core/components/alert-dialog/stories/index.stories.tsx
gloriliale: thanks for catching these!

packages/paste-core/components/alert-dialog/stories/index.stories.tsx
gloriliale: I remember us talking about this now! I think it'll mean I have to change the names of the FCs and therefore edit the tests, but this does seem like a more straightforward way to connect the two. Will do!

packages/paste-core/components/alert-dialog/package.json
```diff
@@ -0,0 +1,45 @@
...
+    "react-dom": "^16.8.6"
+  },
+  "devDependencies": {
+    "@twilio-paste/box": "^4.0.5",
+    "@twilio-paste/modal": "^7.0.1",
+    "@twilio-paste/theme": "^5.1.0",
+    "@twilio-paste/types": "^3.1.1",
+    "prop-types": "^15.7.2",
+    "react": "^16.8.6",
+    "react-dom": "^16.8.6"
```
richbachman, line 43: We'll need to add heading, button, stack, and uid-library to peerDeps and devDeps since we're using them in the component.

packages/paste-core/components/alert-dialog/src/index.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+export const AlertDialog = React.forwardRef<HTMLDivElement, AlertDialogProps>(
+  (
+    {children, isOpen, heading, destructive, primaryLabel, primaryAction, secondaryLabel, secondaryAction, ...props},
+    ref
+  ) => {
+    const headingID = useUID();
+    const bodyID = useUID();
+
+    const primaryStyle = destructive ? 'destructive' : 'primary';
```
richbachman, line 34: Should this be something like `buttonVariant` or `primaryVariant` instead?

packages/paste-core/components/alert-dialog/package.json
gloriliale: Ah of course! Thanks for catching that. Where is it that uid-library is used?

packages/paste-core/components/alert-dialog/src/index.tsx
gloriliale: I like that better, yeah. I'll change it to `primaryVariant`

.changeset/smooth-flies-own.md
SiTaggart: Counter point, as this is a change log for consumers of the package, they aren't concerned with what you did, but what they can now do with it.

packages/paste-core/components/alert-dialog/package.json
richbachman: That library is generating the `headingID` and `bodyID` here: https://github.com/twilio-labs/paste/pull/1484/files#diff-33645d3572edd38a41d0c48139f80384adde5f8f86122ad2747ab42d407774bfR31-R32

.changeset/smooth-flies-own.md
```diff
@@ -0,0 +1,6 @@
+---
+'@twilio-paste/modal': patch
+'@twilio-paste/core': patch
```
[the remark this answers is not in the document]
richbachman: Even though this is just adding the exportable styles to the Modal package?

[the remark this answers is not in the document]
TheSisb: Yup, because it's new functionality that anyone can import.

TheSisb (approved):

SiTaggart (approved):

PR #1499, opened 2021-05-26
feat(date-picker): create package

nkrantz:
- Added date-picker package
- Added `type="time"` to Input, Input-box, Form
- Wrote tests for Date Picker and `returnDateFormatter` func

**Contributing to Twilio**

### DIFF
--- a/.changeset/purple-monkeys-refuse.md
+++ b/.changeset/purple-monkeys-refuse.md
@@ -1,0 +1,6 @@
+---
+'@twilio-paste/date-picker': minor
+'@twilio-paste/core': minor
+---
+
+Created the Date Picker component
--- a/packages/paste-core/components/date-picker/package.json
+++ b/packages/paste-core/components/date-picker/package.json
@@ -1,0 +1,58 @@
+{
+  "name": "@twilio-paste/date-picker",
+  "version": "0.0.0",
+  "category": "user input",
+  "status": "production",
+  "description": "A form component used to select a date.",
+  "author": "Twilio Inc.",
+  "license": "MIT",
+  "main:dev": "src/index.tsx",
+  "main": "dist/index.js",
+  "module": "dist/index.es.js",
+  "types": "dist/index.d.ts",
+  "sideEffects": false,
+  "publishConfig": {
+    "access": "public"
+  },
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "build": "yarn clean && NODE_ENV=production node build.js && tsc",
+    "build:js": "NODE_ENV=development node build.js",
+    "build:props": "typedoc --tsconfig ./tsconfig.json --json ./dist/prop-types.json",
+    "clean": "rm -rf ./dist",
+    "tsc": "tsc"
+  },
+  "dependencies": {
+    "date-fns": "2.21.3"
+  },
+  "peerDependencies": {
+    "@twilio-paste/box": "^4.0.2",
+    "@twilio-paste/design-tokens": "^6.6.0",
+    "@twilio-paste/input": "^3.0.6",
+    "@twilio-paste/input-box": "^4.0.2",
+    "@twilio-paste/style-props": "^3.0.1",
+    "@twilio-paste/styling-library": "^0.3.1",
+    "@twilio-paste/theme": "^5.0.1",
+    "@twilio-paste/types": "^3.1.1",
+    "@twilio-paste/uid-library": "^0.2.1",
+    "prop-types": "^15.7.2",
+    "react": "^16.8.6",
+    "react-dom": "^16.8.6"
+  },
+  "devDependencies": {
+    "@twilio-paste/box": "^4.0.2",
+    "@twilio-paste/design-tokens": "^6.6.0",
+    "@twilio-paste/input": "^3.0.6",
+    "@twilio-paste/input-box": "^4.0.2",
+    "@twilio-paste/style-props": "^3.0.1",
+    "@twilio-paste/styling-library": "^0.3.1",
+    "@twilio-paste/theme": "^5.0.1",
+    "@twilio-paste/types": "^3.1.1",
+    "@twilio-paste/uid-library": "^0.2.1",
+    "prop-types": "^15.7.2",
+    "react": "^16.8.6",
+    "react-dom": "^16.8.6"
+  }
+}
--- a/packages/paste-core/components/date-picker/src/DatePicker.tsx
+++ b/packages/paste-core/components/date-picker/src/DatePicker.tsx
@@ -1,0 +1,28 @@
+import * as React from 'react';
+import * as PropTypes from 'prop-types';
+import type {InputProps} from '@twilio-paste/input';
+import {Input} from '@twilio-paste/input';
+
+export type DatePickerProps = Omit<InputProps, 'type' | 'insertBefore' | 'insertAfter'>;
+
+const DatePicker = React.forwardRef<HTMLInputElement, DatePickerProps>((props, ref) => {
+  return <Input {...props} type="date" ref={ref} />;
+});
+
+DatePicker.displayName = 'DatePicker';
+
+DatePicker.propTypes = {
+  disabled: PropTypes.bool,
+  hasError: PropTypes.bool,
+  id: PropTypes.string,
+  name: PropTypes.string,
+  onBlur: PropTypes.func,
+  onChange: PropTypes.func,
+  onFocus: PropTypes.func,
+  placeholder: PropTypes.string,
+  readOnly: PropTypes.bool,
+  required: PropTypes.bool,
+  value: PropTypes.string,
+};
+
+export {DatePicker};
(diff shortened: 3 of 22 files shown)


packages/paste-core/components/date-picker/package.json
```diff
@@ -0,0 +1,68 @@
...
+  ],
+  "scripts": {
+    "build": "yarn clean && NODE_ENV=production node build.js && tsc",
+    "build:js": "NODE_ENV=development node build.js",
+    "build:props": "typedoc --tsconfig ./tsconfig.json --json ./dist/prop-types.json",
+    "clean": "rm -rf ./dist",
+    "tsc": "tsc"
+  },
+  "dependencies": {
+    "date-fns": "^2.21.3"
```
TheSisb, line 28: We can safely remove `^` as you brought up because it will reduce margin for error in consumer installs.

packages/paste-core/components/date-picker/src/DatePicker.tsx
```diff
@@ -0,0 +1,13 @@
+import * as React from 'react';
+import type {InputProps} from '@twilio-paste/input';
+import {Input} from '@twilio-paste/input';
+
+export type DatePickerProps = Omit<InputProps, 'type'>;
+
+const DatePicker = React.forwardRef<HTMLInputElement, DatePickerProps>((props, ref) => {
+  return <Input type="date" {...props} ref={ref} />;
```
TheSisb, line 8: Let's move `{...props}` to before the `type="date"` to prevent overrides in JS-land.

packages/paste-core/components/date-picker/src/DatePicker.tsx
```diff
@@ -0,0 +1,13 @@
...
+});
+
+DatePicker.displayName = 'DatePicker';
```
TheSisb, line 11: Missing proptypes.

packages/paste-core/components/date-picker/src/utils.ts
```diff
@@ -0,0 +1,18 @@
...
+// into the formatter, we need to create a new Date using the date `value`. However, when `new Date()`
+// is used, 'T00:00:00' is assigned as the time in UTC+0 then converted tothe user's locale
+// (e.g. value becomes the previous day at 7pm in EST/UTC-5), shifting the date by 1 day.
+// We're setting the time to 12pm to add padding on either side to avoid the shift from happening.
+
+// UPDATE: added '-00:00' to the end of the `time` string so
+// that Date would automatically be instantiated in UTC+0.
+
+export const returnDateFormatter = (dateValue: string, dateFormat: string): string => {
+  const time = 'T12:00:00-00:00';
```
TheSisb, line 15: This would break if someone, somehow, passed in a dateValue string with time already concatenated.
Maybe consider a check like:
```js
const time = 'T12:00:00-00:00';

export const returnDateFormatter = (dateString: string, dateFormat: string): string => {
   let dateValue = dateString;
   if (!dateValue.contains("T")) {
     dateValue = new Date(dateValue.concat(time));
   }
   return format(Number(dateValue), dateFormat);
}
```

TheSisb: And then add 1 test

packages/paste-core/components/date-picker/stories/index.stories.tsx
```diff
@@ -0,0 +1,298 @@
...
+// Stories: default, one using onChange with formatter function (copy combobox controlled using state story) and render value, replicate input stories
+
+export const DefaultDatePicker: React.FC = (props) => {
+  const uidDP = useUID();
+  const uidHT = useUID();
+  return (
+    <>
+      <Label htmlFor={uidDP}>Label</Label>
+      <DatePicker id={uidDP} aria-describedby={uidHT} {...props} />
+      <HelpText id={uidHT}>Help text</HelpText>
```
TheSisb, line 22: Let's make these examples a little more realistic because we have plans to use stories in Tests, the doc site, and more in the future.
i.e.: HelpText shouldn't be "help text" :P

packages/paste-core/components/input/src/Input.tsx
```diff
@@ -72,6 +72,9 @@ export const InputElement = styled.input<InputProps>((props) =>
       '-webkit-text-fill-color': props.variant === 'inverse' ? 'colorTextInverseWeaker' : 'colorTextWeaker',
       '-webkit-opacity': '1',
     },
+    '&::-webkit-datetime-edit': {
+      display: 'flex',
```
TheSisb, line 76: Can we leave a comment on why we added this for future maintainers.

TheSisb (commented):
Great stuff! Super close. Small comments and the verbal ones from Zoom then lgtm!

packages/paste-core/components/date-picker/package.json
```diff
@@ -0,0 +1,68 @@
+{
+  "name": "@twilio-paste/date-picker",
+  "version": "0.0.0",
+  "category": "user input",
+  "status": "alpha",
```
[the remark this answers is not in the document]
richbachman: 👍 I think we're going with production for everything now.

packages/paste-core/components/date-picker/package.json
```diff
@@ -0,0 +1,68 @@
...
+    "@twilio-paste/icons": "^5.2.0",
+    "@twilio-paste/input": "^3.0.6",
+    "@twilio-paste/input-box": "^4.0.4",
+    "@twilio-paste/label": "^6.0.3",
+    "@twilio-paste/style-props": "^3.0.1",
+    "@twilio-paste/styling-library": "^0.3.1",
+    "@twilio-paste/text": "^4.0.4",
+    "@twilio-paste/theme": "^5.0.1",
+    "@twilio-paste/types": "^3.1.1",
+    "@twilio-paste/uid-library": "^0.2.1",
```
richbachman, line 44: Are we using anchor, box, combobox, help-text, icons, label, text, or uid-library anywhere other than in stories? If not, we can remove those from here and the devDeps.

packages/paste-core/components/date-picker/stories/index.stories.tsx
```diff
@@ -0,0 +1,298 @@
...
+import {Text} from '@twilio-paste/text';
+import {Anchor} from '@twilio-paste/anchor';
+import {HelpText} from '@twilio-paste/help-text';
+import {useUID} from '@twilio-paste/uid-library';
+import {InformationIcon} from '@twilio-paste/icons/esm/InformationIcon';
+import {Combobox} from '@twilio-paste/combobox';
+import {DatePicker, returnDateFormatter} from '../src';
+
+// Write date picker stories and export to use in test file
+// Stories: default, one using onChange with formatter function (copy combobox controlled using state story) and render value, replicate input stories
```
richbachman, line 13: Reminder to remove this comment.

packages/paste-core/components/date-picker/package.json
nkrantz: Why don't they need to be dependencies if I'm using them in stories?

.changeset/purple-monkeys-refuse.md
```diff
@@ -0,0 +1,6 @@
+---
+'@twilio-paste/date-picker': patch
+'@twilio-paste/core': patch
```
nkrantz, line 3: @TheSisb should these be minor/major changes? What about for Input and Input Box (the only change was adding type 'date' to both) - patch?

packages/paste-core/components/date-picker/package.json
TheSisb: Storybook has been modified in our repository to resolve (read: find) the src files in each of our packages. So it knows how to find Paste dependencies. This makes it easier to manage and use Storybook. This means we can get hot reloads on our changes to code, and also means that we don't need to couple storybook dependencies to the thing we're shipping in each package.

TheSisb (approved):
LGTM!

packages/paste-core/components/date-picker/package.json
```diff
@@ -0,0 +1,50 @@
...
+  },
+  "dependencies": {
+    "date-fns": "2.21.3"
+  },
+  "peerDependencies": {
+    "@twilio-paste/input": "^3.0.6",
+    "@twilio-paste/style-props": "^3.0.1",
+    "@twilio-paste/styling-library": "^0.3.1",
+    "@twilio-paste/theme": "^5.0.1",
+    "@twilio-paste/types": "^3.1.1",
```
SiTaggart, line 35: Be sure to follow the peer dependency tree and list all sub peer dependencies in this package. Details can be found here https://github.com/twilio-labs/paste#peer-deps

packages/paste-core/components/input/src/Input.tsx
```diff
@@ -72,6 +73,15 @@ export const InputElement = styled.input<InputProps>((props) =>
...
       '-webkit-opacity': '1',
     },
+    // Fixes extra height issue in Safari
+    '&::-webkit-datetime-edit': {
+      display: 'flex',
+    },
+
+    // Change to pointer cursor on cal icon
+    '&::-webkit-calendar-picker-indicator:hover': {
+      cursor: props.readOnly || props.disabled ? 'default' : 'pointer',
```
nkrantz, line 83: @richbachman made these changes (along with line 58), just want to get someone else's eyes on it before I merge it in to make sure it's the right approach to solving the cursor fix.

richbachman (approved):

PR #1501, opened 2021-05-26
fix(menu): use button element when not a link

richbachman:
Fixes: https://github.com/twilio-labs/paste/discussions/1490

- [x] `MenuItem` is rendered as a button unless an href is passed
- [x] Fixed `type` imports
- [x] Added tests


.changeset/itchy-mangos-punch.md
```diff
@@ -0,0 +1,6 @@
+---
+'@twilio-paste/menu': patch
+'@twilio-paste/core': patch
+---
+
+MenuItem now renders as a `buttont` by default. In order to render it as an `a`, and `href` needs to be passed.
```
SiTaggart, line 6:
```suggestion
MenuItem now renders as a `button` by default. In order to render it as an `a`, and `href` needs to be passed.
```

packages/paste-core/components/menu/__tests__/index.spec.tsx
```diff
@@ -80,6 +86,18 @@ describe('Menu', () => {
...
+      const renderedMenuItem = screen.getByTestId('example-menu-item');
+      expect(renderedMenuItem.tagName).toBe('BUTTON');
+    });
+
+    it('should render a menu item as an anchor', () => {
+      render(<MenuMock />);
+      const renderedMenuItem = screen.getByTestId('example-menu-anchor');
+      expect(renderedMenuItem.tagName).toEqual('A');
+    });
+
```
SiTaggart, line 100: Worth adding tests that ensure the onClick handler is fired when I:

- click on an element
- press spacebar
- press enter

On both anchors and buttons?

richbachman: Yeah. Good call. Added those.

TheSisb (approved):
Nice followup, thanks!

PR #1509, opened 2021-05-28
feat(customization): prepare customization provider to accept an elements object 

SiTaggart:
In prep for component customization, prepare the customization provider to be able to accept an object of elements that are to be styled.

Much of this comes from the POC https://github.com/twilio-labs/paste/pull/476/

- accept an well typed elements object on the provider
- create a customization consumer as the returned theme needs retyping to add `elements` to it. Otherwise it would appear on the generic theme that everyone uses internally
- create some custom theme and custom CSS paste types
- export some of the styled system css function types to extend from for our custom css types


packages/paste-customization/src/CustomizationProvider.tsx
```diff
@@ -1,26 +1,40 @@
...
+  elements,
   theme,
   disableAnimations = false,
   ...props
 }) => {
-  const customTheme = React.useMemo(() => createCustomTheme(DefaultTheme, theme || {}, customBreakpoints), [
-    theme,
-    customBreakpoints,
-  ]);
+  const starterTheme = baseTheme === 'dark' ? DarkTheme : DefaultTheme;
```
TheSisb, line 32: nit for future maintainability, we should either inline this into line 35 or move this const inside the useMemo callback.

TheSisb (approved):

PR #1521, opened 2021-06-02
chore(modal): fix chromatic snapshots

richbachman:
- Added extra VRT stories for alert dialog and modal to test the individual parts
- Removed existing alert dialog and modal stories from the Chromatic screenshot flow

### DIFF
--- a/packages/paste-core/components/alert-dialog/src/index.tsx
+++ b/packages/paste-core/components/alert-dialog/src/index.tsx
@@ -2,24 +2,12 @@ import * as React from 'react';
 import * as PropTypes from 'prop-types';
 import {useUID} from '@twilio-paste/uid-library';
 import {useTransition} from '@twilio-paste/animation-library';
-import {css, styled} from '@twilio-paste/styling-library';
-import {Box, safelySpreadBoxProps} from '@twilio-paste/box';
-import {Heading} from '@twilio-paste/heading';
-import {Button} from '@twilio-paste/button';
-import {Stack} from '@twilio-paste/stack';
-import {
-  modalHeaderStyles,
-  modalBodyStyles,
-  modalFooterStyles,
-  ModalDialogOverlay,
-  ModalDialogContent,
-} from '@twilio-paste/modal';
-
-const NarrowModalDialogContent = styled(ModalDialogContent)(() =>
-  css({
-    maxWidth: 'size40',
-  })
-);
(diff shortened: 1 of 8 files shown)


TheSisb (approved):

packages/paste-core/components/alert-dialog/src/AlertDialog.tsx
```diff
@@ -0,0 +1,91 @@
...
+import {useUID} from '@twilio-paste/uid-library';
+import {useTransition} from '@twilio-paste/animation-library';
+import {css, styled} from '@twilio-paste/styling-library';
+import {safelySpreadBoxProps} from '@twilio-paste/box';
+import {ModalDialogOverlay, ModalDialogContent} from '@twilio-paste/modal';
+import {AlertDialogHeader} from './AlertDialogHeader';
+import {AlertDialogBody} from './AlertDialogBody';
+import {AlertDialogFooter} from './AlertDialogFooter';
+
+export const NarrowModalDialogContent = styled(ModalDialogContent)(() =>
```
SiTaggart, line 12: If we're now exporting this from the package, I don't love the name. Can it just be `AlertDialogContent`?

packages/paste-core/components/alert-dialog/src/index.tsx
```diff
@@ -1,108 +1,4 @@
...
-  destructive: PropTypes.bool,
-  heading: PropTypes.string.isRequired,
-  isOpen: PropTypes.bool.isRequired,
-  onConfirm: PropTypes.func.isRequired,
-  onConfirmLabel: PropTypes.string.isRequired,
-  onDismiss: PropTypes.func.isRequired,
-  onDismissLabel: PropTypes.string.isRequired,
-};
+export * from './AlertDialog';
+export * from './AlertDialogHeader';
```
SiTaggart, line 2: Actually, lets just expose AlertDialog from the package. The other bits can be private as you directly import them from source in stories and tests.

That way we're not confusing consumers with additional components that aren't actually part of the documented API

SiTaggart (changes requested):

packages/paste-core/components/alert-dialog/src/index.tsx
richbachman: Good call. I'll make that change.

SiTaggart (approved):

Issue #1542, opened 2021-06-09
Toasts appear to be broken?


sebm:
Thanks for helping make Paste better! Note: For urgent requests, reach out at **#help-design-system**

**Describe the bug**
Toasts are not … toasting … 🍞 

**To Reproduce**
https://paste.twilio.design/components/toast/#showing-dismissing-and-timing-out-a-toast
click "Add Toast"

**Expected behavior**
Expect a toast to show up.

**Screenshots**
A toast does not show up.

**Desktop (please complete the following information):**
 - OS: os x 10.15.17
 - Browser: chrome + ff
 - Version: the latest

richbachman:
Hi @sebm. We've been unable to reproduce this issue. Would you mind verifying on your end? Thanks!

sebm:
https://user-images.githubusercontent.com/188161/121573258-0ac74700-c9f3-11eb-8efc-4f21638ca1e8.mov

here's a screencap

sebm:
this is Chrome Version 91.0.4472.77 (Official Build) (x86_64) on OS 11.5 - i upgraded this morning

sebm:
Tried in FF 90.0b5 (64-bit), Safari Version 14.1.1 (15611.2.7.1.6) - same deal in all browsers, with no error feedback.

Interestingly on the storybook, on the "[toast container](https://5e53448165911c0022e68c74-ffhmauvxwg.chromatic.com/?path=/story/components-toast--toast-container-story)" story I can successfully create a toast but on the "[toaster](https://5e53448165911c0022e68c74-ffhmauvxwg.chromatic.com/?path=/story/components-toast--toaster-story)" story I cannot.

richbachman:
Thanks for the information. We think we found the issue, and will be fixing as soon as possible. We'll update this issue when done.

sebm:
Cool thanks @richbachman!

TheSisb:
PR up with a fix: https://github.com/twilio-labs/paste/pull/1557. Gonna try and merge it today.

sebm:
Nice one, thanks @TheSisb . I'd completely forgotten that I'd set reduced motion on my mac.

TheSisb:
I'm glad you forgot 💯  Helped us track this bug down! 🥇

TheSisb:
Fix was released in the latest version.

PR #1557, opened 2021-06-11
fix(toast): toasts now render when prefersReducedMotion is enabled

TheSisb:
Fixes https://github.com/twilio-labs/paste/issues/1542
Bug: Toasts never displayed when the system was set to prefers reduced motion


richbachman:
Code looks good. Wondering now if we need to do something similar with modal and alert dialog?

TheSisb:
> Code looks good. Wondering now if we need to do something similar with modal and alert dialog?

Good question. We don't because the bug presents itself when using the async enter function. It works with object notation.  

Modal uses the object notation:
image

And my fix here is to use the object notation when prefers reduced motion is enabled. To do so, I have to omit height from the equation or it won't work.

This is a bug in React-spring. It is likely fixed in a later version so I added a FIXME to come back to this on v9+

PR #1567, opened 2021-06-14
fix(tabs): center align horizontal tabs always

TheSisb:
All horizontal tab labels are now center aligned. This PR also updates the stories to include inclusive content and VRT the center alignment change.


packages/paste-core/components/tabs/__tests__/tabs.test.tsx
```diff
@@ -2,21 +2,11 @@ import * as React from 'react';
 import {render, screen} from '@testing-library/react';
 // @ts-ignore typescript doesn't like js imports
 import axe from '../../../../../.jest/axe-helper';
-import {HorizontalTabsExample, VerticalTabsExample, StateHookExample} from '../stories/index.stories';
+import {HorizontalTabs, StateHookTabs} from '../stories/index.stories';
```
richbachman, line 5: 🔥

packages/paste-core/components/tabs/stories/index.stories.tsx
```diff
@@ -76,59 +186,63 @@ const useButtonClickTabState = (): TabStateReturn => {
...
-export default {
-  title: 'Components/Tabs',
-  excludeStories: ['HorizontalTabsExample', 'FittedTabsExample', 'VerticalTabsExample', 'StateHookExample'],
-  component: Tabs,
-  subcomponents: {TabList, Tab, TabPanels, TabPanel},
-};
-
-export const HorizontalTabs = (): React.ReactNode => {
-  return <HorizontalTabsExample />;
+// @ts-expect-error story
```
richbachman, line 214: Not sure whats better here. Having the `@ts-expect-error` or having the extra code to exclude the `...Example` stories. Thoughts?

richbachman:
Code and VRT look good. Just that one question about the stories.

packages/paste-core/components/tabs/stories/index.stories.tsx
TheSisb: I think I prefer TS hints to say "this is fine" over adding more code, but I don't really have a strong opinion.

richbachman: I don't really have a strong opinion on this either. We need a tie-breaker @SiTaggart.

SiTaggart: This may not be a true tie-breaker.

In other stories we type the functional component to return `React.ReactNode`, which is kinda icky but gets around the type error.

In other places like tests I'm pretty partial to not having to do really extreme typescript things it's being picky about because they're tests but, I'm not sure if that's "artisan" and won't bite us in the bum. In those instances where we explicitly expect the type error, I always forget `@ts-expect-error` and I think it's probably the best way to go.

Finally, amusing the `.story` annotations are deprecated anyway https://github.com/storybookjs/storybook/blob/next/MIGRATION.md#hoisted-csf-annotations

TheSisb: Lol. I guess this is fine then? We could also create our own re-useable type in the future:
```js
type Story<P> = React.FC<P> & {
  story: {}
  // In the future, the other ones they added to replace story
}
```

richbachman (approved):

SiTaggart (approved):

PR #1599, opened 2021-06-23
feat: add ability for Box and Text to respond to element customization from the customization provider

SiTaggart:
# Customization core API enablement

## Tagging elements 

Theming is limited to the design tokens provided by Paste, and where they are used in each component. You may want to customize specific things about certain components throughout the application outside of that constraint.

To do so we need to “tag” the internal DOM elements we wish to expose to the customer for customizing.

Box and Text will both have an element prop that comes with an overridable default element name “BOX” and “TEXT”

The element prop will also sprout a DOM attribute `data-paste-element=”BOX”` for identification in the DOM

Each Paste component can then use that element prop from Box and Text to tag the internal DOM nodes we wish to expose by giving them custom element names

image

## Elements props on the customization provider

Elements prop on the Customization Provider allows you to Customize components based on their element names.

If you provide an object to elements and a top level key matches the name of an existing element, the styles under that matching key will be merged into the base styles of that element.

```jsx
<CustomizationProvider
  baseTheme="default"
  elements={{
    CUSTOM_BOX: 
    {
      padding: 'space30',
      borderStyle: 'solid',
      borderWidth: 'borderWidth10',
      borderColor: 'pink',
      variants: {
        primary: {
          backgroundColor: 'colorBackgroundWarning'
        }
      }
    }
  }}
>
  <CustomizableBoxExample element="CUSTOM_BOX">
    Custom Box
  </CustomizableBoxExample>
  <CustomizableBoxExample element="CUSTOM_BOX" variant="primary">
    Custom Box
  </CustomizableBoxExample>
</CustomizationProvider>
```

## Consuming the elements prop

When Box and Text renders it can look at the theme object provided to it by Emotion and see if there are any custom elements. If one of the custom elements matches the element prop that is provided to Box from the component composition, then it will grab the resultant object and pass it through the Styled System CSS function. 

This is then “composed” into the Box styles after any style props have been set so that we can take advantage of Emotions style composition model 

Styles are merged using Emotions Composition https://emotion.sh/docs/composition

Deduping of styles is handled by the emotion babel transform at compile time

The object under each element key can be an Emotion CSS object style. The values of each CSS property can be any valid 
Design Token that is applicable to that CSS property based on the theme specification, or any valid CSS value for that CSS property.

Can handle pseudo selectors such as :focus, :hover and ::after too

image

## Handling variants

Components and their elements come in “variations”, such as primary and secondary button, error or warning alerts

The elements prop allows you to write specific customizations for variants of a particular component element by matching the variant name, under the variants key of an element style object.

The variant prop of the component must be passed to the Box or Text that is tagged as an element

When Box and Text render, they can again inspect the provided theme and check for elements that have variants supplied. If they do, it passes that object through the Styled System Variants utility

All the same rules apply for variant styles as they do for element styles

image

## Enabling custom component instances (one offs)

The element prop should always be overridable for all components. 

By giving the component a new element name, you can use the new element name on the elements prop to style a custom, one off version of the component

Internally tagged elements of a component should derive their names from the top level element name `<Box element={`${props.element}_FOO`} />`

The custom element name, when used as a key on the elements prop acts in exactly the same way as any other component from Paste

- CSS Object
- Design tokens
- pseudo selectors
- Variants

Because the element name only matches that single instance, the customization is not global to the application. Only the one matching element

In the same way that the base theme is immutable, so are the base element styles that you are extending from.

Emotion, using composition, will create a new, unique instance of your custom styles, and a unique css classname. Your custom component has no chance of affecting the styles of the base component you’ve extended as they remain unmutated, and will not clash with any other component on the page.

image

# other things included 

- Swapped out lodash.merge for deepmerge
  - Ideally this would be a shared library package between button, customization, text and box but it's an old commonJS module and I'm struggling to get it compiled and usable across all our tools (jest, components, storybook, website, next)
- Replaces old snapshot tests on box 
- Replaces old snapshot tests on text 
- Bunch of component snapshot tests


SiTaggart:
> +6.15 kB (+422%)

Seems wildly excessive for a merge function... 🤔

I have an alternative but it's not as nice.

packages/paste-core/primitives/box/__tests__/StyleFunctions.test.tsx
```diff
@@ -0,0 +1,174 @@
...
+            borderColor: 'colorBorderError',
+          },
+        },
+      },
+    },
+  },
+};
+
+describe('getPseudoStyles', () => {
+  it('should return resolved object styles when called with psuedo selector props set', () => {
```
gloriliale, line 50:
```suggestion
  it('should return resolved object styles when called with pseudo selector props set', () => {
```

packages/paste-customization/src/types/CustomTheme.ts
```diff
@@ -2,5 +2,7 @@ import type {GenericThemeShape} from '@twilio-paste/theme';
 import type {PasteCustomCSS} from './PasteCSS';
 
 export type CustomTheme = Partial<GenericThemeShape> & {
-  elements?: PasteCustomCSS;
+  elements?: {
+    [key: string]: PasteCustomCSS;
```
TheSisb, line 6: is this supposed to be `{[key: string]: PasteCustomCSS} | PasteCustomCSS` or did I read that wrong?

richbachman (approved):
Pulled down, and everything is looking good. Other than outstanding questions, the code, tests, and stories lgtm.

packages/paste-core/components/button/src/DestructiveButton.tsx
```diff
@@ -6,13 +6,13 @@ import type {DirectButtonProps} from './types';
 import {DirectButtonPropTypes} from './proptypes';
 
 // This module can only be referenced with ECMAScript imports/exports by turning on the 'esModuleInterop' flag and referencing its default export
-const merge = require('lodash.merge');
+const merge = require('deepmerge');
```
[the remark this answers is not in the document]
SiTaggart: Yeah, When I was trying to make a library out of it I was getting tempted to just copy paste the source and typescript it. But we'd need tests as well so I backed out for now.

packages/paste-core/primitives/box/__tests__/StyleFunctions.test.tsx
```diff
@@ -0,0 +1,174 @@
...
+        },
+      },
+    },
+  },
+};
+
+describe('getPseudoStyles', () => {
+  it('should return resolved object styles when called with pseudo selector props set', () => {
+    const cssFunc = getPseudoStyles({_hover: {padding: 'space10', textDecoration: 'underline'}});
+    if (typeof cssFunc === 'function') {
```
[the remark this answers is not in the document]
SiTaggart: It's more to get around typing. `getPseudoStyles` returns either a css function or an empty object. I only want to test the css function here, but typescript is all "BUT IT MIGHT BE AN EMPTY OBJECT!!?!!"

packages/paste-customization/src/types/CustomTheme.ts
SiTaggart: Nah, it was me doing a boo boo. Elements should always have a top level key, which is the name of the element.

```
elements={{
  THE_ELEMENT: {...theCSS},
  CARD: {...theCardCSS},
  ALERT: {...theAlertCSS}
}}
```

packages/paste-core/primitives/text/src/index.tsx
```diff
@@ -1,137 +1,61 @@
...
-    pseudoStyles[PseudoPropStyles[pseudoProp]] = props[pseudoProp];
-  });
-
-  return css(pseudoStyles);
-};
-
-/* eslint-disable emotion/syntax-preference */
-// @ts-ignore
-export const Text = styled.span(
+// @ts-ignore can't work out how to stop the styled div color prop from emotion clashing with our color style prop in TextProps
```
gloriliale, line 18: Does this mean that using TS ignore mitigates this clash?

packages/paste-core/primitives/text/__tests__/StyleFunctions.spec.tsx
```diff
@@ -0,0 +1,174 @@
...
+      });
+    }
+    const primaryAlertProps = {
+      'data-paste-element': 'ALERT',
+      variant: 'foo',
+      ...mockTheme,
+    };
+    // @ts-expect-error because I'm not setting the whole theme
+    const alertCSSFun = getCustomElementStyles(primaryAlertProps);
+    if (typeof alertCSSFun === 'function') {
```
gloriliale, line 165: Should this be `alertCSSFunc` to pattern match the above?

gloriliale (approved):
I mostly understand what's going on! And figured out that I was confused right off the block bc of `data-paste-element` vs. `element` before you added the extended explanation. It also would've helped if I'd realized that what I was first looking at were snapshots 😬. Anyway, this is great! People are gonna love it!

packages/paste-core/primitives/box/src/index.tsx
```diff
@@ -10,216 +10,38 @@ import {
...
+  compose(space, layout, flexbox, background, border, boxShadow, position, typography, PasteStyleProps),
+  getPseudoStyles,
+  getCustomElementStyles
+) as StyledComponent<
+  Omit<React.DetailedHTMLProps<React.HTMLAttributes<HTMLElement>, HTMLElement>, 'color'>,
+  BoxProps,
+  Record<string, unknown>
+>;
+
+const Box = React.forwardRef<HTMLElement, BoxProps>(({children, element = 'BOX', ...props}, ref) => {
```
TheSisb, line 38: I'm a little curious about your rationale for these `element` props to be all caps. Is it because these are case sensitive and we want to avoid mis-typing things? Or is it rather because we want these to read as constants?

TheSisb (approved):

packages/paste-core/primitives/box/src/index.tsx
SiTaggart: Going with case insensitive but making sure you can't have duplicates of the same name in different casings, just in case you get weird cross platform behaviors.

packages/paste-core/primitives/text/__tests__/StyleFunctions.spec.tsx
SiTaggart: I'm not sure I follow on this one? Can you expand on it, please?

packages/paste-core/primitives/text/src/index.tsx
SiTaggart: Yes, it's not ideal though.

PR #1603, opened 2021-06-24
chore(truncate): add required title prop to Truncate

andipants12:
Added a required title `prop` to the `Truncate` component API, intended to represent the full truncated text.

### DIFF
--- a/.changeset/tricky-spiders-matter.md
+++ b/.changeset/tricky-spiders-matter.md
@@ -1,0 +1,7 @@
+---
+'@twilio-paste/truncate': major
+'@twilio-paste/core': major
+---
+
+[Truncate] Added the title prop to show full non-truncated text
+BREAKING CHANGE: `title` prop is now required for the Truncate component.
--- a/packages/paste-core/components/truncate/src/index.tsx
+++ b/packages/paste-core/components/truncate/src/index.tsx
@@ -4,6 +4,7 @@ import {Box, BoxProps} from '@twilio-paste/box';
 
 export interface TruncateProps extends Omit<React.HtmlHTMLAttributes<HTMLSpanElement>, 'color'>, Pick<BoxProps, 'as'> {
   children: NonNullable<React.ReactNode>;
+  title: NonNullable<string>;
 }
 
 const Truncate = React.forwardRef<HTMLSpanElement, TruncateProps>(({children, ...props}, ref) => {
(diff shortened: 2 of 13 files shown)


packages/paste-website/src/components/icons-list/IconTile.tsx
```diff
@@ -50,7 +50,7 @@ const IconTile = React.forwardRef<HTMLButtonElement, IconTileProps>(({children,
       {children}
       <Box as="div" display="flex" color="colorTextWeak" marginTop="space30" maxWidth="100%">
         <ScreenReaderOnly>Show details for:</ScreenReaderOnly>
-        <Truncate>{icon.name.replace('Icon', '')}</Truncate>
+        <Truncate title={icon.name}>{icon.name.replace('Icon', '')}</Truncate>
```
richbachman, line 53: Do we need to pass `icon.name` to the replace function in the `title` as well?

andipants12: Wasn't sure about that, thanks for starting a thread. It looks like `icon.name` always has an `Icon` suffix that is removed by the replace function (so I'm guessing "ChatIcon"). 
I used the full icon name thinking that including the suffix would give the title more descriptive clarity, since the goal of the ticket was to use the full text that is being truncated.
Is that repetitive or unnecessary in this case?

richbachman: > ...including the suffix would give the title more descriptive clarity...

That makes sense to me. Keeping Icon there is fine with me since that is the full name of each.

packages/paste-core/components/truncate/src/index.tsx
```diff
@@ -4,6 +4,7 @@ import {Box, BoxProps} from '@twilio-paste/box';
 
 export interface TruncateProps extends Omit<React.HtmlHTMLAttributes<HTMLSpanElement>, 'color'>, Pick<BoxProps, 'as'> {
   children: NonNullable<React.ReactNode>;
+  title: NonNullable<string>;
```
TheSisb, line 7: nit: you can omit the NonNullable here. https://www.typescriptlang.org/docs/handbook/utility-types.html#nonnullabletype

.changeset/tricky-spiders-matter.md
```diff
@@ -0,0 +1,6 @@
+---
+'@twilio-paste/truncate': major
+'@twilio-paste/core': major
+---
+
+[Truncate] Added the title prop to show full non-truncated text
```
richbachman, line 6: Let's add something about this being a breaking change, and what teams will need to change.

richbachman (approved):
Just that small changeset change, then this looks good.

SiTaggart (approved):

PR #1606, opened 2021-06-25
feat(popover): add state hook prop

richbachman:
- [x] Added `state` prop to Popover
- [x] Added tests and stories
- [x] Added example to Popover docs page
- [x] Expanded Disclosure state hook prop examples based on https://github.com/twilio-labs/paste/discussions/1578


packages/paste-website/src/pages/components/disclosure/index.mdx
```diff
@@ -430,19 +459,69 @@ const PopoverExample: React.FC = () => {
...
+
+Whether it's animating or not.
+
+##### `toggle?: () => void`
+
+Toggles the visible state
+
+##### `stopAnimation?: () => void`
+
+Stops animation. It's called automatically if there's a CSS transition.
```
TheSisb, line 510: Should we bother mentioning this since we control the styles and we don't use CSS animations in this disclosure afaik?

packages/paste-website/src/pages/components/disclosure/index.mdx
```diff
@@ -430,19 +459,69 @@ const PopoverExample: React.FC = () => {
...
+##### `animated?: number | boolean`
+
+If true, animating will be set to true when visible is updated. It'll wait
+for stopAnimation to be called or a CSS transition ends. If animated is set
+to a number, stopAnimation will be called only after the same number of
+milliseconds have passed.
+
+##### `animating?: boolean`
+
+Whether it's animating or not.
```
TheSisb, line 502: Does this work with our animations?

packages/paste-website/src/pages/components/disclosure/index.mdx
richbachman: Yeah. Good catch. Since we're using our animation library here, this won't really work.

packages/paste-website/src/pages/components/disclosure/index.mdx
richbachman: Just verified. It doesn't since we're using our animation library.

packages/paste-website/src/component-examples/DisclosureExamples.ts
```diff
@@ -36,3 +36,36 @@ render(
   <StateHookDisclosure />
 )
 `.trim();
+
+export const StateHookToggleExample = `
+const StateHookToggleDisclosure = () => {
+  const [show, setShow] = React.useState(false);
+  const disclosure = useDisclosureState();
+  const toggleDisclosure = () => setShow(!show);
```
andipants12, line 44: NIT: Line 44 might be a good case to use functional updates.
e.g. `const toggleDisclosure = () => setShow(state => !state);` where the `state` argument refers to the most current value of `show`. This is useful to ensure that the state value `show` is not stale.
More information in the [React hooks docs](https://reactjs.org/docs/hooks-reference.html#functional-updates).

packages/paste-core/components/popover/src/PopoverContainer.tsx
```diff
@@ -1,17 +1,34 @@
...
-};
+export interface PopoverContainerProps extends NonModalDialogPrimitivePopoverInitialState {
+  children: NonNullable<React.ReactNode>;
+  state?: PopoverStateReturn;
+}
+
+const PopoverContainer = React.forwardRef<HTMLDivElement, PopoverContainerProps>(
+  ({children, gutter, placement, state, ...initialState}, ref) => {
+    const popover = state || useNonModalDialogPrimitiveState({gutter, modal: true, placement, ...initialState});
+    const value = React.useMemo(() => popover, Object.values(popover));
```
andipants12, line 22: ❓ If `state` is passed would this memoized value re-compute on each re-render of the `PopoverContainer`? Asking because I'm guessing `state` is and object, and I think objects are re-instantiated on all re-renders. Would it make sense to instead memoize the container itself using `React.memo`?

packages/paste-core/components/popover/__tests__/index.spec.tsx
```diff
@@ -1,71 +1,81 @@
...
+      }
+      // expect(popover.getAttribute('hidden')).toBeNull();
+      expect(popover).toBeVisible();
+
+      ButtonTwo.click();
+      popover = screen.queryByTestId('state-hook-popover');
+      if (popover === null) {
+        return;
+      }
+      // expect(popover.getAttribute('hidden')).toBeDefined();
```
SiTaggart, line 67: couple commented out lines here

richbachman: Thanks for catching those.

packages/paste-core/components/popover/src/PopoverContainer.tsx
richbachman: You're right. `state` is an `object`, so yeah memoizing the container makes sense. I'll make that change.

packages/paste-website/src/pages/components/popover/index.mdx
```diff
@@ -142,6 +143,44 @@ available in the [props description here](#popover-container).
...
+to `PopoverContainer` as the `state` prop.
+
+This allows you to use certain returned props from the state hook, including functions like `hide` and `show`.
+
+An example usecase of this might be programmatically providing the user a popover when a certain user action is taken.
+In the example below we are showing a popover when another button is pressed. When pressed the button uses the `show`
+function from the hook. Another button using the `hide` function from the hook is also provided to hide the popover
+when pressed.
+
+It should be noted that when doing so, the state prop takes precident over the other properties that affect
```
gloriliale, line 161:
```suggestion
It should be noted that when doing so, the state prop takes precedent over the other properties that affect
```

gloriliale (approved):
Looks great! Definitely good to include the links to the Reakit Docs in the state examples sections of our docs, helped me understand the hooks more clearly and know that I could customize more actions if I wanted.

packages/paste-core/components/popover/src/PopoverContainer.tsx
```diff
@@ -1,17 +1,35 @@
...
+    const popover = state || useNonModalDialogPrimitiveState({gutter, modal: true, placement, ...initialState});
+    return (
+      <PopoverContext.Provider value={popover}>
+        <Box ref={ref}>{children}</Box>
+      </PopoverContext.Provider>
+    );
+  }
+);
+
+const PopoverContainer = React.memo(BasePopoverContainer);
```
SiTaggart, line 30: I honestly don't know the answer to this, but does wrapping the component in memo do anything weird to the ref, like not forward it? Does it still work?

SiTaggart (approved):
Outside that one question on refs, it looks good to me.

packages/paste-core/components/popover/src/PopoverContainer.tsx
richbachman: Did some searching and didn't find anything saying NOT to do wrap components that use `forwardRef` in `memo()`. I also tested locally, and the `memo` function seems to have no effect on the `ref`. So I think we're good to go.

SiTaggart: 👍 Bonza

PR #1607, opened 2021-06-25
feat(time-picker): create package

gloriliale:
✅ created a Time Picker package
 ✅ wrote tests for Time Picker
 ✅ made stories for Time Picker
- [x] addressed the seconds/milliseconds and API question

### DIFF
--- a/packages/paste-core/components/input/src/Input.tsx
+++ b/packages/paste-core/components/input/src/Input.tsx
@@ -2,9 +2,9 @@ import * as React from 'react';
 import * as PropTypes from 'prop-types';
 import {styled, css} from '@twilio-paste/styling-library';
 import {InputBox} from '@twilio-paste/input-box';
+import type {InputBoxTypes} from '@twilio-paste/input-box';
 import {safelySpreadFormControlProps} from './utils';
 
-export type InputTypes = 'text' | 'email' | 'hidden' | 'number' | 'password' | 'search' | 'tel' | 'date';
 export type InputVariants = 'default' | 'inverse';
 
 export interface InputProps extends React.InputHTMLAttributes<HTMLInputElement> {
@@ -21,14 +21,14 @@ export interface InputProps extends React.InputHTMLAttributes<HTMLInputElement> 
   required?: boolean;
   size?: never;
   style?: never;
-  type: InputTypes;
+  type: InputBoxTypes;
   value?: string;
   variant?: InputVariants;
   width?: never;
 }
 
 interface TypeProps {
-  type: InputTypes;
+  type: InputBoxTypes;
   inputmode?: string | undefined;
   pattern?: string | undefined;
 }
@@ -55,7 +55,7 @@ export const InputElement = styled.input<InputProps>((props) =>
     paddingTop: 'space30',
     resize: 'none',
     width: '100%',
-    cursor: props.type === 'date' && !props.readOnly && !props.disabled ? 'text' : 'auto',
+    cursor: (props.type === 'date' || props.type === 'time') && !props.readOnly && !props.disabled ? 'text' : 'auto',
 
     '&::placeholder': {
       color: props.variant === 'inverse' ? 'colorTextInverseWeak' : 'colorTextWeak',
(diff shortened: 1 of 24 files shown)


packages/paste-core/components/time-picker/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,125 @@
...
+    expect(formatReturnTime(TEST_TIME, 'hh:mm')).toEqual('01:30');
+    expect(formatReturnTime(TEST_TIME, 'HH:mm:ss')).toEqual('13:30:04');
+    expect(formatReturnTime(TEST_TIME, 'hh:mm:ss')).toEqual('01:30:04');
+    expect(formatReturnTime(TEST_TIME, 'HH:mm:ss.SSS')).toEqual('13:30:04.000');
+    expect(formatReturnTime(TEST_TIME, 'HH:mm:ss.SSS')).toEqual('13:30:04.000');
+    expect(formatReturnTime(TEST_TIME, 'hh:mm:ss.SSS aa')).toEqual('01:30:04.000 p.m.');
+    // testing midnight
+    expect(formatReturnTime('00:00', 'hh:mm A')).toEqual('12:00 AM');
+    // test for edge cases (passing in time+date value) to throw an error
+    expect(() => {
```
andipants12, line 31: Have you considered breaking down this `it` statement? In my experience, it's easier to debug test failures when there are less `expect` statements within an `it` block. For example, we could move the edge cases and the expected thrown error statement to a new block.

packages/paste-core/components/time-picker/src/TimePicker.tsx
```diff
@@ -0,0 +1,29 @@
+import * as React from 'react';
+import * as PropTypes from 'prop-types';
+import type {InputProps} from '@twilio-paste/input';
+import {Input} from '@twilio-paste/input';
+
+export type TimePickerProps = Omit<InputProps, 'type' | 'insertBefore' | 'insertAfter' | 'step'>;
+
+// @ts-ignore to remove step from props so it is not used
```
andipants12, line 8: What is the reported TS error here? I'm wondering if we can find a resolution rather than using `ts-ignore`.

packages/paste-core/components/time-picker/src/utils.ts
```diff
@@ -0,0 +1,37 @@
+// eslint-disable-next-line import/no-extraneous-dependencies
```
andipants12, line 1: Is this eslint error still being reported? I see that `date-fns` is in this component's `package.json` `dependencies`. AFAIK this error should only report if it is not listed.

packages/paste-core/components/input/src/Input.tsx
```diff
@@ -4,7 +4,7 @@ import {styled, css} from '@twilio-paste/styling-library';
 import {InputBox} from '@twilio-paste/input-box';
 import {safelySpreadFormControlProps} from './utils';
 
-export type InputTypes = 'text' | 'email' | 'hidden' | 'number' | 'password' | 'search' | 'tel' | 'date';
+export type InputTypes = 'text' | 'email' | 'hidden' | 'number' | 'password' | 'search' | 'tel' | 'date' | 'time';
```
richbachman, line 7: Could these be imported from input-box? That would save some duplication.

nkrantz:
@TheSisb Test 123....

packages/paste-core/components/input/src/Input.tsx
nkrantz: Good catch! Done

packages/paste-core/components/time-picker/src/TimePicker.tsx
nkrantz: I added a better comment to try to address this! Also changed it to a `@ts-expect-error`. It's because we're omitting `step` from the type but still destructuring it from the time picker props.

packages/paste-core/components/time-picker/src/utils.ts
nkrantz: Good catch! Removed from `time-picker` and `date-picker`.

packages/paste-core/components/time-picker/__tests__/index.spec.tsx
nkrantz: Thanks for noticing this! I pulled out the last 2 expects in `time-picker` and `date-picker`.

nkrantz (approved):

richbachman (approved):

PR #1638, opened 2021-07-02
fix(site-nav): Hide Sibling Box from site left nav

zahnster:
* Runs 'Primitives' section of the doc site nav through filteredComponents fn to filter out unwanted page links.


richbachman (approved):

TheSisb (approved):
v nice, didn't realize we had that function! good catch

PR #1641, opened 2021-07-06
feat(skeleton-loader): add skeleton loader package

richbachman:
- [ ] Skeleton loader package
- [ ] Skeleton loader tests
- [ ] Skeleton loader stories
- [ ] Skeleton loader to core
- [ ] Skeleton loader doc page


packages/paste-core/components/skeleton-loader/src/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+      borderTopRightRadius,
+      display,
+      disabled,
+      height = 'sizeIcon20',
+      margin,
+      marginBottom,
+      marginLeft,
+      marginRight,
+      marginTop,
+      marginX = null,
```
TheSisb, line 40: Why do we need to explicitly set these to null?

packages/paste-core/components/skeleton-loader/src/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+      config: {
+        mass: 0.1,
+        tension: 80,
+        friction: 50,
+      },
+    });
+
+    return (
+      <Box
+        {...safelySpreadBoxProps(props)}
```
TheSisb, line 66: Would it be simpler if we didn't safelySpread here and just passed all the props we weren't setting values for? I'm not sure what the concern is here

packages/paste-website/src/pages/components/skeleton-loader/index.mdx
```diff
@@ -0,0 +1,233 @@
...
+
+<content>
+
+## Guidelines
+
+### About Skeleton Loader
+
+Skeleton Loader displays a placeholder of a page's content before the data gets loaded. This helps reduce load-time frustration by making it feel like loading is happening immediately.
+
+It's an alternative to the traditionaly spinner loading methods. Rather than showing an abstract spinner, it creates anticipation of whats to come and reduces congitive load.
```
TheSisb, line 96:
```suggestion
The Skeleton Loader displays a placeholder representing the page's content before the page is ready to display. This component is used to improve [perceived performance](https://developer.mozilla.org/en-US/docs/Learn/Performance/Perceived_performance). This tends to improves user experience by reducing load time frustration and making the page feel more responsive.

This loader is an alternative to the [Spinner component](/components/spinner). Rather than showing an abstract spinner, it frames the content of what is to come which creates anticipation.
```

TheSisb: Took a stab at clarifying and mentioning perceived performance.

packages/paste-website/src/pages/components/skeleton-loader/index.mdx
```diff
@@ -0,0 +1,233 @@
...
+
+The component is designed to be used in place of the content being loaded. For example:
+
+<LivePreview scope={{SkeletonLoader, Text}} noInline>
+  {CompositionExample}
+</LivePreview>
+
+### Accessibility
+
+- The animation can be disabled using the `disabled` prop, or by enabled `prefers-reduced-motion` at the OS level.
```
TheSisb, line 106:
```suggestion
- The animation can be disabled using the `disabled` prop, or by enabling the `prefers-reduced-motion` setting at the OS level.
```

packages/paste-website/src/pages/components/skeleton-loader/index.mdx
```diff
@@ -0,0 +1,233 @@
...
+- The animation should be disabled if content takes longer than 5 seconds to load.
+- Skeleton Loader has the `aria-busy` attribute set to `true` by default. This is to indicate that the content is loading. This can be set to `false` using the `disabled` prop.
```
TheSisb, line 108:
```suggestion
- Skeleton Loader has the `aria-busy` attribute set to `true` by default. This is to indicate that the content is loading. This is reset to `false` when `disabled` is `true`.
```

packages/paste-website/src/component-examples/SkeletonLoaderExamples.ts
```diff
@@ -0,0 +1,338 @@
...
+            <Th width="300px">Contacts</Th>
+            <Th>Date Added</Th>
+            <Th>Last Opened</Th>
+          </Tr>
+        </THead>
+        <TBody>
+          <Tr>
+            <Td>{loaded ? <Text as="span">Adriana Lovelock</Text> : <SkeletonLoader width="113px" />}</Td>
+            <Td>{loaded ? <Text as="span">2020-09-17</Text> : <SkeletonLoader width="86px" />}</Td>
+            <Td>{loaded ? <Text as="span">2020-06-28</Text> : <SkeletonLoader width="86px" />}</Td>
```
TheSisb, line 302: Setting perfect px values is a little unrealistic. Maybe we can set these to randomized (between like 60-100) % values instead?

packages/paste-core/components/skeleton-loader/src/index.tsx
SiTaggart: If the intention is only to allow Layout, Margin, BorderRadius props and HTML attributes as the possible API, then safely spreading props makes sense to me.

It feels awkward but I'm not sure we have a good solution for "I want these, but not those style props, plus all HTML attributes". We only have a solution for "I don't want any style props, but all HTML attributes"

packages/paste-core/components/skeleton-loader/src/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+const StyledAnimatedSkeleton = styled(AnimatedSkeleton)(() =>
+  css({
+    background: `linear-gradient(90deg, transparent, rgba(255, 255, 255, 0.3) 40%, rgba(255, 255, 255, 0.3) 60%, transparent)`,
+  })
+);
+
+export interface SkeletonLoaderProps
+  extends React.HTMLAttributes<HTMLDivElement>,
+    LayoutProps,
+    MarginProps,
```
SiTaggart, line 19: I wonder if this is a good thing or a bad thing?

Should we make you use a layout like stack to set the gap between skeleton boxes?

I get that the reason we do individual is likely so that you can replicate the spacing between a heading and a paragraph to a box, which might be specific to the component you are mimicking.

But if we ever change those, your skeleton spacing in your product is going to go out of sync and look different to the component spacing anyway (unless you are willing to constantly go around nudging all your skeletons every time we adjust spacing). So why not just simplify this component? 🤔

packages/paste-core/components/skeleton-loader/src/index.tsx
```diff
@@ -0,0 +1,114 @@
...
+import {css, styled} from '@twilio-paste/styling-library';
+import {Box, safelySpreadBoxProps} from '@twilio-paste/box';
+import type {LayoutProps, MarginProps, BorderRadiusProps} from '@twilio-paste/style-props';
+
+const AnimatedSkeleton = animated(Box);
+
+const StyledAnimatedSkeleton = styled(AnimatedSkeleton)(() =>
+  css({
+    background: `linear-gradient(90deg, transparent, rgba(255, 255, 255, 0.3) 40%, rgba(255, 255, 255, 0.3) 60%, transparent)`,
+  })
```
SiTaggart, line 13: I wish we had a better solution for these types of situations. They aren't using system values, and compound css values are really difficult to systemize (but potentially not impossible).

Something to think about for the future.

packages/paste-website/src/pages/components/skeleton-loader/index.mdx
```diff
@@ -0,0 +1,233 @@
...
+The component is designed to be used in place of the content being loaded. For example:
+
+<LivePreview scope={{SkeletonLoader, Text}} noInline>
+  {CompositionExample}
+</LivePreview>
+
+### Accessibility
+
+- The animation can be disabled using the `disabled` prop, or by enabled `prefers-reduced-motion` at the OS level.
+- The animation should be disabled if content takes longer than 5 seconds to load.
```
SiTaggart, line 107: I think we should bake this into the animation loop. People won't do it, and potentially won't know.

We can solve it by just making sure the animation loop is 5 seconds on, and 5 seconds off.

packages/paste-core/components/skeleton-loader/src/index.tsx
richbachman: If we don't the other margin props get overridden. Same with size, if not null, it will override height and width.

packages/paste-website/src/component-examples/SkeletonLoaderExamples.ts
richbachman: Are you saying have the width randomly set in the component, or just in this example use percentages instead of px values?

packages/paste-core/components/skeleton-loader/src/index.tsx
richbachman: > I get that the reason we do individual is likely so that you can replicate the spacing between a heading and a paragraph to a box, which might be specific to the component you are mimicking.

That's the only reason I was using the margins on this one, but that could also be done by wrapping a Box around it. But that would also lead to the same issue if the spacing changes, right? Spacing would be different so the skeletons would be off.

That said, I do think I like the simpler is better approach and just use a Box if you need some extra spacing.

packages/paste-core/components/skeleton-loader/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,103 @@
...
+describe('SkeletonLoader', () => {
+  it('should render', () => {
+    const {getByTestId} = render(<Default />);
+    expect(getByTestId('default-skeleton')).toBeDefined();
+    expect(getByTestId('default-skeleton')).toHaveStyleRule('background-color', 'colorBackgroundStrong');
+    expect(getByTestId('default-skeleton')).toHaveStyleRule('border-radius', 'borderRadius20');
+    expect(getByTestId('default-skeleton')).toHaveStyleRule('height', 'sizeIcon20');
+  });
+
+  it('should render layout prop styles', (): void => {
```
gloriliale, line 20: I don't think I've seen us use `(): void` in our tests before. What is it used for here?

gloriliale (approved):
Looks great, and I like that there are many examples on the docs page to clarify its use in many different cases!

packages/paste-website/src/pages/components/skeleton-loader/index.mdx
richbachman: Fought with this a bit today, and then realized that the animation was taking almost 5 seconds to fully finish, so I added a small delay before the loop starts.

packages/paste-core/components/skeleton-loader/__tests__/index.spec.tsx
richbachman: This is just setting the return type to void, which basically means the function doesn't return any data.

It may not be necessary since the `it` function from Jest comes with the void return type anyway, but its good practice to set the function return type. I think we've done it pretty sporadically throughout the monorepo.

packages/paste-website/src/pages/components/skeleton-loader/index.mdx
SiTaggart: Looks pretty neat actually, I dig it

packages/paste-core/components/skeleton-loader/src/index.tsx
```diff
@@ -0,0 +1,99 @@
...
+);
+
+export interface SkeletonLoaderProps
+  extends React.HTMLAttributes<HTMLDivElement>,
+    Omit<LayoutProps, 'verticalAlign'>,
+    BorderRadiusProps {
+  disabled?: boolean;
```
SiTaggart, line 20: The more I read the docs, especially when it's mentioned in the accessibility section, leads me to think that the disabled prop is kind of redundant. 

Animation is disabled by reduce motion and doesn't loop more that 5 seconds. So we're really not disabling anything anymore I don't think. We don't have a test for it either which is either a happy co-incidence or we probably don't really care about it enough for it to be useful.

Thought?

packages/paste-website/src/pages/components/skeleton-loader/index.mdx
```diff
@@ -0,0 +1,227 @@
...
+
+---
+
+<contentwrapper>
+
+<PageAside data={props.data.mdx} />
+
+<content>
+
+## Guidelines
```
SiTaggart, line 90: Similarly to Alerts, we're going to start adding a basic default example to the top of every component page so you see the component up front.

You could probably just get away with reusing the default example.

SiTaggart:
Very nice overall. Don't forget to include that changelog at the bottom of the docs page though.
<details><summary>Browser metadata</summary>





```
Path:      /components/skeleton-loader
Browser:   Edge 91.0.864.64 on Mac OS 10.15.7
Viewport:  1680 x 914 @2x
Language:  en-US
Cookies:   Enabled
```

[Open in BrowserStack](https://live.browserstack.com/dashboard#url=https%3A%2F%2Fdeploy-preview-1641--paste-docs.netlify.app%2Fcomponents%2Fskeleton-loader&os=OS+X&os_version=Catalina&browser=Edge&browser_version=91.0&resolution=1680x914&speed=1&start=true)



</details>

[Open Deploy Preview](https://deploy-preview-1641--paste-docs.netlify.app/components/skeleton-loader) · [Mark as Resolved](https://app.netlify.com/cdp/resolve?deployID=60e61e6f5bf1490007cd2b50&commentID=60e7b4d846bf752818417db8&resolution=resolved)

packages/paste-website/src/pages/components/skeleton-loader/index.mdx
```diff
@@ -0,0 +1,227 @@
...
+### Accessibility
+
+- The animation can be disabled using the `disabled` prop, or by enabling the `prefers-reduced-motion` setting at the OS level.
+- Skeleton Loader has the `aria-busy` attribute set to `true` by default. This is to indicate that the content is loading. This is reset to `false` when `disabled` is `true`.
+
+## Examples
+
+### Default Skeleton Loader
+
+The default Skeleton Loader uses the `sizeIcon20` token for the `height`, and `borderRadius20` for the `border-radius`. The `width` will be 100% unless a different `width` is specified.
```
serifluous, line 113: It'd be helpful to say what someone might use the default Skeleton Loader for (line of text, maybe?), and that way people can use that size as a reference point for scaling up/down their sizing.

packages/paste-website/src/component-examples/SkeletonLoaderExamples.ts
```diff
@@ -0,0 +1,344 @@
...
+`.trim();
+
+export const ButtonsExample = `
+const SkeletonLoaderButtons = () => {
+  const [loaded, setLoaded] = React.useState(false);
+
+  return (
+    <>
+      {loaded ? (
+        <Stack orientation="horizontal" spacing="space40">
```
andipants12, line 151: Is there a benefit to including the layout component in the ternary render? Alternatively we could move the `Stack` outside of the ternary since for the both conditions we are rendering the same container.

richbachman: Good catch. Not on this on. Just got a little too free with the copy/paste here.

richbachman: Oh wait. I actually do need the stack there. If its not, I have to wrap the Buttons and Skeletons in a fragment, which then counts as the only child, so I would lose the margins.

packages/paste-core/components/skeleton-loader/src/index.tsx
richbachman: I think the `disabled` prop is pretty redundant now that we have the animation fine tuned. I'm going to remove it, and if we get feedback about having it, we can always add it back in.

SiTaggart (approved):

packages/paste-website/src/pages/components/skeleton-loader/index.mdx
```diff
@@ -0,0 +1,233 @@
...
+
+<content>
+
+<LivePreview scope={{SkeletonLoader}}>{`<SkeletonLoader />`}</LivePreview>
+
+## Guidelines
+
+### About Skeleton Loader
+
+The Skeleton Loader displays a placeholder representing the page's content before the page is ready to display. This component is used to improve [perceived performance](https://developer.mozilla.org/en-US/docs/Learn/Performance/Perceived_performance). This tends to improves user experience by reducing load time frustration and making the page feel more responsive.
```
gloriliale, line 97: Linking to more details about Perceived Performance is a really good idea!

gloriliale, line 97:
```suggestion
The Skeleton Loader displays a placeholder representing the page's content before the page is ready to display. This component is used to improve [perceived performance](https://developer.mozilla.org/en-US/docs/Learn/Performance/Perceived_performance). This tends to improve user experience by reducing load time frustration and making the page feel more responsive.
```

packages/paste-website/src/pages/components/skeleton-loader/index.mdx
```diff
@@ -0,0 +1,233 @@
...
+### Changing the Skeleton Loader height
+
+Use the `height` prop to change the height of the Skeleton Loader.
+
+<LivePreview scope={{SkeletonLoader}}>{`<SkeletonLoader height="150px" />`}</LivePreview>
+
+### Loading text
+
+<LivePreview scope={{Box, Button, SkeletonLoader, Stack, Text}} noInline>
+  {TextExample}
```
gloriliale, line 129: I really like how these examples are interactive!!

packages/paste-website/src/pages/components/skeleton-loader/index.mdx
```diff
@@ -0,0 +1,233 @@
...
+
+### Default Skeleton Loader
+
+The default Skeleton Loader can be used to represent a line of text. It uses the `sizeIcon20` token for the `height`, and `borderRadius20` for the `border-radius`. The `width` will be 100% unless a different `width` is specified.
+
+<LivePreview scope={{SkeletonLoader}}>{`<SkeletonLoader />`}</LivePreview>
+
+### Changing the Skeleton Loader height
+
+Use the `height` prop to change the height of the Skeleton Loader.
```
gloriliale, line 122: Would it be useful to say here that the height could be in px _or_ an icon size? It wasn't clear to me on first glance. This could be solved by adding another example that uses icon size for the height. Is there any other type of value one could put into that prop or just those two?

packages/paste-website/src/pages/components/skeleton-loader/index.mdx
```diff
@@ -0,0 +1,233 @@
...
+<LivePreview scope={{SkeletonLoader}}>{`<SkeletonLoader height="150px" />`}</LivePreview>
+
+### Loading text
+
+<LivePreview scope={{Box, Button, SkeletonLoader, Stack, Text}} noInline>
+  {TextExample}
+</LivePreview>
+
+### Loading a paragraph
+
```
gloriliale, line 133: Would it be helpful to mention that using Stack with SkeletonLoader is a common combination for representing lines of text before showing the example?

packages/paste-website/src/component-examples/SkeletonLoaderExamples.ts
TheSisb: Just in this example

packages/paste-website/src/pages/components/skeleton-loader/index.mdx
richbachman: Added some clarifying text.

packages/paste-website/src/pages/components/skeleton-loader/index.mdx
richbachman: Added some clarifying text.

packages/paste-core/components/skeleton-loader/src/index.tsx
TheSisb: I more meant that we probably don't have any restrictions on style props for this other than the ones we're setting ourselves, which would override the spread. So like:
```
<Box {...props} backgroundColor="w/e">
```
Would allow passing any styles Box accepts, but makes sure our styles override.

packages/paste-core/components/skeleton-loader/src/index.tsx
TheSisb: Gradient tokens 👓

packages/paste-core/components/skeleton-loader/src/index.tsx
richbachman: This has changed a bit to only pass down Layout and BorderRadius props. Consumers can use our other layout components for any necessary margins. I feel like scoping this down to just those is best for this first iteration because that adds scope to this component, and removes potentially unnecessary complexity. We could always add in complexity in the future if we get that feedback.

packages/paste-core/components/skeleton-loader/src/index.tsx
richbachman: I like the idea of a gradient function. Maybe we could add that as library or shared utility function in the future.

PR #1655, opened 2021-07-08
fix(tooltip): prevent child props being overridden/swallowed by tooltip

SiTaggart:
When cloning the child wrapped by the tooltip, we are not mixing in the child props that are provided by the consumer.

This means there can be cases where provided props are not called on the trigger element when wrapped in a tooltip as
they no longer exist when we clone it.

Spreading the children props on the tooltip before we clone means they are present when we clone using the referenceProps

### DIFF
--- a/packages/paste-core/components/tooltip/__test__/index.spec.tsx
+++ b/packages/paste-core/components/tooltip/__test__/index.spec.tsx
@@ -70,8 +70,13 @@ describe('Tooltip', () => {
           </Tooltip>
         </Theme.Provider>
       );
-      await act(async () => screen.getByRole('button').focus());
-      await act(async () => fireEvent.click(document.activeElement));
+      await act(async () => {
+        await screen.getByRole('button').focus();
+      });
+      await act(async () => {
+        // @ts-expect-error yes, I know activeElement MIGHT be null, but it's not, OK?
+        await fireEvent.click(document.activeElement);
+      });
       expect(focusHandlerMock).toHaveBeenCalled();
       expect(clickHandlerMock).toHaveBeenCalled();
     });
--- a/packages/paste-core/components/tooltip/__test__/index.spec.tsx
+++ b/packages/paste-core/components/tooltip/__test__/index.spec.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import {render, screen} from '@testing-library/react';
+import {act, fireEvent, render, screen} from '@testing-library/react';
 import {Button} from '@twilio-paste/button';
 import {Theme} from '@twilio-paste/theme';
 // @ts-ignore typescript doesn't like js imports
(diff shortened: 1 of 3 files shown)


packages/paste-core/components/tooltip/src/index.tsx
```diff
@@ -49,7 +49,7 @@ const Tooltip = React.forwardRef<HTMLDivElement, TooltipProps>(({baseId, childre
   return (
     <>
       {React.Children.only(
-        <TooltipPrimitiveReference {...tooltip} ref={ref}>
+        <TooltipPrimitiveReference {...tooltip} ref={ref} {...children.props}>
```
TheSisb, line 52: I'm not sure we should destructure `children.props`, because I've never seen that done before. It may be undocumented API and unsafe.  Got any resources that mention this approach I can review?

SiTaggart: I got it directly from the Reakit docs example https://reakit.io/docs/tooltip/#abstracting

I don't think it's anything fancy. React children have props. If you're inside React.Children.only, there is only one child. So, props.children.props returns the props of the child you passed it.

TheSisb: Interesting - I guess it makes sense if we're checking for `React.Children.only`.

packages/paste-core/components/tooltip/__test__/index.spec.tsx
```diff
@@ -57,6 +57,31 @@ describe('Tooltip', () => {
...
+      );
+      await act(async () => {
+        await screen.getByRole('button').focus();
+      });
+      await act(async () => {
+        // @ts-expect-error yes, I know activeElement MIGHT be null, but it's not, OK?
+        await fireEvent.click(document.activeElement);
+      });
+      expect(focusHandlerMock).toHaveBeenCalled();
+      expect(clickHandlerMock).toHaveBeenCalled();
```
TheSisb, line 81: Can we test that the tooltip still appears too in here even though we override the focus and click event handlers?

SiTaggart: Mmmmmm yeah, good call. Added them in this test.

packages/paste-core/components/tooltip/__test__/index.spec.tsx
```diff
@@ -53,7 +53,40 @@ describe('Tooltip', () => {
       if (tooltip === null) {
         return;
       }
-      expect(tooltip.getAttribute('hidden')).toBeDefined();
+      expect(tooltip.getAttribute('hidden')).not.toBeNull();
```
SiTaggart, line 56: swapping to not be null because if it is null, to be defined is still true, always.

TheSisb (approved):

packages/paste-core/components/tooltip/src/index.tsx
andipants12, line 52: Does the tooltip primitive pass props to its children as well? If not I think this might end up causing some unexpected event handling. For example, if the child is an input, I think this would pass `onChange` to the tooltip primitive, instead of the input.

SiTaggart: It does both. `referenceProps` in the render function in the line below is both the tooltip and child props merged together. 

It makes sense, as the tooltip does need to know about what it's being placed on so it can keep track of it in the hook.

I does pass to the child because the tests pass. I could write another testing using a different type of child to prove there's no prop clash.

nkrantz (approved):
Looks good to me!

Tooltip is now spreading its child's props onto the `TooltipPrimitiveReference` to make sure that the child's props and event handlers don't get overridden by the Tooltip's.

richbachman (approved):

PR #1529, opened 2021-06-04
feat(website): add component overview information from airtable

richbachman:
- [x] Added new overview tables with information from AirTable
- [x] Added overview to component and pattern headers
- [x] Promoted remaining alpha and beta packages to production

PR #1661, opened 2021-07-13
feat(website): adding OpenGraph image cards

TheSisb:
This PR adds [OpenGraph](https://ogp.me/) images to the Paste component pages. This works by adding a `<meta property="og:image" content="https://url/.netlify/functions/opengraph/<component/path>" data-react-helmet="true">` tag to the `<head>` of component pages using [React-helmet ](https://www.npmjs.com/package/react-helmet).

This meta tag URL points to a serverless function hosted by [Netlify Functions](https://www.netlify.com/products/functions/). We've added a function that installs Google chrome and [puppeteer](https://github.com/puppeteer/puppeteer) to visit another URL to generate an image response.  You see, the `og:image` meta tag needs to point to an image asset, but we don't want to create these images manually. Instead, we pointed to an API endpoint that can generate an image for us on the fly, using Paste tokens.  How does it do that? We added a page to our doc site at `/opengraph` that accepts GET params to generate a little preview. Try `/opengraph/?path=components/avatar` to see what it looks like for the Avatar component.

So our little serverless function hits a webpage which renders some UI, it takes a screenshot, and returns it.  That way, when you link to a Paste component page, it will unfurl by checking for that meta tag, which will fire this whole operation and display the image we desire.

To run this locally you need Netlify access. Install netlify CLI (https://docs.netlify.com/cli/get-started/) and run `netlify dev` in the root of the Paste monorepo. It should spin up the serverless function and the website.  If you inspect element on a component page and look into the `<head>` tag, at the very bottom you should find a `meta` tag with `og:image`. The URL there you can visit directly in your browser to play around with and test the image preview functionality.

For this PR:
- The screengrab URL https://deploy-preview-1661--paste-docs.netlify.app/.netlify/functions/opengraph/components/alert
- The rendered opengraph card that gets screengrabbed: https://deploy-preview-1661--paste-docs.netlify.app/opengraph/?path=components/alert


TheSisb:
Not sure why the codesandbox build failed, but I don't think it's related to this PR as previous commits passed.

packages/paste-website/src/components/open-graph-card/AssetStatus.tsx
```diff
@@ -0,0 +1,39 @@
+import * as React from 'react';
```
nkrantz, line 1: These files seem like pretty close duplications of the status components Rich created in the [feat(website): add component overview information from airtable](https://github.com/twilio-labs/paste/pull/1529/files#) PR. Is there a reason they're being recreated here? (duped file: `packages/paste-website/src/components/component-status/AssetStatus.tsx`)

nkrantz (commented):
Attempt at summing up what's going on here:

- Detect the component being requested and ensure it's legit
- Depending on type of OS and environment, find and launch Chrome
- Set the viewport size, visit the component's opengraph page, take screenshot (opengraph page is built using queried airtable data and react components)
- Close the browser and return the `png` or a 500 error

netlify.toml
```diff
@@ -10,3 +15,7 @@
 
 [context.production.environment]
   GATSBY_ENVIRONMENT_CONTEXT = "production"
+
+[functions]
+  directory = "packages/paste-website/functions"
+  external_node_modules = ["chrome-aws-lambda", "puppeteer-core", "path-match"]
```
SiTaggart, line 21: Whoops, `path-match` can be removed from here and the package.json. It was from the example I was trying to get to work and have since removed

packages/paste-website/functions/opengraph.ts
```diff
@@ -0,0 +1,114 @@
...
+  return {
+    executablePath: await chromium.executablePath,
+    args: chromium.args,
+    headless: chromium.headless,
+  };
+}
+
+const handler: Handler = async (event) => {
+  const componentRequested = event.path.replace('/.netlify/functions/opengraph/', '');
+  // console.log('component requested:', componentRequested);
```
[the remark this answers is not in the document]
SiTaggart: Yeah, I tend to keep some logging for node things, because they are running on a server you don't get a web inspector to play with, only a console in Netlify. Just kind of useful to see what it's doing when it's invoked. Ideally it'd pass things to a logging service like datadog but that seems a little extra

nkrantz:
Oh also just thought of another question I had - should we add any tests to make sure these images are created successfully? Or to make sure all packages will successfully be found using the package name mapping?

packages/paste-website/src/components/open-graph-card/AssetStatus.tsx
SiTaggart: I wanted to make the icons and text bigger and couldn't be bother to make the existing one do that? 🤷🏼

packages/paste-website/src/constants.ts
```diff
@@ -16,6 +16,12 @@ export const HOMEPAGE_SECTION_OVERFLOW_OFFSET = 150;
 export const STICKY_COLUMN_OFFSET = 20 + PSA_ALERT_HEIGHT + SITE_MASTHEAD_HEIGHT;
 export const STICKY_SIDEBAR_OFFSET = PSA_ALERT_HEIGHT + SITE_MASTHEAD_HEIGHT;
 
+// TODO detect the netlify deployment url we're in otherwise use local
+// export const FUNCTIONS_URL = process.env.DEPLOYMENT_URL ? process.env.DEPLOYMENT_URL : 'http://localhost:9999';
+export const FUNCTIONS_URL = 'https://deploy-preview-1661--paste-docs.netlify.app';
```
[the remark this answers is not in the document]
SiTaggart: yeah that's why it's a TODO. Netlify environment variables are kind of super unhelpful, in that you only get access to then in the build sever, and they don't actually work in the site assets. 

Having said that, this has just reminded me that I created just the very thing we need on line 41 in this file 🤦🏼

SiTaggart:
> should we add any tests

Tests would be nice. We might be able to create a storybook story to VRT the OG image layout  from the page. Some unit tests might also be nice to ensure this doesn't break in the future I guess. Good call out @nkrantz

packages/paste-website/functions/opengraph.ts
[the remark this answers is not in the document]
TheSisb: Yeah I kind of left them commented because this type of thing is tedious to debug if someone happens, so they serve as kind of comments to help debug anything in the future.

packages/paste-website/src/constants.ts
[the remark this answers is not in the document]
TheSisb: Good catch, forgot to fix this.

packages/paste-website/src/pages/opengraph/index.tsx
```diff
@@ -0,0 +1,183 @@
...
+      }
+    }
+  }
+`;
+
+// Some packages can't be found just by humanizing the package name,
+// so I need to manually map them.
+const ManualPackageNameMapping: Record<string, string> = {
+  // This package needs to preserve one of the hyphens in the name.
+  '@twilio-paste/non-modal-dialog-primitive': 'Non-modal Dialog Primitive',
```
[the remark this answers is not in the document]
TheSisb: Yea :( However it is fairly rare and worst case, you just don't get fancy image previews which isn't too big of a deal.

TheSisb:
I added a cypress test and a couple stories for VRT

SiTaggart (commented):
The og: image url seems to have an undefined in it right now https://deploy-preview-1661--paste-docs.netlify.appundefined/.netlify/functions/opengraph/components/date-picker

TheSisb:
Looks like the Netlify env variables kept turning into undefined for some reason. I used a JS solution in RouteUtils instead. This seems like a better solution overall tbh. @SiTaggart

packages/paste-website/src/utils/RouteUtils.ts
```diff
@@ -18,6 +18,13 @@ export function getCurrentPathHash(): string {
   return window.location.hash;
 }
 
+export function getCurrentOrigin(): string {
+  if (!hasWindowObject()) {
+    return '';
+  }
+  return window.location.origin;
+}
+
```
SiTaggart, line 27: I'm now a little worried this might not work. If I use something like https://www.opengraph.xyz/, it doesn't get the image url because JavaScript needs to be downloaded and executed to update the meta tag after the DOM has loaded. So they grab the HTML and don't wait for JS to execute.

So any crawler that doesn't execute JS might not get our images.

But I do assume that most services do cater for the fact the this stuff might be reliant on client side rendering because React is a thing? 😕

SiTaggart (approved):
Outside of my worry for client side meta tag updating, the code looks great. Wonder if we can test Facebook and Twitter at least?

packages/paste-website/src/utils/RouteUtils.ts
SiTaggart: seems OK for FB https://developers.facebook.com/tools/debug/?q=https%3A%2F%2Fdeploy-preview-1661--paste-docs.netlify.app%2Fcomponents%2Fspinner

TheSisb: Twitter claims it isn't working on this site https://cards-dev.twitter.com/validator, but I don't know.  We're using react-helmet to set the tag anyways, so we need JS regardless. Hmm

TheSisb: According to the docs, it may just be because it's on a test deploy environment: https://developer.twitter.com/en/docs/twitter-for-websites/cards/guides/troubleshooting-cards

TheSisb: I made a change that will make the URL return the correct value on SSR https://github.com/twilio-labs/paste/pull/1661/commits/8dc97fa75df0abc52048864a2dbf707195cef591

cypress/integration/e2e/opengraph/index.spec.ts
```diff
@@ -0,0 +1,17 @@
+describe('OpenGraph Cards', () => {
+  it('renders opengraph cards correctly', () => {
+    cy.visit('/opengraph/?path=components/alert');
```
nkrantz, line 3: Hmm is it enough to just test that `alert` works? Just thinking about future components that get added and might not pass `getHumanizedNameFromPackageName()` or get added to `ManualPackageNameMapping`. Not sure what other edge cases might make the card creation fail, but would it be too time-consuming to check that each page works here? I don't think we'd have to test each of these things on each page (h1, svg, area label, font), but just that they work.

I wonder if there's any way to add the opengraph urls to the links that the broken link checker checks?

packages/paste-website/src/components/open-graph-card/AssetStatus.tsx
nkrantz: iiiiiiinteresting... 🤔

packages/paste-website/src/components/open-graph-card/types.ts
```diff
@@ -0,0 +1,11 @@
+export type PackageData = {
+  name: string;
+  description: string;
+  Feature: string;
+  Documentation: boolean;
+  Figma: string;
+  Code: string;
+  Design_committee_review: string | null;
+  Engineer_committee_review: string | null;
```
nkrantz, line 9: Why are these strings and not booleans?

packages/paste-website/src/pages/opengraph/index.tsx
```diff
@@ -0,0 +1,183 @@
...
+  const [packageType, packageName] = path.split('/');
+
+  if (!packageType || !packageName) {
+    return <div>Invalid request</div>;
+  }
+
+  function mergeAirtableDataForNode(node: NodeGQL, type: string): PackageData {
+    const {name, description} = node;
+
+    // If this package is in our manual mapping list, use that.
```
nkrantz, line 112: The comments are super helpful and this logic is really clear 👍 nice job

nkrantz (approved):
Looks great, stories and tests are a nice addition. Used your tips from our code review demo and it was super helpful!

packages/paste-website/src/components/open-graph-card/types.ts
TheSisb: I inherited these types from prior work on component statuses, but it bothered me too. Looks like it's because they're coming from Airtable's API this way and we aren't remapping them anywhere.

cypress/integration/e2e/opengraph/index.spec.ts
TheSisb: I added two VRT stories for Alert and Button to test the visual rendering works as intended, with every possible option for the statuses on top of this Cypress test.  Since this is one endpoint for every component, I believe testing just one endpoint would be alright as all the data is retrieved in the same way.

nkrantz: Makes sense ✔️

PR #1663, opened 2021-07-13
docs(website): add label package doc page

richbachman:
- [x] Added label package doc page
- [x] Removed label page from cypress ignore list

### DIFF
--- a/packages/paste-website/src/pages/components/label/index.mdx
+++ b/packages/paste-website/src/pages/components/label/index.mdx
@@ -1,0 +1,354 @@
+---
+title: Label - Components
+package: '@twilio-paste/label'
+description: The Label component is used to provide a visible, accessible name to any form element.
+slug: /components/label/
+---
+
+import {graphql} from 'gatsby';
+import {useUID} from '@twilio-paste/uid-library';
+import {Label} from '@twilio-paste/label';
+import {Input} from '@twilio-paste/input';
+import {HelpText} from '@twilio-paste/help-text';
+import {Anchor} from '@twilio-paste/anchor';
+import {Box} from '@twilio-paste/box';
+import {Text} from '@twilio-paste/text';
+import {Select, Option} from '@twilio-paste/select';
+import {TextArea} from '@twilio-paste/textarea';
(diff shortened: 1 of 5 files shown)


nkrantz (approved):
Looks great! Just a reminder to update the airtable before merging

packages/paste-website/src/pages/components/label/index.mdx
```diff
@@ -0,0 +1,354 @@
...
+
+<content>
+
+<LivePreview scope={{Input, Label, HelpText}} language="jsx">
+  {`<>
+  <Label htmlFor="email_address" required>Email address</Label>
+  <Input aria-describedby="email_help_text" id="email_address" name="email_address" type="email" placeholder="example@twilio.com" onChange="" required/>
+  <HelpText id="email_help_text">Please enter a valid email.</HelpText>
+</>`}
+</LivePreview>
```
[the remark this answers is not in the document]
richbachman: I think we're going to start working it into being the norm. Simon suggested I add onto the skeleton  loader docs page, so doing the same here.

packages/paste-website/src/pages/components/label/index.mdx
```diff
@@ -0,0 +1,354 @@
+---
+title: Label - Components
+package: '@twilio-paste/label'
+description: Label used to name Paste form fields.
```
SiTaggart, line 4: My only feedback is that we should come up with a better package description for here and in the `package.json` file. 

Let's get the team to have input on it and get consensus first, before we push up an update though. We can do it in the `plz` slack channel on this review item

richbachman: Sounds good. Thanks for kicking that off.

gloriliale (approved):
This is great! I had the exact same question as Nora about the example coming above guidelines, so I'd like to hear more about what we're trying to achieve with that change, but this doc page is solid 👊

packages/paste-website/src/pages/components/label/index.mdx
richbachman: Updated.

SiTaggart (approved):

PR #1675, opened 2021-07-15
Pagination style updates

richbachman:
- [x] Updated styles of `PaginationArrow` and `PaginationNumber`
- [x] Updated anatomy table on `Pagination` docs page

### DIFF
--- a/.changeset/tidy-dragons-tickle.md
+++ b/.changeset/tidy-dragons-tickle.md
@@ -1,0 +1,20 @@
+---
+'@twilio-paste/pagination': patch
+'@twilio-paste/core': patch
+---
+
+[Pagination] Update styles for better dark theme accessibility
+
+`PaginationArrow` hover and focus:
+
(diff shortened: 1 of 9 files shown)


SiTaggart (approved):
Nice changeset, I quite enjoy the detail

gloriliale (approved):
This is a great style update 🎨! Question: did Pagination not used to do anything onFocus?

richbachman:
> This is a great style update 🎨! Question: did Pagination not used to do anything onFocus?

@gloriliale `PaginationArrow` is a little funky. It consists of a button with a couple different child elements (the arrow and the text label). When the button is focused we want the background color of the circle/arrow to change, but not the background color of any text label of the button to change. So we have to pass that focused state down to the arrow but not the text label, so the background color only shows on the arrow. Before the focus style was very simple, so we didn't need to pass any focus states. Let me know if that explanation helps.

PR #1676, opened 2021-07-16
feat(media-object): add element customization

richbachman:
Added the `element` and `variant` prop to `MediaObject`, `MediaFigure`, and `MediaBody`. This enables customizations from the `CustomizationProvider`. Tests and VRT stories are included.


packages/paste-core/layout/media-object/__fixtures__/CustomizableMediaBody.tsx
```diff
@@ -0,0 +1,16 @@
+import * as React from 'react';
+import {MediaBody} from '../src';
+import type {MediaBodyProps} from '../src';
+
+export interface CustomizableMediaBodyProps extends MediaBodyProps {
+  element?: string;
+  variant?: 'primary' | 'secondary';
```
gloriliale, line 7: This might be jumping ahead since you'll be discussing this during show and tell tomorrow, but what is it the variant is doing here? I don't see it present in the previous version of the Media Object component.

packages/paste-core/layout/media-object/src/index.tsx
```diff
@@ -1,21 +1,29 @@
...
+import {Box, safelySpreadBoxProps} from '@twilio-paste/box';
 import type {BoxProps} from '@twilio-paste/box';
-import {Space, isMarginTokenProp} from '@twilio-paste/style-props';
+import {isMarginTokenProp} from '@twilio-paste/style-props';
+import type {Space} from '@twilio-paste/style-props';
 
-export interface MediaObjectProps extends Pick<BoxProps, 'as' | 'marginTop' | 'marginBottom'> {
+export interface MediaObjectProps
+  extends React.HTMLAttributes<any>,
+    Pick<BoxProps, 'as' | 'element' | 'marginTop' | 'marginBottom' | 'variant'> {
```
gloriliale, line 10: Is specifying to only use the props listed? I don't think I've used it before and am very intrigued!

packages/paste-core/layout/media-object/__tests__/index.spec.tsx
```diff
@@ -1,143 +1,214 @@
...
+    expect(screen.getByText('media-body').getAttribute('data-paste-element')).toEqual('MEDIA_BODY');
+  });
+  it('should set a custom element data attribute for MediaBody', () => {
+    render(<MediaBody element="foo">media-body</MediaBody>);
+    expect(screen.getByText('media-body').getAttribute('data-paste-element')).toEqual('foo');
+  });
+});
+
+describe('Customization', () => {
+  it('should add custom styles to MediaObject', (): void => {
```
gloriliale, line 173: Why is it that use `void` here? I think I asked this before and never looked back at what the answer was.

packages/paste-core/layout/media-object/__fixtures__/CustomizableMediaBody.tsx
gloriliale: Ah, is this to make it easier to create different stories? Is that also the reason we made these 'Customizable' components, for the stories?

packages/paste-core/layout/media-object/src/index.tsx
```diff
@@ -27,65 +35,68 @@ const MediaObject = React.forwardRef<HTMLElement, MediaObjectProps>(
...
-  MediaBody.propTypes = {
-    children: PropTypes.node.isRequired,
-  };
-}
+
+MediaBody.propTypes = {
+  children: PropTypes.node.isRequired,
+};
 
 export {MediaObject, MediaFigure, MediaBody};
```
gloriliale, line 102: Not that these MediaObject components have that much going on with them, but since they have 3 parts, would it be useful to split them up for clarity, the way we do for Modal, etc?

gloriliale (approved):
I'm excited to hear you show and tell this soon, and this looks good to me! Left some questions mostly to understand more about this process of adding customization to our components.

packages/paste-core/layout/media-object/__fixtures__/CustomizableMediaBody.tsx
richbachman: We're using variant here to create new variants of MediaObject. With the CustomizationProvider, consumers will be able to create their own variants of Paste components.

And yeah, this file just makes it easier to use the same custom MediaObject for stories and tests.

packages/paste-core/layout/media-object/src/index.tsx
richbachman: Yep exactly that. It's the opposite of `Omit`. https://www.typescriptlang.org/docs/handbook/utility-types.html#picktype-keys

packages/paste-core/layout/media-object/__tests__/index.spec.tsx
richbachman: All good. Linking to the other answer: https://github.com/twilio-labs/paste/pull/1641#discussion_r665720897

packages/paste-core/layout/media-object/__fixtures__/CustomizableMediaBody.tsx
richbachman: Actually we're going to remove variant for this round. We only want to expose the existing component variants if there are any.

packages/paste-core/layout/media-object/src/MediaFigure.tsx
```diff
@@ -0,0 +1,46 @@
...
+      >
+        {children}
+      </Box>
+    );
+  }
+);
+
+MediaFigure.displayName = 'MediaFigure';
+
+MediaFigure.defaultProps = {
```
TheSisb, line 35: Just curious why you opted to using defaultProps here in the standard/classic style, vs setting it in the destructuring on line 15 along with `as` and `element`

packages/paste-core/layout/media-object/stories/index.stories.tsx
```diff
@@ -95,6 +101,257 @@ export const ConstrainedWidth = (): React.ReactNode => {
...
 
+const CustomMediaObjectCollection: React.FC = () => {
+  return (
+    <Stack orientation="vertical" spacing="space60">
+      <CustomizableMediaObject as="div" verticalAlign="center">
+        <CustomizableMediaFigure as="div" spacing="space100">
+          <Box backgroundColor="colorBackgroundDestructiveWeak" size="size20" />
+        </CustomizableMediaFigure>
+        <CustomizableMediaBody as="div">
+          <Text as="p">Some media Object body text</Text>
```
TheSisb, line 112: Since we may use these stories in the future in our theme-designer and doc site, can we come up with more appropriate text for these examples while we're at it please?

This is a really solid improvement too, thanks for cleaning this up!

packages/paste-core/layout/media-object/__tests__/index.spec.tsx
```diff
@@ -1,143 +1,214 @@
 import * as React from 'react';
-import renderer from 'react-test-renderer';
-import {ReactWrapper, mount} from 'enzyme';
+import {matchers} from 'jest-emotion';
+import {render, screen} from '@testing-library/react';
```
TheSisb, line 3: 💯

packages/paste-core/layout/media-object/src/index.tsx
```diff
@@ -1,91 +1,3 @@
...
-if (process.env.NODE_ENV === 'development') {
-  MediaBody.propTypes = {
-    children: PropTypes.node.isRequired,
-  };
-}
-
-export {MediaObject, MediaFigure, MediaBody};
+export * from './MediaObject';
+export * from './MediaFigure';
+export * from './MediaBody';
```
TheSisb, line 3: Nice refactor, 👍

TheSisb (commented):
Really solid refactor, well done!

packages/paste-core/layout/media-object/src/MediaFigure.tsx
richbachman: Good catch. I'll refactor those.

packages/paste-core/layout/media-object/stories/index.stories.tsx
richbachman: Yep. Changing this section up a little after chatting with Simon, but I'll make text edits to the other examples.

andipants12 (approved):

TheSisb (approved):

PR #1681, opened 2021-07-19
chore(typography): remove Typography package

gloriliale:
- Removed the Typography package from core 
- Created Changeset for core as a **minor** update. 

Kinda nervous about deleting something from our repo, so let me know if I missed a step!


TheSisb (commented):
Please make sure to also remove/replace all references to this package.

cmd+shift+f "typography';"

I see one right away in primitives/tabs/stories/index.stories.tsx, didn't dig deeper

gloriliale:
@TheSisb couldn't reply directly to your comment above for some reason, but I searched and that's the only one that came up that I could see related to the Typography component. We use the term 'Typography' in a lot of other places and aren't referring to the component, but there are definitely no other instances of using `@twilio-paste/typography` (a few come up for blocking it in general though, if you think those are worth deleting?)

TheSisb:
If the blocks are no longer necessary, let's go ahead and delete them 👍

andipants12 (approved):

TheSisb (approved):

PR #1684, opened 2021-07-20
fix(alert-dialog): added missing date-fns dependency

richbachman:
While I was working on the `pnpm` spike, I noticed we missed a dependency on `AlertDialog` and `TimePicker`. This adds those.

### DIFF
--- a/.changeset/orange-jeans-destroy.md
+++ b/.changeset/orange-jeans-destroy.md
@@ -1,0 +1,6 @@
+---
+'@twilio-paste/alert-dialog': patch
+'@twilio-paste/core': patch
+---
+
+[AlertDialog]: added missing `@twilio-paste/uid-library` dependency.
--- a/packages/paste-core/components/time-picker/__tests__/index.spec.tsx
+++ b/packages/paste-core/components/time-picker/__tests__/index.spec.tsx
@@ -24,11 +24,11 @@ describe('formatReturnTime()', () => {
     expect(formatReturnTime(TEST_TIME, 'hh:mm:ss')).toEqual('01:30:04');
     expect(formatReturnTime(TEST_TIME, 'HH:mm:ss.SSS')).toEqual('13:30:04.000');
     expect(formatReturnTime(TEST_TIME, 'HH:mm:ss.SSS')).toEqual('13:30:04.000');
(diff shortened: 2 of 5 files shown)


.changeset/orange-jeans-destroy.md
```diff
@@ -0,0 +1,6 @@
+---
+'@twilio-paste/alert-dialog': patch
+'@twilio-paste/core': patch
+---
+
+[AlertDialog]: added missing `date-fns` dependency.
```
gloriliale, line 6: I think these messages got switched- `date-fns` isn't needed for alert dialog 😆

packages/paste-core/components/time-picker/__tests__/index.spec.tsx
```diff
@@ -24,11 +24,11 @@ describe('formatReturnTime()', () => {
     expect(formatReturnTime(TEST_TIME, 'hh:mm:ss')).toEqual('01:30:04');
     expect(formatReturnTime(TEST_TIME, 'HH:mm:ss.SSS')).toEqual('13:30:04.000');
     expect(formatReturnTime(TEST_TIME, 'HH:mm:ss.SSS')).toEqual('13:30:04.000');
-    expect(formatReturnTime(TEST_TIME, 'hh:mm:ss.SSS aa')).toEqual('01:30:04.000 p.m.');
+    expect(formatReturnTime(TEST_TIME, 'hh:mm:ss.SSS aa')).toEqual('01:30:04.000 PM');
```
gloriliale, line 27: Question: were these tests failing that you fixed?

gloriliale (approved):
Good catch, Rich! Thanks for adding this. Once the changeset messages are switched back, I think we're good to go.

.changeset/orange-jeans-destroy.md
richbachman: 🤦 Good catch. Fixed.

packages/paste-core/components/time-picker/__tests__/index.spec.tsx
richbachman: Yep. Those started failing after I added the missing dep.

TheSisb (approved):

andipants12 (approved):

Issue #1673, opened 2021-07-15
Grid - Layout (Vertical Gutter Value)


SinaTaavon:
Thanks for helping make Paste better! Note: For urgent requests, reach out at **#help-design-system**

**Describe the bug**
On the ```grid``` layout system, there is a ```gutter``` option which gives space on the X-Axis but when the ```grid``` is set to ```vertical``` the gutter still applies to the X-Axis, that might be by design but if we could have a ```vGutter``` prop so we can set space on top and bottom of the ```columns```, Alternatively ```gutter``` could apply the space to both axis.

Note: Manual margin gets ignored if ```grid``` is in use rather than ```flex```. That's why having a prop for the vertical gutter would be really helpful.

**To Reproduce**
The issue is visible on the Storybook for paste and no need to reproduce it.

Link to the Storybook's component where the issue is visible (You can change the window size to change it to vertical) https://5e53448165911c0022e68c74-erijdslbuz.chromatic.com/?path=/docs/layout-grid--grid-12-column-and-gutter-support#grid---equal-height-columns

**Expected behaviour**
A prop to pass the vertical value for the ```gutter```. like ```vGutter``` or something similar.
Alternatively ```gutter``` to apply the space to both axis.

**Screenshots**
![image](https://user-images.githubusercontent.com/32191839/125795749-18611be8-4590-4a4f-9d2a-2efaad451354.png)

**Desktop (please complete the following information):**
 - OS: [e.g. iOS] Mac OS 
 - Browser [e.g. chrome, safari] Opera, Chrome, Safari
 - Version [e.g. 22] Opera: 77, Chrome: 91.0.4472.124, Safari: 14.1 

**Additional context**
Overall Paste is looking very Strong candidate to replace Bootstrap and Material UI, for a Beta product I can confidently say that is extremely polished, just a couple of small changes and it will be one of the best frameworks that we have used to this day. 

Thanks for your time.

andipants12:
Hi @SinaTaavon,

Thank you for the bug report and the suggested addition to the `Grid` API. We are discussing solutions as a team and will get back to you early next week with some updates on how we plan to move forward. 

Thank you for raising this issue, the encouraging words, and helping us improve Paste. 🎉

andipants12:
Hello again @SinaTaavon 👋, we have a pull request open now to resolve this issue, which is linked above.

[...]

PR #1685, opened 2021-07-20
fix(grid): vertical grid should use gutter

andipants12:
## Description
Bug fix for [issue 1673](https://github.com/twilio-labs/paste/issues/1673). `Grid` component gutter not applied when `vertical` prop is true.

See a demo codesandbox of these changes [here](https://codesandbox.io/s/fixgrid-verticalhorizontal-gutter-bug-fix-ssi82?file=/src/App.js).

### Reported issue
When the `Grid` props include `{ vertical: true }`, the gutter spacing is not applied.

### Summary of changes
- when `vertical` is `true` gutter spacing is applied to the **top** and **bottom** of the grid container's margin, and padding of each column
- when `vertical` is `false` gutter spacing is applied to the **right** and **left** of the grid container's margin, and padding of each column
- apply gutter padding for each column when `Grid` props include responsive gutter values and responsive vertical values

### DIFF
--- a/packages/paste-core/layout/grid/__tests__/__snapshots__/grid.test.tsx.snap
+++ b/packages/paste-core/layout/grid/__tests__/__snapshots__/grid.test.tsx.snap
@@ -4,6 +4,10 @@ exports[`Grid render should render a Column 1`] = `
 <DocumentFragment>
   .emotion-0 {
   width: 8.333333333333332%;
+  padding-top: space0;
+  padding-bottom: space0;
+  padding-right: space0;
+  padding-left: space0;
 }
 
 <div
@@ -78,8 +82,8 @@ exports[`Grid render should render responsive css 1`] = `
   -ms-flex-wrap: wrap;
   flex-wrap: wrap;
   display: flex;
-  margin-right: spaceNegative10;
-  margin-left: spaceNegative10;
+  margin-right: space0;
+  margin-left: space0;
   min-width: size0;
 }
 
@@ -88,8 +92,8 @@ exports[`Grid render should render responsive css 1`] = `
     -webkit-flex-direction: column;
     -ms-flex-direction: column;
     flex-direction: column;
-    margin-right: spaceNegative20;
-    margin-left: spaceNegative20;
+    margin-right: space0;
+    margin-left: space0;
   }
 }
 
@@ -105,16 +109,20 @@ exports[`Grid render should render responsive css 1`] = `
 
 .emotion-0 {
   width: 16.666666666666664%;
-  padding-left: space10;
-  padding-right: space10;
+  padding-top: space10;
+  padding-bottom: space10;
+  padding-right: space0;
+  padding-left: space0;
   margin-left: 66.66666666666666%;
 }
 
 @media screen and (min-width:40em) {
   .emotion-0 {
     width: 33.33333333333333%;
-    padding-left: space20;
-    padding-right: space20;
+    padding-top: space20;
+    padding-bottom: space20;
+    padding-right: space0;
+    padding-left: space0;
     margin-left: 50%;
   }
 }
@@ -122,32 +130,40 @@ exports[`Grid render should render responsive css 1`] = `
 @media screen and (min-width:52em) {
   .emotion-0 {
     width: 16.666666666666664%;
-    padding-left: space30;
+    padding-top: space0;
+    padding-bottom: space0;
     padding-right: space30;
+    padding-left: space30;
     margin-left: 66.66666666666666%;
   }
 }
 
 .emotion-1 {
   width: 50%;
-  padding-left: space10;
-  padding-right: space10;
+  padding-top: space10;
+  padding-bottom: space10;
+  padding-right: space0;
+  padding-left: space0;
   min-width: 100%;
   margin-left: space0;
 }
 
 @media screen and (min-width:40em) {
   .emotion-1 {
-    padding-left: space20;
-    padding-right: space20;
+    padding-top: space20;
+    padding-bottom: space20;
+    padding-right: space0;
+    padding-left: space0;
     min-width: 100%;
   }
 }
 
 @media screen and (min-width:52em) {
   .emotion-1 {
-    padding-left: space30;
+    padding-top: space0;
+    padding-bottom: space0;
     padding-right: space30;
+    padding-left: space30;
     min-width: 0;
   }
 }
--- a/packages/paste-core/layout/grid/__tests__/__snapshots__/grid.test.tsx.snap
+++ b/packages/paste-core/layout/grid/__tests__/__snapshots__/grid.test.tsx.snap
@@ -82,8 +82,8 @@ exports[`Grid render should render responsive css 1`] = `
   -ms-flex-wrap: wrap;
   flex-wrap: wrap;
   display: flex;
-  margin-right: space0;
-  margin-left: space0;
+  margin-right: auto;
+  margin-left: auto;
   min-width: size0;
 }
 
@@ -92,8 +92,8 @@ exports[`Grid render should render responsive css 1`] = `
     -webkit-flex-direction: column;
     -ms-flex-direction: column;
     flex-direction: column;
-    margin-right: space0;
-    margin-left: space0;
+    margin-right: auto;
+    margin-left: auto;
   }
 }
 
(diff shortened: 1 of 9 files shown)


packages/paste-core/layout/grid/src/Grid.tsx
```diff
@@ -47,12 +51,14 @@ const Grid = React.forwardRef<HTMLDivElement, GridProps>(
...
         ),
-      [children]
+      [children, vertical, equalColumnHeights, gutter]
     );
 
-    const gutterStyles = React.useMemo(() => getGutterStyles(gutter), [gutter]);
+    const gutterStyles = React.useMemo(() => getGutterStyles(gutter, vertical), [gutter, vertical]);
     const flexDirection = React.useMemo(() => getFlexDirection(vertical), [vertical]);
 
+    const finalStyles = {...gutterStyles, marginTop, marginBottom};
```
richbachman, line 60: nit: should this be something like `gridStyles`? I'm having flashbacks of finalVersion, finalFinalVersion, and so on.

andipants12: Yes, that sounds very reasonable. Thanks!

richbachman (approved):
Looks like theres a snapshot that needs to be updated. After that, I think its good to go.

packages/paste-core/layout/grid/src/types.ts
```diff
@@ -1,12 +1,24 @@
...
 export interface GridProps extends SpaceProps {
   as?: keyof JSX.IntrinsicElements;
   children: NonNullable<React.ReactNode>;
   equalColumnHeights?: boolean;
   gutter?: Space;
   vertical?: ResponsiveValue<boolean>;
+  marginTop?: Margin;
+  marginBottom?: Margin;
+  marginRight?: Margin;
+  marginLeft?: Margin;
```
SiTaggart, line 21: We're adding margin props to the Grid API?

packages/paste-core/layout/grid/src/utils.tsx
```diff
@@ -1,16 +1,86 @@
...
+const getSpaceDefault = (responsive = false): 'space0'[] | 'space0' =>
+  responsive ? new Array(3).fill('space0') : 'space0';
+
+const initDefaultSpacing = ({
+  prefix,
+  responsive,
+}: {
+  prefix: string;
+  responsive: boolean;
+}): {[x: string]: Margin | Padding} => ({
```
SiTaggart, line 24: Might just be me, but I'm finding these inline typings on these method are making things pretty hectic to read. Is it worth pulling them out into types and interfaces to clean the syntax up a little?

packages/paste-core/layout/grid/__tests__/__snapshots__/grid.test.tsx.snap
```diff
@@ -78,8 +82,8 @@ exports[`Grid render should render responsive css 1`] = `
   -ms-flex-wrap: wrap;
   flex-wrap: wrap;
   display: flex;
-  margin-right: spaceNegative10;
-  margin-left: spaceNegative10;
+  margin-right: space0;
+  margin-left: space0;
```
SiTaggart, line 86: I can't immediately tell is this is because the test is different or whether it legitimately caught the negative margins being dropped from the outer Grid? 

This might be a bug

packages/paste-core/layout/grid/src/utils.tsx
andipants12: Sure thing, I can make that happen. I think this shape is reused quite a bit too, so it would be worth creating a named type anyway. Thanks!

packages/paste-core/layout/grid/__tests__/__snapshots__/grid.test.tsx.snap
andipants12: Yeah it was a bug in the sense that these should have been defaulting to `auto` for the margin, not `space0`. However, this is a change related to the fix for the spacing not applied for vertical grids. In this case, the `vertical` value is an array of responsive values and for this test, the screen size is `vertical`. Thus, the right and left margins are cleared.

packages/paste-core/layout/grid/src/types.ts
andipants12: Mmm let me look at this again. I know that previously `marginTop` and `marginBottom` were allowed and applied to the grid stylings if provided (regardless of whether it is a vertical grid or not).

packages/paste-core/layout/grid/src/Column.tsx
```diff
@@ -2,29 +2,32 @@ import * as React from 'react';
 import * as PropTypes from 'prop-types';
 import {styled, compose, flexbox, layout, space} from '@twilio-paste/styling-library';
 import {ResponsiveProp} from '@twilio-paste/style-props';
-import {ColumnProps, ColumnStyleProps} from './types';
-import {getStackedColumns, getColumnOffset, getColumnSpan} from './utils';
-
-export const getColumnStyles = (props: ColumnProps): ColumnStyleProps => {
+import type {ColumnProps, ColumnStyleProps} from './types';
```
gloriliale, line 5: To clarify, this file isn't used in the actual Grid right? Is it for testing the utils only?

packages/paste-core/layout/grid/src/Grid.tsx
```diff
@@ -47,12 +51,14 @@ const Grid = React.forwardRef<HTMLDivElement, GridProps>(
               })
             : child
         ),
-      [children]
+      [children, vertical, equalColumnHeights, gutter]
```
TheSisb, line 54: 👍

packages/paste-core/layout/grid/src/Grid.tsx
```diff
@@ -47,12 +51,14 @@ const Grid = React.forwardRef<HTMLDivElement, GridProps>(
...
     );
 
-    const gutterStyles = React.useMemo(() => getGutterStyles(gutter), [gutter]);
+    const gutterStyles = React.useMemo(() => getGutterStyles(gutter, vertical), [gutter, vertical]);
     const flexDirection = React.useMemo(() => getFlexDirection(vertical), [vertical]);
 
+    const gridStyles = {...gutterStyles, marginTop, marginBottom};
```
TheSisb, line 60: nit: can we undo this as it adds some unnecessary overhead to create a new object

andipants12: Sure thing 👍

packages/paste-core/layout/grid/src/Column.tsx
andipants12: Do you mean the `Column.tsx` file? No this is used to create the grid columns.

packages/paste-core/layout/grid/src/types.ts
andipants12: To follow up, _technically_ these margin props were already available on the Grid API because the props interface extends `SpaceProps`, [which has them available](https://github.com/twilio-labs/paste/blob/8b3924075e2ea5b890f5c0a90c3fc9985ec38708/packages/paste-style-props/src/types/space.ts#L32).

TheSisb (commented):
We should update the doc site examples, which currently use both gutter and Boxes in columns with marginTop/Bottom

SiTaggart (approved):
Agreed, any API updates that need to be added to the doc site page should also come in this PR. It might be nice to also add an example to the page. 

We originally didn't because the site wasn't responsive so we just link you to the storybook. But now we have a responsive site, we can just use that storybook example, as an example actually on the doc site

andipants12:
@TheSisb @SiTaggart Thanks for reviewing again! To clarify, should the updates to the doc site be included in this PR or should I create a ticket for that? Also do you mean the Grid page for the docs site? I just checked the [docs page for grid](https://github.com/twilio-labs/paste/blob/main/packages/paste-website/src/pages/layout/grid/index.mdx) and I don't see the boxes with top/bottom margins.

TheSisb:
Let's do it in this same PR if you don't mind. 

These examples all have Box with the sole purpose of giving marginTop/Bottom

- https://paste.twilio.design/layout/grid#stacked-columns-in-a-vertical-layout
- https://paste.twilio.design/layout/grid#nested-grids
- https://paste.twilio.design/layout/grid#basic-page-layout

.changeset/friendly-suits-change.md
```diff
@@ -0,0 +1,5 @@
+---
+'@twilio-paste/grid': patch
```
gloriliale, line 2: I think we also need to include core here, is that right?

PR #1690, opened 2021-07-22
feat(badge): Add Badge component package

andipants12:
**Contributing to Twilio**

### DIFF
--- a/.changeset/olive-points-shave.md
+++ b/.changeset/olive-points-shave.md
@@ -1,0 +1,8 @@
+---
+'@twilio-paste/badge': major
+'@twilio-paste/core': major
+'@twilio-paste/box': minor
+---
+
+[Box] Add "columnGap" and "rowGap" CSS grid props to Box API.
+[Badge] Release Stage 1 of Badge component. A badge is a small chunk of information that describes an object in a visually highlighted way.
--- a/.changeset/olive-points-shave.md
+++ b/.changeset/olive-points-shave.md
@@ -1,5 +1,6 @@
 ---
 '@twilio-paste/badge': major
+'@twilio-paste/core': major
 ---
 
 [Badge] Release Stage 1 of Badge component. A badge is a small chunk of information that describes an object in a visually highlighted way.
--- a/.changeset/olive-points-shave.md
+++ b/.changeset/olive-points-shave.md
@@ -1,8 +1,5 @@
 ---
 '@twilio-paste/badge': major
-'@twilio-paste/core': major
-'@twilio-paste/box': minor
 ---
 
-[Box] Add "columnGap" and "rowGap" CSS grid props to Box API.
 [Badge] Release Stage 1 of Badge component. A badge is a small chunk of information that describes an object in a visually highlighted way.
--- a/.changeset/rude-doors-pump.md
+++ b/.changeset/rude-doors-pump.md
@@ -1,0 +1,6 @@
+---
+'@twilio-paste/core': major
+---
+
+- [Box] Add column-gap and row-gap (CSS grid, flex box) to Box API.
+- [Badge] Add stage 1 Badge component package.
--- a/.changeset/rude-doors-pump.md
+++ b/.changeset/rude-doors-pump.md
@@ -1,5 +1,5 @@
 ---
-'@twilio-paste/core': major
+'@twilio-paste/core': minor
 ---
 
 - [Box] Add column-gap and row-gap (CSS grid, flex box) to Box API.
--- a/.changeset/rude-doors-pump.md
+++ b/.changeset/rude-doors-pump.md
@@ -1,6 +1,0 @@
----
-'@twilio-paste/core': minor
----
-
-- [Box] Add column-gap and row-gap (CSS grid, flex box) to Box API.
-- [Badge] Add stage 1 Badge component package.
--- a/packages/paste-core/components/badge/__tests__/index.spec.tsx
+++ b/packages/paste-core/components/badge/__tests__/index.spec.tsx
@@ -1,17 +1,20 @@
 import * as React from 'react';
 import {render, fireEvent} from '@testing-library/react';
+import {renderHook} from '@testing-library/react-hooks';
 import {matchers} from 'jest-emotion';
+
+import {InformationIcon} from '@twilio-paste/icons/esm/InformationIcon';
 // eslint-disable-next-line @typescript-eslint/ban-ts-comment
 // @ts-ignore typescript doesn't like js imports
 import axe from '../../../../../.jest/axe-helper';
 import {Badge} from '../src';
-import {getVariantStyles} from '../src/utils';
+import {getVariantStyles, hasValidButtonVariantProps, hasValidAnchorVariantProps} from '../src/utils';
+import {useFocusableVariants, useResizeChildIcons} from '../src/hooks';
 
 expect.extend(matchers);
 
 describe('Badge', () => {
   describe('Utils', () => {
-    // @TODO other util tests.
     describe('getVariantStyles', () => {
       it('should return the correct tokens when variant is "success"', () => {
         expect(getVariantStyles('success')).toEqual({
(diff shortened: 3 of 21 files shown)


SiTaggart (commented):
So initial thoughts:

- **Size**: Discuss the thoughts on this. How does this relate to passing the badge an icon? Should the child composition dictate the size just like in a button? Do we actually want massive badges?
- **Badge as buttons**: Have we considered this use case? We use it in our masthead on the dark mode
- **Badges as links**: Similarly, if we have badges as buttons, shouldn't we have badges as links?

andipants12:
@SiTaggart Thank you for the feedback.

> **Size**: Discuss the thoughts on this. How does this relate to passing the badge an icon? Should the child composition dictate the size just like in a button? Do we actually want massive badges?
I would love to discuss all of these questions with the team. My feeling is probably not. Also I did a very simple implementation of the API now and am composing the `Badge` with the children (e.g. in this case, text and icon) pretty independently (meaning that there is a `size` passed to both `Badge` and the icon).

>**Badge as buttons**: Have we considered this use case? We use it in our masthead on the dark mode
I did not explicitly account for this while working on this ticket.

>**Badges as links**: Similarly, if we have badges as buttons, shouldn't we have badges as links?
Same as above.

I'll follow up on Slack as well,  but for the sake of transparency, I felt like this component would benefit from more exploration of the API but didn't think that was accounted for in the ticket pointing. In preparation for our planning tomorrow, I will update the code sandbox with a few more examples, exploration of the mentioned use cases ("as" link and button), as well as a few different possible API design decisions.

Please let me know if I should look into any thing else for this. Thank you!

packages/paste-core/components/badge/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,130 @@
...
+import {matchers} from 'jest-emotion';
+import axe from '../../../../../.jest/axe-helper';
+import {Badge} from '../src';
+import {getVariantStyles} from '../src/utils';
+
+expect.extend(matchers);
+
+describe('Badge', () => {
+  describe('utils', () => {
+    // @TODO add "describe" block for "getComputedTokenNames"
```
TheSisb, line 12: What's this about? A little unclear

packages/paste-core/components/badge/package.json
```diff
@@ -0,0 +1,41 @@
+{
+  "name": "@twilio-paste/badge",
+  "version": "0.0.0",
+  "category": "data display",
+  "status": "alpha",
```
TheSisb, line 5: Is this truly alpha? or production?

packages/paste-core/components/badge/src/index.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+    return (
+      <Box
+        ref={ref}
+        {...safelySpreadBoxProps(props)}
+        as="span"
+        backgroundColor={backgroundColor}
+        borderRadius="borderRadius10"
+        color={color}
+        display="inline-flex"
```
TheSisb, line 22: Why not `display: inline(-block)`?

packages/paste-core/components/badge/stories/index.stories.tsx
```diff
@@ -0,0 +1,246 @@
...
+  'sizeIcon50',
+  'sizeIcon60',
+  'sizeIcon70',
+  'sizeIcon80',
+  'sizeIcon90',
+  'sizeIcon100',
+  'sizeIcon110',
+];
+
+// Do we not want to include controls..?
```
TheSisb, line 27: While we could have 1 story with controls (I think we're still using knobs even), we use Storybook for VRT so we need to have 1 story per view for testing.

packages/paste-core/components/badge/stories/index.stories.tsx
```diff
@@ -0,0 +1,246 @@
...
+    <Badge variant="warning">Warning</Badge>
+    <Badge variant="error">Error</Badge>
+    <Badge variant="success">Success</Badge>
+    <Badge variant="new">New</Badge>
+  </Stack>
+);
+
+// DEFAULT
+export const DefaultBadge: React.FC = () => {
+  const seed = useUIDSeed();
```
TheSisb, line 47: 👍

packages/paste-core/components/badge/stories/index.stories.tsx
```diff
@@ -0,0 +1,246 @@
...
+          <Badge key={seed(size)} variant="default" size={size}>
+            <InformationIcon size={size} decorative />
+            &nbsp;Default
+          </Badge>
+        ))}
+      </Stack>
+      <Stack orientation="horizontal" spacing="space40">
+        {ICON_SIZE_OPTIONS.map((size) => (
+          <Badge key={seed(size)} variant="default" size={size}>
+            {' '}
```
TheSisb, line 69: I think this is extra spacing (repeated in other examples)

TheSisb (commented):
Really clean code, nice job!

packages/paste-core/components/badge/package.json
andipants12: I used a script in `package.json` to generate the component package directory and this is one of the defaults. I know we are filing this under `stage 1` for now, so this should probably be production when it is ready to be released?

packages/paste-core/components/badge/stories/index.stories.tsx
andipants12: > ... we use Storybook for VRT so we need to have 1 story per view for testing.

In the most recent iteration I am moving the individual variants to their own page, though within each variant page there are three different renders. Does that align with our needs for VTR?

packages/paste-core/components/badge/src/index.tsx
andipants12: Oh I apologize, I just saw this! 😓 The containing `span` is `inline-flex` because the inner contents are using `justify-contents` and other flexy things.

packages/paste-core/components/badge/__tests__/index.spec.tsx
andipants12: Ah yes. Previously I had a helper/utility function called `getComputedTokenNames` that was strongly inspired by the equivalent helper in the `Avatar` package. 
We have since decided that for now we will release a single size `Badge` so I removed this utility for now. I left it as a `@TODO` note originally because I had a feeling it would be out of scope for stage 1.

packages/paste-core/components/badge/package.json
gloriliale: yeah, I think the package-maker script pre-dates our term changes. I think this should be production.

packages/paste-core/components/badge/package.json
```diff
@@ -0,0 +1,43 @@
...
+  "description": "TODO update.",
```
gloriliale, line 6: Just wanna make sure this doesn't get missed! Here's what we have down in the newsletter:
>A small chunk of information that describes an object in a visually highlighted way.

packages/paste-core/components/badge/package.json
SiTaggart: mmhmmm production pls

packages/paste-core/components/badge/package.json
SiTaggart: This does go out in the npm registry, so we should probably add something here for now.

packages/paste-core/components/badge/package.json
```diff
@@ -0,0 +1,43 @@
...
+    "build": "yarn clean && NODE_ENV=production node build.js && tsc",
+    "build:js": "NODE_ENV=development node build.js",
+    "build:props": "typedoc --tsconfig ./tsconfig.json --json ./dist/prop-types.json",
+    "clean": "rm -rf ./dist",
+    "tsc": "tsc"
+  },
+  "peerDependencies": {
+    "@twilio-paste/box": "^4.1.1",
+    "@twilio-paste/button": "^7.0.5",
+    "@twilio-paste/style-props": "^3.0.6",
```
SiTaggart, line 30: When listing peer dependencies don't forget to follow the repo guidelines on hoisting them https://github.com/twilio-labs/paste#peer-deps

andipants12: @SiTaggart Could you take a look at this again? It looks like either some deps are omitted from the button example in the README.md or it is out of date. Thank you!

packages/paste-core/components/badge/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,143 @@
...
+
+      it('should return the correct tokens when variant is "new"', () => {
+        expect(getVariantStyles('new')).toEqual({
+          backgroundColor: 'colorBackgroundNew',
+          color: 'colorTextNew',
+        });
+      });
+
+      it('should return the correct tokens when variant is "default"', () => {
+        expect(getVariantStyles('default')).toEqual({
```
SiTaggart, line 50: Nice

packages/paste-core/components/badge/src/types.ts
```diff
@@ -0,0 +1,11 @@
+import type {MouseEventHandler, ReactNode} from 'react';
+
+export type BadgeVariants = 'info' | 'default' | 'warning' | 'error' | 'success' | 'new';
+export type BadgeBaseElements = 'button' | 'span' | 'a';
+export interface BadgeProps {
+  variant: BadgeVariants;
+  children: NonNullable<ReactNode>;
+  as?: BadgeBaseElements;
```
SiTaggart, line 8: didn't we decide that as was a required prop?

packages/paste-core/components/badge/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,143 @@
...
+        <Badge data-testid="badge-6" variant="success">
+          test
+        </Badge>
+      );
+      const badgeElement = getByTestId('badge-6');
+
+      expect(badgeElement).toHaveStyleRule('color', 'colorTextSuccess');
+      expect(badgeElement).toHaveStyleRule('background-color', 'colorBackgroundSuccessWeakest');
+    });
+  });
```
SiTaggart, line 130: I noticed that there aren't any tests covering the focusable variables here. Are they worth adding to ensure the button variant stays and does what a button should do?

packages/paste-core/components/badge/package.json
SiTaggart: I think that's all of them. Really need a script fo this because it would require me to manually check the dep tree myself

packages/paste-core/components/badge/src/hooks.tsx
```diff
@@ -0,0 +1,120 @@
...
+  };
+};
+
+const StyledSpan: React.FC<{children: React.ReactNode}> = ({children, ...props}) => (
+  <Box display="inherit" as="span" alignSelf="center" {...props}>
+    {children}
+  </Box>
+);
+
+export const useWrapChildren = (reactChildren: React.ReactNode): React.ReactNode => {
```
SiTaggart, line 71: Are we taking inspiration from the button component in here? I'm wondering if there is a CSS solution to simplify things. I might have a play tomorrow

packages/paste-core/components/badge/src/types.ts
andipants12: I thought it was that in order for the badge to be a button or anchor that both `as` and `onClick`/`href` (respectively) are required, the same as it is in button. We def decided that `variant` is required.

packages/paste-core/components/badge/src/hooks.tsx
andipants12: Do you mean mapping through the children (if there are children and not a single child) and wrapping them, etc.? 
Not explicitly no. I opted to use `Box` as a `span` and style the span through the props to be more javascripty (since we are using css in js) and composability using the Paste primitives.

Other than being verbose, what do you think is(are) the cost(s) of this compared to pure css? If you end up playing around with this I'd love to sync and hear your thoughts 🧠

packages/paste-core/components/badge/__tests__/index.spec.tsx
andipants12: Yes, I think so. I omitted them because I wasn't confident that it was within scope based on the ticket description. I can add some now for both the anchor and the button variants.

I'll include:
- some interaction tests (prob with `testing-library/user-event` 
- some style checking tests for the focus and hover states

LMK if you'd like to see anything else. ✔️

packages/paste-core/components/badge/src/hooks.tsx
SiTaggart: I'm mainly thinking just small perf things. If we can get the browser to do things for us, instead of executing JavaScript that is handy. 

If we're only really concerned with creating a space between distinct children, and we're not fussed on order, maybe we can get some use out of [Column gap](https://developer.mozilla.org/en-US/docs/Web/CSS/column-gap)?

andipants12:
@SiTaggart @gloriliale I'm out of date with `main`, but I will rebase tomorrow morning once I've confirmed we will be good to merge this.

As of now this has the following updates:
- added `columnGap` and `rowGap` to the Box package
- Badge is using `columnGap` to position it's children
- simplified hooks logic
- tests updated to cover all utility functions and hooks 🔍

packages/paste-core/components/badge/src/hooks.tsx
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  return {
+    styleProps,
+    wrapper: ({children}) => <>{children}</>,
+    spanProps: redactedProps,
+  };
+};
+
+export const useResizeChildIcons = (children: BadgeChildren): BadgeChildren => {
```
SiTaggart, line 77: This is kinda neat

packages/paste-core/components/badge/src/types.ts
```diff
@@ -0,0 +1,13 @@
+import {ButtonProps} from '@twilio-paste/button';
+
+export type NamedChild = React.ReactElement<Record<string, any>, React.NamedExoticComponent>;
+
+export type BadgeChildren = NamedChild | React.ReactText | (React.ReactText | NamedChild)[];
+
+export type BadgeVariants = 'info' | 'default' | 'warning' | 'error' | 'success' | 'new';
+export type BadgeBaseElements = 'button' | 'span' | 'a';
+export interface BadgeProps extends Omit<ButtonProps, 'variant' | 'as'> {
```
SiTaggart, line 9: When we do the docs lets revisit this interface. 

We can do some nice things with conditional types to make sure there's no cross over between variants.

Right now, I think I can add button props to a badge that isn't a button, and we can fix that with conditionals

packages/paste-core/components/badge/src/utils.ts
```diff
@@ -0,0 +1,66 @@
...
+  return clickHandlerIsValid && asPropIsValid;
+};
+export const hasValidAnchorVariantProps = ({as, href}: Partial<BadgeProps>): boolean => {
+  const asPropIsValid = as === 'a';
+  const hrefIsValid = typeof href === 'string';
+  return asPropIsValid && hrefIsValid;
+};
+
+type Props = Record<string, any>;
+export const safelySpreadProps = (props: Props, denylist: string[]): Partial<Props> =>
```
SiTaggart, line 17: This is what prompted my comment on safely spread. Ideally we only have one version and it lives in Box

yarn.lock
```diff
@@ -19744,19 +19744,19 @@ graphql-import@^0.7.1:
     lodash "^4.17.4"
     resolve-from "^4.0.0"
 
-graphql-playground-html@1.6.25:
-  version "1.6.25"
-  resolved "https://registry.yarnpkg.com/graphql-playground-html/-/graphql-playground-html-1.6.25.tgz#2d8fa250cec4036a4f5b7f8ad069c86d6d64c95f"
-  integrity sha512-wMNvGsQ0OwBVhn72VVi7OdpI85IxiIZT43glRx7gQIwQ6NvhFnzMYBIVmcJAJ4UlXRYiWtrQhuOItDXObiR3kg==
+graphql-playground-html@^1.6.29:
```
SiTaggart, line 19747: This is weird, I wonder why it did this 🤔

packages/paste-core/components/badge/src/types.ts
SiTaggart: I still think this should be required.

packages/paste-core/components/badge/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,597 @@
...
+        expect(badgeElement).toHaveStyleRule('color', 'colorTextSuccess');
+        expect(badgeElement).toHaveStyleRule('background-color', 'colorBackgroundSuccessWeakest');
+      });
+    });
+  });
+
+  describe('Accessibility', () => {
+    it('Should have no accessibility violations', async () => {
+      const {container} = render(
+        <Badge data-testid="badge-1" variant="default">
```
SiTaggart, line 589: Ideally we'd have a as button and as anchor example in here too otherwise we're only testing the single variant for accessibility.

(I'm am aware that our coverage in other components is also like this and needs fixing)

SiTaggart (approved):
All this trying to figure out what props are OK in the hook are a bit icky, and it's only because you're having to do it when we introduce the extra wrapper around the badge, because the outer element where you want to spread the props changes.

In hindsight I think I would have stuck with what you had, and avoided the reset button entirely.

Once we have a button and anchor primitive we can refactor a bunch of this. 

Ideally I only want safely spread on the Box in the component on a single outer node. Right now, with the spread element changing from the wrapper to the actual badge it'll make customization tricky when we come to it.

This is perfect for a Stage 1 release.

SiTaggart:
Oh, and needs a change set

packages/paste-core/components/badge/src/hooks.tsx
andipants12: This is a pretty naive implementation, but meant as a patch since we decided to leave out scaling the badge size with icon sizes for stage 1 and requiring the icon be the default size.

I feel like for development this could benefit from some info logging to inform the consumer of the modification and provide an opportunity for them to "fix" on their own, but left it out since this is stage 1. Also it only covers the case where the children are in a 1 dimensional array.

packages/paste-core/components/badge/src/utils.ts
andipants12: Yeah that makes a lot of sense. I really like the idea too of leveraging hooks for side effects, event handling, and state management (internal and interaction). It would not only help our development, but also empower customers to build their own custom components (similar to how we are using Reakit).

packages/paste-core/components/badge/__tests__/index.spec.tsx
andipants12: Yeah, I have some ideas about unit testing that might help us get better coverage without the heavy lift of many many unique `describe` and `it` blocks in test files (which can be laborious to create and maintain). I'll probably work up an example and share in the next show and tell.

yarn.lock
andipants12: I went ahead and reverted this change. Not sure either... though I did run `yarn clean:full`, which _might_ remove the lock file.

gloriliale (approved):

.changeset/olive-points-shave.md
```diff
@@ -0,0 +1,8 @@
+---
+'@twilio-paste/badge': major
+'@twilio-paste/core': major
+'@twilio-paste/box': minor
```
SiTaggart, line 4: Ah, this may not be immediate obvious. 

We actually need a changeset per component, unless you don't mind the same entry in the changelogs of all components in the changeset. 

So for example, I wouldn't want to read about badge changes in the Box changelog, so box should have it's own changelog, which also applies to core.

andipants12: Ahhh okey, that makes sense. So then I will run the changeset command 3 times?

SiTaggart: No just twice.

badge and core

box and core.

.changeset/rude-doors-pump.md
```diff
@@ -0,0 +1,6 @@
+---
+'@twilio-paste/core': major
```
SiTaggart, line 2: Make this minor and you're good to go.

PR #1696, opened 2021-07-27
docs(website): update truncation use case guidelines

TheSisb:
Updates the Truncate docs to be less strict on use cases. Addresses https://github.com/twilio-labs/paste/discussions/1695

### DIFF
--- a/packages/paste-website/src/components/homepage/HomeHero.tsx
+++ b/packages/paste-website/src/components/homepage/HomeHero.tsx
@@ -66,14 +66,19 @@ const HomeHero: React.FC = () => {
             <Column span={5}>
               <NewComponentBanner>
                 <NewComponentBannerBadge>New!</NewComponentBannerBadge>
-                <NewComponentBannerText>Did you know we&apos;ve got an Alert Dialog?!</NewComponentBannerText>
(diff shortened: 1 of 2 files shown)


SiTaggart (approved):

katieporter (approved):

gloriliale (commented):
One thing about this doc page: we have a `title` prop that is required but we don't specify that anywhere, or mention the prop at all until the anatomy table (where it has no description and does not say it's required). Would this be a good opportunity to add a sentence like. "The `title` prop is required and should be the same as the contents of the Truncate component." or at least say it's required in the anatomy table?

TheSisb:
@gloriliale I checked other docs pages and it looks like we only mark non-required props with a `?`, same as typescript.

gloriliale:
> @gloriliale I checked other docs pages and it looks like we only mark non-required props with a `?`, same as typescript.

Ah I see, fair point. I still feel like there's a missing piece of the puzzle on how to use the `title` prop here tho. Could we add something like "This string will be the accessible label and should match the content of Truncate" or something similar to the description part of the table for `title`?

gloriliale (approved):

PR #1697, opened 2021-07-27
chore: set up automatic PR labelling

SiTaggart:
When a PR is opened use GitHub actions to automatically label them based on their area

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -1,0 +1,43 @@
+# https://github.com/actions/labeler
+
+# Add 'repo' label to any root file changes
+"Area: Repo":
+  - "*"
+
+"Area: Primitives":
+  - packages/paste-core/primitives/**/*
+
+"Area: Layout":
+  - packages/paste-core/layout/**/*
+
+"Area: Components":
+  - packages/paste-core/components/**/*
+
+"Area: Core bundle":
+  - packages/paste-core/core-bundle/**/*
+
+"Area: Doc Site":
+  - packages/paste-website/**/*
+
+"Area: Theme":
+  - packages/paste-theme/**/*
+
+"Area: Tokens":
+  - packages/paste-design-tokens/**/*
+
+"Area: Storybook":
+  - .storybook/**/*
+
+"Area: Infrastructure":
+  - .eslint/**/*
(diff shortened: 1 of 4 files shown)


.github/labeler.yml
```diff
@@ -0,0 +1,34 @@
...
+  - .eslint/**/*
+  - .github/**/*
+  - .jest/**/*
+  - .vscode/**/*
+  - tools/**/*
+
+"Type: Documentation":
+  - packages/paste-website/src/pages/**/*
+
+"Type: Tests":
```
gloriliale, line 32: Does this mean that there can be both a Type and Area label for a PR?

.github/labeler.yml
```diff
@@ -0,0 +1,34 @@
+# https://github.com/actions/labeler
+
+# Add 'repo' label to any root file changes
+"Area: Repo":
+  - "*"
+
+"Area: Components":
+  - packages/paste-core/components/**/*
```
TheSisb, line 8: Do we want coverage over layout and primitive components? or core?

TheSisb (approved):
Pretty cool!

gloriliale (approved):
Just had a question about the number of labels you can put on a PR, but this is exciting!

gloriliale (commented):
Question: why is this a separate PR from the mergify one? Don't we have this same info in the mergify PR?

.github/labeler.yml
SiTaggart: Yup, you can have as many labels as you like

.github/labeler.yml
SiTaggart: Oh, maybe. One for each?

- `Area: Components`
- `Area: Primitives`
- `Area: Layout`
- `Area: Core bundle`

SiTaggart:
> Question: why is this a separate PR from the mergify one? Don't we have this same info in the mergify PR?
 
@gloriliale Yeah I mentioned this is in the review request in Slack. I set this up for the mergify PR, but I'm unsure how much of a good idea allowing a third party merge rights is long term. But the labelling seems like it's worth saving. So i created this PR by cherry picking the commit

PR #1669, opened 2021-07-14
feat(website): rate pages

TheSisb:
This PR edits the button over the Table of Contents on component pages to ask users to rate our docs. The data is sent to Google Analytics as our backend. It is all hidden inside a neat popover that responds differently based on like/dislike. Users rating is stored in localstorage.

PR #1720, opened 2021-07-30
fix(popover): make the close button the first focus always

TheSisb:
In [another PR](https://github.com/twilio-labs/paste/pull/1669#issuecomment-887609140), it was observed that the primary focus when a popover opened may be biased or misleading. I make a change in this PR so that we consistently make the close button the primary focus - no matter the contents of a Popover window.


nkrantz (approved):
Nice simple fix 👍

SiTaggart (approved):

PR #1721, opened 2021-07-30
fix(website): waits longer for opengraph card previews

TheSisb:
This PR should hopefully fix the `Invalid Request` preview cards in production.

I had this set to `2` because of Netlify's custom widget on preview deploy URLs. On prod, it is now taking screenshots too quickly and we want to wait until the page is fully loaded.


zahnster (approved):
Looks good 👍🏼

PR #1722, opened 2021-08-02
Feat/stack customization

zahnster:
* Enable stack and stack children to respect element customizations set on the customization provider. 
* Stack and stack children now enable setting an element name on the underlying HTML element, and checks the emotion theme object to determine whether it should merge in custom styles to the ones set by the component author.
* Stories and tests added to support this change

### DIFF
--- a/packages/paste-core/layout/stack/__tests__/stack.test.tsx
+++ b/packages/paste-core/layout/stack/__tests__/stack.test.tsx
@@ -1,8 +1,9 @@
 import * as React from 'react';
-import {render} from '@testing-library/react';
+import {render, screen} from '@testing-library/react';
 import {matchers} from 'jest-emotion';
 import {Card} from '@twilio-paste/card';
 import {getStackDisplay, getStackStyles, getStackChildMargins, Stack, StackOrientation} from '../src';
+import {CustomizationProvider} from '@twilio-paste/customization';
 
 expect.extend(matchers);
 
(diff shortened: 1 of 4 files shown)


packages/paste-core/layout/stack/src/index.tsx
```diff
@@ -81,37 +76,42 @@ export const getStackChildMargins = (orientation: StackOrientation, spacing: Spa
...
-if (process.env.NODE_ENV === 'development') {
-  Stack.propTypes = {
-    orientation: ResponsiveProp(PropTypes.oneOf(['horizontal', 'vertical'])).isRequired,
-    spacing: isSpaceTokenProp,
-  };
-}
+Stack.propTypes = {
+  orientation: ResponsiveProp(PropTypes.oneOf(['horizontal', 'vertical'])).isRequired,
+  spacing: isSpaceTokenProp,
+  element: PropTypes.string,
```
nkrantz, line 114: Question here, looks like `element` wasn't added to Media Object's `propTypes` in that PR. Should it have been, or is this unnecessary?

nkrantz (commented):
Looks great! I'm curious why you were able to add `element = 'STACK'` on line 80 without having to also add `element` to the `StackProps`. Is it because `StackProps` extends `BoxElementProps` which `element` was already added to?

packages/paste-core/layout/stack/src/index.tsx
```diff
@@ -81,37 +76,42 @@ export const getStackChildMargins = (orientation: StackOrientation, spacing: Spa
...
+    const stackStyles = React.useMemo(() => getStackStyles(orientation), [orientation]);
+    const childMargins = React.useMemo(() => getStackChildMargins(orientation, spacing), [orientation, spacing]);
+    const keySeed = useUIDSeed();
+
+    return (
+      <Box element={element} {...safelySpreadBoxProps(props)} {...stackStyles} ref={ref}>
+        {validChildren.map((child, index) => {
+          return (
+            <Box
+              element={`${element}_CHILD`}
```
gloriliale, line 96: I'm not sure I understand why we're giving the children an `element` prop. Wouldn't all of the children have the same name as the parent `element` and therefore all be styled the same?

packages/paste-core/layout/stack/__tests__/stack.test.tsx
```diff
@@ -105,3 +107,89 @@ describe('Stack', () => {
...
+          <p>One</p>
+          <p>Two</p>
+        </Stack>
+      </CustomizationProvider>
+    );
+    const renderedBox = screen.getByTestId('stack-customization');
+    expect(renderedBox).toHaveAttribute('data-paste-element', 'STACK');
+  });
+
+  it('should set a custom element attribute for Stack Children', (): void => {
```
SiTaggart, line 125: The only thing I'd say about these first two is that it's not a custom element name but the default one for the component. 

Custom element name comes in the next two tests when you, as the consumer, are overriding the element name with your own.

packages/paste-core/layout/stack/src/index.tsx
SiTaggart: It creates an element name for the child based on the parent element name, but not exactly the same.

So yes, child elements have the same name, but it is different to the parent. It's potentially easier to see here in the tests https://github.com/twilio-labs/paste/pull/1722/files#diff-160e3be247ff4ca5d35a10f9e526e4cf0aa098b6b43b134a322e93e3c2ab75a4R159

packages/paste-core/layout/stack/src/index.tsx
SiTaggart: Yes, it should have

SiTaggart:
>  Is it because StackProps extends BoxElementProps which element was already added to?

@nkrantz 100% correct. Well spotted

SiTaggart (approved):

nkrantz (approved):

richbachman (approved):

PR #1725, opened 2021-08-02
fix(combobox): let Downshift handle all props to fix out of sync events

SiTaggart:
Address https://github.com/twilio-labs/paste/discussions/1631 where input events might be out of sync with Downshift internal state.

Because downshift isn't handling the events passed by the consumer, the data stored in e.target.value is out of sync with the internal state of Combobox. Meaning the value is the last value not, the current value.

### DIFF
--- a/packages/paste-core/components/combobox/src/Combobox.tsx
+++ b/packages/paste-core/components/combobox/src/Combobox.tsx
@@ -151,6 +151,7 @@ const Combobox = React.forwardRef<HTMLInputElement, ComboboxProps>(
           <ComboboxInputWrapper {...getComboboxProps({role: 'combobox'})}>
             <ComboboxInputSelect
               {...getToggleButtonProps({tabIndex: 0})}
+              // we spread props into `getInputProps` so that Downshift handles events correctly
               {...getInputProps({disabled, required, ref, ...props})}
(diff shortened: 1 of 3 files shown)


TheSisb (approved):

packages/paste-core/components/combobox/src/Combobox.tsx
```diff
@@ -151,11 +151,10 @@ const Combobox = React.forwardRef<HTMLInputElement, ComboboxProps>(
           <ComboboxInputWrapper {...getComboboxProps({role: 'combobox'})}>
             <ComboboxInputSelect
               {...getToggleButtonProps({tabIndex: 0})}
-              {...getInputProps({disabled, required, ref})}
+              {...getInputProps({disabled, required, ref, ...props})}
```
TheSisb, line 154: Might be nice to a comment in code to remind future maintainers why we're spreading props, in absence of a test.

nkrantz (approved):
So now we're not allowing anyone to set `onChange` or `onInput` so that those methods won't get overridden by the Input's methods?

SiTaggart:
> So now we're not allowing anyone to set onChange or onInput because Downshift's versions (onInputValueChange) should be used instead?

Yup. Exactly.

PR #1726, opened 2021-08-02
feat(card): add element customization

richbachman:
- [x] Enable Card to respect element customizations set on the customization provider.
- [x] Card enables setting an element name on the underlying HTML element, and checks the emotion theme object to determine whether it should merge in custom styles to the ones set by the component author.
- [x] Stories and tests added to support this change


packages/paste-core/components/card/__test__/card.test.tsx
```diff
@@ -1,55 +1,63 @@
 import * as React from 'react';
 import {render} from 'react-dom';
+import {matchers} from 'jest-emotion';
+import {render as testRender, screen} from '@testing-library/react';
```
nkrantz, line 4: Why are you importing `render` as `testRender`?

packages/paste-core/components/card/__test__/card.test.tsx
```diff
@@ -65,3 +73,30 @@ describe('Card', () => {
...
+        elements={{CARD: {backgroundColor: 'colorBackground', borderColor: 'colorBorderDestructive'}}}
+      >
+        <Card data-testid="customizable-card">Custom card</Card>
+      </CustomizationProvider>
+    );
+    const renderedCard = screen.getByTestId('customizable-card');
+    expect(renderedCard).toHaveStyleRule('background-color', 'rgb(244,244,246)');
+    expect(renderedCard).toHaveStyleRule('border-color', 'rgb(214,31,31)');
+  });
+});
```
nkrantz, line 102: Should we also test that customization should add custom styles to Card when a custom `element` attribute value is being used? Or is that redundant?

nkrantz (approved):
Nice refactors!

packages/paste-core/components/card/__test__/card.test.tsx
richbachman: Yeah. I can add a test for that also.

packages/paste-core/components/card/src/index.tsx
```diff
@@ -28,15 +35,15 @@ const Card = React.forwardRef<HTMLElement, CardProps>(
...
-  Card.propTypes = {
-    padding: isPaddingTokenProp,
-    paddingTop: isPaddingTokenProp,
-    paddingRight: isPaddingTokenProp,
-    paddingBottom: isPaddingTokenProp,
-    paddingLeft: isPaddingTokenProp,
-  };
-}
+Card.propTypes = {
+  /** element identifier for customization */
```
andipants12, line 39: NIT: Could this comment go up above by the type definition? I am not sure how we could do this, but it would be nice if the `.d.ts` file could have this comment to clarify the purpose of `element`.

richbachman: Yeah I'm into it.

packages/paste-core/components/card/stories/index.stories.tsx
```diff
@@ -1,10 +1,12 @@
 import * as React from 'react';
 import {Heading} from '@twilio-paste/heading';
 import {withKnobs, select} from '@storybook/addon-knobs';
+import {Box} from '@twilio-paste/box';
+import {Paragraph} from '@twilio-paste/paragraph';
 import {Text} from '@twilio-paste/text';
-import {Padding as PaddingProps} from '@twilio-paste/style-props';
+import type {Padding as PaddingProps} from '@twilio-paste/style-props';
```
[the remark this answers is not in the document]
richbachman: Good question. I think it was done this way because of the story also named Padding. These are pretty old stories, so I'll just rename that one, and use the type Padding properly here.

SiTaggart (approved):
Nice test refactor

PR #1733, opened 2021-08-03
feat(anchor): add element customization

TheSisb:
Adds element customization to the Anchor component so it works with CustomizationProvider.

### DIFF
--- a/packages/paste-core/components/anchor/__tests__/anchor.test.tsx
+++ b/packages/paste-core/components/anchor/__tests__/anchor.test.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import {render, screen} from '@testing-library/react';
 import {matchers} from 'jest-emotion';
+import {CustomizationProvider} from '@twilio-paste/customization';
 // @ts-ignore typescript doesn't like js imports
 import axe from '../../../../../.jest/axe-helper';
 import {Anchor} from '../src';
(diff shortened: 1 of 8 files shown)


packages/paste-core/components/anchor/__tests__/anchor.test.tsx
```diff
@@ -179,4 +180,41 @@ describe('Anchor', () => {
...
+      render(
+        <Anchor element="FUNKY_ANCHOR" href="https://twilio.com">
+          This is an anchor that links to Twilio.com with an external target and rel
+        </Anchor>
+      );
+      const renderedAnchor = screen.getByRole('link');
+      expect(renderedAnchor.getAttribute('data-paste-element')).toEqual('FUNKY_ANCHOR');
+    });
+
+    it('should add custom styles to Anchor', (): void => {
```
nkrantz, line 204: Would it be good to also include a test for adding custom styles to Anchor when a custom element data attribute is being used?

packages/paste-core/components/anchor/src/types.ts
```diff
@@ -1,11 +1,15 @@
...
 export type AnchorTabIndexes = 0 | -1;
 export type AnchorTargets = '_self' | '_blank' | '_parent' | '_top';
 export type AnchorVariants = 'default' | 'inverse';
 
-export interface AnchorProps extends React.AnchorHTMLAttributes<HTMLAnchorElement>, LayoutProps, SpaceProps {
+export interface AnchorProps
+  extends React.AnchorHTMLAttributes<HTMLAnchorElement>,
+    LayoutProps,
+    SpaceProps,
+    Pick<BoxProps, 'element'> {
```
nkrantz, line 12: Why did you add `element` from `BoxProps` instead of adding it manually below? Is it so that the comment would carry over without having to duplicate it? ( /** element identifier for customization */)

nkrantz (commented):
Looks like eslint is failing on the hover story, why is that? Other than that, left a couple comment questions but looks great!

packages/paste-core/components/anchor/src/types.ts
TheSisb: It's generally a good practice to have a "single source of truth". This way, if we ever want to change the type of `element`, we won't have to change every instance of it in every package. 

That said, I'm not sold that Box is the best source of truth, maybe it should be our from our types package instead.

packages/paste-core/components/anchor/__tests__/anchor.test.tsx
TheSisb: Good call!

packages/paste-core/components/anchor/src/types.ts
nkrantz: Makes sense, I wonder if we should all align on where we're pulling `element` from since we're doing so many of these changes. I added it directly to `FlexProps` but I can do a refactor before merging.

TheSisb: I don't think we should get too hung up about it for now. It's unlikely to ever change and shouldn't have a material impact on customers. Feel free to redeclare it as you like.

SiTaggart: The way I put is that we're extending or exposing Box functionality via this components API, so it makes sense to me to extend from `BoxProps`. But its a string so it's no big deal

SiTaggart (approved):

richbachman (approved):

PR #1734, opened 2021-08-03
feat(grid): add element customization

richbachman:
- Adds element customization to the Grid and Column components so they work with CustomizationProvider.
- Refactored `Column` to use `Box`.

### DIFF
--- a/packages/paste-core/layout/grid/src/Column.tsx
+++ b/packages/paste-core/layout/grid/src/Column.tsx
@@ -1,9 +1,6 @@
 import * as React from 'react';
 import * as PropTypes from 'prop-types';
-import type {StyledBoxProps} from '@twilio-paste/box';
-import {safelySpreadBoxProps, getCustomElementStyles} from '@twilio-paste/box';
-import {styled, compose, flexbox, layout, space} from '@twilio-paste/styling-library';
-import type {StyledComponent} from '@twilio-paste/styling-library';
+import {Box, safelySpreadBoxProps} from '@twilio-paste/box';
 import {ResponsiveProp} from '@twilio-paste/style-props';
 import type {ColumnProps, ColumnStyleProps} from './types';
 import {getStackedColumns, getColumnOffset, getColumnSpan, getColumnPadding} from './utils';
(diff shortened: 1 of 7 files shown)


packages/paste-core/layout/grid/src/Column.tsx
```diff
@@ -32,29 +35,37 @@ export const getColumnStyles = (props: ColumnProps): ColumnStyleProps => {
...
 const Column = React.forwardRef<HTMLDivElement, ColumnProps>(
-  ({as, children, count, gutter, offset, span, stretchColumnContent, vertical}, ref) => {
+  ({as, children, count, element = 'COLUMN', gutter, offset, span, stretchColumnContent, vertical, ...props}, ref) => {
     const ColumnStyles = React.useMemo(
       () => getColumnStyles({count, gutter, offset, span, stretchColumnContent, vertical}),
       [count, gutter, offset, span, stretchColumnContent, vertical]
     );
     return (
-      <StyledColumn {...ColumnStyles} as={as} ref={ref}>
+      <StyledColumn {...safelySpreadBoxProps(props)} {...ColumnStyles} as={as} data-paste-element={element} ref={ref}>
```
nkrantz, line 55: Why is this `data-paste-element={element}` instead of `element={element}`?

nkrantz (approved):
Looks good 👍

packages/paste-core/layout/grid/src/Column.tsx
richbachman: I originally did this because Column was not using Box. Its just a styled div. So I used `data-paste-element` directly since that what Box is doing. This bothered me some, so I just spent some time refactoring Column to use Box instead. That way I can just use `element` from Box. I'll push this change soon.

nkrantz: Makes sense, nice!

SiTaggart: > This bothered me some, so I just spent some time refactoring Column to use Box instead.

Is also part of the AC to refactor anything not using Box to use Box

SiTaggart (approved):

PR #1742, opened 2021-08-04
feat(avatar): add element customization 

gloriliale:
- [x] Added the element prop to Avatar
- [x] Created stories to show customization
- [x] Created tests to ensure stability


nkrantz (approved):
Woohoo looks great! Other than using `element` from `BoxProps` like Rich suggested in Slack, I think the story would benefit from some spacing and labels for each row of Avatars (default, customized, customized with custom element tag). Nice!!

TheSisb (approved):

PR #1765, opened 2021-08-11
chore(website): Add Figma link to Empty State pattern

katieporter:
Adds a link to the Empty State component in Figma.


nkrantz (approved):
Approving but side note - this link feels super buried, especially if designers are coming to this page specifically for the Figma asset. I know it's the format for other pattern pages too, but would it make sense to move this section up or at least provide some helper paragraph text under "starter kit" and above the links?

katieporter:
> Approving but side note - this link feels super buried, especially if designers are coming to this page specifically for the Figma asset. I know it's the format for other pattern pages too, but would it make sense to move this section up or at least provide some helper paragraph text under "starter kit" and above the links?

That's a great point. I have an unvalidated theory that PDs don't look for Figma files on these docs pages (having a link to a library isn't actually that helpful when what you really want to do is link a library to your project, which happens from the project file itself), so it might not be a super important thing we need to highlight. When we have Codesandboxes though, it's probably more critical.

I'm gonna do a super scrappy research project and drop a question into the ux-product-design channel and in/validate that assumption real quick. 😂

zahnster (approved):
+1 on Nora's comment but this change looks good.

richbachman (approved):

PR #1766, opened 2021-08-12
feat(heading): add element customization

richbachman:
- [x] Enable element customization to the Heading component so it works with CustomizationProvider.
- [x] Enable element customization to the Paragraph component so it works with CustomizationProvider.
- [x] Enable element customization to all List components so they work with CustomizationProvider.
- [x] Added stories and tests


nkrantz (approved):
So much customization 💪

packages/paste-core/components/heading/__tests__/heading.test.tsx
```diff
@@ -121,38 +110,90 @@ describe('Heading', () => {
...
+          elements={{foo: {color: 'colorTextWeak', backgroundColor: 'colorBackground'}}}
+        >
+          <Heading as="h1" variant="heading10" element="foo">
+            Custom heading
+          </Heading>
+        </CustomizationProvider>
+      );
+      const renderedHeading = screen.getByRole('heading');
+      expect(renderedHeading).toHaveStyleRule('background-color', 'rgb(244,244,246)');
+      expect(renderedHeading).toHaveStyleRule('color', 'rgb(96,107,133)');
```
SiTaggart, line 167: for any component that has a variant I should be able to target that specific variant to style.

```
elements={{
  HEADING: {
    color: 'colorTextWeak',
    variants: {
      heading10: {
        color: 'colorTextSuccess',
      }
    }
  }
  foo: {
    color: 'colorTextWeak',
    variants: {
      heading10: {
        color: 'colorTextSuccess',
      }
    }
  }
}}
```

I forget the exact syntax, but it's something like that

packages/paste-core/components/list/src/index.tsx
```diff
@@ -1,108 +1,5 @@
...
-  );
-});
-ListItem.displayName = 'ListItem';
-
-export {OrderedList, UnorderedList, ListItem};
+export * from './List';
+export * from './ListItem';
+export * from './OrderedList';
+export * from './UnorderedList';
+export * from './types';
```
SiTaggart, line 5: Nice refactor

SiTaggart (commented):
I think this one is just missing variants for Heading. Add the functionality and tests around that and you're good

packages/paste-core/components/heading/__tests__/heading.test.tsx
richbachman: Added stories and tests for customizing a variant. Outside of the `CustomizationProvider` bug, this implementation works.

packages/paste-core/components/heading/__tests__/heading.test.tsx
```diff
@@ -1,107 +1,96 @@
...
           <Heading as="h1" variant="heading10">
             This is an H1
           </Heading>
         </Theme.Provider>
-      )
-      .toJSON();
-    expect(tree).toMatchSnapshot();
-  });
+      );
+      expect(asFragment()).toMatchSnapshot();
```
andipants12, line 22: ❓ Not blocking, but maybe we could phase out the snapshot tests as we touch these older test files?

andipants12 (approved):

packages/paste-core/components/heading/__tests__/heading.test.tsx
richbachman: Yeah. I went back and forth on these. I feel like the snapshots make sense here, since we're verifying the element is rendering as a certain heading level, along with certain styles. I could be convinced otherwise though.

SiTaggart: It's be nice if we can. But not a blocker

SiTaggart (approved):

PR #1771, opened 2021-08-13
feat(input): add element customization

richbachman:
- [x] Enable element customization to the Input component so it works with CustomizationProvider.
- [x] Enable element customization to InputBox, Prefix, and Suffix components so they work with CustomizationProvider.
- [x] Added style props to Box to support refactor of Input
- [x] Added stories and tests


packages/paste-core/components/input-box/__tests__/input-box.test.tsx
```diff
@@ -0,0 +1,31 @@
...
+import {CustomizationProvider} from '@twilio-paste/customization';
+import {InputBox} from '../src';
+
+expect.extend(matchers);
+
+describe('HTML attributes', () => {
+  it('should set a element data attribute for InputBox', () => {
+    render(<InputBox element="foo">input-box</InputBox>);
+    expect(screen.getByText('input-box').getAttribute('data-paste-element')).toEqual('foo');
+  });
```
SiTaggart, line 13: I wouldn't mind a test here for in all tests cases that ensures the default element name is set, along side a custom one.

packages/paste-core/components/input-box/src/InputBox.tsx
```diff
@@ -44,19 +46,18 @@ const InputBox = React.forwardRef<HTMLDivElement, InputBoxProps>(
...
+  element: PropTypes.string,
+  disabled: PropTypes.bool,
+  hasError: PropTypes.bool,
+  insertAfter: PropTypes.node,
+  insertBefore: PropTypes.node,
+  readOnly: PropTypes.bool,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  type: PropTypes.oneOf(['text', 'email', 'hidden', 'number', 'password', 'search', 'tel', 'date', 'time']) as any,
+  // eslint-disable-next-line @typescript-eslint/no-explicit-any
+  variant: PropTypes.oneOf(['default', 'inverse']) as any,
```
SiTaggart, line 60: So fun one on these. When ever we have a variant, we should also be passing variant down to Box, and creating a test that we can add custom styles to the variants

packages/paste-core/components/input-box/src/InputChevronWrapper.tsx
```diff
@@ -1,18 +1,34 @@
...
+    <Box
+      alignItems="center"
+      display="inline-flex"
+      position="absolute"
+      pointerEvents="none"
+      right="space30"
+      top="50%"
+      transform="translateY(-50%)"
+      zIndex="zIndex10"
+      element={`${element}_CHEVRON_WRAPPER`}
```
SiTaggart, line 21: We might need to be a little careful here. `element` from Box is optional, and I know we only use these as a compositional element to Input, but I still think we should either:

- Set a default element name in the arguments
- Make `element` required so that it's never NOT passed (preferred)

SiTaggart (commented):
Main areas to address in here are:

- ability to style and customize variants. With tests
- testing the default data attribute element name is present
- ensuring the element name is typed appropriately on the composite components such as chevron wrapper, or suffix / prefix

SiTaggart (commented):
For the components that are using the shared components, that are having to have element added due to types, you  will need to add a change set for each of them. Or at least one changeset that covers all three, combobox, select, and textarea.

If you don't, when someone installs the latest core, those three components won't have updated, and so the types will fail for them. Can just be a patch

richbachman:
> For the components that are using the shared components, that are having to have element added due to types, you will need to add a change set for each of them. Or at least one changeset that covers all three, combobox, select, and textarea.
> 
> If you don't, when someone installs the latest core, those three components won't have updated, and so the types will fail for them. Can just be a patch

@SiTaggart makes sense. Changesets added.

SiTaggart (approved):

nkrantz (approved):

PR #1781, opened 2021-08-16
feat(separator): add element customization

richbachman:
- [x] Enable element customization to the Heading component so it works with CustomizationProvider.
- [x] Added stories and tests

### DIFF
--- a/packages/paste-core/components/separator/__tests__/index.spec.tsx
+++ b/packages/paste-core/components/separator/__tests__/index.spec.tsx
@@ -16,12 +16,12 @@ describe('Separator', () => {
       expect(getByRole('separator')).not.toBeNull();
     });
 
-    it('should render set horizontal orientation', () => {
+    it('should render correctly when the value of orientation prop is horizontal', () => {
       const {getByRole} = render(<Separator orientation="horizontal" />);
       expect(getByRole('separator').getAttribute('aria-orientation')).toBe('horizontal');
     });
 
-    it('should render set vertical orientation', () => {
+    it('should render correctly when the value of orientation prop is vertical', () => {
       const {getByRole} = render(<Separator orientation="vertical" />);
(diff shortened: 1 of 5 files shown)


packages/paste-core/components/separator/__tests__/index.spec.tsx
```diff
@@ -1,76 +1,119 @@
...
+        >
+          <Separator orientation="horizontal" data-testid="custom-separator" />
+        </CustomizationProvider>
+      );
+      const renderedSeparator = screen.getByTestId('custom-separator');
+      expect(renderedSeparator).toHaveStyleRule('border-color', 'rgb(96,107,133)');
+      expect(renderedSeparator).toHaveStyleRule('margin', '1rem');
+    });
+
+    it('should add custom styles to Grid with a custom element data attribute', (): void => {
```
SiTaggart, line 92: I think just the description needs updating here

SiTaggart (approved):

nkrantz (approved):

packages/paste-core/components/separator/__tests__/index.spec.tsx
```diff
@@ -1,76 +1,119 @@
...
-    const tree = renderer
-      .create(
-        <Theme.Provider>
+  describe('Render', () => {
+    it('should render', () => {
+      const {getByRole} = render(<Separator orientation="horizontal" />);
+      expect(getByRole('separator')).not.toBeNull();
+    });
+
+    it('should render set horizontal orientation', () => {
```
andipants12, line 19: NIT: Maybe this could be `should set "aria-orientation" to "horizontal" when the value of orientation prop is "horizontal"`
 Or, for something shorter `should render correctly when the value of orientation prop is horizontal`. The wording feels off here.

andipants12 (approved):
Looks good! Left a comment on the test file, since the wording of the first two `it` statements feels off. Not blocking though.

PR #1784, opened 2021-08-17
fix(box): fix word break in box primitive

rpresb:
Fixing word break in box primitive to avoid text to overflow the boundaries of the container.

- Before
image

- After
image

*Those aren't valid SIDs*

**Contributing to Twilio**


SiTaggart:
Hi there @rpresb. 

Thanks for the PR, however this isn't something we would recommend at the Box level, for all Boxes, as it might introduce adverse side effects to others using the component in other products.

You might only need to wrap you content with a Box and set `wordBreak` like so 

```
<Card>
  <Box wordBreak="break-word">
    <Heading as="h5" variant="heading30">
      Account SID
    </Heading>
    <Anchor href="https://google.com">
      aaassddfgghkljdjdhdhgfhgjghkgkjfurunfmkkjf
    </Anchor>
  </Box>
</Card>
```

Example https://codesandbox.io/s/break-word-example-gibyj?file=/src/index.tsx:438-741

If you need further assistance, please open a [Discussion](https://github.com/twilio-labs/paste/discussions/new)

rpresb:
Thanks, that makes so much sense.

PR #1785, opened 2021-08-17
fix(box): correctly handle variants that are not on the element object

SiTaggart:
When a component was setting a variant that was not present on the element object, it would throw an error as it was trying to merge an object into undefined.

This simply checks that there is a matching variant on the element object, before it merges the styles.

### DIFF
--- a/packages/paste-core/primitives/box/__tests__/StyleFunctions.test.tsx
+++ b/packages/paste-core/primitives/box/__tests__/StyleFunctions.test.tsx
@@ -171,4 +171,27 @@ describe('getCustomElementStyles', () => {
       });
     }
   });
+
+  it('should not throw when variants that are not present on the theme are set on a component', () => {
+    // this just covers a bug where Box would be looking for a variant that is not present on the theme
+    // and rather than gracefully handle this, it would throw an error
+
+    const primaryCardProps = {
+      'data-paste-element': 'CARD',
+      variant: 'noneexistantvariant',
+      ...mockTheme,
+    };
+    // @ts-expect-error because I'm not setting the whole theme
+    const cardCSSFunc = getCustomElementStyles(primaryCardProps);
+    if (typeof cardCSSFunc === 'function') {
+      // eslint-disable-next-line jest/no-conditional-expect
+      expect(cardCSSFunc()).toEqual({
+        color: 'green',
+        borderColor: 'purple',
+        ':hover': {
+          cursor: 'pointer',
+        },
+      });
+    }
+  });
 });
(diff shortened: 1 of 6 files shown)


richbachman:
Do we need to do the same thing for Text here?

SiTaggart:
> Do we need to do the same thing for Text here?

Ooooh smart

packages/paste-core/primitives/box/__tests__/StyleFunctions.test.tsx
```diff
@@ -171,4 +171,27 @@ describe('getCustomElementStyles', () => {
       });
     }
   });
+
+  it('should handle variants that are not present on the theme', () => {
```
andipants12, line 175: NIT: For readability could this include a brief description of how it should handle variants that are not present? For example, does the function return defaults for variants that do not match the theme? Does it throw an error?

packages/paste-core/primitives/box/src/StyleFunctions.ts
```diff
@@ -121,12 +122,11 @@ export const getCustomElementStyles = (props: StyledBoxProps): (() => CSSObject)
...
     if (themeElements[targetElement] != null) {
       const elementOverrides = themeElements[targetElement];
-      const computedStyles = css(elementOverrides)(props);
+      const computedStyles = css(elementOverrides)(props) as PasteCustomCSS;
       const {variants, ...elementStyles} = computedStyles;
       let variantStyles = {};
 
-      if (props.variant != null && variants != null) {
-        // @ts-ignore typing of css function returns a cssObject which doesn't think variants exists
+      if (props.variant != null && variants != null && variants[props.variant] != null) {
```
andipants12, line 129: Could you provide some context for this use case of the non-strict inequality operator?

andipants12 (commented):
Just a few things that I left comments on:
1. One small "nit pick" about being more explicit in an `it` statement about the condition its testing.
2. Clarification on the use of non-strict inequality operators. I wonder, if the goal is to assess whether or not a specific key or property exists on an object, if we could instead use something like `Object.prototype.hasOwnProperty`.

richbachman (approved):

packages/paste-core/primitives/box/src/StyleFunctions.ts
SiTaggart: So this does a little more than it looks like, and it's also Typescript based.

`x != null` catches that `x !== undefined` _and_ `x !== null`, and when checking for those things this is a little safer this way. It's something we do through out the system, and @TheSisb can go into it more if you would like (he introduced it).

So, `props.variant`, and `variants` are optional types, which means they can be `undefined`, but also defined as `null`, so Typescript forces the check here

packages/paste-core/primitives/box/__tests__/StyleFunctions.test.tsx
SiTaggart: Updated, and threw in a more explicit test to expect that it does not throw too. Good suggestion

packages/paste-core/primitives/box/src/StyleFunctions.ts
TheSisb: +1 to what Simon said.  The only time I don't use strict equality (`===`) is when checking against null & undefined, via `!= null`

TheSisb (approved):

PR #1800, opened 2021-08-23
feat(display-pill-group): add package

TheSisb:
Adds a new Paste component called "DisplayPillGroup". This is the pill group component intended to be used outside of forms, for visual purposes.  Another PR will be coming soon with the "FormPillGroup" component package. 

**Why did we split these up?**
- To keep the API clear and composable; it's more obvious to use the FormPillGroup variant when within forms, than if we had named it something else.
- To reduce the bundle size when not using pills in forms.
- For accessibility; each component has different keyboard controls and semantics for screenreaders

### DIFF
--- a/.codesandbox/ci.json
+++ b/.codesandbox/ci.json
@@ -25,6 +25,7 @@
     "/packages/paste-design-tokens",
     "/packages/paste-core/components/disclosure",
     "/packages/paste-core/primitives/disclosure",
+    "/packages/paste-core/components/display-pill-group",
     "/packages/paste-libraries/dropdown",
     "/packages/paste-core/layout/flex",
     "/packages/paste-core/layout/grid",
--- a/packages/paste-core/components/display-pill-group/__tests__/index.spec.tsx
+++ b/packages/paste-core/components/display-pill-group/__tests__/index.spec.tsx
@@ -1,0 +1,126 @@
+import * as React from 'react';
+import {render} from '@testing-library/react';
+import {matchers} from 'jest-emotion';
+import {CustomizationProvider} from '@twilio-paste/customization';
+// @ts-ignore typescript doesn't like js imports
+import axe from '../../../../../.jest/axe-helper';
+import {DisplayPillGroup, DisplayPill} from '../src';
+import {Basic, CustomDisplayPillGroup} from '../stories/index.stories';
+
+expect.extend(matchers);
+
+describe('DisplayPillGroup', () => {
+  // Verifies that the correct aria attributes and semantics are met
+  describe('Rendered shape', () => {
+    it('should render correctly', () => {
+      const {getByTestId, getByText} = render(<Basic />);
+      expect(getByText('Tennis')).toBeDefined();
+
+      const group = getByTestId('display-pill-group');
+      expect(group.getAttribute('aria-label')).toBe('Your favorite sports:');
+      expect(group.tagName).toBe('UL');
+
+      const pillLink = getByTestId('display-pill-anchor');
+      expect(pillLink.getAttribute('href')).toBe('https://google.com');
+      expect(pillLink.getAttribute('rel')).toBe('noreferrer noopener');
+      expect(pillLink.getAttribute('target')).toBe('_blank');
+      expect(pillLink.tagName).toBe('A');
+
+      const pillStandard = getByTestId('display-pill-standard');
+      expect(pillStandard.tagName).toBe('DIV');
+    });
+  });
+
+  // Verifies the component is fully customizable
+  describe('Customization', () => {
+    it('should set an element data attribute for DisplayPillGroup & DisplayPill', () => {
+      const {getByTestId} = render(<Basic />);
+      const group = getByTestId('display-pill-group');
+      expect(group.getAttribute('data-paste-element')).toEqual('DISPLAY_PILL_GROUP');
+
+      const pillLink = getByTestId('display-pill-anchor');
+      expect(pillLink.getAttribute('data-paste-element')).toEqual('DISPLAY_PILL');
+
+      const pillStandard = getByTestId('display-pill-standard');
+      expect(pillStandard.getAttribute('data-paste-element')).toEqual('DISPLAY_PILL');
+    });
+
+    it('should set a custom element data attribute for DisplayPillGroup & DisplayPill', () => {
+      const {getByTestId} = render(
+        <DisplayPillGroup element="CUSTOM_GROUP" data-testid="group" aria-label="Your favorite sports:">
+          <DisplayPill element="CUSTOM_PILL" data-testid="pill">
+            A
+          </DisplayPill>
+        </DisplayPillGroup>
+      );
+      const group = getByTestId('group');
+      expect(group.getAttribute('data-paste-element')).toEqual('CUSTOM_GROUP');
+      const pill = getByTestId('pill');
+      expect(pill.getAttribute('data-paste-element')).toEqual('CUSTOM_PILL');
+    });
+
+    it('should add custom styles to DisplayPillGroup & DisplayPill', () => {
+      const {getByTestId} = render(<CustomDisplayPillGroup />);
+
+      const group = getByTestId('display-pill-group');
+      expect(group).toHaveStyleRule('margin', '0.75rem');
+
+      const pillLink = getByTestId('display-pill-anchor');
+      expect(pillLink).toHaveStyleRule('background-color', 'rgb(231,220,250)');
+
+      const pillStandard = getByTestId('display-pill-standard');
+      expect(pillStandard).toHaveStyleRule('background-color', 'rgb(231,220,250)');
+    });
+    it('should add custom styles to custom element DisplayPillGroup & DisplayPill', () => {
+      const {getByTestId} = render(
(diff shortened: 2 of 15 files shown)


.codesandbox/ci.json
```diff
@@ -25,6 +25,7 @@
     "/packages/paste-design-tokens",
     "/packages/paste-core/components/disclosure",
     "/packages/paste-core/primitives/disclosure",
+    "/packages/paste-core/components/display-pill-group",
```
gloriliale, line 28: What is it that this file is helping us do?

packages/paste-core/components/display-pill-group/package.json
```diff
@@ -0,0 +1,59 @@
+{
+  "name": "@twilio-paste/display-pill-group",
+  "version": "0.0.0",
+  "category": "data display",
+  "status": "production",
+  "description": "Used to display a collection of labels.",
```
gloriliale, line 6: Is "labels" the least confusing language we can use here? Would something like "objects" or "entities" be too esoteric? The term "label" leans too close to Badge to me.

packages/paste-core/components/display-pill-group/stories/index.stories.tsx
```diff
@@ -0,0 +1,77 @@
...
+        href="https://google.com"
+      >
+        <CalendarIcon decorative size="sizeIcon10" />
+        Tennis
+      </DisplayPill>
+      <DisplayPill data-testid="display-pill-blank">Football</DisplayPill>
+      <DisplayPill href="/">Baseball</DisplayPill>
+      <DisplayPill>Basketball</DisplayPill>
+      <DisplayPill>Soccer</DisplayPill>
+    </DisplayPillGroup>
```
gloriliale, line 27: Right now, the behavior I get when I tab through these Display Pill groups is that the focus skips over the Display Pills that don't have hrefs (from Tennis to Baseball, it skips Football, then skips the rest of the Display Pills on the next tab press). Is that what we want?

packages/paste-core/components/display-pill-group/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,126 @@
...
+      expect(group.getAttribute('aria-label')).toBe('Your favorite sports:');
+      expect(group.tagName).toBe('UL');
+
+      const pillLink = getByTestId('display-pill-anchor');
+      expect(pillLink.getAttribute('href')).toBe('https://google.com');
+      expect(pillLink.getAttribute('rel')).toBe('noreferrer noopener');
+      expect(pillLink.getAttribute('target')).toBe('_blank');
+      expect(pillLink.tagName).toBe('A');
+
+      const pillBlank = getByTestId('display-pill-blank');
```
gloriliale, line 29: Nit: terming this as "Blank" at first made me think this pill would have nothing inside of it, which is not the case. Would a more clear way to name this be "plain" or "default"?

packages/paste-core/components/display-pill-group/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,126 @@
...
+      const group = getByTestId('display-pill-group');
+      expect(group).toHaveStyleRule('margin', '0.75rem');
+
+      const pillLink = getByTestId('display-pill-anchor');
+      expect(pillLink).toHaveStyleRule('background-color', 'rgb(231,220,250)');
+
+      const pillBlank = getByTestId('display-pill-blank');
+      expect(pillBlank).toHaveStyleRule('background-color', 'rgb(231,220,250)');
+    });
+    it('should add custom styles to custom element Avatar', () => {
```
gloriliale, line 74:
```suggestion
    it('should add custom styles to custom element Display Pill Group', () => {
```

gloriliale: What is the difference between this set of tests and the ones above for custom styles?

packages/paste-core/components/display-pill-group/src/DisplayPill.tsx
```diff
@@ -0,0 +1,41 @@
...
+interface DisplayPillProps extends BoxElementProps {
+  href?: string;
+  element?: string;
+  children: React.ReactNode;
+}
+
+export const DisplayPill = React.forwardRef<HTMLElement, DisplayPillProps>(
+  ({element = 'DISPLAY_PILL', ...props}, ref) => {
+    return (
+      <Box as="li" listStyleType="none">
```
gloriliale, line 15: What does the `listStyleType` determine? I haven't used it before.

gloriliale (approved):
Yay Pills! These are looking great. I had a few questions/ comments, mostly for my own understanding. I also was wondering (and these might have answers I was away for): 
1. Have we determined that we don't want these Display Pills to be closable within their Display Pill Group?
2. Will people be able to set event handlers on these Display Pills?

.codesandbox/ci.json
SiTaggart: It's telling Codesandbox which packages to upload to it's own package manager, for our PR codesandboxes. 

Those sandboxes in each PR use the latest code on the branch, so Codesandbox uploads a temporary package so it simulates install the package from NPM. We can then use Codesandbox to very the latest packages without having to deploy them.

Each time we add a new package, we need to inform codesand of that additoin

packages/paste-core/components/display-pill-group/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,126 @@
...
+      const group = getByTestId('display-pill-group');
+      expect(group).toHaveStyleRule('margin', '0.75rem');
+
+      const pillLink = getByTestId('display-pill-anchor');
+      expect(pillLink).toHaveStyleRule('background-color', 'rgb(231,220,250)');
+
+      const pillBlank = getByTestId('display-pill-blank');
+      expect(pillBlank).toHaveStyleRule('background-color', 'rgb(231,220,250)');
+    });
+  });
```
SiTaggart, line 116: 🎉

packages/paste-core/components/display-pill-group/src/DisplayPill.tsx
```diff
@@ -0,0 +1,41 @@
+import * as React from 'react';
+import {Box, safelySpreadBoxProps} from '@twilio-paste/box';
+import type {BoxElementProps} from '@twilio-paste/box';
+import {secureExternalLink} from '@twilio-paste/anchor';
+
+interface DisplayPillProps extends BoxElementProps {
```
SiTaggart, line 6: Probably need to export the interfaces here so the consumer can grab your types

packages/paste-core/components/display-pill-group/src/DisplayPillGroup.tsx
```diff
@@ -0,0 +1,27 @@
+import * as React from 'react';
+import {Box, safelySpreadBoxProps} from '@twilio-paste/box';
+
+interface DisplayPillGroupProps {
```
SiTaggart, line 4: Same here, export the interface

SiTaggart (commented):
Couple of super tiny interface exports, other wise this is basically bang on the money 🥇

packages/paste-core/components/display-pill-group/package.json
TheSisb: Yeah I kind of winged this, open to suggestions!

packages/paste-core/components/display-pill-group/stories/index.stories.tsx
TheSisb: That's correct. It only focuses over href pills.

packages/paste-core/components/display-pill-group/src/DisplayPill.tsx
TheSisb: Browsers have default CSS, so things like `li` have bullets by default. `list-style-type` is CSS to change that behavior: https://developer.mozilla.org/en-US/docs/Web/CSS/list-style-type

TheSisb:
>Have we determined that we don't want these Display Pills to be closable within their Display Pill Group?

I was unable to find a use for dismissable display pills, so I opted to exclude them here. If we discover a need for them I can add them at a later date. Afaict, it's only useful for the form pill group.

> Will people be able to set event handlers on these Display Pills?

Yes and no. We spread props down, so any normal event handlers will get passed, but no in that we typically just want this to be a display and not a control like the FormPillGroup.

packages/paste-core/components/display-pill-group/__tests__/index.spec.tsx
TheSisb: One uses the default `element` prop and the other passes in a custom `element` prop for an instanced restyle.

packages/paste-core/components/display-pill-group/package.json
SiTaggart: I liked the advise @katieporter gave in the Badge docs for the description. The format for all description are aligning around:

`A(n) [component name] is a(n) [type of element] that does [function].`

> A Display Pill Group is a collection of Pills that are used to visually represent a collection of entities outside of a data entry form

or soemthing

SiTaggart (approved):

PR #1801, opened 2021-08-23
chore(button): link and destructive link should not be centered

andipants12:
## Description
Since buttons of the "link" variants have no visual containing box, we are updating them to align text to the left, instead of the current default for button which is centered.

This addresses the concerns raised in [this discussion about link variant buttons in a table.](https://github.com/twilio-labs/paste/discussions/1773)

### Summary of changes
- [x] updated `ButtonContent`, the container for the button text and icons, to apply justified contents to the center only for variants that are not links.
- [x] updated `DestructiveLinkButton` and `LinkButton` to override the button text alignment default (from "center" to "left")
- [x] added test coverage for these modifications
- [x] added test coverage for the default styles of all variants (since that was not previously included in the test suite

### DIFF
--- a/.changeset/chilly-eggs-provide.md
+++ b/.changeset/chilly-eggs-provide.md
@@ -1,0 +1,6 @@
+---
+'@twilio-paste/button': patch
+'@twilio-paste/core': patch
+---
+
+[Button] reset and link variants should have left aligned text
--- a/.changeset/chilly-eggs-provide.md
+++ b/.changeset/chilly-eggs-provide.md
@@ -3,4 +3,4 @@
 '@twilio-paste/core': patch
 ---
 
-[Button] reset and link variants should have left aligned text
+[Button] All link variants should have left aligned text.
--- a/packages/paste-core/components/alert/__tests__/__snapshots__/index.spec.tsx.snap
+++ b/packages/paste-core/components/alert/__tests__/__snapshots__/index.spec.tsx.snap
@@ -130,10 +130,6 @@ exports[`Alert Variant error Should render an error alert with dismiss button 1`
   -webkit-text-decoration: inherit;
   text-decoration: inherit;
   opacity: 1;
-  -webkit-box-pack: center;
-  -webkit-justify-content: center;
-  -ms-flex-pack: center;
-  justify-content: center;
 }
 
 .emotion-9 {
@@ -174,6 +170,7 @@ exports[`Alert Variant error Should render an error alert with dismiss button 1`
 .emotion-5 {
   box-sizing: border-box;
   width: auto;
+  text-align: left;
   -webkit-appearance: none;
   -moz-appearance: none;
   appearance: none;
@@ -496,10 +493,6 @@ exports[`Alert Variant neutral Should render a neutral alert with dismiss button
   -webkit-text-decoration: inherit;
   text-decoration: inherit;
   opacity: 1;
-  -webkit-box-pack: center;
-  -webkit-justify-content: center;
-  -ms-flex-pack: center;
-  justify-content: center;
 }
 
 .emotion-9 {
@@ -561,6 +554,7 @@ exports[`Alert Variant neutral Should render a neutral alert with dismiss button
 .emotion-5 {
   box-sizing: border-box;
   width: auto;
+  text-align: left;
   -webkit-appearance: none;
   -moz-appearance: none;
   appearance: none;
@@ -864,10 +858,6 @@ exports[`Alert Variant warning Should render an warning alert with dismiss butto
   -webkit-text-decoration: inherit;
   text-decoration: inherit;
   opacity: 1;
-  -webkit-box-pack: center;
-  -webkit-justify-content: center;
-  -ms-flex-pack: center;
-  justify-content: center;
 }
 
 .emotion-9 {
@@ -908,6 +898,7 @@ exports[`Alert Variant warning Should render an warning alert with dismiss butto
 .emotion-5 {
   box-sizing: border-box;
   width: auto;
+  text-align: left;
   -webkit-appearance: none;
   -moz-appearance: none;
   appearance: none;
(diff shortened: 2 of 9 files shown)


packages/paste-core/components/button/__tests__/button.test.tsx
```diff
@@ -8,375 +8,548 @@ import {shallow, mount} from 'enzyme';
...
+        <Theme.Provider theme="default">
+          <Button variant="destructive_secondary" data-testid="destructive_secondary-styles">
+            Destructive secondary
+          </Button>
+        </Theme.Provider>
+      );
+
+      const button = getByTestId('destructive_secondary-styles');
+      expect(button).not.toHaveStyleRule('text-align', 'left');
+      expect(button).toHaveStyleRule('color', 'rgb(214,31,31)');
```
TheSisb, line 455: I'm not too sure how valuable some of these are, mainly the color styles. Won't VRT in the stories catch changes here?  

As an aside, if we wanted to guarantee we always used the same token (rather than mapping the value), we can remove the `Theme.Provider` wrapper and check the token directly.

packages/paste-core/components/button/src/ResetButton.tsx
```diff
@@ -30,6 +30,7 @@ const ResetButton = React.forwardRef<HTMLButtonElement, DirectButtonProps>(
       <Box
         ref={ref}
         width={fullWidth ? '100%' : 'auto'}
+        textAlign="left"
```
TheSisb, line 33: I'm pretty sure the reset button doesn't have any alignment set, which should default to `inherit` or `initial` iirc. Adding this could cause issues in existing implementations, maybe we should leave it off?

packages/paste-core/components/button/src/index.tsx
```diff
@@ -112,7 +112,13 @@ const handlePropValidation = ({
   }
 };
 
+const variantsWithoutBoundingBox = new Set(['reset', 'link', 'destructive_link']);
```
TheSisb, line 115: 👍

packages/paste-core/components/button/src/index.tsx
```diff
@@ -112,7 +112,13 @@ const handlePropValidation = ({
...
+
 const ButtonContents: React.FC<ButtonContentsProps> = ({buttonState, children, showLoading, variant}) => {
+  const buttonVariantHasBoundingBox = React.useMemo(() => variant && variantsWithoutBoundingBox.has(variant), [
```
TheSisb, line 118: The Set `has()` function will return false for `undefined` and `null`, and this isn't very costly so we can avoid memoizing.

Simply doing `variantsWithoutBoundingBox.has(variant)` inline on line 130 should be enough

gloriliale (approved):
Looks gooood! I also agree with the discussion peeps were having on Slack that we should not have align left for reset button, which I think Shadi's mentioned here as well.

packages/paste-core/components/button/__tests__/button.test.tsx
andipants12: Yes that makes total sense! I did that previously, but saw some test suites that are not checking the token values, but the actual css values. Will make that change now. Thanks!

packages/paste-core/components/alert/__tests__/__snapshots__/index.spec.tsx.snap
```diff
@@ -174,6 +170,7 @@ exports[`Alert Variant error Should render an error alert with dismiss button 1`
 .emotion-5 {
   box-sizing: border-box;
   width: auto;
+  text-align: left;
```
andipants12, line 173: These Alert examples use a button as anchor.

packages/paste-core/components/alert/__tests__/__snapshots__/index.spec.tsx.snap
```diff
@@ -561,6 +554,7 @@ exports[`Alert Variant neutral Should render a neutral alert with dismiss button
...
```
andipants12, line 557: These Alert examples use a button as anchor.

packages/paste-core/components/alert/__tests__/__snapshots__/index.spec.tsx.snap
```diff
@@ -908,6 +898,7 @@ exports[`Alert Variant warning Should render an warning alert with dismiss butto
...
```
andipants12, line 901: These Alert examples use a button as anchor.

packages/paste-core/components/button/src/index.tsx
andipants12: Hmm so I made this change but TS is unsatisfied since `variant` can be `undefined`. Do you mind if I leave as is?

TheSisb: TS > me.  But we can remove the `useMemo` still

packages/paste-core/components/button/src/DestructiveLinkButton.tsx
```diff
@@ -43,6 +43,7 @@ const DestructiveLinkButton = React.forwardRef<HTMLButtonElement, DirectButtonPr
       <Box
         ref={ref}
         width={fullWidth ? '100%' : 'auto'}
+        textAlign="left"
```
TheSisb, line 46: Because the ButtonContents is `display: flex`, we don't actually need these. It makes the alignment 100% dependent on the justify rule in the index.tsx file

TheSisb (changes requested):

TheSisb (approved):

.changeset/chilly-eggs-provide.md
```diff
@@ -0,0 +1,6 @@
+---
+'@twilio-paste/button': patch
+'@twilio-paste/core': patch
+---
+
+[Button] reset and link variants should have left aligned text
```
katieporter, line 6: I don't really know what this is, but should it be updated to just mention link variants and not reset variants?

PR #1803, opened 2021-08-24
feat(form-pill-group): add package

TheSisb:
The second pill group package ([see the first package PR here](https://github.com/twilio-labs/paste/pull/1800)).

The FormPillGroup is intended to be used within form elements.

**Why did we split these up?**

- To keep the API clear and composable; it's more obvious to use the FormPillGroup variant when within forms, than if we had named it something else.
- To reduce the bundle size when not using pills in forms.
- For accessibility; each component has different keyboard controls and semantics for screenreaders

### DIFF
--- a/packages/paste-core/components/form-pill-group/__tests__/index.spec.tsx
+++ b/packages/paste-core/components/form-pill-group/__tests__/index.spec.tsx
@@ -1,0 +1,191 @@
+import * as React from 'react';
+import {render, fireEvent} from '@testing-library/react';
+import {matchers} from 'jest-emotion';
+import {CustomizationProvider} from '@twilio-paste/customization';
+// @ts-ignore typescript doesn't like js imports
+import axe from '../../../../../.jest/axe-helper';
+import {useFormPillState, FormPillGroup, FormPill} from '../src';
+import {Basic, SelectAndRemove, CustomFormPillGroup} from '../stories/index.stories';
+
+expect.extend(matchers);
+
+const CustomElementFormPillGroup: React.FC = () => {
+  const pillState = useFormPillState();
+
+  return (
+    <form>
+      <FormPillGroup
+        {...pillState}
+        element="CUSTOM_PILL_GROUP"
+        data-testid="form-pill-group"
+        aria-label="Your favorite sports:"
+      >
+        <FormPill {...pillState} element="CUSTOM_PILL" data-testid="form-pill">
+          Tennis
+        </FormPill>
+      </FormPillGroup>
+    </form>
+  );
+};
+
+describe('FormPillGroup', () => {
+  describe('Rendered shape', () => {
+    it('has the correct aria attributes and semantic HTML', () => {
+      const {getByTestId, getByText} = render(<Basic />);
+      expect(getByText('Tennis')).toBeDefined();
+
+      const group = getByTestId('form-pill-group');
+      expect(group.getAttribute('aria-label')).toBe('Your favorite sports:');
+      expect(group.tagName).toBe('DIV');
+      expect(group.getAttribute('role')).toBe('listbox');
+
+      const pill = getByTestId('form-pill');
+      expect(pill.tagName).toBe('BUTTON');
+      expect(pill.getAttribute('role')).toBe('option');
+      expect(pill.getAttribute('aria-selected')).toBe('false');
+
+      const pillSelected = getByTestId('form-pill-selected');
+      expect(pillSelected.getAttribute('aria-selected')).toBe('true');
+    });
+  });
+
+  describe('Selecting and Removing', () => {
+    it('can select and navigate pills', () => {
+      const {getByTestId} = render(<SelectAndRemove />);
+
+      // Get the first pill
+      const firstPill = getByTestId('form-pill-0');
+      // Click it and make sure it selected
+      fireEvent.click(firstPill);
+      expect(firstPill.getAttribute('aria-selected')).toBe('true');
+
+      // Make sure it deselects on click
+      fireEvent.click(firstPill);
+      expect(firstPill.getAttribute('aria-selected')).toBe('false');
+
+      // Make sure it selects on Enter key
+      fireEvent.keyDown(firstPill, {key: 'Enter', code: 'Enter'});
+      expect(firstPill.getAttribute('aria-selected')).toBe('true');
+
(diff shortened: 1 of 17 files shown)


packages/paste-core/components/form-pill-group/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,121 @@
...
+        <FormPill {...pillState} element="CUSTOM_PILL" data-testid="form-pill">
+          Tennis
+        </FormPill>
+      </FormPillGroup>
+    </form>
+  );
+};
+
+describe('FormPillGroup', () => {
+  // Verifies that the correct aria attributes and semantics are met
```
andipants12, line 32: Nit: Instead of comments here, could this information be included in the `describe` and/or `it` blocks? These comments provide helpful context, which I think could be super useful to see in the test suite command line output.

packages/paste-core/components/form-pill-group/src/FormPill.tsx
```diff
@@ -0,0 +1,80 @@
...
+  ({element = 'FORM_PILL', selected = false, ...props}, ref) => (
+    <Box
+      {...safelySpreadBoxProps(props)}
+      ref={ref}
+      element={element}
+      as="button"
+      type="button"
+      role="option"
+      cursor="pointer"
+      backgroundColor={selected ? 'colorBackgroundPrimaryWeakest' : 'colorBackground'}
```
SiTaggart, line 25: When selected can you also add `aria-selected="true"` to the pill?

packages/paste-core/components/form-pill-group/src/FormPill.tsx
```diff
@@ -0,0 +1,80 @@
...
+import {CompositeItem} from '@twilio-paste/reakit-library';
+import type {CompositeStateReturn} from '@twilio-paste/reakit-library';
+import {PillCloseIcon} from './PillCloseIcon';
+
+interface FormPillStylesProps {
+  selected?: boolean;
+  element?: string;
+  children: React.ReactNode;
+  onKeyDown?: (event: React.KeyboardEvent<HTMLButtonElement>) => void;
+  onClick?: () => void;
```
andipants12, line 12: Could we instead change this to something like one of the following?

`onClick?: HTMLButtonElement['onClick'];`
`onClick?: `MouseEventHandler<HTMLButtonElement>;`
`onClick?: (event: React.ClickEvent<HTMLButtonElement>) => void;`

I ask because in case our consumers want to use the event in their `onClick` callback they pass to this component. As it is typed, I think this would discourage that.

packages/paste-core/components/form-pill-group/src/PillCloseIcon.tsx
```diff
@@ -0,0 +1,22 @@
...
+
+export const PillCloseIcon: React.FC<PillCloseIconProps> = ({onClick}) => {
+  return (
+    <Box
+      onClick={onClick}
+      _hover={{
+        color: 'colorTextLinkDark',
+      }}
+    >
+      <ScreenReaderOnly>. Press delete or backspace to remove</ScreenReaderOnly>
```
SiTaggart, line 18: :chefkissingfingers:

packages/paste-core/components/form-pill-group/src/FormPill.tsx
```diff
@@ -0,0 +1,80 @@
...
+      {props.children}
+    </Box>
+  )
+);
+
+export interface FormPillProps extends CompositeStateReturn {
+  selected?: boolean;
+  element?: string;
+  children: React.ReactNode;
+  onSelect?: () => void;
```
andipants12, line 52: Same concern here, as above on [line 12](https://github.com/twilio-labs/paste/pull/1803/files#diff-f5db32f3b3a094b9a7bd26d634033f83ac1284d421faeebb634fd5f3f90a8502R12). I wonder if we should type these with the event, or if we are not overriding the types from global handlers, maybe type them as such.

packages/paste-core/components/form-pill-group/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,121 @@
...
+      expect(getByText('Tennis')).toBeDefined();
+
+      const group = getByTestId('form-pill-group');
+      expect(group.getAttribute('aria-label')).toBe('Your favorite sports:');
+      expect(group.tagName).toBe('DIV');
+      expect(group.getAttribute('role')).toBe('listbox');
+
+      const pill = getByTestId('form-pill');
+      expect(pill.tagName).toBe('BUTTON');
+      expect(pill.getAttribute('role')).toBe('option');
```
SiTaggart, line 45: Once we have `aria-selected` reflecting the selected state of a pill, it would be nice to have a test that ensures we're selecting pills correctly.

Do you also think it would be worthwhile just having one test that moves focus around using the arrow keys to ensure the keyboard nav continues to work? I know that's quite well tested in Composite, but we might want to test our implementation just in case we mess something up in the future

packages/paste-core/components/form-pill-group/src/FormPill.tsx
TheSisb: This component (and prop) isn't actually exposed to the user. This is used by the `FormPill` in the same file, which passes in the `onSelect` as an `onClick`

packages/paste-core/components/form-pill-group/src/FormPill.tsx
TheSisb: I typed this in this particular way because I don't actually pass any arguments when this is called. The `events` object that is typically passed (what I believe you're referencing), is not provided.

andipants12: Oh so we are not exposing the events argument? Makes sense then. FWIW I find it a little confusing that we are using the event handler names but they're not actually event handlers, just callbacks. Not blocking, just my two cents.

packages/paste-core/components/form-pill-group/__tests__/index.spec.tsx
TheSisb: Added a bunch of tests for this behavior.

SiTaggart (approved):
🎉

andipants12 (approved):

PR #1816, opened 2021-08-28
feat: textarea, date-picker and time-picker customization

SiTaggart:
Add customization to 

- Textarea
- date-picker
- time-picker

Also allow Box to render as another component, so that we can use Box for Textarea when it's using another component underneath to do the height expansion.


packages/paste-core/components/date-picker/__tests__/index.spec.tsx
```diff
@@ -84,6 +85,97 @@ describe('DatePicker', () => {
...
+    const initialProps = {
+      id: 'datepicker',
+      name: 'dp',
+      defaultValue: '2020-03-21',
+      'data-testid': 'datepickerid',
+    };
+    it('should add the element dom attibute', () => {
+      const {container} = render(<DefaultDatePicker {...initialProps} />);
+      expect(container.querySelector('[data-paste-element="DATEPICKER"]')).toBeInTheDocument();
+      expect(screen.getByTestId('datepickerid').getAttribute('data-paste-element')).toEqual('DATEPICKER_ELEMENT');
```
nkrantz, line 98: Why is `data-paste-element` both `DATEPICKER` and `DATEPICKER_ELEMENT`?  I'm confused where the `_ELEMENT` suffix comes from since I can't find where you're adding it in `DatePicker.tsx`

nkrantz (approved):

packages/paste-core/components/date-picker/__tests__/index.spec.tsx
SiTaggart: Yeah, so it's passed down to `Input`, and inside `Input` is[ where it gets split out.](https://github.com/twilio-labs/paste/blob/main/packages/paste-core/components/input/src/Input.tsx#L95) 

By default Input outputs `INPUT` and `INPUT_ELEMENT`, but if you override it with an element name, you get a custom name

packages/paste-core/components/textarea/stories/textarea.stories.tsx
```diff
@@ -418,3 +421,298 @@ export const TextareaDisabledInsertBeforeAndAfterInverse = (): React.ReactNode =
...
+export const CustomizedTextarea: React.FC = () => {
+  const activeTheme = useTheme();
+  const uidTwo = useUID();
+  const uidTwoVariant = useUID();
+  const uidThree = useUID();
+  const uidThreeVariant = useUID();
+  return (
+    <CustomizationProvider
+      baseTheme="default"
+      theme={activeTheme}
```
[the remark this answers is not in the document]
SiTaggart: So when we display these stories for VRT we place them side by side, one in default and one in dark mode. 

The problem is the theme is decided at the storybook level, so the dark mode rendered version is using the default theme because we hard code the base theme to be default.

By grabbing the active theme from context, which is supplied by storybook, we can just provide that theme to the customization provider so each side is using the correct theme.

packages/paste-core/components/textarea/src/TextArea.tsx
```diff
@@ -26,65 +25,57 @@ export interface TextAreaProps extends React.TextareaHTMLAttributes<HTMLTextArea
...
-      '-webkit-text-fill-color': props.variant === 'inverse' ? 'colorTextInverseWeaker' : 'colorTextWeaker',
-      '-webkit-opacity': '1',
-    },
-  })
-);
-/* eslint-enable */
+const TextAreaElement = React.forwardRef<HTMLTextAreaElement, TextAreaProps>((props, ref) => {
+  return (
+    <Box
+      as={TextareaAutosize}
```
TheSisb, line 31: Can this be a `Styled(TextareaAutosize)` private component that we compose instead?

SiTaggart: We need the element `<textarea />` to be a Box so we can apply the custom styles to it from the theme, so I'm not sure if we can do that? I'd still need to make Box as `StyledTextareaAutosize` for it to pick up the styles

TheSisb: I guess I'm not understanding why this wouldn't be the same:

```js
const StyledTextareaAutosize = styled(TextareaAutosize)(css({
      appearance="none"
      backgroundColor="transparent"
      borderWidth="borderWidth0"
      borderRadius="borderRadius20"
      boxShadow="none"
      color="inherit"
      display="block"
      fontFamily="inherit"
      fontSize="fontSize30"
      fontWeight="fontWeightMedium"
      lineHeight="lineHeight20"
      maxHeight="size30"
      outline="none"
      paddingBottom="space30"
      paddingLeft="space40"
      paddingRight="space40"
      paddingTop="space30"
      resize="vertical"
      width="100%"
      _placeholder={{
        color: props.variant === 'inverse' ? 'colorTextInverseWeak' : 'colorTextWeak',
        fontStyle: 'italic',
      }}
      _focus_placeholder={{
        color: props.variant === 'inverse' ? 'colorTextInverseWeak' : 'colorTextWeak',
      }}
      _disabled={{
        color: props.variant === 'inverse' ? 'colorTextInverseWeaker' : 'colorTextWeaker',
        cursor: 'not-allowed',
        // Fixes value color in Safari
        '-webkit-text-fill-color': props.variant === 'inverse' ? 'colorTextInverseWeaker' : 'colorTextWeaker',
        '-webkit-opacity': '1',
      }}
}));


export const TextareaElement = React.forwardRef<HTMLTextAreaElement, TextAreaProps>((props, ref) => {
  return (
    <StyledTextareaAutosize
       async
      rows={3}
      spellCheck
      size={null}
      height={null}
      ref={ref}
   />
)

SiTaggart: OK, we now use the box utility on the style component instead. Smaller change, no Box as other component

TheSisb (approved):

PR #1817, opened 2021-08-30
feat(table): add element customization

andipants12:
## Description
Add `element` to the table package components, to allow for customization.

### Summary of changes
- [x] add `element` to the table API (defaults to name of components, allow string values)
- [x] Create tests that ensure the defined custom element API remains.
- [x] Create tests that ensure each tagged element merges the custom styles.
- [x] Add `CustomTable` story to use in VRT


packages/paste-core/components/table/src/Tr.tsx
```diff
@@ -1,26 +1,30 @@
...
-      _last={{borderWidth: 'borderWidth0'}}
-    />
-  );
-});
+const Tr = React.forwardRef<HTMLTableRowElement, TrProps>(
+  ({element = 'TR', verticalAlign = 'middle', ...props}, ref) => {
+    const {striped} = React.useContext(TableContext);
+    return (
+      <Box
+        {...striped}
```
andipants12, line 12: Technically this is a boolean, so spreading it is jus spreading an empty object. Probably should remove this at some point, but leaving since it's out of scope.

packages/paste-core/components/table/src/types.ts
```diff
@@ -1,6 +1,8 @@
 import * as PropTypes from 'prop-types';
-import {ValueOf} from '@twilio-paste/types';
-import {WidthOptions, isWidthTokenProp} from '@twilio-paste/style-props';
+import type {BoxProps} from '@twilio-paste/box';
```
andipants12, line 2: IMHO the `PropType` declarations should be removed since they're not TS types, mixing in the const declarations and the type declarations made this file super confusing to read.

gloriliale: I agree, I was confused by this as well. I'm of the mind that each of the PropTypes should be in their respective files instead of this one. We do this in most cases already, and I don't think we're gaining anything from abstracting them to this file, since as Andi said, they're not TS types.

gloriliale (approved):
Lookin' good! I agree with you about the PropTypes placement and left a comment about that. Other than that, solid customization 😎

packages/paste-core/components/table/src/types.ts
SiTaggart: For larger components I don't mind them being in a file separate to the component. So for convention I don't mind it on smaller ones too.

I don't like that some time we put prop types into types. I want separate files for both.

Ideal folder strcuture:

```
src/ 
  - index.tsx
  - types.ts
  - proptypes.ts
  - Component.tsx
  - SubComponent.tsx
```

SiTaggart: I don't mind you pulling them out here in this PR if you want to. It's a small change for this component

packages/paste-core/components/table/stories/index.stories.tsx
```diff
@@ -1006,3 +1007,74 @@ export const Truncation = (): React.ReactNode => {
...
+    elements={{
+      TABLE: {fontFamily: 'fontFamilyCode'},
+      TH_HEAD_LARGE: {fontSize: 'fontSize50'},
+      TH: {fontWeight: 'fontWeightBold'},
+      TH_BODY: {fontWeight: 'fontWeightMedium'},
+      TH_FOOTER: {textDecoration: 'underline'},
+      TR_DESTRUCTIVE: {color: 'colorTextLinkDestructive', fontWeight: 'fontWeightBold'},
+      TR_NEUTRAL: {color: 'colorTextNeutral', fontFamily: 'fontFamilyText'},
+      TR: {color: 'colorTextWeak'},
+      TD_FOOT_HIGHLIGHT: {
```
SiTaggart, line 1023: I see what you did here, you mixed default element names with custom element names. Neat condensed story

packages/paste-core/components/table/src/types.ts
andipants12: Okey cool, I'll pull it out into it's own folder. ⛏️

packages/paste-core/components/table/stories/index.stories.tsx
andipants12: I was hoping folks would leave feedback on this. Thanks, I figured it's a good starting point for this customization story, since there are so many components that go into composing a table. Figured it's maybe more robust to combine them.

packages/paste-core/components/table/stories/index.stories.tsx
```diff
@@ -1006,3 +1007,74 @@ export const Truncation = (): React.ReactNode => {
     </Table>
   );
 };
+
+export const CustomTable = (): React.ReactNode => (
+  <CustomizationProvider
```
SiTaggart, line 1012: The only thing you might want to consider here, which is something I've started doing, is grabbing the active theme  from context and providing it to the CustomizationProvider.

This way when the story is in side by side for VRT (one side default, one side dark), the theme being rendered is dynamic and matches the side by side.

```
const activeTheme = useTheme();
<CustomizationProvider
  theme={activeTheme}
  ...
```

packages/paste-core/components/table/src/types.ts
```diff
@@ -14,70 +16,84 @@ export interface TableContextProps {
 
 export interface TableProps extends React.TableHTMLAttributes<HTMLTableElement> {
   children: NonNullable<React.ReactNode>;
+  element?: BoxProps['element'];
```
nkrantz, line 23: Is this the same as writing `Pick<BoxProps, 'element'>`?

packages/paste-core/components/table/stories/index.stories.tsx
andipants12: Ohh yeah good point! Thanks,  I'll make the change right now.

packages/paste-core/components/table/src/types.ts
andipants12: Not quite. `Pick<BoxProps, 'element'>` represents an object with only the `element` key picked from `BoxProps`, while `BoxProps['element']` gives you just the type of the key `element` from box props.

andipants12: Technically this could be done by extending the interface again with `Partial<Pick<BoxProps, 'element'>>`, but IMHO this is hard to read and the intellisense is confusing, since it doesn't extract the type, rather it will show the chaining of utility types.

SiTaggart: Kind of, you definitely could have done this and achieved the same thing, as it takes the key, the requirement and value.

```
export interface TableProps extends React.TableHTMLAttributes<HTMLTableElement>, Pick<BoxProps, 'element'> {
  ...
```

`element: BoxProps['element'];` is useful if you want to change the requirement of element, from optional to required

SiTaggart (approved):
Very nice

PR #1818, opened 2021-08-30
feat(toast): allow onDismiss on toast push method

TheSisb:
Allow passing an onDismiss prop to Toasts with the `useToaster` API


packages/paste-core/components/toast/src/useToaster.ts
```diff
@@ -3,21 +3,51 @@ import {uid} from '@twilio-paste/uid-library';
...
+      /**
+       * Clear all timeouts from the toaster when the component unmounts
+       */
+      toasts.forEach((toast) => {
+        if (toast.timeOutId) {
+          window.clearTimeout(toast.timeOutId);
+        }
+      });
+    };
+  }, []);
```
SiTaggart, line 22: Mmmmm so this is the small potential memory leak. If you trash the DOM, navigate away, there can be timeouts still running.

So we clean up on unmount in this effect.

packages/paste-core/components/toast/src/useToaster.ts
```diff
@@ -3,21 +3,51 @@ import {uid} from '@twilio-paste/uid-library';
...
+
   const pop = (id: ToasterToast['id']): void => {
+    if (!isMounted.current) {
+      return;
+    }
```
SiTaggart, line 27: How likely is this to happen? Method called when unmounted?

SiTaggart (approved):

packages/paste-core/components/toast/src/useToaster.ts
TheSisb: Correct

packages/paste-core/components/toast/src/useToaster.ts
TheSisb: It was happening in the tests. I somehow was calling `pop` after the component had been unmounted, I think by holding a reference to it. The added safety doesn't cost much, as this isn't hot code.

SiTaggart: 👍🏼

packages/paste-core/components/toast/stories/index.stories.tsx
```diff
@@ -231,6 +231,9 @@ export const ToasterStory = (): React.ReactNode => {
       variant,
       message: messageText,
       ...(toastTimeout !== '0' && {dismissAfter: Number.parseInt(toastTimeout, 10)}),
+      onDismiss: () => {
+        console.log('dismissed!');
+      },
```
nkrantz, line 236: Would it be helpful to have the story do something visible on the UI rather than console logging, so that it'll be easy to see this happening if we use it as an example on the website?

nkrantz (approved):
Looks good just left a comment about the story

packages/paste-core/components/toast/stories/index.stories.tsx
TheSisb: I don't think it's necessary for this particular prop. It's a pretty niche requirement of Toasts.

PR #1826, opened 2021-08-31
chore(button): use left text alignment for link button variants

andipants12:
**Contributing to Twilio**


packages/paste-core/components/button/src/styles.ts
```diff
@@ -13,6 +13,7 @@ const ResetStyles: BoxStyleProps = {
   transition: 'background-color 100ms ease-in, box-shadow 100ms ease-in, color 100ms ease-in',
   fontFamily: 'fontFamilyText',
   fontWeight: 'fontWeightSemibold',
+  textAlign: 'unset',
```
gloriliale, line 16: Why is it that we want to set it to `unset` specifically, as opposed to inherit, or left?

gloriliale (approved):
Just had a question about the use of `unset`, but all looks good otherwise

packages/paste-core/components/button/src/styles.ts
andipants12: I'm curious what the rest of the team thinks about this, but it's currently using `unset` because it conditionally behaves like `initial` or `inherit`. You can read more in detail on the [MDN page for `unset`](https://developer.mozilla.org/en-US/docs/Web/CSS/unset).

Also, here is a sandbox that demonstrates how this works in a table. Because there is a parent with text alignment defined (e.g. the table), the button text is aligned according to that value: https://developer.mozilla.org/en-US/docs/Web/CSS/unset

gloriliale: Could you link the code part of the sandbox? I only see the result!

andipants12: Opps! Yeah here it is: https://codesandbox.io/s/button-text-alignment-in-tables-r4utl

SiTaggart: We only want left alignded text for link style variants, right?

TheSisb: Yes, only left aligned for links.

.changeset/brown-parents-cross.md
```diff
@@ -0,0 +1,6 @@
+---
+'@twilio-paste/button': patch
+'@twilio-paste/core': patch
+---
+
+[Button]: Unset default text alignment for buttons.
```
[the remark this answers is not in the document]
andipants12: Updated 😸

SiTaggart (approved):

PR #1835, opened 2021-09-01
chore(tokens): update text-icon-color tokens

nkrantz:
- Text tokens are updated as per [this doc](https://docs.google.com/document/d/1-40c2ZzUGdCa8rU5FZ7FHfNBo64PVw2F0kNgnQeQDtY/edit#)
- Background tokens are updated as per the doc
- Update Alert and Toast to use the new icon tokens in the doc

**Contributing to Twilio**


katieporter (approved):

packages/paste-design-tokens/tokens/global/text-color.yml
```diff
@@ -247,23 +247,23 @@ props:
...
+    value: "{!palette-orange-70}"
     comment: Color for warning text.
     uicontrol_contrast_pairing:
       # - color-background
       # - color-background-body
       - color-background-body-inverse
       # - color-background-row-striped
   color-text-warning-strong:
-    value: "{!palette-orange-70}"
+    value: "{!palette-orange-80}"
```
SiTaggart, line 266: With these increased color contrast ratios are there any of these contrast pairings that now work and can be uncommented?

nkrantz: > With these increased color contrast ratios are there any of these contrast pairings that now work and can be uncommented?

Uncommented and tests are passing 👍 

> I'm also assuming the these are truly text color tokens now and should also be having text contrast pairings

Trying to add these in but every text contrast pairing test is failing with the new tokens and I'm not sure why

packages/paste-core/components/alert/__tests__/__snapshots__/index.spec.tsx.snap
```diff
@@ -77,8 +77,8 @@ exports[`Alert Variant error Should render an error alert 1`] = `
       >
         <span
           className="emotion-0"
-          color="colorTextError"
-          data-paste-element="ICON"
+          color="colorTextIconError"
+          data-paste-element="BOX"
```
andipants12, line 81: Did you install and build after rebasing with main? I saw this snapshot change from `ICON` to `BOX` too before I installed the most up to date package versions.

andipants12 (commented):
Left a note to check on the data attribute changes in the snapshots, which I think might be outdated. Other than that, looks good. 😄

packages/paste-core/components/alert/__tests__/__snapshots__/index.spec.tsx.snap
nkrantz: I did not install after rebasing! Will do 👍 

I updated snapshots so those should be good next time I push up

andipants12 (approved):

TheSisb (approved):

PR #1840, opened 2021-09-02
feat(tabs): add element customization

andipants12:
## Description
Add `element` to the tab package components, to allow for customization.

### Summary of changes
- [x] add `element` to the table API (defaults to name of components, allow string values)
- [x] Create tests that ensure the defined custom element API remains.
- [x] Create tests that ensure each tagged element merges the custom styles.
- [x] Add stories with customization for VRT

### DIFF
--- a/packages/paste-core/components/tabs/__tests__/tabs.test.tsx
+++ b/packages/paste-core/components/tabs/__tests__/tabs.test.tsx
@@ -7,10 +7,33 @@ import {CustomizationProvider} from '@twilio-paste/customization';
 import axe from '../../../../../.jest/axe-helper';
 import {HorizontalTabs, StateHookTabs} from '../stories/index.stories';
 import {Tabs, Tab, TabList, TabPanels, TabPanel} from '../src';
+import {getElementName} from '../src/utils';
 
 expect.extend(matchers);
 
 describe('Tabs', () => {
+  describe('Utils', () => {
+    describe('getElementName', () => {
+      const mockFallbackName = 'TEST_ELEMENT_FALLBACK';
+
+      it('should return the correct string when element is undefined', () => {
+        expect(getElementName('horizontal', mockFallbackName)).toEqual('HORIZONTAL_TEST_ELEMENT_FALLBACK');
+        expect(getElementName('vertical', mockFallbackName)).toEqual('VERTICAL_TEST_ELEMENT_FALLBACK');
+      });
+      it('should return the correct string when element is null', () => {
+        expect(getElementName('horizontal', mockFallbackName, null)).toEqual('HORIZONTAL_TEST_ELEMENT_FALLBACK');
+        expect(getElementName('vertical', mockFallbackName, null)).toEqual('VERTICAL_TEST_ELEMENT_FALLBACK');
+      });
+      it('should return the correct string when element is defined', () => {
+        expect(getElementName('horizontal', mockFallbackName, 'CUSTOM_NAME')).toEqual('CUSTOM_NAME');
+        expect(getElementName('vertical', mockFallbackName, 'CUSTOM_NAME')).toEqual('CUSTOM_NAME');
+      });
+      it('should return the correct string when element is an empty string', () => {
+        expect(getElementName('horizontal', mockFallbackName, '')).toEqual('');
+        expect(getElementName('vertical', mockFallbackName, '')).toEqual('');
+      });
+    });
+  });
   describe('Render', () => {
     it('relevant html and aria attributes', () => {
       const [tabOneId, tabTwoId, tabThreeId, panelOneId, panelTwoId, panelThreeId] = [...new Array(6)].map(
--- a/packages/paste-core/components/tabs/__tests__/tabs.test.tsx
+++ b/packages/paste-core/components/tabs/__tests__/tabs.test.tsx
@@ -1,9 +1,14 @@
 import * as React from 'react';
+import {matchers} from 'jest-emotion';
 import {render, screen} from '@testing-library/react';
+import {CustomizationProvider} from '@twilio-paste/customization';
+// eslint-disable-next-line @typescript-eslint/ban-ts-comment
 // @ts-ignore typescript doesn't like js imports
 import axe from '../../../../../.jest/axe-helper';
 import {HorizontalTabs, StateHookTabs} from '../stories/index.stories';
 import {Tabs, Tab, TabList, TabPanels, TabPanel} from '../src';
+
+expect.extend(matchers);
 
 describe('Tabs', () => {
   describe('Render', () => {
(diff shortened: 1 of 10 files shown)


SiTaggart:
I have deep regret that orientation wasn't pulled out as their own components of Tabs and Vertical Tabs. I suspect people will want to design them independently from each other. Pseudo code: 

```
<div data-paste-element="HORIZONTAL_TABS">
  <div data-paste-element="HORIZONTAL_TAB_LIST">
    <div data-paste-element="HORIZONTAL_TAB">
      ...
    </div>
    <div data-paste-element="HORIZONTAL_TAB">
      ...
    </div>
  </div>
  <div data-paste-element="HORIZONTAL_TAB_PANELS">
    <div data-paste-element="HORIZONTAL_TAB_PANEL">
      ...
    </div>
    <div data-paste-element="HORIZONTAL_TAB_PANEL">
      ...
    </div>
  </div>
</div>
```

```
<div data-paste-element="VERTICAL_TABS">
  <div data-paste-element="VERTICAL_TAB_LIST">
    <div data-paste-element="VERTICAL_TAB">
      ...
    </div>
    <div data-paste-element="VERTICAL_TAB">
      ...
    </div>
  </div>
  <div data-paste-element="VERTICAL_TAB_PANELS">
    <div data-paste-element="VERTICAL_TAB_PANEL">
      ...
    </div>
    <div data-paste-element="VERTICAL_TAB_PANEL">
      ...
    </div>
  </div>
</div>
```

2 cents  🪙

packages/paste-core/components/tabs/src/Tabs.tsx
```diff
@@ -1,42 +1,55 @@
...
+    return <Box element={`HORIZONTAL_${element}`}>{returnValue}</Box>;
   }
 );
 
 if (process.env.NODE_ENV === 'development') {
   Tabs.propTypes = {
+    element: PropTypes.string,
     selectedId: PropTypes.string,
-    orientation: PropTypes.oneOf(['horizontal', 'vertical', null]),
+    orientation: PropTypes.oneOf(['horizontal', 'vertical', undefined]),
```
andipants12, line 52: AFAICT `orientation` should never be `null`, but it can be undefined on the `Tabs` component API.

packages/paste-core/components/tabs/src/TabsContext.tsx
```diff
@@ -1,11 +1,12 @@
 import * as React from 'react';
-import {TabPrimitiveState} from '@twilio-paste/tabs-primitive';
-import {Variants} from './types';
+import type {TabPrimitiveState} from '@twilio-paste/tabs-primitive';
+import type {Variants} from './types';
 
-interface TabState extends TabPrimitiveState {
+interface TabState extends Omit<TabPrimitiveState, 'orientation'> {
   variant?: Variants;
+  orientation: 'horizontal' | 'vertical';
```
andipants12, line 7: Just using the type value of `Orientation` since that specific type is not exported from reakit this is the simplest way to type this.

packages/paste-core/components/tabs/__tests__/tabs.test.tsx
```diff
@@ -92,6 +97,317 @@ describe('Tabs', () => {
...
+            <TabPanel element="PUG" data-testid="tab-panel-3">
+              Tab 3
+            </TabPanel>
+          </TabPanels>
+        </Tabs>
+      );
+
+      const outerDiv = screen.getByTestId('tab-list').parentElement as HTMLElement;
+
+      expect(outerDiv.getAttribute('data-paste-element')).toEqual('HORIZONTAL_HORSE');
```
TheSisb, line 195: I don't think we should prefix the direction for custom elements because it breaks the convention of element overwriting defaults and may come as a frustrating surprise to developers. Curious about your rationale though

packages/paste-core/components/tabs/src/TabList.tsx
```diff
@@ -1,35 +1,40 @@
...
 }
 
-const HorizontalTabList: React.FC = ({children}) => (
+const HorizontalTabList: React.FC<{element?: BoxProps['element']}> = ({children, element}) => (
   <Box
     display="flex"
     borderBottomWidth="borderWidth10"
     borderBottomColor="colorBorderWeak"
     borderBottomStyle="solid"
+    element={`HORIZONTAL_${element}`}
```
TheSisb, line 25: I now understand why HORIZONTAL/VERTICAL are prefixing customized element props in the tests. In this case I would do something like `element={element != null ? element : "HORIZONTAL_TAB_LIST"}`

packages/paste-core/components/tabs/src/TabPanels.tsx
```diff
@@ -1,17 +1,31 @@
...
-    <Box {...safelySpreadBoxProps(props)} width="100%" ref={ref}>
-      {children}
-    </Box>
-  );
-});
+const TabPanels = React.forwardRef<HTMLDivElement, TabPanelsProps>(
+  ({children, element = 'TAB_PANELS', ...props}, ref) => {
+    const {orientation} = React.useContext(TabsContext);
+    return (
+      <Box {...safelySpreadBoxProps(props)} element={`${orientation.toUpperCase()}_${element}`} width="100%" ref={ref}>
```
TheSisb, line 17: Maybe a shared package util function to help with these element names:

```js
function getElementName(orientation: 'horizontal' | 'vertical', element?: string, fallback: string) {
  if (element == null) {
   return `${orientation.toUpperCase()}_${fallback}`;
  }
  return element;
}
```

And this code would be 
```js
element={getElementName(orientation, element, 'TAB_PANELS')}
```

packages/paste-core/components/tabs/src/Tabs.tsx
```diff
@@ -1,42 +1,55 @@
...
-    const tab = state || useTabPrimitiveState({orientation, ...initialState});
-    const value = React.useMemo(() => ({...tab, variant}), [...Object.values(tab), variant]);
+  ({children, element = 'TABS', orientation = 'horizontal', state, variant, ...initialState}, ref) => {
+    // If returned state from primitive has orientation set to undefined, use the default "horizontal"
+    const {orientation: tabOrientation = orientation, ...tab} =
+      state || useTabPrimitiveState({orientation, ...initialState});
+
+    const value = React.useMemo(() => ({...tab, orientation: tabOrientation, variant}), [
+      ...Object.values(tab),
+      orientation,
```
TheSisb, line 31: should this be `tabOrientation`?

andipants12: Yes, thank you! I'm kinda surprised I didn't get a lint error here 🤔

packages/paste-core/components/tabs/src/TabPanels.tsx
andipants12: Yeah that sounds like a good idea. Is that something that we should provide as a named export, or have it be "internal" to the `/tabs`?

packages/paste-core/components/tabs/src/utils.ts
```diff
@@ -0,0 +1,5 @@
+export const getElementName = (
```
andipants12, line 1: @TheSisb I am going to add one or two tests for this now, but please let me know what you think! For now this is not available as a named export from the package.

packages/paste-core/components/tabs/__tests__/tabs.test.tsx
```diff
@@ -1,11 +1,39 @@
...
+      });
+      it('should return the correct string when element is null', () => {
+        expect(getElementName('horizontal', mockFallbackName, null)).toEqual('HORIZONTAL_TEST_ELEMENT_FALLBACK');
+        expect(getElementName('vertical', mockFallbackName, null)).toEqual('VERTICAL_TEST_ELEMENT_FALLBACK');
+      });
+      it('should return the correct string when element is defined', () => {
+        expect(getElementName('horizontal', mockFallbackName, 'CUSTOM_NAME')).toEqual('CUSTOM_NAME');
+        expect(getElementName('vertical', mockFallbackName, 'CUSTOM_NAME')).toEqual('CUSTOM_NAME');
+      });
+      it('should return the correct string when element is an empty string', () => {
```
nkrantz, line 31: Why do we want it to return an empty string when that's passed in? What would be the use case for that?

nkrantz (approved):
Looks great, really thorough testing. Left a question about your util func but not a blocker

packages/paste-core/components/tabs/src/TabPanels.tsx
TheSisb: Internal to tabs is fine

TheSisb (approved):

packages/paste-core/components/tabs/__tests__/tabs.test.tsx
andipants12: The `element` is typed to be a `string` which can include empty strings. I can't think of great example for a use case off the top of my head, but I think returning an empty string, if that is provided as the `element` prop is appropriate given the typing.

This is the case as well for the other components that have been updated for customization. 
In those instances, we are using default values assigned where the props are destructured. For example:

```
const Component = ({ element =  'EXAMPLE', children }) => (<Box element={element}>{children}</Box>)
```

Note that the default value is only used if element is implicitly undefined (e.g. the prop `element` is not set) or explicitly undefined (e.g. `element: undefined`). Even though both undefined and `''` are technically falsy values, if we pass in an empty string as the value of `element`, the data attribute will be set to an empty string.

I think this could be improved with development-only console warning logs might be useful, to help folks catch this if it is done unintentionally.

PR #1852, opened 2021-09-09
feat(help text): add customization

nkrantz:
- Enable customization for help text
- Add tests for html attribute and customization
- Add customization story

**Contributing to Twilio**

### DIFF
--- a/packages/paste-core/components/help-text/__tests__/helptext.test.tsx
+++ b/packages/paste-core/components/help-text/__tests__/helptext.test.tsx
@@ -1,6 +1,11 @@
 import * as React from 'react';
 import {shallow} from 'enzyme';
+import {CustomizationProvider} from '@twilio-paste/customization';
+import {render, screen} from '@testing-library/react';
+import {matchers} from 'jest-emotion';
 import {HelpText} from '../src';
+
+expect.extend(matchers);
 
 describe('HelpText variant prop', () => {
   const container = shallow(<HelpText variant="error" />);
@@ -10,7 +15,7 @@ describe('HelpText variant prop', () => {
   });
 
   it('should have colorTextError', () => {
-    expect(container.find('Text').prop('color')).toEqual('colorTextError');
+    expect(container.prop('color')).toEqual('colorTextError');
   });
 });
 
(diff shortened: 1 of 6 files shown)


zahnster (approved):
Looks great! 🚢

packages/paste-core/components/help-text/__tests__/helptext.test.tsx
```diff
@@ -21,3 +26,83 @@ describe('HelpText marginTop prop', () => {
     expect(container.find('Flex').prop('marginTop')).toEqual('space0');
   });
 });
+
+describe('HelpText HTML attributes', () => {
+  it('should set element data attribute for flex wrapper and help text', (): void => {
+    const {container} = render(<HelpText data-testid="help_text">This is help text.</HelpText>);
+    expect(container.querySelector('[data-paste-element="HELP_TEXT_WRAPPER"]')).toBeInTheDocument();
```
andipants12, line 33: This might be a nit pick, but I wonder if it is more useful to check not only for the existence for the correct `paste-element` data attributes, but also that they are on the correct elements. That would mean that we should find the target nodes in the rendered tree, and confirm that the attribute is on the correct node.
The extra constraint for the test potentially gives us more security against future regressions.

gloriliale: @andipants12 would the test on line 35 (`expect(screen.getByTestId('help_text').getAttribute('data-paste-element')).toEqual('HELP_TEXT');` be more secure in the way you're describing? I've also refactored this component so that it only needs one `element` attribute, so this Wrapper-specific one will be soon deleted.

packages/paste-core/components/help-text/src/HelpText.tsx
```diff
@@ -1,10 +1,9 @@
 import * as React from 'react';
 import * as PropTypes from 'prop-types';
-import {ValueOf} from '@twilio-paste/types';
+import type {ValueOf} from '@twilio-paste/types';
 import {Box} from '@twilio-paste/box';
-import {Flex} from '@twilio-paste/flex';
-import {Text, safelySpreadTextProps} from '@twilio-paste/text';
-import {TextColor} from '@twilio-paste/style-props';
+import type {TextProps} from '@twilio-paste/text';
```
SiTaggart, line 5: We're not using `Text` anymore are we? Can this be `BoxProps` take from `box`?

We might even then be able to remove the `@twilio-paste/text` peer dependency if it's not used anymore?

gloriliale: True! I just changed replaced `TextProps` references to `BoxProps`, and removed dependencies to Text and Flex (since we removed both)

SiTaggart (approved):

packages/paste-core/components/help-text/src/HelpText.tsx
```diff
@@ -17,63 +16,70 @@ export const HelpTextVariants = {
...
-  let textColor = 'colorTextWeak' as TextColor;
-  if (variant === 'error') {
-    textColor = 'colorTextError';
-  } else if (variant === 'error_inverse') {
-    textColor = 'colorTextErrorWeak';
-  } else if (variant === 'inverse') {
-    textColor = 'colorTextInverseWeak';
-  }
+    let textColor = 'colorTextWeak' as TextColor;
+    if (variant === 'error') {
```
TheSisb, line 47: On line 28 we use constants to check the variant, but here we're using strings. Can we use the constants again?

SiTaggart: Yeah, completely refactored this to be a little safer

packages/paste-core/components/help-text/__tests__/helptext.test.tsx
```diff
@@ -10,14 +15,84 @@ describe('HelpText variant prop', () => {
...
+    expect(screen.getByTestId('help_text').getAttribute('data-paste-element')).toEqual('foo');
+  });
+});
+
+describe('Customization', () => {
+  it('should customize help text default and error variant', (): void => {
+    render(
+      <CustomizationProvider
+        baseTheme="default"
+        elements={{
```
TheSisb, line 55: Now that I merged the global constant `TestTheme` we should set `theme={TestTheme` here (with a `@ts-ignore`) so that this test doesn't break on token changes.  Same for the other test below on 78

TheSisb (approved):

packages/paste-core/components/help-text/__tests__/helptext.test.tsx
SiTaggart: Oh yeah, for sure

nkrantz: Updated

PR #1853, opened 2021-09-10
test: add global 'TestTheme' jest variable

TheSisb:
This PR adds a global variable to the jest VM called `TestTheme` which we can use when testing CustomizationProvider without concerns for all our tests breaking when we change a token.

The TestTheme object is just a copy of our current defaultTheme json, to make migration easy.


.jest/globals/TestTheme.js
```diff
@@ -0,0 +1,595 @@
+module.exports = {
+  shadows: {
```
TheSisb, line 2: Theme copied from old theme designer demo made by @serifluous 
https://twilio.slack.com/archives/GGQD89JRL/p1607718136194300

nkrantz (approved):
Looks good, just wondering, is the `baseTheme` prop still necessary if you're providing an entire `theme`?

Thinking there might be a couple of customization components (help text, disclosure) that _may_ get merged in before this one so I'm just making a note here to remember to go back and check next week (I can do this).

jest.config.js
```diff
@@ -1,4 +1,9 @@
+const TestTheme = require('./.jest/globals/TestTheme');
+
 module.exports = {
+  globals: {
```
andipants12, line 4: 👍

packages/paste-core/components/date-picker/__tests__/index.spec.tsx
```diff
@@ -107,6 +107,8 @@ describe('DatePicker', () => {
       const {container} = render(
         <CustomizationProvider
           baseTheme="default"
+          // @ts-expect-error global test variable
```
andipants12, line 110: 💡 We could prevent this error from being reported by adding an interface for `TestTheme` on our globals object.

andipants12, line 110: 💡 We could prevent this error from being reported by adding an interface for `TestTheme` on our globals object.

andipants12 (approved):
Left a suggestion about extending our globals interface to include the `TestTheme` so we don't have to manually declare the TS error expected. Not blocking though. 🚀

SiTaggart (approved):

TheSisb:
> Looks good, just wondering, is the baseTheme prop still necessary if you're providing an entire theme?

Looks like it's always been optional and will default to the `default` theme, but it's not going to change anything if I don't mention it because CustomizationProvider always includes a base theme in its current implementation.

>  We could prevent this error from being reported by adding an interface for TestTheme on our globals object.

Couldn't figure this out, gonna ship for now and we can fast-follow later.

PR #1858, opened 2021-09-10
feat(spinner): add new animated spinner

andipants12:
Our existing Spinner is looking quite dated, so this PR refreshes the appearance of it.  There are no breaking changes, so Paste consumers will be happy to see a refreshed Spinner in their apps without any effort.


.changeset/light-beans-speak.md
```diff
@@ -0,0 +1,6 @@
+---
+'@twilio-paste/spinner': major
+'@twilio-paste/core': minor
+---
+
+[Spinner]: Add new spinner component (TODO update this.)
```
gloriliale, line 6: Maybe we could say something like "Update visual design of the spinner component"? And if we're not introducing a breaking change to the API the consumers use, would this be a major change, or just a minor?

andipants12: Sure thing, this is just a draft for now, hence the "TODO" comment.

packages/paste-core/components/spinner/src/keyframes.tsx
```diff
@@ -0,0 +1,50 @@
+import {keyframes} from '@twilio-paste/styling-library';
+
+// const originalKeyframes = `
```
andipants12, line 3: FYI these are the original keyframes. Check out [this codesandbox](https://codesandbox.io/s/keyframe-exploration-4r6yj) to see the visual difference.

packages/paste-core/components/spinner/src/index.tsx
```diff
@@ -1,70 +1,50 @@
...
+  r: 45,
+};
 
-export interface SpinnerProps extends LoadingIconProps {
+export interface SpinnerProps extends IconWrapperProps {
   title: string;
   size?: IconSize;
   delay?: number;
+  decorative: boolean;
+  color?: TextColor;
```
TheSisb, line 23: color and size are already defined in IconWrapperProps iirc

packages/paste-core/components/spinner/src/index.tsx
```diff
@@ -1,70 +1,50 @@
...
 }
 
-const Spinner = React.forwardRef<HTMLDivElement, SpinnerProps>(
-  ({as, size, color, decorative, title, delay = 250}, ref) => {
+export const Spinner = React.forwardRef<HTMLDivElement, SpinnerProps>(
+  ({size = 'sizeIcon30', color = 'currentColor', title, as, display, decorative, delay = 250}, ref) => {
+    const titleId = `spinner-${useUID()}`;
+    const {borderWidths: {borderWidth40 = '8px'} = {}} = useTheme() || {};
```
TheSisb, line 29: What's the concern around `useTheme()` returning null? I think we should assume it works and throw if it doesn't here.

packages/paste-core/components/spinner/src/styled.tsx
```diff
@@ -0,0 +1,45 @@
...
+  };
+});
+
+export const AnimatedStyledCircle = styled.circle<{show: boolean}>(({show}) => {
+  return {
+    ...circleStyleProps,
+    animation: `1.4s linear infinite both ${CircleKeyframes}`,
+    strokeDasharray: '285',
+    stroke: 'currentColor',
+    '@media screen and (prefers-reduced-motion: reduce)': {
```
TheSisb, line 30: I don't think we need this, since Paste's global styles does it for us now. You can test it by setting prefers-reduced-motion to on in your macbook accessibility settings.

packages/paste-core/components/spinner/src/styled.tsx
```diff
@@ -0,0 +1,45 @@
...
+    opacity: show ? 1 : 0,
+  };
+});
+
+export const StyledSvg = styled.svg(() => ({
+  height: '100%',
+  width: '100%',
+  display: 'block',
+  animation: `6s ease-in-out infinite both ${SvgKeyframes}`,
+  '@media screen and (prefers-reduced-motion: reduce)': {
```
TheSisb, line 42: This too

packages/paste-core/components/spinner/src/index.tsx
andipants12: Today when I was running the unit tests, I noticed that if the theme provider is not used, `useTheme()` returns `undefined`. I don't know if we have consumers using our spinner, but not using theme, but I think it is possible.

packages/paste-core/components/spinner/src/styled.tsx
andipants12: Ah ok cool! Wasn't sure about that. Thanks, will remove.

.changeset/light-beans-speak.md
andipants12: IMHO since it's a significantly net new component, `major` feels right here, even though the API is the same.

SiTaggart: I would like to reserve Majors for breaking changes, things that mean the consumer needs to do stuff, just so it's a really clear message on what a major means to a consumer

andipants12: Ok, I'll change to a minor. Thanks folks.

loreina (commented):
This spinner feels slower than the one in this [codesandbox](https://2z3lg.sse.codesandbox.io/spin). Maybe we can pair on this when you're back? It might just be playing around with some keyframes values.

packages/paste-core/components/spinner/src/index.tsx
TheSisb: Pretty sure none of our components work without Theme.Provider visually. The tests that don't wrap it are so that the CSS values aren't modified from `space20` to the corresponding rem value, to make testing consistent even when tokens change.

andipants12: Totally. It's just that without the default values this causes the tests to throw an error.

packages/paste-core/components/spinner/src/constants.ts
```diff
@@ -0,0 +1,11 @@
+type CircleGeometryProps = Pick<React.SVGProps<SVGCircleElement>, 'cx' | 'cy' | 'r'>;
+
+const circleRadius = 45;
+export const circleGeometry: CircleGeometryProps = {
+  cx: 50,
+  cy: 50,
+  r: circleRadius,
+};
+export const circleCircumference = Math.floor(Math.PI * 2 * circleRadius);
```
TheSisb, line 9: Can we add some comments here so that future maintainers know how to reason through this code please?

packages/paste-core/components/spinner/src/styled.tsx
```diff
@@ -0,0 +1,27 @@
...
+    transformOrigin: 'center',
+    opacity: 0.25,
+  };
+});
+
+export const AnimatedStyledCircle = styled.circle<{show: boolean}>(({show}) => {
+  return {
+    transformOrigin: 'center',
+    animation: `${1.5}s ease-in-out infinite both ${CircleKeyframes}`,
+    strokeDasharray: circleCircumference + 1,
```
TheSisb, line 17: Can we leave a comment about why we need the `+1`?  We also have the `strokeDashArray` export in the constant file you can use

packages/paste-core/components/spinner/src/index.tsx
andipants12: Are we confident that folks using our components are definitely using the theme? I'm hesitant to remove as I am concerned about breaking our downstream consumers' apps. If we are super confident it won't cause any issues  for consumers I can remove it.

TheSisb: 100% confident. Everything breaks without it, not just this component.

packages/paste-core/components/spinner/src/styled.tsx
```diff
@@ -0,0 +1,27 @@
+import {styled} from '@twilio-paste/styling-library';
+
+import {SvgKeyframes, CircleKeyframes} from './keyframes';
+import {strokeDasharray} from './constants';
+
+export const StyledCircleTrack = styled.circle(() => {
```
TheSisb, line 6: nit: since you don't need any props, this can be simplified to
```
styled.circle({
  transformOrigin: 'center',
  opacity: 0.25,
})
```
same on line 22 (reduces by 1 function call each)

packages/paste-core/components/spinner/src/constants.ts
```diff
@@ -0,0 +1,21 @@
...
+ */
+export const circleCircumference = Math.PI * 2 * circleRadius;
+
+/* We set the stroke dash array value to describe how to draw the stroke:
+ *  If the stroke dash array is set to "1" the stroke will be drawn as discrete dashes whose length is equal to the length of the stroke segments.
+ *  By setting the stroke dash array to be the circle circumference, we are drawing a stroke that looks continuous but is actually discrete (e.g. countable).
+ *  We can then use the offset to "hide" parts of the total stroke, calculated as a percentage of the circle circumference.
+ *  We can think of the stroke then as being measured in the same "units" that describe the circle's geometry, which makes styling the SVG simpler.
+ */
+export const strokeDasharray = circleCircumference;
```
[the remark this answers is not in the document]
andipants12: Yeahhhh we can 😅 Based on the previous feedback I was trying to be super explicit and show that the circumference is the stroke dash array.

TheSisb (approved):

loreina (approved):
approved on chromatic! this looks amazing, great job @andipants12 🎉

packages/paste-core/components/spinner/src/index.tsx
```diff
@@ -1,70 +1,42 @@
...
   title: string;
-  size?: IconSize;
   delay?: number;
+  decorative: boolean;
 }
 
-const Spinner = React.forwardRef<HTMLDivElement, SpinnerProps>(
-  ({as, size, color, decorative, title, delay = 250}, ref) => {
+export const Spinner = React.forwardRef<HTMLDivElement, SpinnerProps>(
+  ({size, color = 'currentColor', title, as, display, decorative, delay = 250}, ref) => {
```
gloriliale, line 17: Just curious- the previous version of spinner didn't have a `display` prop. What's the benefit from having it now?

SiTaggart: Technically this is a correction. We were extending the interface of LoadingIconProps, which does allow display, but not actually passing display down to the spinner anywhere. So typescript allowed it but it didn't do anything

PR #1860, opened 2021-09-13
feat(disclosure): add element customization

gloriliale:
Added customization capability to the disclosure package. Lingering question of why testing for backgroundColor doesn't seem to be working, which is commented out of the test file.

### DIFF
--- a/packages/paste-core/components/disclosure/__tests__/disclosure.test.tsx
+++ b/packages/paste-core/components/disclosure/__tests__/disclosure.test.tsx
@@ -232,6 +232,8 @@ describe('Disclosure', () => {
       render(
         <CustomizationProvider
           baseTheme="default"
+          // @ts-expect-error global test variable
+          theme={TestTheme}
           elements={{
             DISCLOSURE: {padding: 'space100'},
             DISCLOSURE_CONTENT: {color: 'colorTextErrorStrong'},
@@ -256,6 +258,8 @@ describe('Disclosure', () => {
       render(
         <CustomizationProvider
           baseTheme="default"
+          // @ts-expect-error global test variable
+          theme={TestTheme}
           elements={{
             MY_DISCLOSURE: {padding: 'space100'},
             MY_DISCLOSURE_CONTENT: {color: 'colorTextErrorStrong'},
--- a/packages/paste-core/components/disclosure/__tests__/disclosure.test.tsx
+++ b/packages/paste-core/components/disclosure/__tests__/disclosure.test.tsx
@@ -1,11 +1,15 @@
 import * as React from 'react';
 import {render, screen, fireEvent} from '@testing-library/react';
 import {Theme} from '@twilio-paste/theme';
+import {CustomizationProvider} from '@twilio-paste/customization';
+import {matchers} from 'jest-emotion';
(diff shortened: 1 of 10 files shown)


packages/paste-core/components/disclosure/__tests__/disclosure.test.tsx
```diff
@@ -164,6 +195,87 @@ describe('Disclosure', () => {
...
+
+    it('should set a custom element data attribute for custom named Disclosure components', () => {
+      render(
+        <Theme.Provider theme="default">
+          <MockCustomElementDisclosure />
+        </Theme.Provider>
+      );
+
+      const renderedDisclosureHeading = screen.getByTestId('disclosure-heading');
+      const renderedDisclosure = renderedDisclosureHeading.parentElement.parentElement;
```
TheSisb, line 223: It may be better to give this a testid

TheSisb (approved):

packages/paste-core/components/disclosure/__tests__/disclosure.test.tsx
gloriliale: I tried to, put we don't spread BoxProps on the Disclosure prop, so it doesn't show up. I should've asked if that's on purpose or not actually. I assumed that it was for a specific reason, and this is my workaround for grabbing Disclosure. If not, I could just do safelySpreadBoxProps onto Disclosure and then use testId.

nkrantz (approved):
Lookin goooood! I think we have been adding 3 examples to the customization stories, one normal, one customized, and one customized with a custom `element`. Not sure if that's necessary, but just wanted to call it out in case.

Also, I also have this written down on my list, but just want to make a note that once the Test Theme is merged in, we'll have to go back and update the Disclosure and Help Text tests to use it.

packages/paste-core/components/disclosure/__tests__/disclosure.test.tsx
gloriliale: Talked with Simon about it and implemented the the latter option (use safelySpreadBoxProps on Disclosure and getByTestId in tests). Thanks for pushing on this, @TheSisb!

packages/paste-core/components/disclosure/src/DisclosureHeading.tsx
```diff
@@ -56,6 +60,7 @@ const StyledDisclosureHeading = React.forwardRef<HTMLDivElement, StyledDisclosur
         >
           <Box
             as="span"
+            element={`${element}_ICON`}
```
SiTaggart, line 63: There's a story for icon (which is weird if you had to add this @TheSisb, was it actually working?) but no test to make sure we don't accidentally remove it.

Can we add a test for DISCLOSURE_HEADING_ICON?

SiTaggart (changes requested):

packages/paste-core/components/disclosure/src/DisclosureHeading.tsx
TheSisb: I just added the story for icon, there wasn't one before.
I'll add a test.

SiTaggart (approved):

PR #1878, opened 2021-09-21
feat(label): add element customization

nkrantz:
- Refactor Label to use 1 Box with all style rules
- Add ability to customize Label and Label Required Dot
- Create stories for VRT
- Write tests

**Contributing to Twilio**


packages/paste-core/components/label/src/Label.tsx
```diff
@@ -64,26 +69,21 @@ const Label = React.forwardRef<HTMLLabelElement, LabelProps>(
...
+        lineHeight="lineHeight30"
+        color={textColor}
+        cursor={disabled ? 'not-allowed' : 'pointer'}
         ref={ref}
       >
         <MediaObject verticalAlign="top">
           {required && (
             <MediaFigure spacing="space20">
-              <RequiredDot />
+              <RequiredDot element={element} />
```
SiTaggart, line 83: I'm not sure I know why I prefer this, but I think I would treat Required Dot as a stand alone component that deals with setting it's own element name.

Then here, because it's a composed use case, I might provide a composed element name that overrides the default. `element={`${element}_REQUIRED_DOT`}`.

That way if I were to use Required Dot on it's own, as it is exported so you can, I would get an element name by default.

```
<RequiredDot />
```

would result in `data-paste-element="REQUIRED_DOT"`. Right now I would get `data-paste-element="_REQUIRED_DOT"`, because element is not provided.

packages/paste-core/components/label/__tests__/label.test.tsx
```diff
@@ -28,3 +32,39 @@ describe('Label required prop', () => {
     expect(screen.getByText('Required:')).not.toBeNull();
   });
 });
+
+describe('Customization', () => {
```
TheSisb, line 36: Very clean 👍

TheSisb (approved):

andipants12 (approved):

PR #1880, opened 2021-09-21
feat(pagination): add element customization

TheSisb:
Adds element customization to pagination, including stories and tests.  Uses stories in the tests and applies minor code innovations to clean up large files.


packages/paste-core/components/pagination/stories/customization.stories.tsx
```diff
@@ -0,0 +1,157 @@
...
+          <NumbersAndLabel />
+        </CustomizationProvider>
+      }
+    />
+  );
+};
+
+// eslint-disable-next-line import/no-default-export
+export default {
+  title: 'Components/Pagination/Customization',
```
TheSisb, line 155: This creates a subfolder under Pagination for Customization stories.

packages/paste-core/components/pagination/__tests__/customization.spec.tsx
```diff
@@ -0,0 +1,211 @@
+import * as React from 'react';
```
TheSisb, line 1: I split customization tests into its own file because the main test file is huge.

packages/paste-core/components/pagination/__tests__/customization.spec.tsx
```diff
@@ -0,0 +1,212 @@
...
+    const {getByTestId} = render(<WrapperAndArrows />);
+    const pagination = getByTestId('pagination');
+    const paginationItems = getByTestId('pagination-items');
+    const paginationBackArrow = getByTestId('pagination-back-arrow');
+
+    expect(pagination.getAttribute('data-paste-element')).toEqual('PAGINATION');
+
+    expect(paginationItems.getAttribute('data-paste-element')).toEqual('PAGINATION_ITEMS');
+    expect((paginationItems.firstChild as HTMLElement).getAttribute('data-paste-element')).toEqual(
+      'PAGINATION_ITEMS_ITEM'
```
SiTaggart, line 61: How do we feel about this items_item name? `PAGINATION_ITEMS_ITEM`.

I think that's the only thing that stands out to be, feels a bit cumbersome. Would you consider `PAGINATION_LIST` and `PAGINATION_LIST_ITEM`?

SiTaggart (approved):

packages/paste-core/components/pagination/__tests__/customization.spec.tsx
TheSisb: I went back and forth on this. My hunch is that it's easier to reason about element names if they match the component name, so I went with this approach rather than changing it up.

andipants12 (approved):

nkrantz (approved):

PR #1886, opened 2021-09-23
feat(select):  add element customization

andipants12:
**Contributing to Twilio**

### DIFF
--- a/.changeset/empty-spiders-rest.md
+++ b/.changeset/empty-spiders-rest.md
@@ -3,4 +3,8 @@
 '@twilio-paste/core': patch
 ---
 
+<<<<<<< HEAD
 [InputBox] Remove duplicated prefix in chevron icon wrapper element name, and update name to better reflect the component's purpose.
+=======
+[InputBox] Remove duplicative prefix in chevron icon wrapper element name, and update name too better reflect the component's purpose.
+>>>>>>> 44ecb168... chore(input-box): remove duplicated suffix in element string
--- a/.changeset/empty-spiders-rest.md
+++ b/.changeset/empty-spiders-rest.md
@@ -3,8 +3,4 @@
 '@twilio-paste/core': patch
 ---
 
-<<<<<<< HEAD
-[InputBox] Remove duplicated prefix in chevron icon wrapper element name, and update name to better reflect the component's purpose.
-=======
 [InputBox] Remove duplicative prefix in chevron icon wrapper element name, and update name too better reflect the component's purpose.
->>>>>>> 44ecb168... chore(input-box): remove duplicated suffix in element string
--- a/.changeset/empty-spiders-rest.md
+++ b/.changeset/empty-spiders-rest.md
@@ -1,0 +1,6 @@
+---
+'@twilio-paste/input-box': patch
+'@twilio-paste/core': patch
+---
+
+[InputBox] Remove duplicated prefix in chevron icon wrapper element name, and update name to better reflect the component's purpose.
--- a/.changeset/five-lobsters-juggle.md
+++ b/.changeset/five-lobsters-juggle.md
@@ -1,0 +1,6 @@
+---
+'@twilio-paste/combobox': patch
+'@twilio-paste/core': patch
+---
+
+[Combobox]: Update customization tests and stories to use more explicit a more explicit element name for the chevron wrapper. No consumer change required.
--- a/packages/paste-core/components/input-box/__tests__/input-chevron-wrapper.test.tsx
+++ b/packages/paste-core/components/input-box/__tests__/input-chevron-wrapper.test.tsx
@@ -8,13 +8,13 @@ expect.extend(matchers);
 
 describe('HTML attributes', () => {
   it('should set a element data attribute for InputChevronWrapper', () => {
-    render(<InputChevronWrapper element="INPUT">input-chevron</InputChevronWrapper>);
+    render(<InputChevronWrapper element="INPUT_CHEVRON_WRAPPER">input-chevron</InputChevronWrapper>);
     expect(screen.getByText('input-chevron').getAttribute('data-paste-element')).toEqual('INPUT_CHEVRON_WRAPPER');
   });
 
   it('should set a custom element data attribute for InputChevronWrapper', () => {
-    render(<InputChevronWrapper element="foo">input-chevron</InputChevronWrapper>);
-    expect(screen.getByText('input-chevron').getAttribute('data-paste-element')).toEqual('foo_CHEVRON_WRAPPER');
+    render(<InputChevronWrapper element="FOO_CHEVRON_WRAPPER">input-chevron</InputChevronWrapper>);
+    expect(screen.getByText('input-chevron').getAttribute('data-paste-element')).toEqual('FOO_CHEVRON_WRAPPER');
   });
 });
 
@@ -27,7 +27,7 @@ describe('Customization', () => {
           INPUT_CHEVRON_WRAPPER: {backgroundColor: 'colorBackground'},
         }}
       >
-        <InputChevronWrapper element="INPUT">input-chevron</InputChevronWrapper>
+        <InputChevronWrapper element="INPUT_CHEVRON_WRAPPER">input-chevron</InputChevronWrapper>
       </CustomizationProvider>
     );
     const renderedInputChevronWrapper = screen.getByText('input-chevron');
@@ -39,10 +39,10 @@ describe('Customization', () => {
       <CustomizationProvider
         baseTheme="default"
         elements={{
-          foo_CHEVRON_WRAPPER: {backgroundColor: 'colorBackground'},
+          FOO_CHEVRON_WRAPPER: {backgroundColor: 'colorBackground'},
         }}
       >
-        <InputChevronWrapper element="foo">input-chevron</InputChevronWrapper>
+        <InputChevronWrapper element="FOO_CHEVRON_WRAPPER">input-chevron</InputChevronWrapper>
       </CustomizationProvider>
     );
     const renderedInputChevronWrapper = screen.getByText('input-chevron');
--- a/packages/paste-core/components/input-box/src/InputChevronWrapper.tsx
+++ b/packages/paste-core/components/input-box/src/InputChevronWrapper.tsx
@@ -21,7 +21,7 @@ export const InputChevronWrapper = React.forwardRef<HTMLDivElement, InputChevron
       top="50%"
       transform="translateY(-50%)"
       zIndex="zIndex10"
-      element={`${element}_CHEVRON_WRAPPER`}
+      element={element}
       ref={ref}
     >
       {children}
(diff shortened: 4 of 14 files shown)


packages/paste-core/components/select/stories/select.stories.tsx
```diff
@@ -726,3 +728,89 @@ export const SelectOverflowLongValue = (): React.ReactNode => {
...
+        },
+        SELECT_ICON_WRAPPER: {
+          transform: 'rotate(90deg) translateX(-50%) translateY(-20%)',
+        },
+        CUSTOM_SELECT_ICON: {
+          color: 'colorTextInverseWeak',
+        },
+      }}
+    >
+      <Box maxWidth="size40" paddingX="space40" paddingY="space80">
```
andipants12, line 779: FYI I didn't create a component to share between the tests and stories since I did not want to create scope creep from modifying the unit tests.

.changeset/empty-spiders-rest.md
```diff
@@ -0,0 +1,6 @@
+---
+'@twilio-paste/input-box': patch
+'@twilio-paste/core': patch
+---
+
+[InputBox] Remove duplicative prefix in chevron icon wrapper element name, and update name too better reflect the component's purpose.
```
gloriliale, line 6:
```suggestion
[InputBox] Remove duplicative prefix in chevron icon wrapper element name, and update name to better reflect the component's purpose.
```

packages/paste-core/components/input-box/src/InputChevronWrapper.tsx
```diff
@@ -21,7 +21,7 @@ export const InputChevronWrapper = React.forwardRef<HTMLDivElement, InputChevron
       top="50%"
       transform="translateY(-50%)"
       zIndex="zIndex10"
-      element={`${element}_CHEVRON_WRAPPER`}
+      element={element}
```
gloriliale, line 24: This was a good catch! Especially when the term is already so long 😅

packages/paste-core/components/select/__tests__/option.test.tsx
```diff
@@ -1,14 +1,19 @@
...
 import axe from '../../../../../.jest/axe-helper';
 import {Option} from '../src';
 import {createAttributeMap} from '../test-utils';
 
+expect.extend(matchers);
+
 interface MockOptionProps extends React.OptionHTMLAttributes<HTMLOptionElement> {
   suffix?: string;
   value?: string;
+  element?: string;
```
gloriliale, line 16: Is it possible to use non-mocked props for this? or do we want them mocked?

packages/paste-core/components/select/__tests__/option.test.tsx
```diff
@@ -74,4 +79,63 @@ describe('Form | Option', () => {
...
+            OPTION_GROUP: {
+              color: 'colorText',
+            },
+          }}
+        >
+          <MockOption suffix="custom-styles" />
+        </CustomizationProvider>
+      );
+
+      expect(screen.getByTestId('option-custom-styles')).toHaveStyleRule('color', 'rgb(3,11,93)');
```
gloriliale, line 117: It might make sense to test for both of the stylings included in the custom theme, to make sure that both OPTION and OPTION_GROUP are behaving as expected. Thoughts?

gloriliale: Oh I think I see that this is being tested in the next spec, got it!

packages/paste-core/components/select/src/types.ts
```diff
@@ -1 +1,5 @@
+import type {BoxElementProps} from '@twilio-paste/box';
```
gloriliale, line 1: this is cool, should this be our preferred way of adding Element into our component types?

gloriliale (approved):
This looks customizable to me! Great job catching the places where we could make improvements in the other packages. Is the new Element type in types one that we can now use for customization anywhere??

packages/paste-core/components/input-box/__tests__/input-chevron-wrapper.test.tsx
```diff
@@ -8,13 +8,13 @@ expect.extend(matchers);
 
 describe('HTML attributes', () => {
   it('should set a element data attribute for InputChevronWrapper', () => {
-    render(<InputChevronWrapper element="INPUT">input-chevron</InputChevronWrapper>);
-    expect(screen.getByText('input-chevron').getAttribute('data-paste-element')).toEqual('INPUT_CHEVRON_WRAPPER');
+    render(<InputChevronWrapper element="INPUT_ICON_WRAPPER">input-chevron</InputChevronWrapper>);
+    expect(screen.getByText('input-chevron').getAttribute('data-paste-element')).toEqual('INPUT_ICON_WRAPPER');
```
SiTaggart, line 12: I would like to keep the element name as `x_CHEVRON_WRAPPER` in these type of inputs. Mainly because it's not just the input icon, it's specifically the input chevron, the select chevron, the combobox chevron, etc

packages/paste-core/components/select/__tests__/select.test.tsx
```diff
@@ -171,13 +185,229 @@ describe('Form | Select', () => {
...
+      );
+    });
+  });
+
+  describe('Customization', () => {
+    const CustomizationWrapper: React.FC<{children: React.ReactNode}> = ({children}) => (
+      <CustomizationProvider
+        // @ts-expect-error global test variable
+        theme={TestTheme}
+        elements={{
```
nkrantz, line 229: Ah now I see you're testing a lot of style rules down here in Select, nice!

nkrantz (approved):
Nice work Andi!

.changeset/five-lobsters-juggle.md
```diff
@@ -0,0 +1,6 @@
+---
+'@twilio-paste/combobox': patch
+'@twilio-paste/core': patch
+---
+
+[Combobox]: Small fix for element naming pattern.
```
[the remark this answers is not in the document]
SiTaggart: It matters, the changelog this adds to is public record. It should read to inform the consumer of the library what happened

packages/paste-core/components/select/__tests__/option.test.tsx
```diff
@@ -74,4 +79,63 @@ describe('Form | Option', () => {
...
+    });
+
+    it('should add custom styles to Option with a custom element data attribute', () => {
+      testRender(
+        <CustomizationProvider
+          // @ts-expect-error global test variable
+          theme={TestTheme}
+          elements={{
+            OPTION: {
+              color: 'colorTextLinkDarker',
```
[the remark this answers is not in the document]
SiTaggart: Tricky one, but in these cases what we're testing is that the ability to override or add styles to the target element node. We're also testing the name too as that's a public API. So the element name can't change, and the functionality targeting that name can't change.

In the box and text utils where we do the style merging is where we do a little more specific testing of the style props. 

But even then we're reliant on styled system and the css method doing their part to merge the styles into emotion. Which also have their own tests.

So, the end of my long story is each layer of the stack has its own tests, and up here at the top of the stack we don't necessarily need to retest things that are happening down the stack. So testing one style property is perfectly fine and should provide the coverage we need.

.changeset/five-lobsters-juggle.md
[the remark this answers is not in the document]
andipants12: Yeah I have a todo list item to look up some examples of really good change log notes because my drafts definitely leave room for improvement. Will push an update to this.

packages/paste-core/components/input-box/__tests__/input-chevron-wrapper.test.tsx
andipants12: Okey, that's fair. I will update 😸 I'm assuming the same for `Select` then too @SiTaggart?

packages/paste-core/components/select/__tests__/option.test.tsx
andipants12: I should probably rename this component function here, as it's not actually mocking anything, rather just creating a render of the component that we want to test configured for the test. 

That said, the interface for that component's props is called `MockOptionProps` because it's using a naming convention where the props interface is named by simply adding `Props` as a suffix to the component name.

Does that answer the question? Would renaming the props interface here make this more clear?
Would renaming the component `TestOption` and the props interface `TestOptionProps` make this more clear?
@gloriliale

packages/paste-core/components/select/__tests__/optiongroup.test.tsx
```diff
@@ -1,12 +1,17 @@
...
+import {CustomizationProvider} from '@twilio-paste/customization';
 // @ts-ignore typescript doesn't like js imports
 import axe from '../../../../../.jest/axe-helper';
 import {OptionGroup, Option} from '../src';
 import {createAttributeMap} from '../test-utils';
 
+expect.extend(matchers);
+
 interface MockOptionGroupProps extends React.OptgroupHTMLAttributes<HTMLOptGroupElement> {
+  element?: string;
```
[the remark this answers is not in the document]
andipants12: Not anything technically significant. I decided to simply declare the `element` type here since the rest of the "mock" test component interfaces are defined that way. Do you think this creates a level of confusion?

packages/paste-core/components/select/src/types.ts
andipants12: Yeah I've seen this done a few places in slightly different ways. Makes me wonder if it's worth creating a type in the box component package that is just `Element`

`export type Element = BoxElementProps['element'];`

Functionally it's the same as declaring this in each file, or skipping the declaration and just using the definition, but I think it might provide a layer of consistency and clarity. @SiTaggart @TheSisb Curious what you two think. 👀

packages/paste-core/components/input-box/__tests__/input-chevron-wrapper.test.tsx
SiTaggart: Yup

.changeset/five-lobsters-juggle.md
```diff
@@ -0,0 +1,6 @@
+---
+'@twilio-paste/combobox': patch
+'@twilio-paste/core': patch
+---
+
+[Combobox]: Update customization tests and stories to use more explicit a more explicit element name for the chevron wrapper. No consumer change required.
```
SiTaggart, line 6: Are this and input-box changelogs 100% accurate now?

SiTaggart (approved):
Outside of the question of changelogs, I think we're good

.changeset/five-lobsters-juggle.md
```diff
@@ -0,0 +1,6 @@
+---
+'@twilio-paste/combobox': patch
+'@twilio-paste/core': patch
+---
+
+[Combobox]: Add suffix to Chevron Icon wrapper. This ensures that the element string value for the combobox's chevron wrapper will always include the suffix "_CHEVRON_WRAPPER", and use either the default element name, or the consumer provided custom element name. No consumer change required.
```
andipants12, line 6: @SiTaggart I just updated this one, because I think the previous change log wasn't super descriptive of the actual change.

PR #1887, opened 2021-09-23
feat(button, base-radio-checkbox): add element customization

nkrantz:
**Contributing to Twilio**

### DIFF
--- a/.changeset/olive-lamps-sort.md
+++ b/.changeset/olive-lamps-sort.md
@@ -1,0 +1,6 @@
+---
+'@twilio-paste/base-radio-checkbox': patch
+'@twilio-paste/core': patch
+---
+
+[BaseRadioCheckbox] Enable Component to respect element customizations set on the customization provider. Component now enables setting an element name on the underlying HTML element and checks the emotion theme object to determine whether it should merge in custom styles to the ones set by the component author.
--- a/packages/paste-core/components/button/__tests__/customization.test.tsx
+++ b/packages/paste-core/components/button/__tests__/customization.test.tsx
@@ -1,0 +1,89 @@
+import * as React from 'react';
+import {matchers} from 'jest-emotion';
+import {render} from '@testing-library/react';
+import {CustomizationProvider} from '@twilio-paste/customization';
+import type {PasteCustomCSS} from '@twilio-paste/customization';
+import {AnyButton} from '../stories/customization.stories';
+
+expect.extend(matchers);
+
+const customButtonStyles = {
+  backgroundColor: 'colorBackgroundBusy',
+  borderRadius: 'borderRadius0',
+  borderWidth: 'borderWidth40',
+  borderColor: 'colorBorderError',
+  fontWeight: 'fontWeightLight',
+  variants: {
+    destructive: {backgroundColor: 'colorBackgroundDestructiveDarkest', color: 'colorTextWeaker'},
+    link: {padding: 'space40', borderRadius: 'borderRadiusCircle'},
+    destructive_link: {
+      padding: 'space40',
+      borderRadius: 'borderRadiusCircle',
+      backgroundColor: 'colorBackgroundDestructiveLight',
+      color: 'colorTextWarningStrong',
+      fontWeight: 'fontWeightBold',
+    },
+  },
+} as PasteCustomCSS;
+
+describe('Button customization', () => {
+  it('should set element data attribute on Button', () => {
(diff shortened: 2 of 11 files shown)


packages/paste-core/components/button/__tests__/customization.test.tsx
```diff
@@ -0,0 +1,70 @@
+import * as React from 'react';
```
nkrantz, line 1: I'm getting a `tabIndex` error on the customization tests. Is there something about nesting the Button within the CustomizationProvider that makes it think I'm setting the `tabIndex` to `-2`?
```
console.error
    Warning: Failed prop type: Invalid prop `tabIndex` of value `-2` supplied to `Button`, expected one of [0,-1].
        in Button
```

packages/paste-core/components/button/src/styles.ts
```diff
@@ -73,3 +74,22 @@ export const SizeStyles: {[key in ButtonSizes]: BoxStyleProps} = {
...
+    link: {padding: 'space40', borderRadius: 'borderRadiusCircle'},
+    destructive_link: {
+      padding: 'space40',
+      borderRadius: 'borderRadiusCircle',
+      backgroundColor: 'colorBackgroundDestructiveLight',
+      color: 'colorTextWarningStrong',
+      fontWeight: 'fontWeightBold',
+    },
+  },
+} as PasteCustomCSS;
```
SiTaggart, line 95: For test fixtures or mocks, it's better to not include them in source `src` files because they get compiled with the button package and delivered to customers via the compiled JS. 

Better to have a fixtures folder, or for something things small, just a constant in the test file itself if it's not shared. I also don't mind them being duplicate between stories and tests, they can even be different styles between the two assets

packages/paste-core/components/button/__tests__/customization.test.tsx
```diff
@@ -0,0 +1,70 @@
...
+import {CustomizationProvider} from '@twilio-paste/customization';
+import {AnyButton} from '../stories/customization.stories';
+import {customButtonStyles} from '../src/styles';
+
+expect.extend(matchers);
+
+describe('Button customization', () => {
+  it('should set element data attribute on Button', () => {
+    const {getByTestId} = render(<AnyButton variant="primary" />);
+    const button = getByTestId('button_for_customization');
```
SiTaggart, line 13: The preference from React testing library is to use `getByRole` over test ID where possible. I guess it makes sense because you're ensuring semantics too. 

They also prefer you to `render` and then use `screen` to query the DOM (`screen.getByTestId()`). I haven't figured out if that is more convention that specific improvement, because its multiple ways to do the same thing, but one is better?

Neither are required. Just a note for the future

packages/paste-core/components/button/__tests__/customization.test.tsx
SiTaggart: I can't see anything. Have you used the react testing library debugger to see whats going on?

packages/paste-core/components/button/src/styles.ts
nkrantz: Makes sense

packages/paste-core/components/button/__tests__/customization.test.tsx
nkrantz: Good to know, I copied the structure from the pagination PR here but I will keep this in mind going forward

packages/paste-core/components/button/__tests__/customization.test.tsx
nkrantz: Added to the test that was causing the prop types error to expect it and now I'm not seeing the error

.changeset/olive-lamps-sort.md
```diff
@@ -0,0 +1,6 @@
+---
+'@twilio-paste/base-radio-checkbox': patch
+'@twilio-paste/core': patch
+---
+
+[Base-radio-checkbox] Enable Component to respect element customizations set on the customization provider. Component now enables setting an element name on the underlying HTML element and checks the emotion theme object to determine whether it should merge in custom styles to the ones set by the component author.
```
andipants12, line 6: Super nit: Should this be hyphenated or should it read like a sentence? We've been capitalizing these so I think maybe this could be `BaseRadioCheckbox`?

nkrantz: Good catch! Updated

SiTaggart (approved):

zahnster (approved):
🎉 Excellent!

PR #1893, opened 2021-09-27
feat(data-grid): adding DataGrid component package

TheSisb:
This PR adds a new DataGrid component and package to Paste. The Paste DataGrid is a fully accessible and composable component. This PR releases it in `alpha` status, indicating there may be future breaking changes although unlikely. This component is built on top of Reakit's Tabbable and Composite, as well as @SiTaggart's codesandbox and knowledge.  It is very exciting as it is the first fully composable DataGrid component I've seen in React.

This work required many changes to other packages, clearly listed in individual commits and changeset files.


packages/paste-core/components/datagrid/src/utils/modded-reakit.ts
```diff
@@ -0,0 +1,158 @@
+/**
+ * Lightly modified functions copied from Reakit
+ * https://github.com/reakit/reakit/tree/master/packages/reakit-utils/src
+ * Used for focus management and tabbable controls in the datagrid
+ */
```
SiTaggart, line 5: I'm not sure I saw them but it might be useful have some commentary on what we lightly modified.

packages/paste-core/components/datagrid/src/utils/cell-management.ts
```diff
@@ -0,0 +1,72 @@
+import {getActiveElement} from './modded-reakit';
+
+export const isHTMLElement = (node: Node): boolean => {
+  return node instanceof HTMLElement;
+};
```
SiTaggart, line 5: We should probably have some unit tests around these if possible

packages/paste-core/components/datagrid/src/utils/modded-reakit.ts
SiTaggart: Also can we steal any tests reakit had for these for our test suite? Just for coverage and safety

packages/paste-core/components/datagrid/src/DataGrid.tsx
```diff
@@ -0,0 +1,148 @@
...
+  isCell,
+  delayedSetFocusable,
+} from './utils';
+
+interface DataGridProps extends TableProps {
+  'aria-label': string;
+  element?: BoxElementProps['element'];
+}
+
+export const DataGrid = React.forwardRef<HTMLTableElement, DataGridProps>(({element = 'DATAGRID', ...props}, ref) => {
```
SiTaggart, line 25: It's not been a rule and I don't really know how to make it a rule, but where possible I've been trying to add more TSDoc comments on everything, including component exports so that you get a little more info in your VSCode when using them. 

I think it would be a nice pro touch to expand this across the system.

packages/paste-core/components/datagrid/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,162 @@
+import * as React from 'react';
+import {Button} from '@twilio-paste/button';
+import {fireEvent, render, screen} from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+// @ts-ignore typescript doesn't like js imports
+import axe from '../../../../../.jest/axe-helper';
+
+import {ComposableCells} from '../stories/ComposableCells';
+
+describe('Datagrid', () => {
```
SiTaggart, line 10: I think generally this looks like it covers everything we need, just needs a little tidy up.

I can't think of any interaction cases outside of these right now.

How do you feel about some tests that assert the markup? Specifically which aria attributes are set on what cells, just to ensure our markup output stays as expected?

packages/paste-core/components/datagrid/src/DataGrid.tsx
```diff
@@ -0,0 +1,148 @@
...
+  const dataGridId = `datagrid-${useUID()}`;
+  const lastFocusedElement = React.useRef<Element | null>(null);
+  const compositeState = useCompositeState({unstable_virtual: false});
+  const [actionable, setActionable] = React.useState<boolean>(false);
+
+  const handleMouseDown = React.useCallback(() => {
+    if (!actionable) {
+      setActionable(true);
```
andipants12, line 33: Since the state change is being computed using the current state, pretty sure this would need to use a [functional update](https://reactjs.org/docs/hooks-reference.html#functional-updates "React docs - useState - functional updates") like above.

```
setActionable((isActionable) => {
    if (!isActionable) {
        return true; // 
    }
});
```

💡 You can actually simplify this by removing the conditional, since React will only apply the update if the value has changed, [which is mentioned in the API reference](https://reactjs.org/docs/hooks-reference.html#usestate). Check out the terminal logs in this sandbox for an example of that: https://codesandbox.io/s/usestate-functional-and-same-state-update-skipped-iu4kk?file=/src/App.js

packages/paste-core/components/datagrid/src/DataGrid.tsx
```diff
@@ -0,0 +1,148 @@
...
+  /**
+   * - Reset dataGrid to navigational on blur
+   * - Sets the last focused element before blurring to be the active tab stop (line 43)
+   */
+  const handleBlur = React.useCallback(
+    (event) => {
+      const isDataGridBlurred = !event.currentTarget.contains(event.relatedTarget);
+
+      if (isDataGridBlurred && actionable) {
+        setActionable(false);
```
andipants12, line 56: This one should use the functional update.

packages/paste-core/components/datagrid/src/DataGridCell.tsx
```diff
@@ -0,0 +1,88 @@
...
+  React.useEffect(() => {
+    if (cellRef.current) {
+      toggleActionable(getAllFocusableIn(cellRef.current), cellRef.current, dataGridState.actionable);
+    }
+  }, [dataGridState.actionable]);
+
+  return (
+    <CompositeItem
+      {...props}
+      {...dataGridState}
```
andipants12, line 79: Is this memoized? If not it might be a candidate for it, do we have any performance info on the new component?

packages/paste-core/components/datagrid/src/DataGridContext.tsx
```diff
@@ -0,0 +1,14 @@
+import * as React from 'react';
+import type {CompositeStateReturn} from '@twilio-paste/reakit-library';
+
+interface DataGridState extends Partial<CompositeStateReturn> {
+  actionable: boolean;
+  setActionable: (value: React.SetStateAction<boolean>) => void;
+}
+
+const NOOP: () => void = () => {};
```
andipants12, line 9:
```suggestion
const NOOP: VoidFunction = () => {};
```

Would make this easier to read.

packages/paste-core/components/datagrid/stories/ComposableCells.tsx
```diff
@@ -0,0 +1,81 @@
...
+          <DataGridHeader>Actions</DataGridHeader>
+        </DataGridRow>
+      </DataGridHead>
+      <DataGridBody>
+        {TableBodyData.map((row, rowIndex) => (
+          <DataGridRow key={`row-${rowIndex}`}>
+            {row.map((col, colIndex) => {
+              if (colIndex === 0 || colIndex === 1) {
+                return (
+                  <DataGridCell key={`col-${colIndex}`}>
```
andipants12, line 49: Would prefer to use `useUIDSeed` to create a UID seed and use those for the keys instead. In my experience that works super well for things like this, especially when a set is reordered, etc.

andipants12 (commented):
Didn't make it all the way through, but I got to the stories. Lots of great stuff here 🙌

packages/paste-core/components/datagrid/src/DataGridCell.tsx
TheSisb: This is just context, I'm not sure if / how we should memoize it?

packages/paste-core/components/datagrid/src/DataGrid.tsx
TheSisb: I need to unconditionally set this to false, I'm not using the previous state to infer the new state so it can remain as it is.

packages/paste-core/components/datagrid/src/DataGridContext.tsx
TheSisb: Cool! TIL

packages/paste-core/components/datagrid/stories/ComposableCells.tsx
TheSisb: While you can use `uid-library` for react keys, I actually recommend not doing so. This is because React keys aren't similar to HTML `id`s; they don't need to be globally unique. They only need to be contextually unique to this list. Furthermore it is too expensive to couple and call for something that a React implementation detail imo.

packages/paste-utils/src/hooks/useMutationObservable.ts
```diff
@@ -0,0 +1,74 @@
...
+ * This custom hooks abstracts the usage of the MutationObserver with React components.
+ * Watch for changes being made to the DOM tree and trigger a custom callback.
+ * @param {Element} targetEl DOM element to be observed
+ * @param {Function} cb callback that will run when there's a change in targetEl or any
+ * child element (depending on the provided options)
+ * @param {Object} options
+ * @param {Object} options.config check \[options\](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver/observe)
+ * @param {number} [options.debounceTime=0] a number that represents the amount of time in ms
+ * that you which to debounce the call to the provided callback function
+ */
```
SiTaggart, line 30: 🥇

packages/paste-utils/src/hooks/useMutationObservable.ts
```diff
@@ -0,0 +1,74 @@
...
+export function useMutationObservable(targetEl: Element, cb: Callback, options: Options = DEFAULT_OPTIONS): void {
+  const [observer, setObserver] = useState<MutationObserver | null>(null);
+
+  /* Create an observer instance */
+  useEffect(() => {
+    if (!cb || typeof cb !== 'function') {
+      // eslint-disable-next-line no-console
+      console.warn(`You must provide a valida callback function, instead you've provided ${cb}`);
```
SiTaggart, line 38:
```suggestion
      console.warn(`You must provide a valid callback function, instead you've provided ${cb}`);
```

packages/paste-utils/src/hooks/useMergeRefs.ts
```diff
@@ -0,0 +1,51 @@
+/* Credit to Chakra-UI
```
SiTaggart, line 1: classy 👍🏼

packages/paste-utils/src/hooks/useMergeRefs.ts
```diff
@@ -0,0 +1,51 @@
...
+ * React hook that merges react refs into a single memoized function
+ *
+ * @example
+ * import React from "react";
+ * import { useMergeRefs } from `@chakra-ui/hooks`;
+ *
+ * const Component = React.forwardRef((props, ref) => {
+ *   const internalRef = React.useRef();
+ *   return <div {...props} ref={useMergeRefs(internalRef, ref)} />;
+ * });
```
SiTaggart, line 38: :chef_kiss:

packages/paste-core/core-bundle/src/datagrid.tsx
```diff
@@ -0,0 +1 @@
+export * from '@twilio-paste/datagrid';
```
SiTaggart, line 1: Did the package name change by any chance? I think this one needs deleting

packages/paste-core/components/data-grid/stories/components/PaginatedDataGrid.tsx
```diff
@@ -0,0 +1,255 @@
...
+} from '@twilio-paste/pagination';
+import {DataGrid, DataGridHead, DataGridRow, DataGridHeader, DataGridBody, DataGridCell} from '../../src';
+import {TableHeaderData, PaginatedTableBodyData} from './constants';
+
+const getRange = (start: number, end: number): number[] => {
+  return [...new Array(end - start + 1)].map((_, index) => index + start);
+};
+
+/* Calculates the correct display of the pagination numbers */
+/* TODO: move this into the Pagination package as an export with tests */
```
SiTaggart, line 22: You thinking like a hook potentially or something like that? It's pretty neat

packages/paste-core/components/data-grid/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,250 @@
...
+    it('handles entering actionable mode from a cell without focusable children correctly', () => {
+      const {getByTestId} = render(<ComposableCellsDataGrid />);
+      const headerCell = getByTestId('header-1');
+      const firstInputCell = getByTestId('input-0-0')?.parentElement?.parentElement;
+      const secondInput = getByTestId('input-0-1');
+      if (firstInputCell == null) {
+        throw new Error('cannot find firstInputCell');
+      }
+
+      // I added this particular sequence because it was a reproducable bug in my manual tests
```
SiTaggart, line 191: Nice, useful

packages/paste-core/components/data-grid/src/table/Th.tsx
```diff
@@ -0,0 +1,51 @@
...
+        _focus={{
+          outline: 'none',
+          boxShadow: 'shadowFocusInset',
+        }}
+      />
+    );
+  }
+);
+
+Th.displayName = 'Th';
```
SiTaggart, line 46: This display name, and the TD and TR alike, might clash with the table ones. Lets say for exxample you for what ever reason add a table and a data grid in the same page. Should these be somehting more related to Datagrid so the disaplay names don't clash?

nkrantz (approved):
Looks great! Comments were super helpful and your code (especially tests) were really easy and clear to read. One question about the utils package: will we use that going forward for other util functions? How will we know what functions to add to this package, versus adding util functions to the package they are used for?

Also, just wanted to note again that some element props are `DATA_GRID` and some are `DATAGRID_[something]`. I left one comment about it but saw it in more places so just wanted to call out.

This review brought to you by @gloriliale and me!

packages/paste-core/core-bundle/src/datagrid.tsx
TheSisb: It did, great catch!

packages/paste-core/components/data-grid/stories/components/PaginatedDataGrid.tsx
TheSisb: Haven't decided on the form yet, but _something_ that lets people use Pagination the way we intended. This was complicated code to write, would be hard for teams to get it right consistently

packages/paste-core/components/data-grid/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,250 @@
...
+      expect(dataGrid.getAttribute('aria-label')).toBeDefined();
+      expect(dataGrid.getAttribute('role')).toBe('grid');
+    });
+  });
+
+  describe('Keyboard navigation', () => {
+    it('moves with arrow keys in navigational mode', () => {
+      const {getByTestId} = render(<ComposableCellsDataGrid />);
+      const headerCell = getByTestId('header-1');
+      const firstInputCell = getByTestId('input-0-0')?.parentElement?.parentElement;
```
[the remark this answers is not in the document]
TheSisb: Exactly. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

packages/paste-core/components/data-grid/src/DataGrid.tsx
```diff
@@ -0,0 +1,161 @@
...
+  );
+
+  const dataGridState = {
+    ...compositeState,
+    actionable,
+  };
+
+  return (
+    <DataGridContext.Provider value={dataGridState}>
+      <Box
```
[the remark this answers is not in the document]
TheSisb: This is required for horizontal scrolling to work.

packages/paste-core/components/data-grid/src/DataGridHeaderSort.tsx
```diff
@@ -0,0 +1,59 @@
...
+ * DataGrid sort button component. Must be placed in a DataGridHeader
+ *
+ * @param {"ascending" | "descending" | "none"} direction - provided to determine which icon to render
+ * @param {function} onClick - callback for when the button is clicked
+ * @param {string} element - customization element
+ */
+export const DataGridHeaderSort: React.FC<DataGridHeaderSortProps> = ({
+  direction,
+  onClick,
+  element = 'DATAGRID_HEADER_SORT',
```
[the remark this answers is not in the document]
TheSisb: Good catch. I didn't do a great job renaming this package and some stuff slid through. Ill fix it up.

packages/paste-core/components/data-grid/src/table/Th.tsx
TheSisb: It shouldn't be an issue even if they clash, and currently thead/tbody/tfoot/table are just using the Table component so if I change these, the other components will still match Tables. I think it will be fine as is.

SiTaggart (approved):

PR #1894, opened 2021-09-28
chore(spinner): add element customization

andipants12:
**Contributing to Twilio**

### DIFF
--- a/packages/paste-core/components/spinner/__tests__/index.test.tsx
+++ b/packages/paste-core/components/spinner/__tests__/index.test.tsx
@@ -12,16 +12,14 @@ import axe from '../../../../../.jest/axe-helper';
 expect.extend(matchers);
 
 const TestWrapper = (elements?: Record<string, any>): RenderOptions['wrapper'] => ({children}) => (
-  <Theme.Provider theme="default">
-    <CustomizationProvider
-      // @ts-expect-error global test variable
-      theme={TestTheme}
-      elements={elements}
-      data-testid="test-wrapper"
-    >
-      {children}
-    </CustomizationProvider>
-  </Theme.Provider>
+  <CustomizationProvider
+    // @ts-expect-error global test variable
+    theme={TestTheme}
+    elements={elements}
+    data-testid="test-wrapper"
+  >
+    {children}
+  </CustomizationProvider>
 );
 
 describe('Spinner', () => {
--- a/packages/paste-core/components/spinner/__tests__/index.test.tsx
+++ b/packages/paste-core/components/spinner/__tests__/index.test.tsx
@@ -1,0 +1,119 @@
+import * as React from 'react';
+import {matchers} from 'jest-emotion';
+import {Theme} from '@twilio-paste/theme';
+import {render, screen} from '@testing-library/react';
+import type {RenderOptions} from '@testing-library/react';
+import {CustomizationProvider} from '@twilio-paste/customization';
+
+import {Spinner} from '../src';
+// @ts-ignore typescript doesn't like js imports
+import axe from '../../../../../.jest/axe-helper';
+
+expect.extend(matchers);
+
+const TestWrapper = (elements?: Record<string, any>): RenderOptions['wrapper'] => ({children}) => (
+  <Theme.Provider theme="default">
+    <CustomizationProvider
+      // @ts-expect-error global test variable
+      theme={TestTheme}
+      elements={elements}
+      data-testid="test-wrapper"
+    >
+      {children}
+    </CustomizationProvider>
+  </Theme.Provider>
+);
+
+describe('Spinner', () => {
+  describe('HTML Attribute', () => {
+    it('should set an element data attribute for Spinner (default)', () => {
+      render(<Spinner color="colorText" title="Default name customization" decorative={false} />, {
+        wrapper: TestWrapper(),
+      });
+
+      const nodeNamedSpinner = screen.getByTestId('test-wrapper').firstChild as HTMLElement;
+
+      expect(nodeNamedSpinner.getAttribute('data-paste-element')).toEqual('SPINNER');
+    });
+
+    it('should set an element data attribute for Spinner', () => {
+      render(
+        <Spinner element="CUSTOM_SPINNER_A" color="colorText" title="Unique name customization" decorative={false} />,
+        {wrapper: TestWrapper()}
+      );
+
+      const nodeNamedSpinner = screen.getByTestId('test-wrapper').firstChild as HTMLElement;
+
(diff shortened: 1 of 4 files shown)


packages/paste-core/components/spinner/src/index.tsx
```diff
@@ -14,7 +14,10 @@ export interface SpinnerProps extends IconWrapperProps {
 }
 
 export const Spinner = React.forwardRef<HTMLDivElement, SpinnerProps>(
-  ({size, color = 'currentColor', title, as, display, decorative, delay = 250}, ref) => {
+  (
+    {size, color = 'currentColor', title, as, display, decorative, delay = 250, element = 'SPINNER_ICON_WRAPPER'},
```
TheSisb, line 18: Suggest matching component name
```suggestion
    {size, color = 'currentColor', title, as, display, decorative, delay = 250, element = 'SPINNER'},
```
The rest of this file is fine, just this one change

packages/paste-core/components/spinner/stories/index.stories.tsx
```diff
@@ -178,3 +180,76 @@ export const InFullscreenOverlay: React.ReactNode = () => {
...
+              element="CUSTOM_SPINNER_ICON_WRAPPER_B"
+              color="colorText"
+              title="Loading content"
+              decorative={false}
+            />
+
+            <Spinner color="colorText" title="Loading content" decorative={false} />
+
+            {/* Note: color passed here is ignored due to customization rules above */}
+            <Spinner
```
TheSisb, line 232: Is this a dupe of line 213?

nkrantz (changes requested):
Should Spinner have `PropTypes`? I can't find where they live, but I think `element` should be added to those in addition to `SpinnerProps`.

Also, it doesn't look like there's a test file for this package. Was that intentional? Either way, customization should probably have some tests, yeah?

Looking good otherwise!

packages/paste-core/components/spinner/src/index.tsx
```diff
@@ -14,7 +14,10 @@ export interface SpinnerProps extends IconWrapperProps {
 }
```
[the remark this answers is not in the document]
andipants12: Yes, thanks! Can't believe I missed that. 💯

packages/paste-core/components/spinner/stories/index.stories.tsx
andipants12: Yeah I was trying to demonstrate customizability between instances, and shared customization. Do you think this should be removed?

andipants12: I went ahead and removed them FYI

packages/paste-core/components/spinner/src/index.tsx
```diff
@@ -40,3 +52,14 @@ export const Spinner = React.forwardRef<HTMLDivElement, SpinnerProps>(
     );
   }
 );
+
+if (process.env.NODE_ENV === 'development') {
```
nkrantz, line 56: Still think we should ask about whether we should be adding this in but I didn't have time to bring up at 16th minute today

nkrantz (approved):

packages/paste-core/components/spinner/src/index.tsx
andipants12: Sure I think that's a good thing to discuss, I don't think it should block this PR though.

packages/paste-core/components/spinner/__tests__/index.test.tsx
```diff
@@ -0,0 +1,119 @@
...
+import {CustomizationProvider} from '@twilio-paste/customization';
+
+import {Spinner} from '../src';
+// @ts-ignore typescript doesn't like js imports
+import axe from '../../../../../.jest/axe-helper';
+
+expect.extend(matchers);
+
+const TestWrapper = (elements?: Record<string, any>): RenderOptions['wrapper'] => ({children}) => (
+  <Theme.Provider theme="default">
```
TheSisb, line 15: You don't need the Theme.Provider when immediately wrapping with Customization.Provider

packages/paste-core/components/spinner/__tests__/index.test.tsx
```diff
@@ -0,0 +1,119 @@
...
+    <CustomizationProvider
```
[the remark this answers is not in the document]
TheSisb: It can, but it is unnecessary in this example

packages/paste-core/components/spinner/src/index.tsx
TheSisb: We should be

nkrantz: We should be including `if (process.env.NODE_ENV === 'development') {` ? A few of the customization PRs have been removing this line.

andipants12: I think even if we decide to remove it for whatever reason, it shouldn't necessarily be done with the customization tickets since it's not part of the AC. That can lead us to scope creep, and loss of context/history.  I'll start a thread in our team channel now, but will resolve this convo.

packages/paste-core/components/spinner/__tests__/index.test.tsx
```diff
@@ -0,0 +1,119 @@
...
+import type {RenderOptions} from '@testing-library/react';
+import {CustomizationProvider} from '@twilio-paste/customization';
+
+import {Spinner} from '../src';
+// @ts-ignore typescript doesn't like js imports
+import axe from '../../../../../.jest/axe-helper';
+
+expect.extend(matchers);
+
+const TestWrapper = (elements?: Record<string, any>): RenderOptions['wrapper'] => ({children}) => (
```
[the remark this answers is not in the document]
andipants12: Yeah, record is super useful! It's a [utility type](https://www.typescriptlang.org/docs/handbook/utility-types.html#recordkeys-type) for `type Meow = { [x: string]: any }`

packages/paste-core/components/spinner/__tests__/index.test.tsx
```diff
@@ -0,0 +1,119 @@
...
+);
+
+describe('Spinner', () => {
+  describe('HTML Attribute', () => {
+    it('should set an element data attribute for Spinner (default)', () => {
+      render(<Spinner color="colorText" title="Default name customization" decorative={false} />, {
+        wrapper: TestWrapper(),
+      });
+
+      const nodeNamedSpinner = screen.getByTestId('test-wrapper').firstChild as HTMLElement;
```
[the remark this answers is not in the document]
andipants12: It could be yeah, but in this case I am saying that I expect the first child to be this thing, so if the tree changes this test will fail. I prefer that since typically unit tests should fail when changes are made, otherwise we could end up with sneaky false positives or missed bugs.

PR #1895, opened 2021-09-29
feat(breadcrumb): add element customization

andipants12:
**Contributing to Twilio**

### DIFF
--- a/packages/paste-core/components/breadcrumb/__tests__/breadcrumb.spec.tsx
+++ b/packages/paste-core/components/breadcrumb/__tests__/breadcrumb.spec.tsx
@@ -1,6 +1,7 @@
 import * as React from 'react';
 import {matchers} from 'jest-emotion';
 import {render, screen} from '@testing-library/react';
+import {CustomizationProvider} from '@twilio-paste/customization';
 // @ts-ignore typescript doesn't like js imports
 import axe from '../../../../../.jest/axe-helper';
 import {Breadcrumb, BreadcrumbItem} from '../src';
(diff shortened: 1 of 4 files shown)


packages/paste-core/components/breadcrumb/__tests__/breadcrumb.spec.tsx
```diff
@@ -66,4 +67,206 @@ describe('Breadcrumb', () => {
...
+
+      const node2 = breadcrumbItem2.firstChild as HTMLElement;
+      const separator2 = breadcrumbItem2.lastChild as HTMLElement;
+
+      expect(node2.getAttribute('data-paste-element')).toEqual('BREADCRUMB_ITEM_ANCHOR');
+      expect(separator2.getAttribute('data-paste-element')).toEqual('BREADCRUMB_SEPARATOR');
+
+      const node3 = breadcrumbItem3.firstChild as HTMLElement;
+
+      // fix this test!!!!!!
```
TheSisb, line 109: TODO remove?

packages/paste-core/components/breadcrumb/__tests__/breadcrumb.spec.tsx
```diff
@@ -66,4 +67,206 @@ describe('Breadcrumb', () => {
       expect(results).toHaveNoViolations();
     });
   });
+
+  describe.only('HTML attributes', () => {
```
TheSisb, line 71: Hmm this describe block name is a little misleading because we're checking for data-attributes specifically for customization. Maybe consider moving this into the customization block?

packages/paste-core/components/breadcrumb/src/index.tsx
```diff
@@ -14,39 +15,51 @@ const BreadcrumbSeparator: React.FC = () => (
     paddingLeft="space20"
     paddingRight="space20"
     role="presentation"
+    element={`${element}_SEPARATOR`}
```
TheSisb, line 18: Since this component isn't being exported, and is directly passed from BreadcrumbItem, you only need to pass `element={element}`

packages/paste-core/components/breadcrumb/src/index.tsx
```diff
@@ -14,39 +15,51 @@ const BreadcrumbSeparator: React.FC = () => (
...
+  element?: BoxElementProps['element'];
   href?: string;
   last?: boolean;
 }
 
 const BreadcrumbItem = React.forwardRef<HTMLAnchorElement, BreadcrumbItemProps>(
-  ({children, href, last, ...props}, ref) => {
+  ({children, element, href, last, ...props}, ref) => {
+    const separatorElementName = element || 'BREADCRUMB';
+    const elementName = element || 'BREADCRUMB_ITEM';
```
TheSisb, line 34: The same behavior can be achieved more succinctly by doing `element = 'BREADCRUMB_ITEM',` on line 32

packages/paste-core/components/breadcrumb/src/index.tsx
```diff
@@ -14,39 +15,51 @@ const BreadcrumbSeparator: React.FC = () => (
...
   children: NonNullable<React.ReactNode>;
+  element?: BoxElementProps['element'];
   href?: string;
   last?: boolean;
 }
 
 const BreadcrumbItem = React.forwardRef<HTMLAnchorElement, BreadcrumbItemProps>(
-  ({children, href, last, ...props}, ref) => {
+  ({children, element, href, last, ...props}, ref) => {
+    const separatorElementName = element || 'BREADCRUMB';
```
TheSisb, line 33: I'd recommend doing this inline on line 62 like this:
```{!last && <BreadcrumbSeparator element={`${element}_SEPARATOR`} />}```
The same way ANCHOR and TEXT are being handled

File: packages/paste-core/components/breadcrumb/src/index.tsx [hunk 6]
```diff
@@ -14,39 +15,51 @@ const BreadcrumbSeparator: React.FC = () => (
...
+    const elementName = element || 'BREADCRUMB_ITEM';
     return (
       <Box
         {...safelySpreadBoxProps(props)}
         alignItems="center"
         as="li"
         color="colorText"
         display="inline-flex"
+        element={elementName}
```
TheSisb, line 42:
```suggestion
        element={element}
```

packages/paste-core/components/breadcrumb/__tests__/breadcrumb.spec.tsx
andipants12: Yes for sure, thank you!

packages/paste-core/components/breadcrumb/src/index.tsx
andipants12: So you are saying we should move the `_SEPARATOR` and the string interp to line 62?

packages/paste-core/components/breadcrumb/__tests__/breadcrumb.spec.tsx
andipants12: This is how most of the customization tests have been done, was following the patterns we established. I agree though that the describe block is confusing, I think though it warrants a separate ticket/effort to address in all the customization tests.

packages/paste-core/components/breadcrumb/src/index.tsx
andipants12: Sure, I can make that change. Curious if this is a preference in style or if there is a functional benefit? I've gotten in the habit of declaring constants for these kinds of things rather than using inline string interpolation when there are conditionals since it provides a bit more clarity on specifically why the interpolation is there. I know there is technically memory spent on declaring the constant, but I think it's pretty minimal.

packages/paste-core/components/breadcrumb/src/index.tsx
andipants12: I intentionally did not use a default here since so I could conditionally pass the consumer provided element string, if it is defined, and if not use `BREADCRUMB_SEPARATOR` and `BREADCRUMB_ITEM` as defaults for the breadcrumb item and the separator component respectively.
Notice that the separator is named `BreadcrumbSeparator` and we have been using default names that match the component name.

I could have alternatively added some conditional logic like:
```
<BreadcrumbSeparator element={element === 'BREADCRUMB_ITEM' ? 'BREADCRUMB_SEPARATOR' : `${element}_SEPARATOR`}
```

I chose not to do that though, since inline ternary operators like this are less readable.

packages/paste-core/components/breadcrumb/src/index.tsx
TheSisb: Yup

packages/paste-core/components/breadcrumb/src/index.tsx
TheSisb: The closer the variables are to where they are used, without sacrificing readability, the better generally.
1- Avoids a const declaration, which lets readers of this code not need to figure out what this line does as they read through the code.
2- Since line 62 is conditional, `{!last && ...` then the string concat doesn't need to run if it isn't used. Using locality this way provides immediate benefits.
3- We do it on lines 47 and 54 :P

packages/paste-core/components/breadcrumb/src/index.tsx
TheSisb: if `element="BREADCRUMB_ITEM"`, then the separator will be `BREADCRUMB_ITEM_SEPARATOR` which is fine.
if the user passes and element (`element="FOO"`), then the separator will be `FOO_SEPARATOR` which is also fine.

This is how we handle it everywhere

packages/paste-core/components/breadcrumb/__tests__/breadcrumb.spec.tsx
TheSisb: np

packages/paste-core/components/breadcrumb/stories/index.stories.tsx
```diff
@@ -49,3 +51,81 @@ export const AnchorsWithTruncation = (): React.ReactNode => {
...
+        BREADCRUMB_ITEM: {fontWeight: 'fontWeightMedium'},
+        BREADCRUMB_ITEM_ANCHOR: {
+          textDecoration: 'underline wavy',
+          color: 'colorTextInverseWeaker',
+          ':hover': {color: 'colorLinkStronger'},
+        },
+        BREADCRUMB_ITEM_TEXT: {letterSpacing: '0.25rem'},
+        BREADCRUMB_SEPARATOR: {
+          color: 'colorTextBrandHighlight',
+        },
```
SiTaggart, line 71: Are we open to the following public API?

```
<nav data-paste-element="BREADCRUMB">
  <ol data-paste-element="BREADCRUMB_LIST">
    <li data-paste-element="BREADCRUMB_ITEM">
      <a href="" data-paste-element="BREADCRUMB_ANCHOR" />
      <span  data-paste-element="BREADCRUMB_SEPARATOR" />
    </li>
    <li data-paste-element="BREADCRUMB_ITEM">
      <a href="" data-paste-element="BREADCRUMB_ANCHOR" />
      <span  data-paste-element="BREADCRUMB_SEPARATOR" />
    </li>
    <li data-paste-element="BREADCRUMB_ITEM">
      <span data-paste-element="BREADCRUMB_TEXT" />
    </li>
  </ol>
</nav>
```

Then work backwards to achieve it?

packages/paste-core/components/breadcrumb/src/index.tsx
andipants12: Oh but the separator is called `BreadcrumbSeparator`. I think yesterday in focus time we said that it should not include `Item`. If that's not a requirement, I'm happy to change.

packages/paste-core/components/breadcrumb/stories/index.stories.tsx
andipants12: @SiTaggart I included `Item` in the anchor and text element names respectfully because they are included currently in the `BreadcrumbItem` component. Are you saying that we should not apply that convention rigidly here?

andipants12: FWIW I think I'm getting thrown a bit here because:
1. the component name is `BreadcrumbItem` so I am using `BREADCRUMB_ITEM` as the default element name, and passing it to all nodes rendered by the component (that we want to target)
2. It makes sense to me that the `li` is the only node that has `_ITEM` in the name, because it is literally an item.

SiTaggart: We don't need to expose our internal implementation details in a public API, yeah. Design for the customer experience.

packages/paste-core/components/breadcrumb/src/index.tsx
```diff
@@ -57,44 +72,50 @@ BreadcrumbItem.displayName = 'BreadcrumbItem';
...
-    <Box {...safelySpreadBoxProps(props)} aria-label="breadcrumb" as="nav" ref={ref}>
-      <Box alignItems="center" as="ol" display="inline-flex" listStyleType="none" margin="space0" padding="space0">
-        {validChildren.map((child, index) =>
-          React.cloneElement(child as React.ReactElement<any>, {
-            last: childrenCount === index + 1,
-            key: keySeed(`breadcrumb-${index}`),
-          })
-        )}
+    return (
+      <Box {...safelySpreadBoxProps(props)} aria-label="breadcrumb" as="nav" element={element} ref={ref}>
```
nkrantz, line 98: Wouldn't this Box end up having the same `data-paste-element` value as the BreadCrumbSeparator? `DEFAULT_ELEMENT_NAME`?

nkrantz (approved):
Looks good, I was a bit confused by how things were working in `index.tsx` but I think I understand. Just left a comment about element naming.

packages/paste-core/components/breadcrumb/__tests__/breadcrumb.spec.tsx
```diff
@@ -66,4 +67,228 @@ describe('Breadcrumb', () => {
       expect(results).toHaveNoViolations();
     });
   });
+
+  describe('Customization', () => {
+    describe('HTML attributes', () => {
```
TheSisb, line 72: This describe block may be redundant at this point

packages/paste-core/components/breadcrumb/__tests__/breadcrumb.spec.tsx
```diff
@@ -66,4 +67,228 @@ describe('Breadcrumb', () => {
...
+      describe('Parent element name is defined (Breadcrumb) and element name is defined (BreadcrumbItem)', () => {
```
TheSisb, line 73: This is kind of confusing me :/ 
I think it's trying to say `Breadcrumb and BreadcrumbItem are being passed an element prop`?

Also you don't need the describe block, just put this string in the `it()` directly.

packages/paste-core/components/breadcrumb/src/index.tsx
TheSisb: Separator does a concat

nkrantz: Oh I see that now

TheSisb (approved):

PR #1899, opened 2021-10-01
Customization/ga

gloriliale:
Incomplete!

Customization of Inline Control Group, Checkbox (halfway through), Radio Group (only started).

Putting this on the chopping block before I continue work on Monday. Thanks!


packages/paste-core/components/checkbox/src/Checkbox.tsx
```diff
@@ -136,7 +162,11 @@ const Checkbox = React.forwardRef<HTMLInputElement, CheckboxProps>(
             </MediaObject>
           </BaseRadioCheckboxLabelText>
         </BaseRadioCheckboxLabel>
-        {helpText && <BaseRadioCheckboxHelpText helpTextId={helpTextId}>{helpText}</BaseRadioCheckboxHelpText>}
+        {helpText && (
+          <BaseRadioCheckboxHelpText element={`${element}_HELP_TEXT`} helpTextId={helpTextId}>
```
SiTaggart, line 166: We're just going with the things unique to Checkbox and Radio button right? Label, Help Text and Error will remain the base so that can be customized as a whole form for each input type

packages/paste-core/components/checkbox/__tests__/checkbox.test.tsx
```diff
@@ -224,6 +238,66 @@ describe('Checkbox event handlers', () => {
   });
 });
 
+describe('Customization', () => {
+  it('Should set an element data attribute for Checkbox', (): void => {
```
andipants12, line 242: FWIW you don't need to type the assertion callback in an `it` statement.

🪙 🪙 Might be good to remove, since it adds unnecessary noise.

packages/paste-core/components/checkbox/stories/checkbox.stories.tsx
```diff
@@ -643,3 +644,78 @@ export const CheckboxDisclaimerDisabled = (): React.ReactNode => {
 CheckboxDisclaimerDisabled.story = {
   name: 'Checkbox Disclaimer - Disabled',
 };
+
+export const CustomizedCheckboxGroup = (): React.ReactNode => {
+  const [checked1, setChecked1] = React.useState(true);
```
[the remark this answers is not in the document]
gloriliale: That's a very good point! Will pay more attention to this moving forward haha.

packages/paste-core/components/checkbox/__tests__/checkbox.test.tsx
gloriliale: ah i see! hmm I'd seen in it one of Rich's PR's a while back and added in bc of that, but I can remove these easily. Any other opinions on the typing here?

PR #1900, opened 2021-10-01
feat(menu): customization

andipants12:
**Contributing to Twilio**


packages/paste-core/components/menu/src/constants.ts
```diff
@@ -2,3 +2,5 @@ export const MenuItemVariants = {
   DEFAULT: 'default',
   GROUP_ITEM: 'group_item',
 } as const;
+
+export const DEFAULT_ELEMENT_NAME = 'MENU';
```
nkrantz, line 6: This is a good idea!

packages/paste-core/components/menu/src/MenuGroup.tsx
```diff
@@ -1,35 +1,44 @@
...
+          element={`${element}_GROUP`}
+          role="presentation"
+          aria-label={label}
+          textDecoration="none"
+          ref={ref}
+        >
+          <Box display="flex" alignItems="center" paddingX="space70" paddingY="space30" cursor="default">
+            {React.isValidElement(icon) ? (
+              <Box flexShrink={0} size="sizeIcon30">
+                {React.cloneElement(icon, {color: 'colorTextIcon', parentElement: element})}
```
nkrantz, line 24: What is setting `parentElement` equal to `element` on the Icon doing here?

packages/paste-core/components/menu/src/MenuItem.tsx
```diff
@@ -3,49 +3,53 @@ import {secureExternalLink} from '@twilio-paste/anchor';
...
-      _disabled={{cursor: 'not-allowed'}}
-      ref={ref}
-    >
-      {props.children}
-    </Box>
-  );
-});
+export const StyledMenuItem = React.forwardRef<HTMLDivElement | HTMLAnchorElement, MenuItemProps>(
+  ({element = `${DEFAULT_ELEMENT_NAME}_ITEM`, ...props}, ref) => {
+    const woof = props.href ? 'ANCHOR' : 'BUTTON';
```
nkrantz, line 11: I love it, but do we maybe want to call `woof` something else before merging 😄

nkrantz (commented):
Some cool stuff going on here! Just left a few questions for ya 🎉

packages/paste-core/components/menu/src/MenuItem.tsx
andipants12: yess we do in fact 🤣 classic andipants at work, thanks for the catch!

packages/paste-core/components/menu/src/MenuGroup.tsx
andipants12: Oh I actually need to remove this. Originally I had thought we should pass the parent element name and conditionally replace the item default with it, if it's a sub menu, etc. but we decided to not do this. 👍

packages/paste-core/components/menu/src/MenuButton.tsx
```diff
@@ -6,9 +6,9 @@ import {Button} from '@twilio-paste/button';
 
 export type MenuButtonProps = MenuPrimitiveButtonProps & ButtonProps;
 
-const MenuButton = React.forwardRef<HTMLButtonElement, MenuButtonProps>((props, ref) => {
+const MenuButton = React.forwardRef<HTMLButtonElement, MenuButtonProps>(({element = 'MENU_BUTTON', ...props}, ref) => {
```
SiTaggart, line 9: Seems to be the only place where you're not using the constant element name. My gut said it felt a little unnecessary to begin with, but it's something I don't really mind about either way.

We should however either; use it everywhere or use it nowhere.

packages/paste-core/components/menu/src/constants.ts
TheSisb: I'm not sure this provides that much value personally, since the default element name is just a copy of the component and the string concatenation adds developer complexity.

packages/paste-core/components/menu/stories/customization.stories.tsx
```diff
@@ -0,0 +1,186 @@
...
+        </Menu>
+      </CustomizationProvider>
+    );
+  }
+);
+
+export const WithDefaultElementName: React.FC = () => {
+  return (
+    <Box display="flex" flexDirection="row" justifyContent="space-between">
+      <BaseMenu menuButtonVariant="primary" element="MENU" />
```
TheSisb, line 167: We should remove the element prop for the default to truly kick in

TheSisb (commented):
Still missing some tests I think?

packages/paste-core/components/menu/src/MenuGroup.tsx
```diff
@@ -1,35 +1,44 @@
...
-            as="span"
-            color="colorText"
-            role="presentation"
-            fontWeight="fontWeightBold"
-            paddingLeft={icon != null ? 'space20' : undefined}
-          >
-            {label}
-          </Text>
+const MenuGroup = React.forwardRef<HTMLDivElement, MenuGroupProps>(
+  ({label, icon, children, element = DEFAULT_ELEMENT_NAME, ...props}, ref) => {
```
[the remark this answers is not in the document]
andipants12: It's called menu group below 👇

packages/paste-core/components/menu/src/MenuItem.tsx
```diff
@@ -3,49 +3,52 @@ import {secureExternalLink} from '@twilio-paste/anchor';
...
-      }}
-      _disabled={{cursor: 'not-allowed'}}
-      ref={ref}
-    >
-      {props.children}
-    </Box>
-  );
-});
+export const StyledMenuItem = React.forwardRef<HTMLDivElement | HTMLAnchorElement, MenuItemProps>(
+  ({element = `${DEFAULT_ELEMENT_NAME}_ITEM`, ...props}, ref) => {
```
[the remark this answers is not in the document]
andipants12: This is `MENU_ITEM` I was just using a constant for the `MENU` prefix. Might remove it though, I feel like it's less useful here and just obfuscates the default name.

packages/paste-core/components/menu/src/MenuGroup.tsx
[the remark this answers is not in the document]
TheSisb: Ah yeah, I mean like `element="MENU_GROUP"` right here exactly, and then `element={element}` below since we don't need to join strings for any other reason

packages/paste-core/components/menu/src/MenuItem.tsx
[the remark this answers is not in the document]
TheSisb: Yup exactly

packages/paste-core/components/menu/src/MenuGroup.tsx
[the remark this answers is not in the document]
TheSisb: Ah yeah, I mean like element="MENU_GROUP" right here exactly, and then element={element} below since we don't need to join strings for any other reason

packages/paste-core/components/menu/src/constants.ts
nkrantz: I was thinking that it would be useful if we ever decided to change the default element name, since we'd only have to make that change in 1 place. I guess maybe that's not going to be too common? I followed a similar pattern in Toast (adding the default element name to the constants folder), but open to not doing it that way.

packages/paste-core/components/menu/__tests__/customization.spec.tsx
```diff
@@ -0,0 +1,399 @@
+import * as React from 'react';
```
andipants12, line 1: These are the new tests, not sure why it's hiding this diff, maybe it's just large enough 🤷‍♀️ @TheSisb

packages/paste-core/components/menu/src/SubMenuButton.tsx
```diff
@@ -1,24 +1,31 @@
...
-        {props.children && <MediaBody as="span">{props.children}</MediaBody>}
-        <MediaFigure as="span" align="end" spacing="space20">
-          <ChevronRightIcon decorative size="sizeIcon30" />
-        </MediaFigure>
-      </MediaObject>
-    </MenuPrimitiveButton>
-  );
-});
+const SubMenuButton = React.forwardRef<HTMLButtonElement, SubMenuButtonProps>(
+  ({element = 'SUBMENU', ...props}, ref) => {
```
TheSisb, line 12: prefer: `SUBMENU_BUTTON`. If a user passes `element="ABC"`, then it will not need to concat on the wrapper, which makes using this and customizing it much more obvious.

packages/paste-core/components/menu/stories/customization.stories.tsx
```diff
@@ -0,0 +1,231 @@
...
+      destructive_link: {
+        padding: 'space40',
+        borderRadius: 'borderRadiusCircle',
+        backgroundColor: 'colorBackgroundDestructiveLight',
+        color: 'colorTextWarningStrong',
+        fontWeight: 'fontWeightBold',
+      },
+    },
+  },
+  [`SUB${element}_BUTTON`]: {
```
TheSisb, line 38: Does this still work with the changes? It shouldn't afaict..

packages/paste-core/components/menu/stories/customization.stories.tsx
```diff
@@ -0,0 +1,231 @@
...
+
+    const subMenu = useMenuState({baseId: `${menuButtonVariant}-menu-submenu`});
+    const onClick = React.useCallback(() => {
+      menu.hide();
+    }, [menu.hide]);
+    return (
+      <>
+        <MenuButton
+          {...menu}
+          element={getElementName(element, 'BUTTON')}
```
TheSisb, line 117: Since we're always passing _some_ element here, we're never checking the default. So if someone in the future changed the default value in the component, our tests won't catch it. Do you think we should test for that?

packages/paste-core/components/menu/src/SubMenuButton.tsx
```diff
@@ -1,24 +1,31 @@
...
+    // MenuPrimitiveButton from reakit types `as` as HTML element names, but accepts components. any prevents type errors
+    return (
+      <MenuPrimitiveButton {...props} as={StyledMenuItem as any} element={`${element}_BUTTON`} ref={ref}>
```
TheSisb, line 15: This becomes `element={element}`

packages/paste-core/components/menu/src/SubMenuButton.tsx
```diff
@@ -1,24 +1,31 @@
...
+        <MediaObject as="span" verticalAlign="center" element={`${element}_MEDIA_OBJECT`}>
+          {props.children && (
+            <MediaBody as="span" element={`${element}_MEDIA_BODY`}>
```
TheSisb, line 18: This _should_ be SUBMENU_BUTTON_MEDIA_BODY, because it's the media body in the SubMenuButton component, not the actual SUBMENU that appears as a result of clicking this.

packages/paste-core/components/menu/src/SubMenuButton.tsx
```diff
@@ -1,24 +1,31 @@
...
+              {props.children}
+            </MediaBody>
+          )}
+          <MediaFigure as="span" align="end" spacing="space20" element={`${element}_MEDIA_FIGURE`}>
+            <ChevronRightIcon decorative size="sizeIcon30" element={`${element}_ICON`} />
```
TheSisb, line 23: Same comment as above, this becomes by default SUBMENU_BUTTON_ICON

TheSisb (changes requested):

packages/paste-core/components/menu/stories/customization.stories.tsx
andipants12: This actually does pass the default because it only applies the suffix if element is not undefined. If undefined, it passes undefined, and the component uses its' default.

packages/paste-core/components/menu/__tests__/customization.spec.tsx
```diff
@@ -0,0 +1,464 @@
...
+import type {RenderOptions} from '@testing-library/react';
+import {CustomizationProvider} from '@twilio-paste/customization';
+
+import {initStyles, BaseMenu} from '../stories/customization.stories';
+
+expect.extend(matchers);
+
+const PASTE_ELEMENT = 'data-paste-element';
+
+const initWrapper = (element: string): RenderOptions['wrapper'] => ({children}) => (
```
nkrantz, line 13: What is `RenderOptions`? I can't find it in testing library docs unless I'm just missing it 👀

packages/paste-core/components/menu/__tests__/customization.spec.tsx
```diff
@@ -0,0 +1,464 @@
...
+
+        expect(nodeCalledMenuSeparator).toHaveStyleRule('border-style', 'inset');
+        expect(nodeCalledMenuSeparator).toHaveStyleRule('opacity', '0.5');
+        expect(nodeCalledMenuSeparator).toHaveStyleRule('border-width', '8px');
+        expect(nodeCalledMenuSeparator).toHaveStyleRule('border-color', 'rgb(225,227,234)');
+      });
+
+      it('Should apply correct style rules when menu button variant is "secondary"', () => {
+        render(<BaseMenu menuButtonVariant="secondary" />, {
+          wrapper: initWrapper('menu'),
```
nkrantz, line 114: I always learn something from your PRs!

nkrantz (approved):

PR #1903, opened 2021-10-02
feat(modal): customization

andipants12:
**Contributing to Twilio**

### DIFF
--- a/packages/paste-core/components/modal/__tests__/customization.spec.tsx
+++ b/packages/paste-core/components/modal/__tests__/customization.spec.tsx
@@ -9,6 +9,7 @@ import {BaseModal, initStyles} from '../stories/customization.stories';
 expect.extend(matchers);
 
 jest.mock('@twilio-paste/modal-dialog-primitive', () => {
+  // Mocking the portal as a div so it renders within the body of the rendered test fragment, rather than using Portal behavior.
   const actual = jest.requireActual('@twilio-paste/modal-dialog-primitive');
   const {forwardRef: mockForwardRef} = jest.requireActual('react');
   const MockModalDialogPrimitiveOverlay = mockForwardRef(
--- a/packages/paste-core/components/modal/__tests__/customization.spec.tsx
+++ b/packages/paste-core/components/modal/__tests__/customization.spec.tsx
@@ -1,0 +1,162 @@
+import * as React from 'react';
+
+import {render, screen} from '@testing-library/react';
+import {CustomizationProvider} from '@twilio-paste/customization';
+
+import {BaseModal, initStyles} from '../stories/customization.stories';
+
+jest.mock('@twilio-paste/modal-dialog-primitive', () => {
+  const actual = jest.requireActual('@twilio-paste/modal-dialog-primitive');
+  const {forwardRef: mockForwardRef} = jest.requireActual('react');
+  const MockModalDialogPrimitiveOverlay = mockForwardRef(
+    (
+      {
+        children,
+        'data-paste-element': dataPasteElement,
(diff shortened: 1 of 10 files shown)


nkrantz (commented):
It's looking great so far! Missing tests and changeset but component names are good.

packages/paste-core/components/modal/src/Modal.tsx
```diff
@@ -123,22 +129,30 @@ const Modal = React.forwardRef<HTMLDivElement, ModalProps>(
         {transitions(
           (styles, item) =>
             item && (
-              <ModalDialogOverlay
+              <Box
+                // @ts-expect-error Render overlay as box for customization
```
SiTaggart, line 133: I actually don't _mind_ this approach, but if there is some push back you can achieve it with styled components. We did it in Textarea https://github.com/twilio-labs/paste/blob/main/packages/paste-core/components/textarea/src/TextArea.tsx#L71

The custom css function is exported from Box, and you can pass it to the styled component manually to merge in the custom styles from `element`. You just need to manually apply the `data-paste-element` attributes on the node you want to target. 

I don't have a preference on either way. This is just knowledge share

packages/paste-core/components/modal/src/Modal.tsx
```diff
@@ -147,16 +161,14 @@ const Modal = React.forwardRef<HTMLDivElement, ModalProps>(
...
-  };
-}
+Modal.propTypes = {
+  children: PropTypes.node.isRequired,
+  isOpen: PropTypes.bool.isRequired,
+  onDismiss: PropTypes.func.isRequired,
+  allowPinchZoom: PropTypes.bool,
+  size: PropTypes.oneOf(['default', 'wide'] as Sizes[]).isRequired,
+  initialFocusRef: PropTypes.object as any,
+  ariaLabelledby: PropTypes.string.isRequired,
```
SiTaggart, line 171: Should we be adding `element` to all these propTypes, across all the components in the PR?

andipants12: Ah yes we should, that was an oversight. Will add it now. 👍

packages/paste-core/components/modal/__tests__/customization.spec.tsx
```diff
@@ -0,0 +1,162 @@
...
+    (
+      {
+        children,
+        'data-paste-element': dataPasteElement,
+        style,
+        className,
+      }: {children: any; 'data-paste-element': string; style: any; className: string},
+      ref: any
+    ) => (
+      <div
```
andipants12, line 21: Mocking the react portal for this test.

nkrantz (approved):
I like the helper functions you made! This is looking good 👍

packages/paste-core/components/modal/__tests__/customization.spec.tsx
```diff
@@ -0,0 +1,165 @@
...
+import {render, screen} from '@testing-library/react';
+import {CustomizationProvider} from '@twilio-paste/customization';
+import {matchers} from 'jest-emotion';
+
+import {BaseModal, initStyles} from '../stories/customization.stories';
+
+expect.extend(matchers);
+
+jest.mock('@twilio-paste/modal-dialog-primitive', () => {
+  const actual = jest.requireActual('@twilio-paste/modal-dialog-primitive');
```
SiTaggart, line 12: I guess I'm a little unsure on why this needs to be mocked, so providing some sort of commentary on the why might be helpful in a code comment.

SiTaggart (approved):

PR #1915, opened 2021-10-06
Customization/popover

zahnster:
Adds element customization to Popover, including stories and tests.


packages/paste-core/components/popover/__tests__/index.spec.tsx
```diff
@@ -78,4 +84,128 @@ describe('Popover', () => {
...
+              <Text as="span">This is the Twilio styled popover that you can use in all your applications.</Text>
+            </Popover>
+          </PopoverContainer>
+        </CustomizationProvider>
+      );
+
+      const popoverComp = screen.getByTestId('popover');
+      const popoverButton = screen.getByTestId('popover-button');
+
+      expect(popoverComp.querySelector('[data-paste-element="POPOVER"]')).toHaveStyleRule(
```
TheSisb, line 146: You might be able to just merge this `it()` block with the previous one, and check for the style and data-paste-element in the same block.  Would simplify the test file by not needing the repetitive render boilerplate.

PR #1952, opened 2021-10-14
Chore/yarn3

zahnster:
* Updates project to Yarn3
* Updates dependencies and scripts to be compatible

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -30,6 +30,5 @@
     X-Shenanigans = "none"
 
 [build.environment]
-  YARN_VERSION = "3.0.2"
   YARN_ENABLE_GLOBAL_CACHE = "true"
   YARN_GLOBAL_FOLDER = "/opt/buildhome/.yarn_cache"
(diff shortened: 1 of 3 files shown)


SiTaggart:
So the reason yarn install fails is because the "fake" core bundle packages are build artifacts. We don't commit them. 

So a fresh git clone won't have those package.json files available, so yarn tries to unlink them by removing the entries in the lockfile

image

SiTaggart:
OK, seems like most things are starting to work. Outstanding issues include:

- [ ] likely need to finally fork https://github.com/preactjs/compressed-size-action for Package size check. It only support Yarn v1 and hard codes the `--frozen-lockfile` flag.
- [ ] Netlify build command is taking president from the Netlify UI, not the `netlify.toml` file. We need to not run `yarn bootstrap`, but current build rely on the UI command, hence trying to override it in a branch with out cause disruption for everyone trying to deploy a current build.
- [ ] Netlify cache doesn't see to cache the node_modules for yarn v3.

SiTaggart:
Definitely seeing some pretty hefty gains from this already 🎉 

before: 
image

after:
image

PR #1965, opened 2021-10-22
chore: update to react 17

shleewhite:
* Updates react, react-dom, @types/react to version 17
* Adds `yarn test:react-16` script
* Updates all tests that use Enzyme `mount` to use  react testing library `render` instead
* Adds `build_and_react16_tests` github check

**Contributing to Twilio**


gloriliale (approved):
Looks great! And the turnaround time for getting those Enzyme tests switched was so quick! 🎉  

Two questions: I thought that we wanted to keep tests running in React 16, but we changed them from using Enzyme bc it clashed with React 17. If that's correct, then what is it that we're keeping React 16 for? And why aren't we keeping it as a dev Dependency?

shleewhite:
Also open 

> Looks great! And the turnaround time for getting those Enzyme tests switched was so quick! 🎉
> 
> Two questions: I thought that we wanted to keep tests running in React 16, but we changed them from using Enzyme bc it clashed with React 17. If that's correct, then what is it that we're keeping React 16 for? And why aren't we keeping it as a dev Dependency?

@gloriliale Totally fair questions!

React 16 is not a dev dependency because we will build Paste with 17 moving forward, but it still is in the monorepo in a couple ways:
* React 16 is now [a peer dependency for most of our packages](https://github.com/twilio-labs/paste/blob/2bf793941015328a59afb76f2510002773f012e1/packages/paste-core/components/avatar/package.json#L37) so that consumers can use either version
* It is installed in [the root package as an alias](https://github.com/twilio-labs/paste/blob/2bf793941015328a59afb76f2510002773f012e1/package.json#L204) so it can be used for testing. Since people can still use paste in react 16, we should test it too bc there might be 16 specific bugs.

It's a way to move away from React 16 without breaking our consumers or forcing them to update right now.

.changeset/six-carrots-swim.md
```diff
@@ -0,0 +1,74 @@
...
+'@twilio-paste/reakit-library': patch
+'@twilio-paste/styling-library': patch
+'@twilio-paste/uid-library': patch
+'@twilio-paste/nextjs-template': patch
+'@twilio-paste/theme': patch
+'@twilio-paste/types': patch
+'@twilio-paste/utils': patch
+---
+
+Upgrade to React 17
```
SiTaggart, line 74: For this changelog entry, what would I want to read as a consumer of Paste?

packages/paste-core/components/combobox/__tests__/virtualization.spec.tsx
```diff
@@ -161,19 +160,19 @@ describe('Combobox - Virtualization', () => {
...
       expect(third).toHaveAttribute('aria-setsize', '5');
       expect(third).toHaveAttribute('aria-posinset', '3');
 
-      expect(fourth).toHaveAttribute('transform', 'translateY(80px)');
+      expect(fourth).toHaveAttribute('transform', 'translateY(116px)');
       expect(fourth).toHaveAttribute('aria-setsize', '5');
       expect(fourth).toHaveAttribute('aria-posinset', '4');
 
-      expect(fifth).toHaveAttribute('transform', 'translateY(116px)');
+      expect(fifth).toHaveAttribute('transform', 'translateY(152px)');
```
SiTaggart, line 175: Did we ever determine the cause of these calculation differences?

packages/paste-core/components/alert/__tests__/index.spec.tsx
```diff
@@ -15,75 +13,75 @@ describe('Alert', () => {
...
-          <Theme.Provider theme="console">
-            <Alert onDismiss={onDismissMock} variant="neutral">
-              This is an alert
-            </Alert>
-          </Theme.Provider>
-        )
-        .toJSON();
-      expect(tree).toMatchSnapshot();
+      const {asFragment} = render(
+        <Theme.Provider theme="console">
```
andipants12, line 78: NIT: We recently started using a testing theme, essentially mocking our theme so it is constant throughout time, and token changes do not require test updates.

To use it you would replace the `ThemeProvider` wrapper with:
```
<CustomizationProvider theme={TestTheme}>
    {/* the component being tested */}
</CustomizationProvider>
```

andipants12 (changes requested):
This is awesome 🙌  I left some suggested changes in the test files where we replaced enzyme with testing library.

packages/paste-core/components/combobox/__tests__/virtualization.spec.tsx
SiTaggart: So it's basically a corrected test result.

packages/paste-core/components/alert/__tests__/index.spec.tsx
```diff
@@ -15,75 +13,75 @@ describe('Alert', () => {
...
+      const {getByRole} = render(
         <Alert onDismiss={eventHandlerMock} variant="neutral">
           This is an alert
         </Alert>
       );
-      wrapper.find('button').simulate('click');
+
+      const button = getByRole('button');
+      fireEvent.click(button);
       expect(eventHandlerMock).toHaveBeenCalledTimes(1);
```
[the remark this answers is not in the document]
SiTaggart: The reason they are using `waitFor` in that example is because the request for the content is async, not the event

packages/paste-core/components/combobox/__tests__/virtualization.spec.tsx
shleewhite: Yeah, I think these are the correct values, they also match the values rendered in storybook.

packages/paste-core/components/alert/__tests__/index.spec.tsx
```diff
@@ -15,77 +13,57 @@ describe('Alert', () => {
   describe('Dismiss button', () => {
     it('Should add a dismiss button when onDismiss is passed as a function to call', () => {
       const eventHandlerMock: jest.Mock = jest.fn();
-      const wrapper: ReactWrapper = mount(
+      const {getByRole} = render(
```
andipants12, line 16: Not blocking, but you can call `screen.getByRole` instead of destructuing it from the render return.

andipants12 (approved):
LGTM 🥳

packages/paste-core/components/alert/__tests__/index.spec.tsx
```diff
@@ -1,7 +1,5 @@
 import * as React from 'react';
-import renderer from 'react-test-renderer';
-import {render, screen} from '@testing-library/react';
-import {ReactWrapper, mount} from 'enzyme';
+import {render, screen, fireEvent} from '@testing-library/react';
```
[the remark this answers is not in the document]
shleewhite: Updated!

.changeset/six-carrots-swim.md
shleewhite: Updated!

SiTaggart (approved):

PR #1970, opened 2021-10-25
chore(website): add customization interactive image slider

andipants12:
## Description
Replace static image of slider with an interactive one for medium and large screen sizes on the customization landing page.

- [x] build image slider on top of input range element
- [x] component should render a useful accessibility tree
- [x] the slider should render only for large and medium screen sizes


packages/paste-website/src/components/customization-landing-page/image-slider/ImageSlider.tsx
```diff
@@ -0,0 +1,185 @@
...
+        maxWidth="640px"
+        cursor={shouldMeasure ? 'grabbing' : 'grab'}
+        aria-label="Controls the width of the customizable image."
+        // @ts-expect-error "htmlFor" is an allowed attribute for Label elements.
+        htmlFor={uidSeed(INPUT_ID)}
+        color="colorTextWeaker"
+        _hover={{color: 'colorTextWeak'}}
+        onMouseDown={() => {
+          setShouldMeasure(true);
+          document.addEventListener('mouseup', handleMouseUp);
```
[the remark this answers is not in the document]
andipants12: Ah so you are saying calling `removeEventListener` when `isDragging` changes from `true` to `false` is not necessarily 1:1 because of batched state updates? So is it "safer" to remove here to ensure proper clean up?

Also curious why the `removeEventListener` call in the `useEffect` unmount logic was removed. I added that thinking that before the component unmounts we should do cleanup to be safe. Is that risky or just unncessary?

packages/paste-website/src/components/customization-landing-page/image-slider/ImageSlider.tsx
```diff
@@ -0,0 +1,174 @@
...
+  // Minimum change defines the absolute distance threshold for value to change.
+  // Calculated by half the width of the SVG, default is half of the absolute width.
+  const minimumChange = svgWidth ? svgWidth / 2 : DEFAULT_MIN_CHANGE;
+
+  // Width of SVG Clip used to conditionally show/hide part of the each image
+  // SVG clip is calculated from right to left, since the dynamic value here is the width.
+  // const clip = (value / MAX_VALUE) * containerWidth;
+  const clip = (value / MAX_VALUE) * containerWidth;
+
+  const handleMouseMove = React.useCallback(
```
andipants12, line 45: Could you explain the decision to memoize these but not the `onMouseDown` handler on line 103?

TheSisb: I would say it was a _lack_ of a decision :P

packages/paste-website/src/components/customization-landing-page/image-slider/ImageSlider.tsx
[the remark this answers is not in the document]
TheSisb: > Ah so you are saying calling removeEventListener when isDragging changes from true to false is not necessarily 1:1 because of batched state updates? So is it "safer" to remove here to ensure proper clean up?

Not quite. Ignore `isDragging` for the time being. Previously this code called `addEventListener` more frequently than `removeEventListener`.  `removeEventListener` was only being called in a `useEffect` (so on unmount, aka once), and `addEventListener` was called 0..n times every time someone `mouseDown`

> Also curious why the removeEventListener call in the useEffect unmount logic was removed. I added that thinking that before the component unmounts we should do cleanup to be safe. Is that risky or just unncessary?

Since we're now binding to the document.onMouseUp event, we don't need it.

andipants12:
@TheSisb I'm going to update the branch against main. LMK if you still have feedback that I should address. Thanks so much for helping out on Friday!

packages/paste-website/src/components/customization-landing-page/image-slider/ImageSlider.tsx
andipants12: Ha ok, I'll leave them for now since I feel like it's pretty negligible. I am going too remove the commented out line 42 also, but then that is my last update. 🥳

TheSisb (approved):

Issue #1935, opened 2021-10-11
barreled-to-unbarreled codemod does not fix barreled combobox imports


ghelton:
Thanks for helping make Paste better! Note: For urgent requests, reach out at **#help-design-system**

**Describe the bug**
Running barreled-to-unbarreled codemod reverts unbarreled combobox imports

**To Reproduce**
1. Run `npx @twilio-paste/codemods barreled-to-unbarreled src/`
2. Select `Typescript`
4. Results in no change

**Expected behavior**
```
-import { Combobox, ComboboxProps, UseComboboxState } from '@twilio-paste/core';
+import { Combobox, ComboboxProps, UseComboboxState } from '@twilio-paste/core/combobox';
```

SiTaggart:
Thanks @ghelton, will look into it.

TheSisb:
From my tests, the library _worked_ but it was missing a bunch of new components since we hadn't updated it in a while. I made a change so that we notice to update it in the future and pushed a PR with the current updates. Should land soon. Thanks for letting us know!

ghelton:
Awesome thanks!

PR #1989, opened 2021-11-01
chore(codemods): update mappings for new Paste components

TheSisb:
Updates the mappings.json file for newly added Paste components, so this properly unbarrels all the components now.
 
I also remove the mapping folder from .gitignore, so we see the change and remember to version this library in the future.

Fixes https://github.com/twilio-labs/paste/issues/1935


SiTaggart (approved):
Hmm there is definitely a few places now where we need to keep up to date some form log log of whats in core.

There's a gitignore in core, theres a codesandbox config, there's this list of files, there's the repo cache for local package name to file path mapping, this cache.

Something we might what to centralise in the future so it's co-ordinated

TheSisb:
> Hmm there is definitely a few places now where we need to keep up to date some form log log of whats in core.

I had the same thought while fixing this. All the current lists are ever so _slightly_ different, but I think there's value in a centralized _thing_.  Will keep noodling on it.

andipants12 (approved):

PR #1990, opened 2021-11-01
fix(data grid): make 'striped' prop work

TheSisb:
Previously the `striped` prop wasn't doing anything. This fixes the `striped` prop and updates the docs so it works correctly on Data Grid.

 Updated the Loading story to use `striped={false}` so we can VRT it on one of our existing examples.

### DIFF
--- a/packages/paste-core/components/data-grid/src/DataGrid.tsx
+++ b/packages/paste-core/components/data-grid/src/DataGrid.tsx
@@ -24,133 +24,137 @@ export interface DataGridProps extends TableProps {
  * DataGrid wrapper component.
  *
  * @param {string} aria-label - for screen readers
+ * @param {boolean} striped - zebra striping on table rows
  * @param {string} element - customization element
  */
-export const DataGrid = React.forwardRef<HTMLTableElement, DataGridProps>(({element = 'DATA_GRID', ...props}, ref) => {
-  const dataGridId = `data-grid-${useUID()}`;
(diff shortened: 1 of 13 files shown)


packages/paste-core/components/data-grid/src/DataGrid.tsx
```diff
@@ -24,133 +24,137 @@ export interface DataGridProps extends TableProps {
...
  * @param {string} aria-label - for screen readers
+ * @param {boolean} striped - zebra striping on table rows
  * @param {string} element - customization element
  */
-export const DataGrid = React.forwardRef<HTMLTableElement, DataGridProps>(({element = 'DATA_GRID', ...props}, ref) => {
-  const dataGridId = `data-grid-${useUID()}`;
-  const lastFocusedElement = React.useRef<Element | null>(null);
-  const compositeState = useCompositeState({unstable_virtual: false});
-  const [actionable, setActionable] = React.useState<boolean>(false);
+export const DataGrid = React.forwardRef<HTMLTableElement, DataGridProps>(
```
TheSisb, line 30: hide whistespace changes when viewing this file
image

shleewhite (approved):
LGTM

andipants12 (approved):

katieporter:
Little bit of a nit pick, but would it be hard to make the `striped` prop default to `false` rather than `true` to match Table?

PR #1999, opened 2021-11-03
chore: streamline CI workflows

SiTaggart:
Streamline the build process in CI:

- reorganise workflows around events. `on_pull_request`, `on_merge_to_main` and `on_release_publish`.
- each job that requires a build, now is dependent on a single build for each workflow event. (except package size which does it's own builds)
- each build is uploaded as an artifact and shared between subsequent jobs.
- reduces the change of a failed build, but keeps parallel jobs to make debugging easier.
- renamed a few jobs with nicer names
- requires a change in the require CI checks for Pull Requests going forward


.github/workflows/on_merge_to_main.yml
```diff
@@ -0,0 +1,181 @@
...
+        id: node_modules_cache_id
+        with:
+          path: |
+            node_modules
+            */*/node_modules
+          key: ${{ runner.os }}-${{ hashFiles('**/yarn.lock') }}
+
+      - name: Install Dependencies
+        if: steps.yarn_cache_id.outputs.cache-hit != 'true' || steps.node_modules_cache_id.outputs.cache-hit != 'true'
+        run: yarn install --immutable
```
TheSisb, line 51: Is it possible to move this "setup" into a `job` that gets run within the other jobs, to avoid the duplication and to make these files easier to reason about?
https://docs.github.com/en/actions/learn-github-actions/reusing-workflows#creating-a-reusable-workflow

.github/workflows/on_merge_to_main.yml
```diff
@@ -0,0 +1,181 @@
+name: Merge to main
```
TheSisb, line 1: Love this refactor around events

.github/workflows/on_pull_request.yml
```diff
@@ -1,9 +1,12 @@
 name: PR Checks
 
-on: pull_request
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
```
TheSisb, line 5: Should we move `ready_to_review` into another file so that we only run the jobs that weren't ran on draft?

.github/workflows/on_merge_to_main.yml
```diff
@@ -0,0 +1,181 @@
...
+          version: yarn version
+          publish: yarn release
+          commit: "chore(release): version packages"
+        env:
+          GITHUB_TOKEN: ${{ secrets.GH_TOKEN }}
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+          # due to this bug in yarn https://github.com/yarnpkg/berry/issues/3606 we need to manually turn off immutable installs to update the lockfile
+          YARN_ENABLE_IMMUTABLE_INSTALLS: false
+
+  chromatic:
```
TheSisb, line 129: Why do we run chromatic on main again?

.github/workflows/on_pull_request.yml
```diff
@@ -212,24 +288,152 @@ jobs:
...
+        uses: actions/download-artifact@v2
+        with:
+          name: compiled-js-and-types
+          path: packages/
 
       - name: Run prettier
         run: yarn prettier
 
-  build_and_danger:
+  vrt:
```
TheSisb, line 300: Should we call this `chromatic` to match the `on merge to main` workflow?

.github/workflows/on_pull_request_package_size.yml
```diff
@@ -1,9 +1,12 @@
-name: Package Size
+name: Package size
 
-on: pull_request
+on:
+  pull_request:
+    types: [opened, synchronize, reopened, ready_for_review]
```
TheSisb, line 5: Maybe we can remove `ready_for_review` here since nothing should change in the code between draft -> rdy that isn't already caught

package.json
```diff
@@ -55,7 +55,7 @@
     "build:props": "lerna run build:props --ignore @twilio-paste/website --concurrency 2",
     "pre-push": "concurrently \"yarn:lint\" \"yarn:test\" \"yarn:prettier\" \"yarn:type-check\"",
     "prerelease": "yarn build && yarn lint && yarn test",
-    "release": "yarn prerelease && yarn changeset publish",
+    "release": "yarn changeset publish",
```
TheSisb, line 58: Can I get some info around this change please?

.github/workflows/on_merge_to_main.yml
SiTaggart: OMG I was looking for just this!

.github/workflows/on_merge_to_main.yml
SiTaggart: To create the baseline images for branches to compare against.

.github/workflows/on_pull_request_package_size.yml
SiTaggart: I think I'm going to just scope it to on: `pull_request` and not care about types.

package.json
SiTaggart: Yup, so here's my logic. On merge to main two things need to happen:

1. Either version or publish on the packages
2. Run chromatic to create a new main branch baseline

To do either, you must build. There shouldn't be a need to build more than once, that's wasteful. If the build fails neither chromatic nor versioning is worth running.

If the build succeeds, but versioning fails, we still need a new chromatic baseline otherwise active branches are out of syn

If the build succeeds, but chromatic fails, we still need versioning to run.

So to keep all three independent, the build needs to run once, and be shared. The problem is if the release must run `prerelease` that runs the build again. The easiest way is to make release independent, and make prerelease the first job to both versioning and chromatic.

.github/workflows/on_pull_request.yml
SiTaggart: I can try it. `ready_for_review` is a special event. My worry is, what I tend to do it push new code or a rebase, and then make it ready for review, and I'm not sure if that would cancel all the other jobs that are legit needed. But I won't know until I try it.

.github/workflows/on_merge_to_main.yml
SiTaggart: OK, so this is far less useful than it seems. 

You can't share parts of a workflow. Only an entire job. Meaning you can't extend a shared workflow by adding steps to it. You can only use a shared workflow as a top level job, and that job cannot have steps. And you can't call another shared workflow from a shared workflow. 

You might be able to do that with a custom action maybe? But that seems like more work than the scope of this fix

.github/workflows/on_pull_request.yml
SiTaggart: OK, so this won't necessarily work either. Mainly because you end up needing two different named checks for Chromatic because it has to live in its own workflow file with its own name. 

`ready_for_review` only ever fires  when you hit that button, so you need a chromatic job for `opened, synchronize, reopened` which are the default events for every other thing you do in a PR.

TheSisb (approved):

PR #2007, opened 2021-11-04
chore(cypress): add roadmap page baseline test

andipants12:
## Description
Add baseline test for the Roadmap page to verify that it was built correctly and renders with populated release tables.

### Summary of changes
- [x] add test that verifies the roadmap page renders with correct expected release tables
- [x] separate e2e tests from the site-wide integration tests for more clarity


cypress/integration/e2e/roadmap-page.spec.ts
```diff
@@ -0,0 +1,38 @@
+describe('Roadmap Page', () => {
+  beforeEach(() => {
+    cy.server();
+    cy.route('/page-data/roadmap/page-data.json').as('roadmapDataRequest');
+    cy.visit('/roadmap');
+    cy.wait('@roadmapDataRequest');
+
+    return cy
+      .get<{response: {body: {result: {data: {BPM: Record<string, any>[]; Roadmap: Record<string, any>[]}}}}}>(
```
andipants12, line 9: this type is super gross... going to move it to a central `types.ts` file.

cypress/integration/e2e/roadmap-page.spec.ts
```diff
@@ -0,0 +1,38 @@
+describe('Roadmap Page', () => {
+  beforeEach(() => {
+    cy.server();
+    cy.route('/page-data/roadmap/page-data.json').as('roadmapDataRequest');
```
andipants12, line 4: I'm using the interceptor here because AFAIK:
1. if all went well at build time, this route should respond with the JSON data used to generate the static page
2. So we can check that it's not empty (e.g. there was a problem with the query and it did not return the expected content), as well as that we are rendering the expected count of release tables, and release names.

cypress/integration/e2e/roadmap-page.spec.ts
TheSisb: Can just define it at the start of this file until we use it more widely.

PR #2011, opened 2021-11-05
fix(popover): remove popovercontainer Box wrapper

TheSisb:
Fixes https://github.com/twilio-labs/paste/discussions/1978 and https://github.com/twilio-labs/paste/discussions/2006 and https://github.com/twilio-labs/paste/discussions/1870

Removes the wrapping Box in PopoverContainer. PopoverContainer is simply a context provider, like Theme.Provider, and that wrapping Box just added styling issues.

### DIFF
--- a/packages/paste-core/components/popover/src/PopoverContainer.tsx
+++ b/packages/paste-core/components/popover/src/PopoverContainer.tsx
@@ -1,5 +1,4 @@
 import * as React from 'react';
-import {Box} from '@twilio-paste/box';
 import type {
   NonModalDialogPrimitiveStateReturn,
   NonModalDialogPrimitivePopoverInitialState,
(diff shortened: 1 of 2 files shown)


packages/paste-core/components/popover/src/PopoverContainer.tsx
```diff
@@ -16,17 +15,16 @@ export interface PopoverContainerProps extends NonModalDialogPrimitivePopoverIni
   state?: PopoverStateReturn;
 }
 
-const BasePopoverContainer = React.forwardRef<HTMLDivElement, PopoverContainerProps>(
-  ({children, gutter, placement, state, ...initialState}, ref) => {
-    const popover = state || useNonModalDialogPrimitiveState({gutter, modal: true, placement, ...initialState});
-    return (
-      <PopoverContext.Provider value={popover}>
-        <Box ref={ref}>{children}</Box>
```
nkrantz, line 24: What does removing the `ref` here do exactly?

nkrantz (approved):

packages/paste-core/components/popover/src/PopoverContainer.tsx
TheSisb: All our components forward refs, so I think it was just doing what other components do. But this isn't a standard component, it's similar to Theme.Provider and is just a React Context.  

Removing or adding ref would ordinarily breaking change, but since this isn't a standard component and ref here is arguably useless, and since reporting shows no one is passing ref to this --- it should be harmless to remove and mark as a minor change.

shleewhite (approved):

SiTaggart (approved):
Should be pretty low impact, easy fix

PR #2012, opened 2021-11-05
fix(website): broken roadmap link

TheSisb:
This PR fixes broken links on the Roadmap page due to a missing `#` character that would route the user to a broken page rather than a page hash.

I also improve the broken link checker, providing the calling route to the HTTP header object so we can see from which file the error originated:
image
Additionally, I removed the contents of the IGNORE_LIST in the broken link checker since that list is now stale.

### DIFF
--- a/cypress/integration/e2e/link-checker/index.spec.ts
+++ b/cypress/integration/e2e/link-checker/index.spec.ts
@@ -10,16 +10,7 @@
  */
 
 // FIXME: these are all broken links we should do something about
-const IGNORE_LIST = [
-  // Left these in there because they're being called from the sidebar nav.
-  // That will need to be refactored to pull from AirTable instead of packages.
-  'primitives/sibling-box',
-  'components/badge',
-  'components/display-pill-group',
-  'components/form-pill-group',
-  'patterns/data-export', // this randomly started failing and the page doesn't exist.
-  '/__/', // I don't know where this is being picked up
-];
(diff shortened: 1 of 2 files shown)


cypress/integration/e2e/link-checker/index.spec.ts
```diff
@@ -10,16 +10,7 @@
...
-  // Left these in there because they're being called from the sidebar nav.
-  // That will need to be refactored to pull from AirTable instead of packages.
-  'primitives/sibling-box',
-  'components/badge',
-  'components/display-pill-group',
-  'components/form-pill-group',
-  'patterns/data-export', // this randomly started failing and the page doesn't exist.
-  '/__/', // I don't know where this is being picked up
-];
+const IGNORE_LIST = [];
```
nkrantz, line 13: Side note: would be good for us to get in the practice of including "add link to the ignore list" to ticket AC of all new components/packages, before the docs are written

nkrantz (approved):

shleewhite (approved):

cypress/integration/e2e/link-checker/index.spec.ts
SiTaggart: Maybe a better approach would be to fix the logic on the docsite that creates a broken link in the first place

SiTaggart (approved):

PR #2020, opened 2021-11-09
fix(pill-group): fix how pill groups wrap

TheSisb:
Our pill groups weren't wrapping when they overflowed, so this small fix let's them wrap and adds a small gap between rows.


shleewhite (approved):

SiTaggart (approved):
Loving them gap properties 🎉

PR #2023, opened 2021-11-10
fix: only prevent chromatic running on version merge commits

SiTaggart:
Was overly protective on the merge to main workflow steps. We actually only want to prevent versioning merge commits running Chromatic. 

The other steps should still run.


andipants12 (approved):
I'm not super well versed in these workflows, but lgtm!

TheSisb (approved):

PR #2027, opened 2021-11-10
chore: use ultra-runner for builds

TheSisb:
Experimenting with ultra-runner for builds


SiTaggart:
Same error, different runner 
image

PR #2029, opened 2021-11-11
Badge ref fix

shleewhite:
If the Badge is focusable, the ref is passed to the wrapper component instead of the span.

Also export the BadgeProps type.

**Contributing to Twilio**

### DIFF
--- a/packages/paste-core/components/badge/src/utils.ts
+++ b/packages/paste-core/components/badge/src/utils.ts
@@ -1,66 +1,11 @@
-import type {TextColorOptions, BackgroundColorOptions} from '@twilio-paste/style-props';
+import {safelySpreadBoxProps} from '@twilio-paste/box';
+import type {BadgeProps} from './types';
 
-import type {BadgeVariants, BadgeProps} from './types';
+export const isFocusableElement = ({as}: Partial<BadgeProps>): boolean => as !== 'span';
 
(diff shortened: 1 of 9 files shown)


packages/paste-core/components/badge/src/utils.ts
```diff
@@ -1,66 +1,11 @@
-import type {TextColorOptions, BackgroundColorOptions} from '@twilio-paste/style-props';
+import {safelySpreadBoxProps} from '@twilio-paste/box';
+import type {BadgeProps} from './types';
 
-import type {BadgeVariants, BadgeProps} from './types';
+export const isFocusableElement = ({as}: Partial<BadgeProps>): boolean => as !== 'span';
```
TheSisb, line 4: Hm in case we add new variants here (i.e. div) can we prefer to be explicit and do `as === 'a' || as === 'button'` please?  The function name is `isFocusableElement` but this function expects knowledge of another component's internals. Best to make this function stand on its own for future proofyness

TheSisb (approved):

packages/paste-core/components/badge/src/utils.ts
shleewhite: Sure! Fixed.

nkrantz (approved):

TheSisb (approved):

PR #2035, opened 2021-11-12
chore: log build errors to help debug intermittant issue

SiTaggart:
I'm not sure if this will actually be any better, but if possible log the error we get in esBuild to help identify the root of this intermittent build bug.


zahnster (approved):
🚢

shleewhite (approved):

PR #2040, opened 2021-11-15
chore(cypress): add component test commands and script for test files

andipants12:
**Contributing to Twilio**


cypress/support/dual-commands.ts
```diff
@@ -0,0 +1,7 @@
+/**
```
andipants12, line 1: Just a place holder with a doc description and link to cypress documentation on dual commands.

cypress/support/override-commands.ts
```diff
@@ -0,0 +1,7 @@
+/**
```
andipants12, line 1: Just a place holder with a doc description and link to cypress documentation on overrides.

shleewhite (approved):
LGTM

cypress/integration/e2e/components/index.mdx.spec.ts
```diff
@@ -0,0 +1,14 @@
...
+    cy.visit('/components/index.mdx');
+  });
+
+  it('should render the index.mdx component page correctly', () => {
+    cy.pageHeaderShouldBeVisible('Index Mdx');
+
+    cy.checkInPageNavigationLinks();
+    cy.checkPageAside();
+    cy.checkLivePreviews();
+    cy.checkChangelogRevealer();
```
TheSisb, line 12: The fact that this function doesn't fail on the overview page is concerning... There's no changelog on this page.

cypress/integration/e2e/components/index.mdx.spec.ts
```diff
@@ -0,0 +1,14 @@
+describe('Index.mdx component documentation page', () => {
+  beforeEach(() => {
+    cy.visit('/components/index.mdx');
```
TheSisb, line 3: This isn't a valid route. Did you mean `/components`?

cypress/support/parent-commands.ts
```diff
@@ -0,0 +1,168 @@
...
+  cy.get('@overviewTableHeaders').each((header, index) => {
+    // using include here to do a loose match of headers
+    cy.wrap(header).should('include.text', headers[index]);
+  });
+
+  // verify at least one row is rendered.
+  cy.get('@tableRows').its('length').should('be.greaterThan', 0);
+});
+
+Cypress.Commands.add('checkInPageNavigationLinks', () => {
```
TheSisb, line 110: I'm not a big fan of globals in code and kind of wish these were regular, imported functions. I feel as if these global commands shouldn't have page specific content, and instead should be high level commands like checking the URL or messing with localstorage. JS things, not DOM things basically.

cypress/support/parent-commands.ts
```diff
@@ -0,0 +1,168 @@
...
+// Cypress.Commands.add('checkComponentHeader', () => {});
+
+Cypress.Commands.add('checkPageAside', () => {
+  cy.getDocsPageContentArea().getInFixedContainer('[data-cy="page-aside"]').as('pageAside');
+  cy.get('@pageAside').find('[data-cy="table-of-contents"]').shouldBeVisible();
+  cy.get('@pageAside').contains('button', 'Rate this page').shouldBeVisible();
+  cy.get('@pageAside').find('[data-cy="page-aside-anchor"]').shouldBeVisible();
+});
+
+Cypress.Commands.add('checkLivePreviews', () => {
```
TheSisb, line 139: Some JSDocs on all of these would be really handy, this seems like the meat and potatos file.

cypress/integration/e2e/components/index.mdx.spec.ts
andipants12: It actually doesn't pass. The test suite just doesn't run the rest of the commands if the before/beforeEach hook fails, which is intended behavior.

Screen Shot 2021-11-16 at 12 17 44 PM

I just fixed the visit command and the whole test is failing as expected. This was just generated by the script, and isn't actually the test for the overview page. going to delete it.

cypress/support/parent-commands.ts
andipants12: There are JSDocs above, with the TS declarations.

andipants12: If you look at the branch in an editor you should see the intellisense.

tools/utils/cypress-test-generators/TEMPLATE.js
```diff
@@ -0,0 +1,19 @@
...
+  });
+
+  it('should render the ${componentName} component page correctly', () => {
+    cy.pageHeaderShouldBeVisible('${startCase(componentName)}');
+
+    cy.checkInPageNavigationLinks();
+    cy.checkPageAside();
+    cy.checkLivePreviews();
+    cy.checkChangelogRevealer();
+  });
```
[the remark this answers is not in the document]
andipants12: This is definitely possible, but I feel pretty strongly that we should favor individual test spec files. That tends to obfuscate the the test condition and can make it difficult to track down the root of an issue if tests are flakey or do not pass (as cypress does not necessarily log specifically where the error occurred). Additionally, with a single test like this, if one page has a flakey test and fails, all tests cascade to failure.

We could make this abstraction in the future, but for now I think it makes sense to hold off.

cypress/support/parent-commands.ts
andipants12: This is a pretty standard pattern with cypress. These custom commands could also be functions, but I wrote them as custom commands since that's what was discussed in the kickoff. Happy to iterate on that, but for the purpose of these tickets that are in flight I'd prefer to stick to the current methods.

TheSisb (approved):

PR #2041, opened 2021-11-15
feat(spacing-update): update spacing for site docs

zahnster:
* Updates spacing and sizing properties on the docs site to match updated comps.

### DIFF
--- a/packages/paste-website/src/components/paste-mdx-provider/index.tsx
+++ b/packages/paste-website/src/components/paste-mdx-provider/index.tsx
@@ -91,7 +91,7 @@ const MDXPoviderComponents = {
   img: (props: React.ComponentProps<'img'>): React.ReactElement => <img {...props} />, // eslint-disable-line jsx-a11y/alt-text
   sup: (props: React.ComponentProps<'sup'>): React.ReactElement => <StyledSup {...props} />,
   content: (props: Omit<React.ComponentProps<'div'>, 'color' | 'ref'>): React.ReactElement => (
-    <Box as="div" maxWidth="size70" {...props} />
+    <Box as="div" maxWidth="size70" minWidth="0" {...props} />
   ),
   contentwrapper: (props: Omit<React.ComponentProps<'div'>, 'color' | 'ref'>): React.ReactElement => (
     <Box as="div" display={['block', 'block', 'flex']} {...props} />
--- a/packages/paste-website/src/components/paste-mdx-provider/index.tsx
+++ b/packages/paste-website/src/components/paste-mdx-provider/index.tsx
@@ -91,7 +91,7 @@ const MDXPoviderComponents = {
   img: (props: React.ComponentProps<'img'>): React.ReactElement => <img {...props} />, // eslint-disable-line jsx-a11y/alt-text
   sup: (props: React.ComponentProps<'sup'>): React.ReactElement => <StyledSup {...props} />,
   content: (props: Omit<React.ComponentProps<'div'>, 'color' | 'ref'>): React.ReactElement => (
-    <Box as="div" minWidth="0" {...props} />
+    <Box as="div" {...props} />
   ),
   contentwrapper: (props: Omit<React.ComponentProps<'div'>, 'color' | 'ref'>): React.ReactElement => (
     <Box as="div" display={['block', 'block', 'flex']} {...props} />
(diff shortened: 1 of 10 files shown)


shleewhite (approved):

packages/paste-website/src/layouts/DefaultLayout.tsx
```diff
@@ -58,8 +58,10 @@ const DefaultLayout: React.FC<DefaultLayoutProps> = ({children, pageContext, loc
       </Box>
       <Box
         paddingTop={['space40', 'space40', 'space100']}
-        paddingX={['space50', 'space70', 'space200']}
+        paddingX={['space70', 'space200', 'space200']}
         backgroundColor="colorBackgroundBody"
+        maxWidth="73.5rem" // custom spacing value: size100(64rem), + (space200(4.75rem) * 2)
```
loreina, line 63: quick maths

packages/paste-website/src/components/paste-mdx-provider/index.tsx
```diff
@@ -83,13 +83,13 @@ const MDXPoviderComponents = {
   strong: (props: React.ComponentProps<'strong'>): React.ReactElement => <strong {...props} />,
   del: (props: React.ComponentProps<'del'>): React.ReactElement => <del {...props} />,
   hr: (props: SeparatorProps): React.ReactElement => (
-    <Separator {...props} orientation="horizontal" verticalSpacing="space70" />
+    <Separator {...props} orientation="horizontal" verticalSpacing="space100" />
```
loreina, line 86: I updated the Figma specs so we're using `space100` on top/bottom of the Separator

loreina (approved):

packages/paste-website/src/layouts/DefaultLayout.tsx
katieporter: What's this used for? We tried to use sizing tokens for everything but maybe we missed something?

katieporter:
Ahhh this update looks so good! 😭  No comments from me on the overall updates, though I did notice a couple of individual instances that look a wee bit off:
- Could all of the overview pages (Patterns, Components, Primitives, Layout) span the full width of the content area since they don't have sidebars (like how the Token List page works)?
- The Articles section is also a little funny now...but I am thinking about potentially do an actual design pass during Tweek Week. 🤔  For now, could the Articles overview span the full-width of the page, and could we add a sidebar to Nora's blog post so it doesn't have the funky white space (I don't want to make the article span the full width since the character count per line would start to get hard to read).

packages/paste-website/src/layouts/DefaultLayout.tsx
zahnster: @katieporter padding will take away from an element's available width. In the comps, 1024px was declared for the width of the content, but we also have padding on that element, so we have to add it to the total width for the element.

The rendering sizes for this ends up like this:

Screen Shot 2021-11-18 at 1 39 13 PM

katieporter: Ahhh gotcha, thanks for explaining. I didn't realize we were using padding. Can we use margins instead so we can stick to using token values, or is there a specific reason we chose to pad it?

zahnster: We use auto margins to center the element, so we're sort of restricted there.

We could change the box model sizing for the element. This might actually be an appropriate use for the property. 🤔

katieporter:
It looks like the width of the content area might not be flexible—on smaller screen sizes, it maintains the same width which pushes the sidebar out of the viewport:
image

TheSisb:
re katie's observed bug, this is only happening on some pages. I've debugged it down to our `code` html elements. We need to give them a `white-space: pre-wrap;` CSS rule

image

zahnster:
> re katie's observed bug, this is only happening on some pages. I've debugged it down to our `code` html elements. We need to give them a `white-space: pre-wrap;` CSS rule

I tried implementing this and it did resolve for cases where CodeBlock was causing the overflow, but it did not resolve other issues around CodeSandbox embeds and the ReactLivePreview components causing overflow stretch.

I decided to take a slightly different approach and re-introduce the max-width for the main content. For pages that are single col, they don't need either <contentwrapper> or <content> so I removed them to allow them to go max-width.

katieporter (approved):

PR #1956, opened 2021-10-19
fix (website): remove ScrollAnchorIntoView

shleewhite:
Updates the website layout so we can let the browser handle in-page links instead of using ScrollAnchorIntoView.

**Contributing to Twilio**

> All third-party contributors acknowledge that any contributions they provide will be made under the same open-source license that the open-source project is provided under.

- [ ] I acknowledge that all my contributions will be made under the project's license.

PR #2042, opened 2021-11-16
feat(website): Update TableOfContents styles

shleewhite:
Updates the TableOfContents component to the latest designs. Separates out the sub-components into their own files.

**Contributing to Twilio**

### DIFF
--- a/packages/paste-website/src/components/shortcodes/PageAside.tsx
+++ b/packages/paste-website/src/components/shortcodes/PageAside.tsx
@@ -2,7 +2,6 @@ import * as React from 'react';
 import {Box} from '@twilio-paste/box';
 import {TableOfContents} from './table-of-contents';
 import {FeedbackPopover} from './feedback-popover';
-import {STICKY_COLUMN_OFFSET} from '../../constants';
 
 interface PageAsideProps {
   data: {
@@ -27,10 +26,10 @@ const PageAside: React.FC<PageAsideProps> = ({data}) => {
       marginLeft="space140"
       minWidth="size20"
       id="page-aside"
-      data-cy="page-aside"
       display={['none', 'none', 'block']}
+      data-cy="page-aside"
     >
-      <Box position="sticky" top={STICKY_COLUMN_OFFSET}>
+      <Box position="sticky" top="space130">
         <FeedbackPopover />
         <TableOfContents headings={data.headings} />
       </Box>
--- a/packages/paste-website/src/components/Roadmap/RoadmapAside.tsx
+++ b/packages/paste-website/src/components/Roadmap/RoadmapAside.tsx
@@ -3,7 +3,8 @@ import {Box} from '@twilio-paste/box';
 import {Anchor} from '@twilio-paste/anchor';
 import {FeedbackPopover} from '../shortcodes/feedback-popover';
 import {STICKY_COLUMN_OFFSET} from '../../constants';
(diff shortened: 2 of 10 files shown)


loreina (commented):
This is some sweet stuff! 🎉 

One thing I noticed is that the list item hover area is the entire box but I can only click the text to redirect to that section. Could we either update the hover area to only the text or have the entire box clickable to redirect? I would lean towards the second one as that's how we have it setup in the side navigation
Kapture 2021-11-16 at 11 55 51

katieporter:
Weee this looks amazing! I just noticed one thing in addition to Loreina's comment, and this is actually a bug with the existing Table of Contents, so if it's out of scope to fix, that's okay:

What happens is, when there is a section that houses a lot of H4–H6 headings (which don't appear in the TOC nav), the TOC nav stops highlighting the parent section. A good example of this is on `/components/combobox#examples`. Once you start scrolling through the examples, there is no longer an active nav item in the TOC.

Is is possible to keep the parent item selected in cases like that? If that's way out of scope to fix right now, could we make a followup ticket to address it?

packages/paste-website/src/components/shortcodes/PageAside.tsx
```diff
@@ -23,7 +22,7 @@ interface PageAsideProps {
 const PageAside: React.FC<PageAsideProps> = ({data}) => {
   return (
     <Box order={2} marginLeft="space140" minWidth="size20" id="page-aside" display={['none', 'none', 'block']}>
-      <Box position="sticky" top={STICKY_COLUMN_OFFSET}>
+      <Box position="sticky" top="space130">
```
gloriliale, line 25: Was there a reason we had this as a constant before?

shleewhite: Yeah, I think the way the flex boxes were set up before [the scroll fix pr](https://github.com/twilio-labs/paste/pull/1956)  - top had to account for the header/banner height otherwise it would go below them. Now, its structured better so the value doesn't need to do that.

shleewhite:
> Weee this looks amazing! I just noticed one thing in addition to Loreina's comment, and this is actually a bug with the existing Table of Contents, so if it's out of scope to fix, that's okay:
> 
> What happens is, when there is a section that houses a lot of H4–H6 headings (which don't appear in the TOC nav), the TOC nav stops highlighting the parent section. A good example of this is on `/components/combobox#examples`. Once you start scrolling through the examples, there is no longer an active nav item in the TOC.
> 
> Is is possible to keep the parent item selected in cases like that? If that's way out of scope to fix right now, could we make a followup ticket to address it?

@katieporter I'll make a new ticket for that now!

shleewhite:
> This is some sweet stuff! 🎉
> 
> One thing I noticed is that the list item hover area is the entire box but I can only click the text to redirect to that section. Could we either update the hover area to only the text or have the entire box clickable to redirect? I would lean towards the second one as that's how we have it setup in the side navigation 

@loreina I fixed it! Now the whole thing is clickable.

zahnster (approved):
This looks great, and yay for more VRT stories!

nkrantz (approved):
This is some good stuff!

packages/paste-website/src/components/shortcodes/PageAside.tsx
gloriliale: okay cool! thanks for explaining!

TheSisb (approved):

PR #2044, opened 2021-11-16
fix(react17-testing): update testing library

zahnster:
* Poking at the React17 test failure
* Updating testing library deps


packages/paste-core/components/data-grid/__tests__/index.spec.tsx
```diff
@@ -35,22 +36,28 @@ describe('Data Grid', () => {
...
       const {getByTestId} = render(<ComposableCellsDataGrid />);
+      const wrapper = getByTestId('data-grid');
       const headerCell = getByTestId('header-1');
-      const firstInputCell = getByTestId('input-0-0')?.parentElement?.parentElement;
+      const firstRowFirstInputCell = getByTestId('input-0-0');
+      const firstRowSecondInputCell = getByTestId('input-0-1');
+      const secondRowFirstInputCell = getByTestId('input-1-0');
+      const firstInputCell = firstRowFirstInputCell?.parentElement?.parentElement;
+
       if (firstInputCell == null) {
```
andipants12, line 56: Not blocking, just info: I think this line could be removed if on lines 51, 52, 53 we modified them to look liike:

```
it('has proper keyboard navigation behavior', async () => {
    const firstRowFirstInputCell = await screen.findByTestId('input-0-0');
});
 ```

[More context in this article.](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-waitfor-to-wait-for-elements-that-can-be-queried-with-find)

File: yarn.lock [hunk 2]
```diff
@@ -6398,6 +6398,19 @@ __metadata:
   languageName: node
   linkType: hard
 
+"@jest/types@npm:^27.2.5":
```
andipants12, line 6401: Did we upgrade jest in this PR as well? If not, should we roll back this to the 26 version?

TheSisb (approved):
Andi's comment raises concerns over a mismatched yarn lockfile, but otherwise LGTM. Good job digging into this @andipants12 and thanks for the PR @zahnster

zahnster:
Regarding the Lockfile update, there's an upstream dependency that pulls @jest/types@v27 in.
@testing-library/react -> @testing-library/dom -> pretty-format -> @jest/types

You can see a dependency visualization here:
https://npm.anvaka.com/#/view/2d/%2540testing-library%252Freact

packages/paste-core/components/data-grid/__tests__/index.spec.tsx
```diff
@@ -35,22 +36,24 @@ describe('Data Grid', () => {
...
+    it('has proper keyboard navigation behavior', async () => {
       const {getByTestId} = render(<ComposableCellsDataGrid />);
+      const wrapper = getByTestId('data-grid');
       const headerCell = getByTestId('header-1');
-      const firstInputCell = getByTestId('input-0-0')?.parentElement?.parentElement;
-      if (firstInputCell == null) {
-        throw new Error('cannot find firstInputCell');
-      }
+      const firstRowFirstInputCell = await getByTestId('input-0-0');
+      const firstRowSecondInputCell = await getByTestId('input-0-1');
```
andipants12, line 52:
```suggestion
      const firstRowSecondInputCell = await findByTestId('input-0-1');
```

andipants12:
```suggestion
      const firstRowSecondInputCell = await findByTestId('input-0-1');
```

packages/paste-core/components/data-grid/__tests__/index.spec.tsx
```diff
@@ -35,22 +36,24 @@ describe('Data Grid', () => {
...
+      const secondRowFirstInputCell = await getByTestId('input-1-0');
```
andipants12, line 53:
```suggestion
      const secondRowFirstInputCell = await findByTestId('input-1-0');
```

andipants12 (approved):
I left suggested changes where we should probably use `findBy` instead of `getBy`, but other than that lgtm. Approving so I don't end up blocking this fix. 🙌

zahnster:
> I left suggested changes where we should probably use `findBy` instead of `getBy`, but other than that lgtm. Approving so I don't end up blocking this fix. 🙌

Curious on the reasons for this. I do see 752 matches in our codebase for 'getByTestId', I'm not seeing any matches for 'findByTestId'. Seems like they all alias to the same thing? https://testing-library.com/docs/queries/bytestid/

We can note this for a possible future improvement, but I am inclined to merge this now to help with other incoming PRs.

PR #2046, opened 2021-11-16
fix(website): component import codeblock missing quotes

TheSisb:
Fixing inconsistent and missing quotes on website's pattern import codeblocks. If copy pasted (which is the intended use), can be confusing to debug.


nkrantz (approved):
Nice improvement

zahnster (approved):
👍🏼

PR #2051, opened 2021-11-16
chore: upgrade esbuild

TheSisb:
Upgrading esbuild to the latest version. Noticed [this thread](https://github.com/evanw/esbuild/issues/359) that might relate to our flaky builds, so hoping that this upgrade will fix it.

v10.2 is the latest esbuild version that doesn't break our builds.


andipants12 (approved):
LGTM 👍

PR #2056, opened 2021-11-17
chore(cypress): turn on new cypress tests in ci

andipants12:
**Contributing to Twilio**

### DIFF
--- a/cypress/support/parent-commands.ts
+++ b/cypress/support/parent-commands.ts
@@ -79,6 +79,8 @@ declare namespace Cypress {
      * - on click of header, the changelog content is visible.
      */
     checkChangelogRevealer(): void;
+
+    abortPrefetchRequests(target: string): void;
   }
 }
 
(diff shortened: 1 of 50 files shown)


cypress/support/parent-commands.ts
```diff
@@ -85,7 +85,7 @@ declare namespace Cypress {
 Cypress.Commands.add('getDocsPageContentArea', () => cy.get('#paste-docs-content-area'));
 
 Cypress.Commands.add('pageHeaderShouldBeVisible', (headerText) => {
-  cy.contains('h1', headerText).shouldBeVisible();
+  cy.contains('h1', headerText, {matchCase: false}).shouldBeVisible();
```
andipants12, line 88: Turning on loose text matching since we are already matching the header with the `h1` element.

package.json
```diff
@@ -70,7 +70,7 @@
     "test:tools": "jest ./tools/ -c ./tools/jest.config.js",
     "test:tools-16": "USE_REACT_16=true jest ./tools/ -c ./tools/jest.config.js",
     "test:coverage": "yarn pre-test && jest --coverage",
-    "test:website": "start-server-and-test 'yarn serve:website' http://localhost:9000 'yarn cypress run --record --spec ./cypress/integration/**/*.spec.ts'",
+    "test:website": "start-server-and-test 'yarn serve:website' http://localhost:9000 'yarn cypress run --record --spec ./cypress/integration/**/**/*.spec.ts'",
```
andipants12, line 73: Include all specs in ci tests.

shleewhite (approved):

shleewhite:
The code looks ok, but I am worried that 2 of the tests timed out on cypress.. This happened before with the sidebar and we ended up having to change the test bc cypress wouldn't scroll to find the sidebar links. 

Is it always the same tests that time out (Tabs and Stack)? Or would it fluctuate from run to run?

TheSisb:
Also concerned about the two flakes
image

Worth considering: how much additional protection are we getting by testing every single page for the same thing when they all follow the same structure, VS how much we're waiting or spending on running those tests.

andipants12:
@TheSisb @shleewhite I'll look at these now. FWIW I ran these tests 10 times locally and they all passed (including the flakey tests check). With regard to the fix up for the sidebar, I have a separate branch with an upgrade to cypress v8 and that is not an ideal solution for the problem. I added a cypress command to find items in a fixed container, which also works for sidebar. Will post an update when I have one.

andipants12:
Closing in favor of: https://github.com/twilio-labs/paste/pull/2063

PR #2057, opened 2021-11-18
Fix/scroll jump

zahnster:
This change introduces two fixes:

1. Our Checkbox component has an absolutely positioned (hidden) input, but the Checkbox container doesn't declare relative position, leaving the positioning of the checkbox element ambiguous. This causes the 'jump' when our checkboxes are clicked in our website's layout. The change to Checkbox declares position relative on the Checkbox container so the input has a declared relative container.
2. Since our website uses flex / div-based scrolling, we don't want elements hidden outside the layout to invoke an overflow state on the body. Adding overflow: hidden will suppress these elements from triggering an overflow scroll on the site body.

### DIFF
--- a/.changeset/four-cycles-collect.md
+++ b/.changeset/four-cycles-collect.md
@@ -1,0 +1,6 @@
+---
+'@twilio-paste/checkbox': patch
+'@twilio-paste/core': patch
+---
+
+Declare relative position on checkbox to fix focus position bug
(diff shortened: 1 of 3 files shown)


.changeset/four-cycles-collect.md
```diff
@@ -0,0 +1,6 @@
+---
+'@twilio-paste/checkbox': patch
+'@twilio-paste/core': patch
+---
+
+Declare relative position on checkbox to fix focus position bug
```
TheSisb, line 6: Can you prefix this message with `[Checkbox]` for consistency please?

TheSisb (approved):
Just need to update the changeset message - ill change the PR title

shleewhite (approved):

PR #2060, opened 2021-11-19
chore(cypress): add spec for docs site search

andipants12:
## Description
Add basic cypress test suite to verify our site search functionality. A new test suite has been added under the `/integrations` folder to test the site search on the website homepage.

### Testing the following expected behavior:
- site search input should be focusable and visible (for non-mobile screen size, mobile not tested)
- `POST` request to Algolia should be triggered with correct query parameters when search text is entered
- `POST` request should respond with success status code (`200`)
- the number of suggestion options rendered should match the number of query matches in the response (for hits greater than or equal to 0 matches)

**Contributing to Twilio**


cypress.json
```diff
@@ -7,5 +7,7 @@
     "runMode": 2,
     "openMode": 0
   },
-  "blockHosts": "*.google-analytics.com"
+  "blockHosts": "*.google-analytics.com",
+  "testFiles": ["**/**/*.spec.ts"],
+  "experimentalNetworkStubbing": true
```
andipants12, line 12: Allow us to try out `route2` "stubbing" which has become their new intercept tool (called `.intercept`).

cypress.json
```diff
@@ -7,5 +7,7 @@
     "runMode": 2,
     "openMode": 0
   },
-  "blockHosts": "*.google-analytics.com"
+  "blockHosts": "*.google-analytics.com",
+  "testFiles": ["**/**/*.spec.ts"],
```
andipants12, line 11: Let's tell cypress which files to include here to make our test script a little more readable.

cypress/support/parent-commands.ts
```diff
@@ -79,6 +79,9 @@ declare namespace Cypress {
      * - on click of header, the changelog content is visible.
      */
     checkChangelogRevealer(): void;
+
+    // @TODO add notes/docs here.
```
andipants12, line 83: @andipants12 Do this plz, and thank you.

cypress.json
SiTaggart: Given that we do want to upgrade cypress in the very near future, do you think it is worthwhile starting to use a configuration option that is removed in [version 6](https://docs.cypress.io/guides/references/changelog#6-0-0), which enables the use of a feature that is also removed in [version 7](https://docs.cypress.io/guides/references/changelog#7-0-0)?

cypress/integration/site-search/index.spec.ts
```diff
@@ -0,0 +1,95 @@
...
+
+const addQueryMetaData = ({query, result}) => {
+  const [results] = result.results;
+
+  return {results: [{...results, query, params: `query=${query}&hitsPerPage=5`}]};
+};
+const extractRequestParams = (parsedBody) => Cypress._.get(parsedBody, 'requests[0].params');
+const extractHits = (xhr) => Cypress._.get(xhr, 'response.body.results[0].hits', []);
+
+const testSearch = (searchPhrase) => {
```
SiTaggart, line 11: Most of this test section seems to cover the majority of what we want to expect, perfectly. I can focus, I can type, I get options shown in the dropdown. It tests the website UI that I expect to interact with as a user.

Some parts though, I still worry are a little too tightly coupled to the implementation. Like, how Algolia works. We don't have any control over the request, nor the response. Honestly, that's entirely their concern. If they change anything about those things in a version bump of the aloglia plugin, this test will likely break which makes it a little brittle for me.

cypress.json
andipants12: No I don't. Truthfully, I think the best path forward would be to wait to add this test until we update. I added this as a work around just in case folks felt strongly that the ticket for site search test coverage be closed before the end of the year.

cypress/support/parent-commands.ts
```diff
@@ -79,6 +79,13 @@ declare namespace Cypress {
      * - on click of header, the changelog content is visible.
      */
     checkChangelogRevealer(): void;
+
+    /**
+     * Asserts that:
+     * - parsed search string has expected keys that map to the correct values
+     * - uses [URLSearchParams](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams)
+     */
+    verifyExpectedQuery(queryString: string, expectedParams: Record<string, string | boolean>): void;
```
SiTaggart, line 88: Do you still need this command for this changeset?

package.json
```diff
@@ -70,7 +70,7 @@
     "test:tools": "jest ./tools/ -c ./tools/jest.config.js",
     "test:tools-16": "USE_REACT_16=true jest ./tools/ -c ./tools/jest.config.js",
     "test:coverage": "yarn pre-test && jest --coverage",
-    "test:website": "start-server-and-test 'yarn serve:website' http://localhost:9000 'yarn cypress run --record --spec \"./cypress/integration/**/**/*.spec.ts\"'",
+    "test:website": "start-server-and-test 'yarn serve:website' http://localhost:9000 'yarn cypress run --record'",
```
SiTaggart, line 73: This is way better!

cypress.json
SiTaggart: I assume you don't need this for this changeset anymore?

cypress/integration/site-search/index.spec.ts
```diff
@@ -0,0 +1,39 @@
+describe('Docs website search', () => {
+  let searchTerm = '';
+  before(() => {
+    // Start an XHR server with the built-in cypress utilities.
+    cy.server();
+    cy.route('**/**/page-data.json').as('pageDataResponse');
+    cy.route('**/**/app-data.json').as('appDataResponse');
+    cy.visit('/');
+    // Wait for the JS assets to load since they are necessary for the search functionality.
+    cy.wait(['@pageDataResponse', '@appDataResponse']);
```
SiTaggart, line 10: Running locally, removing the wait here has no impact on the test run. Inspecting the network requests I can't see any `app-data.json` files, and `page-data.json` is only page content unrelated to search. 

I think it's super safe to remove them.

cypress/integration/site-search/index.spec.ts
```diff
@@ -0,0 +1,39 @@
...
+  });
+
+  beforeEach(() => {
+    cy.server();
+    cy.route({url: 'https://bh4d9od16a-dsn.algolia.net/1/indexes/*/**', method: 'POST'}).as('searchRequest');
```
SiTaggart, line 15: If you really want to wait for the request it looks like you can wildcard the whole domain, and the alias will still work as expected, which is cool.

`https://**.algolia.net/**`

We're at least a little less dependent on the url structure that way.

cypress/support/parent-commands.ts
andipants12: Hmm no, I do not. Good catch. Will remove 👍

cypress/integration/site-search/index.spec.ts
andipants12: @SiTaggart Without these wait statements, we get an error from cypress that the input cannot be found because it has been temporarily removed from the dom. I didn't dig too much into it, but AFAIK it's related to the state/status of these requests, since if you block these requests, the components that require JS for interaction do not work.
Unfortunately, this is definitely tightly coupled to a gatsby internal, and I know we should avoid that.

I noticed on the navigation tests, we have an arbitrary wait for some milieseconds, which is a patch for this kind of issue, but is [regarded as an antipattern by the cypress folks](https://docs.cypress.io/guides/references/best-practices#Unnecessary-Waiting).

cypress/integration/site-search/index.spec.ts
andipants12: Ohhh nice! Yeah I was playing around with the route matching with this older version and was struggling a bit to generalize. I'll make that change. Thanks!

cypress/support/index.js
```diff
@@ -16,12 +16,14 @@
 // Import commands.js using ES2015 syntax:
 import './commands';
 
+import {hydrationFinishedEvent} from '../../packages/paste-website/gatsby-browser.js';
```
SiTaggart, line 19: Just need to delete this and you're golden

SiTaggart (approved):

gloriliale (approved):
Approved!

PR #2063, opened 2021-11-19
chore(cypress): upgrade cypress to v8 and turn on tests

andipants12:
## Description
This PR turns on the newest addition to our End to End (E2E) tests and introduces some optimizations to the Github Actions job (GHAJ) that runs them.

### Summary of changes
- [x] Updated cypress and pinned to version 8.4.0 (^8.6.0 has known minor issues)
- [x] Added  a `strategy` configuration for the Cypress GHAJ that allows our tests to run on 3 machines and to not fail fast if a single test case or suite fails ([which causes the dashboard to hang](https://github.com/cypress-io/github-action/issues/48)),
- [x] Added command `abortPrefetchRequests` to prevent prefetches on page load that slow down our tests
- [x] Added this command invocation to (most) of the existing page spec files
- [x] Updated the Sidebar Navigation test to resolve some flakiness (see notes in diff)
- [x] Introduced looser matching for custom command `headerShouldBeVisible` to decrease the tests' coupling to the formatted content
- [x] Added config option for scroll behavior in the change revealer test
- [x] Added `ensureScrollable` config option to `getInFixedContainer` to cause tests to error out if the container it is called on is not in fact scrollable.

**Contributing to Twilio**


zahnster (approved):
Nice! 👍🏼  Code changes are solid. See below for updated commentary.

zahnster (changes requested):
Looking at the perf, we should get some parallelization in here so this doesn't increase our CI times so much.

SiTaggart:
For visibility I want to surface this comment from the now closed PR https://github.com/twilio-labs/paste/pull/2056#issuecomment-973181708

> Worth considering: how much additional protection are we getting by testing every single page for the same thing when they all follow the same structure, VS how much we're waiting or spending on running those tests.

I too have concerns about testing every page with the same duplicate tests. What mechanism is in place to make sure new pages get new tests? is it written down? Is there some kind of check? Human error will be high on this and pages will get missed

andipants12:
> What mechanism is in place to make sure new pages get new tests? is it written down? Is there some kind of check?

My thinking was that it will be pretty easy to add the script to compare documented pages to tested pages and run the generation on some interval. That is not currently implemented and I have not yet documented all this, but it is still technically a work in progress.

> how much additional protection are we getting by testing every single page for the same thing when they all follow the same structure?

In an ideal world where all of our website pages constructed using the same utilities (e.g components etc) that makes sense. In it's current state though, that's not really the case and we are leaving a lot of room for human error in the source code. Because of that I think there is a benefit to testing pages individually.

.github/workflows/on_pull_request.yml
```diff
@@ -203,165 +203,174 @@ jobs:
...
-
-      - name: Run Chromatic
-        uses: chromaui/action@v1
-        with:
-          projectToken: ${{ secrets.CHROMA_APP_CODE }}
-          token: ${{ secrets.GITHUB_TOKEN }}
-          buildScriptName: "build:storybook"
-          autoAcceptChanges: "main"
-          exitOnceUploaded: true
+  # react16_tests:
```
andipants12, line 206: Temporarily commented-out these workflow jobs so as to not over-run them on GitHub Actions. Should be un-commented before this PR is merged.

SiTaggart:
I don't think we're quite there as we're tripling the build minutes, where each job is building the website. It's about 25 minutes of website building.

Can we share the website build artefacts? 

image

andipants12:
> Can we share the website build artefacts?

Funny you should ask, I have a branch where I added a job to build the website after the other assets are built, and the tests just load them from the cache. 
I didn't include it here because I was worried about too much scope creep. Should I go ahead and add it in?

cypress.json
```diff
@@ -7,5 +7,6 @@
     "runMode": 2,
     "openMode": 0
   },
-  "blockHosts": "*.google-analytics.com"
+  "blockHosts": "*.google-analytics.com",
+  "nodeVersion": "system"
```
andipants12, line 11: The default sets the node version to the package that is bundled with Cypress. [See here for more info](https://docs.cypress.io/guides/references/configuration#Node-version)

This means that we were not necessarily running our tests with Node v12 as specified in our workflow. This instructs Cypress to use the version on the CI machine(s) (or local machine).

.github/workflows/on_pull_request.yml
```diff
@@ -59,58 +59,11 @@ jobs:
             packages/paste-design-tokens/types/**/*
             packages/paste-codemods/tools/.cache/mappings.json
 
-  type_check:
-    name: Type check
+  build_website:
```
andipants12, line 62: ℹ️ Job that will build the website immediately after the other package assets are finished building. This will then be saved to the cache to be used by the test job later.

.github/workflows/on_pull_request.yml
```diff
@@ -151,60 +103,25 @@ jobs:
...
+          name: compiled-website
           path: |
-            node_modules
-            */*/node_modules
-          key: ${{ runner.os }}-${{ hashFiles('**/yarn.lock') }}
-
-      - name: Install Dependencies
-        if: steps.yarn_cache_id.outputs.cache-hit != 'true' || steps.node_modules_cache_id.outputs.cache-hit != 'true'
-        run: yarn install --immutable
+            packages/paste-website/.cache/mappings.json
```
andipants12, line 120: TBH not sure if we actually need this, but it threw an error when I omitted all of the `.cache` files.

.github/workflows/on_pull_request.yml
```diff
@@ -401,18 +471,26 @@ jobs:
...
-        run: yarn build:website
+      - name: Verify Cypress
         env:
-          AIRTABLE_APIKEY: ${{ secrets.AIRTABLE_APIKEY }}
-          AIRTABLE_BASEID: ${{ secrets.AIRTABLE_BASEID }}
-          GATSBY_DOCSEARCH_APIKEY: ${{ secrets.GATSBY_DOCSEARCH_APIKEY }}
+          # make sure every Cypress install prints minimal information
+          CI: 1
+        # print Cypress and OS info
+        run: |
```
andipants12, line 486: ℹ️  This will verify the cypress installation before running our tests, as well as print some logs about the environment the tests will run in. This can be helpful to ensure that we are using the system config that we expect.

.github/workflows/on_pull_request.yml
```diff
@@ -401,18 +471,26 @@ jobs:
...
 
-      - name: Build website
-        run: yarn build:website
+      - name: Verify Cypress
         env:
-          AIRTABLE_APIKEY: ${{ secrets.AIRTABLE_APIKEY }}
-          AIRTABLE_BASEID: ${{ secrets.AIRTABLE_BASEID }}
-          GATSBY_DOCSEARCH_APIKEY: ${{ secrets.GATSBY_DOCSEARCH_APIKEY }}
+          # make sure every Cypress install prints minimal information
+          CI: 1
```
andipants12, line 484: ℹ️  CI flag that limits logging output.

.github/workflows/on_pull_request.yml
```diff
@@ -401,18 +471,26 @@ jobs:
...
 
-      - name: Download build artifacts
+      - name: Download website build artifact
         uses: actions/download-artifact@v2
         with:
-          name: compiled-js-and-types
-          path: packages/
+          name: compiled-website
+          path: packages/paste-website/
+          if-no-files-found: error
```
andipants12, line 479: ℹ️ I noticed that if the save-cache step _fails_ it just throws a warning, and does not error out. This will cause the job to fail and the action to stop running if no files are found on download.

.github/workflows/on_pull_request.yml
```diff
@@ -349,19 +261,177 @@ jobs:
...
-    needs: build
+    needs: [build, build_website]
     runs-on: ubuntu-latest
+    strategy:
+      # when one test fails, DO NOT cancel the other
+      # containers, because this will kill Cypress processes
+      # leaving the Dashboard hanging ...
+      # https://github.com/cypress-io/github-action/issues/48
+      fail-fast: false
+      matrix:
```
andipants12, line 432: ℹ️ this is a config specific to github actions to allow parallelization. Note that there are separate required configuration options for the cypress command itself (which are noted in a comment in the `package.json` file)

.github/workflows/on_pull_request.yml
```diff
@@ -151,60 +103,25 @@ jobs:
         with:
           name: compiled-js-and-types
           path: packages/
+          if-no-files-found: error
```
andipants12, line 106: ℹ️ I noticed that if the save-cache step _fails_ it just throws a warning, and does not error out. This will cause the job to fail and the action to stop running if no files are found on download. (also added to the job `build_website`

package.json
```diff
@@ -70,7 +70,8 @@
     "test:tools": "jest ./tools/ -c ./tools/jest.config.js",
     "test:tools-16": "USE_REACT_16=true jest ./tools/ -c ./tools/jest.config.js",
     "test:coverage": "yarn pre-test && jest --coverage",
-    "test:website": "start-server-and-test 'yarn serve:website' http://localhost:9000 'yarn cypress run --record --spec ./cypress/integration/**/*.spec.ts'",
+    "test:website": "start-server-and-test 'yarn serve:website' http://localhost:9000 'yarn cypress run --record --parallel --group cypress-tests --spec \"./cypress/integration/**/**/*.spec.ts\"'",
```
andipants12, line 73: ℹ️ Parallelized tests in cypress are available in a CI environment when:
- the run command has `--parallel=true`
- the run command has `--record=true`
- the CI environment config specifies the number of machines to run
Also note that we added a grouping label to this command.
It could be helpful in the future to split out our tests into separate jobs and group them by type/category.

🪙 🪙 The biggest value add here is that it could help us analyze our testing data in the future via the cypress dashboard & api.

cypress.json
SiTaggart: Those docs state the default value for `nodeVersion` is `system`. You're just resetting the value to be the default value. You can remove it

SiTaggart: Also states its deprecated and will be removed 
image

cypress/support/parent-commands.ts
```diff
@@ -129,11 +131,32 @@ Cypress.Commands.add('checkInPageNavigationLinks', () => {
 // @TODO Check ComponentHeader <--- waiting for changes to this component to be merged.
 // Cypress.Commands.add('checkComponentHeader', () => {});
 
+Cypress.Commands.add('abortPrefetchRequests', (target) => {
```
SiTaggart, line 134: Do you mind explaining what this is doing?

cypress/support/parent-commands.ts
```diff
@@ -129,11 +131,32 @@ Cypress.Commands.add('checkInPageNavigationLinks', () => {
...
+    },
+  }).as('pageData');
+  cy.route({
+    method: 'GET',
+    url: '/**/**/app-data.json',
+    onRequest: (xhr) => {
+      xhr.xhr.abort();
+    },
+  }).as('appData');
+  cy.route('http://api.github.com/repos/twilio-labs/paste', {}).as('githubData');
```
SiTaggart, line 152: Also this line

.github/workflows/on_pull_request.yml
TheSisb: We need this file. I might refactor it later, but this file needs to be commit and shipped.

TheSisb:
> Updated cypress and pinned to version 8.4.0 (^8.6.0 has known minor issues)

Just curious, what minor issues? Didn't see that documented anywhere.

> Added command abortPrefetchRequests to prevent prefetches on page load that slow down our tests

Is this safe? This seems... sketchy. And very coupled to Gatsby, which we have hopes to migrate off of in the next year. How much faster do the tests run with/without this? Some perf numbers would help me better understand the evaluation of this trade off.

> Added config option for scroll behavior in the change revealer test

I couldn't grok this sentence :/  Could I get more words explaining what this means please?




> Updated the Sidebar Navigation test to resolve some flakiness (see notes in diff)

I'm still seeing this PR fail on that and on the tabs page:
image
Is this catching a legitimate bug? I'm concerned about adding flakiness to our already flaky test suite...

.github/workflows/on_pull_request.yml
andipants12: Yeah that sounds reasonable. FWIW this isn't being used to build the website on release, only for the pull requests CI.

cypress/support/parent-commands.ts
andipants12: Oh yes, I totally forgot to add the docs to this command. Will add.

cypress/integration/sidebar-navigation/index.spec.ts
```diff
@@ -1,30 +1,35 @@
...
 
-    cy.get('[data-cy=content-button]').click({force: true}).should('have.attr', 'aria-expanded', 'true');
-    cy.get('[data-cy=content-list]').should('be.visible');
+    cy.get('[data-cy="content-button"]').click().shouldHaveAttribute('aria-expanded', 'true');
+    cy.getInFixedContainer('[data-cy="content-list"]').should('be.visible');
 
-    cy.get('[data-cy=patterns-button]').click({force: true}).should('have.attr', 'aria-expanded', 'true');
-    cy.get('[data-cy=patterns-list]').should('be.visible');
+    cy.get('[data-cy="patterns-button"]').click().shouldHaveAttribute('aria-expanded', 'true');
+    cy.getInFixedContainer('[data-cy="patterns-list"]').should('be.visible');
```
SiTaggart, line 33: I don't recall these ever being flakey, and now they are. This leads me to wonder if this `getInFixedContainer` is the cause of flakiness.

cypress/support/parent-commands.ts
```diff
@@ -79,13 +79,15 @@ declare namespace Cypress {
...
+
+    abortPrefetchRequests(target: string): void;
   }
 }
 
 Cypress.Commands.add('getDocsPageContentArea', () => cy.get('#paste-docs-content-area'));
 
 Cypress.Commands.add('pageHeaderShouldBeVisible', (headerText) => {
-  cy.contains('h1', headerText).shouldBeVisible();
+  cy.contains('h1', headerText, {matchCase: false}).should('be.visible');
```
SiTaggart, line 90: I'm starting to wonder why these test assertions are about visibility. We don't do anything about it being visible, we are trying to assert they exist.

cypress/support/child-commands.ts
```diff
@@ -30,10 +21,6 @@ declare namespace Cypress {
   }
 }
 
-Cypress.Commands.add('shouldBeVisible', {prevSubject: 'element'}, (subject) => {
```
SiTaggart, line 33: So these commands are confusing me.

It looks like you've replaced all usage of `shouldBeVisible` with the native `.should('be.visible')`, but then replaced `.should('have.attr', 'aria-expanded', 'true');` with `shouldHaveAttribute` in other places.

It looks like you pulled out the one command that was being used, and then started using another one for a different purpose.

I'm also not entirely convinced that 

`.shouldHaveAttribute('aria-expanded', 'true')`

is much, if not any, improvement over the native

`.should('have.attr', 'aria-expanded', 'true');`

Maaaaaybe lets' just delete this file and commands entirely?

cypress.json
andipants12: @SiTaggart I'm going to remove this for now, but FWIW added [this line to print the node version cypress is using](https://github.com/twilio-labs/paste/pull/2063/files#diff-9079843739331d8f7ae3e87cd86d1711e6a9fc3d30730886dd06d44d95808df5R493) it printed out the version that is bundled with the cypress version.

I wonder if the above snippet from the docs is true for more recent versions of cypress, but we are currently a few major version behind.

PR #2065, opened 2021-11-19
fix(site-layout): add position relative on scrolling container

zahnster:
* Our scrolling element was lacking a declaration of position: relative, leaving certain focus events tied into the body in Chrome, and causing rendering issues.
* This change introduces that css declaration. It also removes a previous attempt to fix this bug by suppressing overflow on the site body.


andipants12 (approved):
LGTM

PR #2070, opened 2021-11-22
chore(website): update inline code design

nkrantz:
Update styles of inline code for mdx ([ticket](https://issues.corp.twilio.com/browse/DSYS-3128))

**Contributing to Twilio**

### DIFF
--- a/packages/paste-website/src/components/Typography.tsx
+++ b/packages/paste-website/src/components/Typography.tsx
@@ -31,13 +31,16 @@ export const InlineCode: React.FC = (props) => {
     <Box
       as="code"
       backgroundColor="colorBackground"
+      borderWidth="borderWidth10"
+      borderColor="colorBorderWeaker"
+      borderStyle="solid"
       borderRadius="borderRadius20"
-      color="colorText"
+      color="inherit"
       fontFamily="fontFamilyCode"
       fontSize="fontSize20"
       lineHeight="lineHeight20"
-      paddingY="space20"
-      paddingX="space30"
+      paddingY="space10"
+      paddingX="space20"
       {...props}
     />
   );
--- a/packages/paste-website/stories/inline-code.stories.tsx
+++ b/packages/paste-website/stories/inline-code.stories.tsx
@@ -1,0 +1,41 @@
+import * as React from 'react';
+import {Anchor} from '@twilio-paste/core/anchor';
+import {Box} from '@twilio-paste/core/box';
+import {InlineCode} from '../src/components/Typography';
+import {Callout, CalloutText, CalloutTitle} from '../src/components/callout';
+
+export const RegularInline = (): React.ReactNode => <InlineCode>string</InlineCode>;
+
+export const AnchorInline = (): React.ReactNode => (
+  <InlineCode>
+    <Anchor href="#">string</Anchor>
+  </InlineCode>
+);
+
+export const ExternalAnchorInline = (): React.ReactNode => (
+  <InlineCode>
+    <Anchor href="#" showExternal>
+      string
+    </Anchor>
+  </InlineCode>
+);
+
+export const InParagraphInline = (): React.ReactNode => (
+  <Box width="400px">
+    <Callout>
+      <CalloutTitle as="h4">Remember!</CalloutTitle>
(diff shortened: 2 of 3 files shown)


packages/paste-website/src/components/Typography.tsx
```diff
@@ -31,8 +31,11 @@ export const InlineCode: React.FC = (props) => {
...
       backgroundColor="colorBackground"
+      borderWidth="borderWidth10"
+      borderColor="colorBorderWeaker"
+      borderStyle="solid"
       borderRadius="borderRadius20"
-      color="colorText"
+      color="inherit"
       fontFamily="fontFamilyCode"
       fontSize="fontSize20"
       lineHeight="lineHeight20"
```
gloriliale, line 41: The padding here seems like it could use some tweaking. In the avatar page, I see this example of two inline code snippets touching:
image

It looks like the padding in the figma example is also less extreme.

packages/paste-website/stories/inline-code.stories.tsx
```diff
@@ -0,0 +1,25 @@
...
+      string
+    </Anchor>
+  </InlineCode>
+);
+
+// eslint-disable-next-line import/no-default-export
+export default {
+  title: 'Website/Inline-Code',
+  component: RegularInline,
+};
```
gloriliale, line 25: It miiiight be worth adding an example that has two inline code snippets stacked on top of each other to see what it looks like without having to search through the docs page, and maybe one for inside of a callout (though I guess that would be most relevant for when the callout PR is merged in...)

packages/paste-website/src/components/Typography.tsx
```diff
@@ -31,8 +31,11 @@ export const InlineCode: React.FC = (props) => {
     <Box
       as="code"
       backgroundColor="colorBackground"
+      borderWidth="borderWidth10"
```
andipants12, line 34: Just checked this out on the quick start page and it LGTM.

One question: Now that the inline text is in a badge-like container, do we still need a [colon here](https://deploy-preview-2070--paste-docs.netlify.app/getting-started/quickstart#scripts)?

Screen Shot 2021-11-23 at 9 35 29 AM

andipants12 (approved):
I left a comment that it be a good idea to check in with folks about whether or not we need to modify the composition of he inline text with other text or elements, but nothing blocking. 👍

shleewhite:
Not sure whats going on, but it seems like a lot of the external anchors are still a little wonky. I can see it in the changelogs and [the core upgrade guide](https://deploy-preview-2070--paste-docs.netlify.app/core/upgrade-guide)

Screen Shot 2021-11-23 at 10 20 22 AM
Screen Shot 2021-11-23 at 10 22 33 AM

packages/paste-website/src/components/Typography.tsx
nkrantz: Thanks for catching this! I adjusted the padding to account for the extra border pixels and added a story to catch this example.

packages/paste-website/stories/inline-code.stories.tsx
nkrantz: truuuue

packages/paste-website/src/components/Typography.tsx
nkrantz: Hmm, I don't think the colon is necessary, but what would replace it? Might be confusing not having anything there to differentiate the script from the definition, since the inline code tags are used both before and after the colon.

nkrantz:
> Not sure whats going on, but it seems like a lot of the external anchors are still a little wonky. I can see it in the changelogs and the core upgrade guide

@shleewhite I spent a little too long trying to figure out where the changelog pages are built, because I'm _pretty_ sure this just happens when the `<InlineCode>` tag goes inside of the `<Anchor>` tag. To get the "external" icon to show up within the code snippet, like it does in the new story, the `<Anchor>` must go inside of the `<InlineCode>`. That's my guess as to why those pages are wonky. I'm going to fix the core upgrade guide one because that's happening right in the mdx.

SiTaggart (approved):

PR #2087, opened 2021-12-01
fix(label): default cursor when label is a legend

shleewhite:
For the label component - when the label has `as="legend"` and it is not `disabled`, the cursor should be default.

**Contributing to Twilio**


TheSisb (approved):

SiTaggart (approved):

gloriliale (approved):
woop woop!

PR #2093, opened 2021-12-07
feat(theme): add dataVisualization tokens to theme object

TheSisb:
- [x] Adds data visualization colors to the `color` bucket in the `@twilio-paste/design-tokens` package.
- [x] Exposes the new tokens on the `theme` object under the `dataVisualization` key.
- [x] Tests the new data visualization tokens against each other and against `color-background-body`


packages/paste-theme/src/types/GenericThemeShape.ts
```diff
@@ -1,130 +1,151 @@
...
+export type FontSizesKeys = keyof typeof FontSizes;
+export type FontWeightsKeys = keyof typeof FontWeights;
+export type FontsKeys = keyof typeof Fonts;
+export type SizingKeys = keyof typeof Sizings;
+export type LineHeightsKeys = keyof typeof LineHeights;
+export type RadiiKeys = keyof typeof Radii;
+export type BoxShadowsKeys = keyof typeof BoxShadows;
+export type SpacingsKeys = keyof typeof Spacings;
+export type TextColorsKeys = keyof typeof TextColors;
+export type ZIndicesKeys = keyof typeof ZIndices;
```
SiTaggart, line 31: Sensible 👍🏼

packages/paste-theme/src/usePasteHighchartsTheme.tsx
```diff
@@ -0,0 +1,20 @@
...
+import {ThemeContext} from '@twilio-paste/styling-library';
+import type {ThemeShape} from './types/themes';
+
+type HighchartsOptions = {
+  colors: string[];
+};
+
+// Needs more iteration, but currently returns the colors array used when
+// setting a theme in highCharts. See https://jsfiddle.net/gh/get/library/pure/highcharts/highcharts/tree/master/samples/highcharts/members/setoptions
+const usePasteHighchartsTheme = (): HighchartsOptions => {
```
SiTaggart, line 11: Is there anything specific to hightcharts here? I'm wondering if we're too tied to Hightcharts, could we get away with a more generic name?

packages/paste-design-tokens/tokens/global/palette-color.yml
```diff
@@ -44,3 +44,34 @@ props:
   color-brand-highlight:
     value: "{!amaranth}"
     comment: "Twilio brand red"
+  # Data Viz Colors
+  color-data-viz-1:
+    value: "{!palette-blue-80}"
+    comment: Color used in a sequence for data vizualizations.
```
katieporter, line 50: vizualizations -> visualizations

katieporter:
Would we want to add contrast checking to these tokens to ensure they are always >2:1 against the colors they will sit beside in our recommended sequence and/or color groups?

packages/paste-design-tokens/tokens/global/palette-color.yml
katieporter: Also, what would you think of updating the description just a touch to: "Color used for data visualizations. Must be used in a sequence."

We're starting to work on refining our token definitions and having them all follow a similar format, and a lot of our other tokens use the first sentence for a definition and an optional second sentence for usage restrictions.

packages/paste-theme/src/usePasteHighchartsTheme.tsx
SiTaggart: OK, yeah, part one answers part two for me then. 

Based on justification of the second, I think I tend to agree with you there. Because it's specific to highcharts, not putting it into the theme package probably makes sense there. If we switch libraries, we can just drop the highcharts package as a separate entity

packages/paste-design-tokens/tokens/global/palette-color.yml
TheSisb: Works for me!

packages/paste-theme/src/generateThemeFromTokens.ts
```diff
@@ -47,10 +48,24 @@ export const generateThemeFromTokens = ({
...
   zIndices,
+  colors,
 }: GenerateThemeFromTokensArgs): GenericThemeShape => {
   // default breakpoints
   const breakpoints = [sizings.size40, sizings.size100, sizings.size120];
 
+  // Get only the dataViz color tokens from the 'colors' token bucket
+  const dataVisualization = Object.keys(colors)
+    .filter((colorToken) => {
+      return colorToken.includes('DataViz');
```
shleewhite, line 59: should this be `.includes('DataVisualization')` ?

packages/paste-color-contrast-utils/src/utils.ts
```diff
@@ -196,3 +207,29 @@ export const getContrastRatingsOfTokensWithUIControlContrastRequirements = (
...
+ *
+ * @param {Partial<GenericTokensShape>} tokens
+ * @return {*}  {TokenPairContrastRating[]}
+ */
+export const getContrastRatingsOfTokensWithDataVisualizationContrastRequirements = (
+  tokens: Partial<GenericTokensShape>
+): TokenPairContrastRating[] => {
+  // always use the Default raw JSON to get the pairings as other themes won't inherit them automatically
+  const defaultThemeRawJSON = DefaultRawTokenJSON.props;
+  const tokenWithUIControlContrastRequirements = getTokensWithDataVisualizationContrastRequirements(
```
SiTaggart, line 226: This maybe a copy pasta variable name? TokenWithDataVisualizationContrastRequirements?

SiTaggart (approved):
Minor variable name question, their wise looks legit 👍

packages/paste-color-contrast-utils/__fixtures__/tokensWithExtraPairings.raw.json
```diff
@@ -163,6 +163,16 @@
       "comment": "Dark shade of destructive link text to be used in interactions",
       "originalValue": "{!palette-red-70}",
       "name": "color-text-link-destructive-dark"
+    },
+    "color-data-visualization-1": {
```
shleewhite, line 167: I'm not super familiar with the color contrast tools, but why do we only need to test one data viz color?

andipants12 (approved):

nkrantz (approved):
I might've missed the conversation on this, but is there any risk in having all of the comments be the same for these tokens? Would it potentially be helpful to include more nuance, like where in the sequence they should be used?

packages/paste-color-contrast-utils/__fixtures__/tokensWithExtraPairings.raw.json
TheSisb: We're testing all the color contrasts in the other test file. This particular fixture is for a unit test to make sure the function that pulls out tokens that need dataviz color contrast checking works correctly.

TheSisb:
> I might've missed the conversation on this, but is there any risk in having all of the comments be the same for these tokens? Would it potentially be helpful to include more nuance, like where in the sequence they should be used?

These first data visualization tokens are part of a sequential set that will be used as a collection altogether. There should be no risk in having them use the same comment, and updating the comment is also very easy in the future should we desire to do so.

gloriliale (approved):

shleewhite (approved):

PR #2105, opened 2021-12-17
chore: add TestTheme and emotion style matchers as globals

SiTaggart:
~I have no idea why adding these to a file in a new `globals` directory in the type root `./@types/` works but adding them to the already existing `./@types/index.d.ts` does not.~

Reverted to just including a special types declaration file for Jest globals.


shleewhite (approved):

andipants12 (approved):
LGTM 👍

PR #2106, opened 2021-12-29
fix: default imports for JSON for webpack 5 compatibility

stanlemon:
Addresses #2097, fixing JSON default imports in paste to be compatible with webpack 5 without throwing warnings during compilation.

**Contributing to Twilio**

> All third-party contributors acknowledge that any contributions they provide will be made under the same open-source license that the open-source project is provided under.

- [x] I acknowledge that all my contributions will be made under the project's license.

PR #2107, opened 2021-12-30
fix: default imports for JSON for webpack 5 compatibility

SiTaggart:
Taking over from https://github.com/twilio-labs/paste/pull/2106, thanks for the contribution @stanlemon

This is so all the tests runs and checks pass that are dependent on secrets.

### DIFF
--- a/.changeset/fluffy-beers-help.md
+++ b/.changeset/fluffy-beers-help.md
@@ -1,0 +1,5 @@
+---
+'@twilio-paste/color-contrast-utils': minor
+---
+
+Fix default imports for JSON to play nice with webpack 5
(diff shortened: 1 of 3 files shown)


zahnster (approved):
👍🏼  (thank you very much for the assist, @stanlemon)

TheSisb (approved):

stanlemon:
I ran into another instance of this at https://github.com/twilio-labs/paste/blob/main/packages/paste-style-props/src/constants.ts#L1 - seems to be the only other one though. I can try to put together a PR later today or this weekend.

zahnster:
> I ran into another instance of this at https://github.com/twilio-labs/paste/blob/main/packages/paste-style-props/src/constants.ts#L1 - seems to be the only other one though. I can try to put together a PR later today or this weekend.

No need - thanks to your help with this instance, I'm getting a PR together as we speak. Thanks again for your help!

PR #2118, opened 2022-01-07
Fix: style props default import

zahnster:
* Following #2106 / #2107, we noticed there was one more instance of a JSON default import in the style-props package.
* This change addresses that remaining instance, updating the import and the tsconfig to allow esModuleInterop


SiTaggart (approved):
Gah, I knew there was another one. I just couldn't remember where.

gloriliale (approved):
lgtm!

PR #2120, opened 2022-01-07
fix(airtable-update): update gastby refs to match updated airtable

zahnster:
* Updates Airtable name for BPM table in Gatsby config


gloriliale (approved):
yaaaasssss

SiTaggart (approved):
I didn't actually test that this was the fix, but I think it is

shleewhite (approved):

PR #2123, opened 2022-01-11
chore(website): setup VRT with Applitools for the token list page

andipants12:
## Description
Add setup for basic visual regression tests using `@applitools/eyes-cypress`. 👀 This is a very minimal set up that is comparing the token page's snapshot at low level. 

### Summary of changes:
- [x] add env variables for Applitools eyes to the cypress Github workflow (so the VRT runs in CI)
- [x] add custom commands that are simple wrappers around `@applitools/cypress-eyes` commands with default options set
- [x] manually add setup for `@applitools/eyes-cypress` (according to [instructions from the docs](https://github.com/applitools/eyes.sdk.javascript1/tree/master/packages/eyes-cypress#manual-configuration))
- [x] add simple eyes check to the tokens list page test suite
- [x] push commits to verify: content change does not trigger a required approval, css change does not trigger required approval, but dom change does!

### Verifications
- [x] [test setting new baseline](https://twilioeyes.applitools.com/app/test-results/00000251759081417945/00000251759081417773/steps/1?accountId=ulncwB59jE_UReq3X0V3RA~~&mode=step-editor)
- [x] test layout change
- [x] test content change

**Contributing to Twilio**


.gitignore
```diff
@@ -88,6 +88,9 @@ docs
 cypress/videos
 cypress/screenshots
 
+# Applitools
+secrets.applitools.json
```
andipants12, line 92: Dunno if this is how we want to do this, or with shell variables, but we should discuss.

applitools.config.js
```diff
@@ -1,6 +1,5 @@
 module.exports = {
   serverUrl: 'https://twilioeyesapi.applitools.com',
-  batchId: process.env.GITHUB_HEAD_SHA,
```
andipants12, line 3: Removed these to allow me to run the eyes checks locally. This is essentially replaced with a bit more involved logic checking to see if the cypress run is in the context of a local run, or a CI run (e.g. via github workflows)

cypress.json
```diff
@@ -5,7 +5,7 @@
   "projectId": "pe4h41",
   "retries": {
     "runMode": 2,
-    "openMode": 2
+    "openMode": 0
```
andipants12, line 8: @andipants12 Remember to change this back please 👍 
Set temporarily to `0` to expedite local testing with expected failures.

cypress/integration/e2e/components/alert-dialog.spec.ts
```diff
@@ -1,4 +1,4 @@
-describe('Alert-dialog component documentation page', () => {
+describe('Alert-dialog component documentation page', function () {
```
andipants12, line 1: ❓ If we use anon functions instead of arrow functions, we can leverage Mocha's context and pull the test name directly from `this`, rather than having to hard-code test names. What do folks think?

cypress/plugins/index.js
```diff
@@ -12,10 +12,33 @@
 // This function is called when a project is opened or re-opened (e.g. due to
 // the project's config changing)
 
+const secrets = require('../../secrets.applitools.json');
+const getBranchName = require('../../tools/utils/getBranchName');
+
+if (!process.env.APPLITOOLS_API_KEY) {
```
andipants12, line 18: Basically, this is either using the env variable when it is defined in the context of a github workflow (CI), or getting the key from a non-committed "secrets" file. See previous note/discussion question about using "secret" files in the repo, declarations in some bash shell, etc.

tools/utils/getBranchName.js
```diff
@@ -0,0 +1,11 @@
+const {exec} = require('child_process');
+
+module.exports = (callback) => {
```
andipants12, line 3: Super ruff around the edges script to get the branch name in a local env or if git is defined... pretty specifically intended for the cypress/eyes integration. Curious what folks think of the basic principle here, of:
1. figuring out the branch name 
2. using the branch name and the environment to name applitools branches

cypress/support/commands/parent-commands.ts
```diff
@@ -6,82 +10,6 @@
  * @see https://on.cypress.io/custom-commands#Parent-Commands
  */
```
andipants12, line 12: ℹ️  Moved these to a directory `@types` in the cypress dir so leverage TS's default behavior of looking for and successfully finding type declarations in `@types` folders.

.github/workflows/on_pull_request.yml
```diff
@@ -155,101 +155,101 @@ jobs:
...
-
-      - name: Download build artifacts
-        uses: actions/download-artifact@v2
-        with:
-          name: compiled-js-and-types
-          path: packages/
-
-      - name: Run tests
-        run: yarn test:react-16
+  # tests:
```
andipants12, line 158: ℹ️ A bunch of these jobs are commented out to allow us a bit more freedom to test the cypress/applitools changes.

.github/workflows/on_pull_request.yml
```diff
@@ -418,6 +418,11 @@ jobs:
         run: yarn test:website
         env:
           CYPRESS_RECORD_KEY: ${{ secrets.CYPRESS_RECORD_KEY }}
+          APPLITOOLS_API_KEY: ${{ secrets.APPLITOOLS_API_KEY }}
```
andipants12, line 421: @SiTaggart @TheSisb I couldn't find a record of this API key being removed. Can either of you confirm that it is still available on this `secrets` object?

.github/workflows/on_pull_request.yml
```diff
@@ -418,6 +418,11 @@ jobs:
...
+          APPLITOOLS_BATCH_NAME: "${{ github.event.pull_request.head.repo.full_name }}: pull request from ${{ github.event.pull_request.head.ref }} - ${{ github.event.pull_request.title }} (${{ github.event.pull_request.head.sha }})"
+          GITHUB_HEAD_SHA: ${{ github.event.pull_request.head.sha }}
```
andipants12, line 423: Lines 422 and 423 were ported from the github workflow for storybook VRT when it was still using applitools. 
Source: https://github.com/twilio-labs/paste/pull/1259/files#diff-4177c38233ccde70569c4fb02d752aaec30c3279d95a00c8af8f2382e7c3bc95

.github/workflows/on_pull_request.yml
```diff
@@ -418,6 +418,11 @@ jobs:
...
+          GITHUB_BRANCH_NAME: ${{ github.event.pull_request.head.ref }}
+          USE_CYPRESS_EYES: true
```
andipants12, line 425: ❓ Saw some folks using a similar technique to conditionally run/not run the applitools checks and thought it could be useful. What do folks think?

.gitignore
SiTaggart: We tend to use `.env` files for secrets. 

There's not "one" file, but one for each package that needs one.

Adding to the root env file would be fine, I think thats where it used to be

.github/workflows/on_pull_request.yml
SiTaggart: You should be able to view what secrets are set under the repo settings in GitHub. I don't think access to that is restricted by role if you're a maintainer of the repo, but I could be wrong

cypress/plugins/index.js
SiTaggart: If you use a `.env` file for the secret you can just use `process.env.APPLITOOLS_API_KEY` and call it a day. Might simplify things a little for you.

cypress/plugins/index.js
```diff
@@ -12,10 +12,33 @@
...
+}
+
 /**
  * @type {Cypress.PluginConfig}
  */
 module.exports = (on, config) => {
   // `on` is used to hook into various events Cypress emits
   // `config` is the resolved Cypress config
+
+  if (process.env.GITHUB_BRANCH_NAME != null) {
```
SiTaggart, line 29: How bad would it be if we just rely on the Branch name provided by the GitHub actions context? 

It's not like we have a use case for running locally and if you really need to test it works we can just use a static fall back of `local`

cypress/support/commands/@types/index.d.ts
```diff
@@ -0,0 +1,79 @@
+// eslint-disable-next-line @typescript-eslint/no-namespace
+declare namespace Cypress {
```
SiTaggart, line 2: For me, this certainly makes way more sense being in here now. 👏🏼

cypress/plugins/index.js
andipants12: Yeah for sure, that makes sense to me. I would love to sync with folks during focus time tomorrow on how we should leverage things like branch name, batch name, etc. 

I'll go ahead and shuffle things around tomorrow to rely only on the name from the GH actions, defaulting to `local` if that is not available. Thanks!

.github/workflows/on_pull_request.yml
andipants12: Yeah I don't have access.

cypress/plugins/index.js
```diff
@@ -12,10 +12,25 @@
...
 
+const result = require('dotenv').config();
+
 /**
  * @type {Cypress.PluginConfig}
  */
 module.exports = (on, config) => {
-  // `on` is used to hook into various events Cypress emits
-  // `config` is the resolved Cypress config
+  if (config.env.USE_CYPRESS_EYES == null) {
```
andipants12, line 21: This is wip, just testing for now.

.github/workflows/on_pull_request.yml
shleewhite: Why would we conditionally run them?

cypress/integration/token-list/index.spec.ts
```diff
@@ -8,4 +9,22 @@ describe('Token list filter', () => {
...
     cy.get('#border-colors').should('not.exist');
   });
+
+  describe('Visual regression tests', () => {
+    /* TODO:
+      - full page check or multiple section/region checks
+      - need to add check for branch name/batch id for CI runs
+    */
+    it('Basic VRT', () => {
+      // ℹ️ We are able to reference `this.title` because we have bound this describe block to the Cypress context.
```
shleewhite, line 19: Love these comments, theyre super helpful!

shleewhite (approved):
Looks good to me!

cypress/support/commands/parent-commands.ts
```diff
@@ -166,3 +94,37 @@ Cypress.Commands.add('getInFixedContainer', (selector) => {
       return cy.get('@target').scrollIntoView({offset: {top: (height as number) / 2, left: 0}, ensureScrollable: true});
     });
 });
+
+Cypress.Commands.add('openTwilioEyes', (overrides = {}) => {
```
nkrantz, line 98: Where do we run these commands?

nkrantz (approved):
There's a lot going on here, and it would be cool to walk through it with you now that I've studied it myself, but approving! Just left a couple comments/questions. Thanks for leaving the comments, those are really helpful.

.github/workflows/on_pull_request.yml
andipants12: Except for the purposes of developing on or debugging the applitools configuration and tests, we do not need to run them locally. This allows us a simple way to run or skip all applitools tests.

cypress/support/commands/parent-commands.ts
```diff
@@ -166,3 +94,37 @@ Cypress.Commands.add('getInFixedContainer', (selector) => {
...
+  }
+});
+
+Cypress.Commands.add('checkTwilioEyes', (overrides = {}) => {
+  const checkOptions = {ignoreDisplacements: true, useDom: true, ...overrides};
+  if (eyesAreEnabled()) {
+    cy.eyesCheckWindow(checkOptions);
+  }
+});
+Cypress.Commands.add('closeTwilioEyes', () => {
```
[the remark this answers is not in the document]
andipants12: Good catch, thanks! It's not required, as cypress eyes actually logs for us, but it's not super clear and I wanted to add these to help folks interpret the logs. 
I added this into the tokens test file. 👍

.github/workflows/on_pull_request.yml
SiTaggart: OK, well rest assured the secret for applitools is still there

SiTaggart (commented):
Is the applitools test running? I can see the github check is reporting no tests being run still. 

I'll be able to give better overall feedback once that's happening. For example I suspect you don't want the batch name to be the git sha, but without it running to see what the output is it's hard to articulate why.

andipants12:
@SiTaggart 

> I'll be able to give better overall feedback once that's happening. For example I suspect you don't want the batch name to be the git sha, but without it running to see what the output is it's hard to articulate why.

@zahnster Just pointed out that CI check to me yesterday. Currently the applitools test is running _with_ the cypress tests,  contained [within a new `describe` block in the tokens list test file](https://github.com/twilio-labs/paste/pull/2123/files#diff-cd1a2c2cdc96b89d19bf3418b21cf23a043af131c53adf8bb3a98fd228bc9a3cR21).

I haven't pushed again recently (about to) but you should be able to see the [most recent result in the Applitools Dashboard](https://twilioeyes.applitools.com/app/test-results/00000251760191351399/00000251760191351196/steps/1?accountId=ulncwB59jE_UReq3X0V3RA~~&mode=step-editor).

🪙 🪙 I'm going to do a quick draft in a doc of how to run, configure, check the test result, etc.

SiTaggart:
So, the applitools check used to work, when Rich set it up for cypress for the first time we were seeing the check report back the test runs. Lets investigate what that is, I suspect it's the batch name and id needing to match the PR correct so that applitools knows what PR to update

.github/workflows/on_pull_request.yml
```diff
@@ -418,6 +418,11 @@ jobs:
         run: yarn test:website
         env:
           CYPRESS_RECORD_KEY: ${{ secrets.CYPRESS_RECORD_KEY }}
+          APPLITOOLS_API_KEY: ${{ secrets.APPLITOOLS_API_KEY }}
+          APPLITOOLS_BATCH_NAME: "${{ github.event.pull_request.head.repo.full_name }}: pull request from ${{ github.event.pull_request.head.ref }} - ${{ github.event.pull_request.title }} (${{ github.event.pull_request.head.sha }})"
```
andipants12, line 422: ❗ This was pulled from the previous set up of Applitools with storybook. Need to:
1. look up the Github workflow integration with Applitols/cypress-eyes
2. confirm that this is the required batch name setup
3. Update as necessary (and maybeeee leave a comment/note)

cypress/integration/token-list/index.spec.ts
```diff
@@ -8,4 +17,43 @@ describe('Token list filter', () => {
...
+      const testSuiteName = this.title;
+
+      cy.log('[Applitools]: checking if eyes are enabled');
+      if (eyesAreEnabled()) {
+        cy.log('[Applitools]: Eyes are enabled, proceed with eyes check.');
+        const openParams: Partial<Eyes.Open.Options> = {
+          ...DEFAULT_OPEN_PARAMS,
+          testName: testSuiteName,
+          batchName: getBatchName(),
```
andipants12, line 36: TODO: This needs (may need) to be updated when you figure out the requirements from Applitools for the batch name and github workflow integration.

andipants12:
Hey @SiTaggart, I have an update for ya: 👇 

1. I just pushed a commit that intentionally changed the DOM of the tokens page, to check and see if we detect a difference. The other two changes (CSS change and content change) did not trigger a detection, which is good!
2. The GH Actions integration is 90% working 🙌 After the cypress tests finish, the applitools check is correctly updated. I tested it by adding a new VRT page also, and it correctly determined that there were two passing images, one of them new.
3. Before the cypress tests finish, the CI report checks indicate that applitools is passing with no tests, check out the images below.

This is before the cypress tests have finished and/or started 👇 

Here is the applitools process:
Screen Shot 2022-01-25 at 9 32 00 AM

Here is the cypress process:
Screen Shot 2022-01-25 at 9 32 18 AM

Any idea if that is how it worked before as well?

PR #2127, opened 2022-01-12
feat(logo-link): make entire logo clickable, add hover

zahnster:
* Makes full Paste logo on the website clickable
* Adds hover effect for logo swapping
* Updates logo UI to match Figma design

Screen Shot 2022-01-12 at 10 12 50 AM

### DIFF
--- a/packages/paste-website/src/components/site-wrapper/site-header/SiteHeaderLogo.tsx
+++ b/packages/paste-website/src/components/site-wrapper/site-header/SiteHeaderLogo.tsx
@@ -5,18 +5,32 @@ import {Text} from '@twilio-paste/text';
 import {Box} from '@twilio-paste/box';
 import {MediaObject, MediaBody, MediaFigure} from '@twilio-paste/media-object';
 import {styled} from '@twilio-paste/styling-library';
-// import {useTheme} from '@twilio-paste/theme';
-// import {PasteIconInverse} from '../../icons/PasteIconInverse';
+import {useTheme} from '@twilio-paste/theme';
+import {PasteIcon} from '../../icons/PasteIcon';
 import {PasteIconPride} from '../../icons/PasteIconPride';
 
-const StyledLink = styled(Link)`
-  text-decoration: none;
-  color: inherit;
+interface StyledLinkProps {
+  href: string;
+  onClick?: () => void;
+  onMouseEnter?: () => void;
+  onMouseLeave?: () => void;
+}
 
-  &:visited {
-    color: inherit;
-  }
-`;
+const StyledLink: React.FC<StyledLinkProps> = ({children, ...props}) => {
+  return (
+    <Box
+      {...props}
+      as="a"
+      textDecoration="none"
+      color="colorTextInverse"
+      _hover={{textDecoration: 'underline'}}
+      _focus={{boxShadow: 'shadowFocus', borderRadius: 'borderRadius10'}}
+      _focusVisible={{outline: 'none'}}
+    >
+      {children}
+    </Box>
+  );
+};
 
 interface SiteHeaderLogoProps {
   title?: string;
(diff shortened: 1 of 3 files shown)


loreina (approved):
This looks good to me! 🎉 

A small nit pick: is it possible for the focus ring to wrap the logo svg tightly? There's just a wee bit of margin on the top/bottom
Current | Design spec 
--- | ---
image | image

It'd also be cool to have the hover transition and ease in, the switch from the multicoloured logo to red logo feels a lil jarring but totally just me being a ✨ designer ✨ and not at all necessary

katieporter:
+1 to Loreina's comments. Otherwise looks great!

andipants12 (approved):

packages/paste-website/src/components/site-wrapper/site-header/SiteHeaderLogo.tsx
```diff
@@ -1,22 +1,34 @@
...
+      textDecoration="none"
+      color="colorTextInverse"
+      _hover={{textDecoration: 'underline'}}
+      _focus={{boxShadow: 'shadowFocus', borderRadius: 'borderRadius10'}}
+      _focusVisible={{outline: 'none'}}
+    >
+      {children}
+    </Box>
+  );
+};
```
SiTaggart, line 31: lovely refactor of a styled component

packages/paste-website/src/components/site-wrapper/site-header/SiteHeaderLogo.tsx
```diff
@@ -38,39 +55,38 @@ const SiteHeaderLogo: React.FC<SiteHeaderLogoProps> = ({title, subtitle}) => {
...
+        href="/"
+        onClick={() =>
+          trackCustomEvent({
+            category: 'Top Navigation',
+            action: 'click-paste-logo',
+            label: 'Paste logo',
+          })
+        }
+        onMouseEnter={() => setPasteIcon(pasteRedIcon)}
+        onMouseLeave={() => setPasteIcon(pastePrideIcon)}
```
SiTaggart, line 68: If you didn't want to mess around with state you could probably use `backgroundImage="url()"` and on `_hover={{backgroundImage: 'url()'}}`.

TheSisb: I'm not sure how Gatsby would bundle that so I could be wrong, but the css hover approach may cause a flicker on hover as the image is loaded for the first time unless a sprite is used.

packages/paste-website/src/components/site-wrapper/site-header/SiteHeaderLogo.tsx
```diff
@@ -1,22 +1,34 @@
...
 
-  &:visited {
-    color: inherit;
-  }
-`;
+const StyledLink: React.FC<StyledLinkProps> = ({children, ...props}) => {
+  return (
+    <Box
+      {...props}
+      as="a"
```
TheSisb, line 21: If we use the gatsby link we can benefit from JS routing here which won't cause a full page refresh. Not sure if this is a big deal, so consider it a nit being called out.

TheSisb (approved):

packages/paste-website/src/components/site-wrapper/site-header/SiteHeaderLogo.tsx
shleewhite: I was wondering about this, some places on the site we use Gatsby Link and sometimes we just use anchor. Do we want to have a stance on this as we're refactoring so many parts of the site?

SiTaggart: lol, we do kind of have a stance, we just made some mistakes and missed somethings.

If we're internal linking, we should be using the `SiteLink` component which is a wrapper around Gatsby Link, or if we don't want it to be blue, a custom link using gatsby link.

packages/paste-website/src/components/site-wrapper/site-header/SiteHeaderLogo.tsx
zahnster: Since we wanted to transition between the two states, I left this approach, with some modifications.

packages/paste-website/src/components/site-wrapper/site-header/SiteHeaderLogo.tsx
zahnster: Updated to use Gatsby's link component, but opted to not use `SiteLink` as the UI effects didn't match between how we treat normal text links and how we wanted to treat this header link.

PR #2128, opened 2022-01-12
test(combobox): help debugging

andipants12:
**Contributing to Twilio**

### DIFF
--- a/packages/paste-core/components/combobox/src/Combobox.tsx
+++ b/packages/paste-core/components/combobox/src/Combobox.tsx
@@ -138,6 +138,8 @@ const Combobox = React.forwardRef<HTMLInputElement, ComboboxProps>(
       iconColor = 'colorTextInverseWeak';
     }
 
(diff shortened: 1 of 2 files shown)


packages/paste-core/components/combobox/src/Combobox.tsx
```diff
@@ -138,6 +138,8 @@ const Combobox = React.forwardRef<HTMLInputElement, ComboboxProps>(
       iconColor = 'colorTextInverseWeak';
     }
 
+    const {ref: toggleButtonRef, ...toggleButtonProps} = getToggleButtonProps({tabIndex: 0, ref});
```
andipants12, line 141: Unlike `getInputProps`, `getToggleButtonProps` is _not_ an impure function. This means that we do not need to call it and spread it inline with the component. We can pass the forwarded ref to `getToggleButtonProps`, and pass the returned ref to `getInputProps`.

PR #2126, opened 2022-01-12
fix compile bug caused by react-router-dom

younessssssss:
update version of react-router-dom
update index.tsx

**Contributing to Twilio**

> All third-party contributors acknowledge that any contributions they provide will be made under the same open-source license that the open-source project is provided under.

- [ ] I acknowledge that all my contributions will be made under the project's license.

PR #2136, opened 2022-01-13
fix(cra-template): upgrade react-router-dom

TheSisb:
Taking over from #2126, thanks for the contribution @younessssssss

This is so all the tests runs and checks pass that are dependent on secrets.

### DIFF
--- a/.changeset/old-beds-invent.md
+++ b/.changeset/old-beds-invent.md
@@ -1,0 +1,5 @@
+---
+'@twilio-paste/cra-template': minor
+---
+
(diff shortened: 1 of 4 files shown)


SiTaggart:
Ah, my logic was a bit naive for that DangerJS core package changeset check. https://github.com/twilio-labs/paste/blob/main/.danger/changesets-that-need-core-check.ts#L16

I only check to see if it's core or icons, but really there are a bunch of packages that aren't in the core bundle that need to be excluded from this check

shleewhite (approved):

SiTaggart (approved):

PR #2151, opened 2022-01-24
fix(breadcrumb): spread props on child Anchor or Text element

gloriliale:
In Breadcrumb:
- Before: props were spread onto the Box as list item element. 
- In this PR: props are spread onto either the Anchor or Text element inside of the Box list item. This was done to match the component's presentation: BreadcrumbItem looks like an Anchor, so a data-testid on BreadcrumbItem should map onto the Anchor component.

Docs site:
- Section added to the Breadcrumb docs to indicate how to use routing, which is analogous to Anchor routing. Along these same lines, there's a link to a codesandbox that uses Breadcrumb alongside React Router. Would love eyes on this codesanbox, since the one for Anchor routing has had like +11k views 👀 

Open questions
- [ ] Will this structure change break anyone's stuff? Is it a Breaking Change (right now its changeset has it be a minor change)? Will we only know this through reporting?

### DIFF
--- a/packages/paste-core/components/breadcrumb/src/index.tsx
+++ b/packages/paste-core/components/breadcrumb/src/index.tsx
@@ -1,9 +1,10 @@
 import * as React from 'react';
 import * as PropTypes from 'prop-types';
 import {Box, safelySpreadBoxProps} from '@twilio-paste/box';
-import type {BoxElementProps} from '@twilio-paste/box';
+import type {BoxProps, BoxElementProps} from '@twilio-paste/box';
 import {Anchor} from '@twilio-paste/anchor';
-import {Text} from '@twilio-paste/text';
+import type {AnchorProps} from '@twilio-paste/anchor';
+import {Text, safelySpreadTextProps} from '@twilio-paste/text';
 import {useUIDSeed} from '@twilio-paste/uid-library';
 
 const BreadcrumbSeparator: React.FC<{element: BoxElementProps['element']}> = ({element}) => (
@@ -21,22 +22,28 @@ const BreadcrumbSeparator: React.FC<{element: BoxElementProps['element']}> = ({e
   </Text>
 );
 
-export interface BreadcrumbItemProps extends React.HTMLAttributes<HTMLLIElement> {
+type BreadcrumbItemBaseProps = Pick<BoxProps, 'element'> & {
   children: NonNullable<React.ReactNode>;
-  element?: BoxElementProps['element'];
   parentElement?: BoxElementProps['element'];
-  href?: string;
   last?: boolean;
-}
+};
+
+type BreadcrumbItemAsSpanProps = React.HTMLAttributes<HTMLSpanElement> &
+  BreadcrumbItemBaseProps & {
+    href?: never;
+  };
+
+type BreadcrumbItemAsAnchorProps = AnchorProps & BreadcrumbItemBaseProps;
+
+type BreadcrumbItemProps = BreadcrumbItemAsSpanProps | BreadcrumbItemAsAnchorProps;
 
 const DEFAULT_ELEMENT_NAME = 'BREADCRUMB';
 
-const BreadcrumbItem = React.forwardRef<HTMLAnchorElement, BreadcrumbItemProps>(
+const BreadcrumbItem = React.forwardRef<HTMLSpanElement | HTMLAnchorElement, BreadcrumbItemProps>(
   ({children, element, parentElement, href, last, ...props}, ref) => {
     const elementName = element || parentElement || DEFAULT_ELEMENT_NAME;
     return (
       <Box
-        {...safelySpreadBoxProps(props)}
         alignItems="center"
         as="li"
         color="colorText"
@@ -46,11 +53,17 @@ const BreadcrumbItem = React.forwardRef<HTMLAnchorElement, BreadcrumbItemProps>(
         lineHeight="lineHeight20"
       >
         {href ? (
-          <Anchor element={`${elementName}_ANCHOR`} href={href} ref={ref}>
+          <Anchor
+            {...(props as Partial<BreadcrumbItemAsAnchorProps>)}
+            element={`${elementName}_ANCHOR`}
+            href={href}
+            ref={ref as React.ForwardedRef<HTMLAnchorElement>}
+          >
             {children}
           </Anchor>
         ) : (
           <Text
+            {...safelySpreadTextProps(props)}
             aria-current="page"
             as="span"
             element={`${elementName}_TEXT`}
@@ -73,7 +86,7 @@ if (process.env.NODE_ENV === 'development') {
   BreadcrumbItem.propTypes = {
     children: PropTypes.node.isRequired,
     element: PropTypes.string,
-    href: PropTypes.string,
+    href: PropTypes.string as PropTypes.Validator<string>,
     last: PropTypes.bool,
   };
 }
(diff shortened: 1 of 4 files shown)


packages/paste-website/src/pages/components/breadcrumb/index.mdx
```diff
@@ -154,6 +154,20 @@ You may choose to use a page Heading as the current page in the breadcrumb.
...
+### Breadcrumb with a router
+
+Using breadcrumb with a router is analogous to
+[using Anchor with a router](/components/anchor/#using-anchor-with-a-router), since
+our BreadcrumbItem component creates links using [Anchor](/components/anchor) under the hood.
+
+<iframe
+  src="https://codesandbox.io/embed/using-breadcrumb-with-react-router-t24en?fontsize=14&hidenavigation=1&theme=dark"
+  style="width:100%; height:500px; border:0; border-radius: 4px; overflow:hidden;"
+  title="Using Anchor with React Router"
```
andipants12, line 166:
```suggestion
  title="Using BreadcrumbItem as anchor with React Router"
```

packages/paste-website/src/pages/components/breadcrumb/index.mdx
```diff
@@ -154,6 +154,20 @@ You may choose to use a page Heading as the current page in the breadcrumb.
...
 </LivePreview>
 
+### Breadcrumb with a router
+
+Using breadcrumb with a router is analogous to
+[using Anchor with a router](/components/anchor/#using-anchor-with-a-router), since
+our BreadcrumbItem component creates links using [Anchor](/components/anchor) under the hood.
+
+<iframe
+  src="https://codesandbox.io/embed/using-breadcrumb-with-react-router-t24en?fontsize=14&hidenavigation=1&theme=dark"
```
andipants12, line 164: In the deploy preview it looks like this is a duplicate sandbox of the anchor demo, except that the sandbox title is different. Is this intentional or should we show a demo/example of using `Breadcrumb`?

packages/paste-website/src/pages/components/breadcrumb/index.mdx
```diff
@@ -154,6 +154,20 @@ You may choose to use a page Heading as the current page in the breadcrumb.
...
+  style="width:100%; height:500px; border:0; border-radius: 4px; overflow:hidden;"
+  title="Using Anchor with React Router"
+  allow="accelerometer; ambient-light-sensor; camera; encrypted-media; geolocation; gyroscope; hid; microphone; midi; payment; usb; vr; xr-spatial-tracking"
+  sandbox="allow-forms allow-modals allow-popups allow-presentation allow-same-origin allow-scripts"
+></iframe>
+
```
andipants12, line 170: I don't think there is enough spacing between the sandbox and the next header 

Screen Shot 2022-01-24 at 11 17 23 AM

andipants12 (changes requested):

packages/paste-core/components/breadcrumb/src/index.tsx
```diff
@@ -46,11 +45,12 @@ const BreadcrumbItem = React.forwardRef<HTMLAnchorElement, BreadcrumbItemProps>(
         lineHeight="lineHeight20"
       >
         {href ? (
-          <Anchor element={`${elementName}_ANCHOR`} href={href} ref={ref}>
+          <Anchor {...safelySpreadBoxProps(props)} element={`${elementName}_ANCHOR`} href={href} ref={ref}>
```
SiTaggart, line 48: As a note worth mentioning, anywhere you are using one of our own components, you _should_ in _theory_ be able to skip using `safelySpreadBoxProps`, as the component you are using, `Anchor` in this case, is already doing it at it's level. 

We therefore don't need to use it twice to render the same DOM element.

Good rule of thumb, use `safelySpreadBoxProps` on a `<Box />` only

packages/paste-website/src/pages/components/breadcrumb/index.mdx
gloriliale: ah thanks, good catch!

packages/paste-website/src/pages/components/breadcrumb/index.mdx
gloriliale: They look nearly identical, but this is the one w/ breadcrumb (there just aren't any breadcrumbs on the Home page since it's a home page).

It would probably be more useful if what I linked was the specific Page 2 where there's the most breadcrumbs. Is there a way to do that in a Codesandbox embed?

packages/paste-website/src/pages/components/breadcrumb/index.mdx
gloriliale: V true, just changed it to have a separator between these two. thanks!

packages/paste-core/components/breadcrumb/src/index.tsx
gloriliale: o0o0o a good note, thanks @SiTaggart! I'd like to change it to`{...props}` for consistency, but when I try to do that, I get some typing errors. I'll see if anyone's available to help me with this.

gloriliale: Done!

packages/paste-website/src/pages/components/breadcrumb/index.mdx
gloriliale: Figured out how to change the initial path, so upon opening this docs page it should show the breadcrumbs. Lemme know if this looks good to you @andipants12!

packages/paste-core/components/breadcrumb/src/index.tsx
```diff
@@ -73,7 +84,7 @@ if (process.env.NODE_ENV === 'development') {
   BreadcrumbItem.propTypes = {
     children: PropTypes.node.isRequired,
     element: PropTypes.string,
-    href: PropTypes.string,
+    href: PropTypes.string.isRequired,
```
andipants12, line 87: Is this a required prop? On line 57 it looks like it's possible it is undefined.

andipants12 (commented):
Left one final comment about the `PropType` declarations for the breadcrumb component. 👀

packages/paste-core/components/breadcrumb/src/index.tsx
gloriliale: It throws an error if we don't add `.isRequired` that looks like this:
```
Type 'Requireable<string>' is not assignable to type 'Validator<string>'.
  Types of property '[nominalTypeHack]' are incompatible.
    Type '{ type: string | null | undefined; } | undefined' is not assignable to type '{ type: string; } | undefined'.
      Type '{ type: string | null | undefined; }' is not assignable to type '{ type: string; }'.
        Types of property 'type' are incompatible.
          Type 'string | null | undefined' is not assignable to type 'string'.
            Type 'undefined' is not assignable to type 'string'.ts(2322)
```

I'm not exactly sure what to make of it so just went with it tbh

SiTaggart: Actually, this is kind of tricky on second thought. React proptypes is trying to infer the type based on the typescript types. But it can't understand that it is conditional, it's likely taking the last one it sees, or when it is an anchor and required, it trumps everything. I naively just made it required to satisfy the compiler, but thats not accurate at all.

You might have to remove it entirely, or `@ts-expect-error` as we're still figuring out conditionally typed refs

shleewhite: Looking at [Badge prop types](https://github.com/twilio-labs/paste/blob/91a8a4943a6aa7ec0c17619ff662892b89c22cc6/packages/paste-core/components/badge/src/index.tsx#L64), we're able to say `href` isn't required when it is required sometimes...

zahnster (approved):
🙏🏼 🎉 🚢

SiTaggart (approved):

PR #2119, opened 2022-01-07
fix(combobox): multiselect combobox example doesn't reselect item when it is selected and then removed

shleewhite:
Github Discussion: https://github.com/twilio-labs/paste/discussions/2115

**Bug:** When you select and then immediately unselect an item, if you try to select it again noting happens.
**Cause:** 
* We have a function called `handleSelectItemOnClick` that calls the downshift function to add the item to the list of selected items and removes it from the list of combobox items
* `handleSelectItemOnClick`  is passed to the `useComboboxPrimitive`  as the value for  `onSelectedItemChange`

So the problem is that when the user adds an item, removes it, and then adds it again, `handleSelectItemOnClick` isn’t being called because the item isn’t changing.

This bug happens in storybook and the website because it is the same code.

- [x] Fixes all the downshift console errors 
- [x] Uses a function to filter a static list of items to improve typeahead
- [x] Uses `onStateChange` instead of `selectedItemChange` to match best practices (https://downshift.netlify.app/use-multiple-selection)
- [x] Fixes the multiselect bug in storybook and website examples

**Contributing to Twilio**

> All third-party contributors acknowledge that any contributions they provide will be made under the same open-source license that the open-source project is provided under.

- [x] I acknowledge that all my contributions will be made under the project's license.

PR #2155, opened 2022-01-25
fix(combobox/combobox-primitive): fix multiselect example and ref merge bug

andipants12:
## Description
Continuation of [PR 2119](https://github.com/twilio-labs/paste/pull/2119) from @shleewhite.

#### Issue 
A customer [reported that our multiple selection example is not working as expected](https://github.com/twilio-labs/paste/discussions/2115).

#### Cause
Downshift does not automatically "reset" the `useCombobox` selected item state. The just-selected, then removed, item cannot be selected again immediately after because it is still set as the currently selected Item.

#### Fix
Manually reset the selected item state for the multiple selection examples.

#### Issue 
We were seeing `console.error` statements indicating that the `ref` properties from various prop getters were not being correctly passed to indented components. This was specifically happening for the controlled component whenever we passed our [combobox its state via our `state` property](https://paste.twilio.design/components/combobox/#usecombobox-state-hook).
> downshift_error

#### Cause
We are currently conditionally using a "default" state for the combobox, by allowing our customers to pass in their own state that matches the API specifications of our component and the managing hook. Due to internal implementation details of the `useCombobox` hook, this caused the observed side effects.

#### Fix
Remove default hook render if `state` property is received by the component.

## Summary of changes:
- [x] manually set `selectedItem` to null after selection in multi primitive storybook
- [x] conditionally render the combobox with default state or with consumer-provided state
- [x] update website docs examples for: pill form, combobox component, combobox primitives
- [x] add missing lodash types


package.json
```diff
@@ -233,5 +233,8 @@
       "pre-push": "if git-branch-is main; then yarn pre-push; fi"
     }
   },
-  "packageManager": "yarn@3.1.0"
+  "packageManager": "yarn@3.1.0",
+  "devDependencies": {
```
andipants12, line 237: ooops going to remove this.

packages/paste-core/primitives/combobox/stories/mutliselect.stories.tsx
```diff
@@ -0,0 +1,136 @@
...
+  const formPillState = useFormPillState();
+
+  const {
+    getSelectedItemProps,
+    getDropdownProps,
+    addSelectedItem,
+    removeSelectedItem,
+    selectedItems,
+  } = useMultiSelectPrimitive({
+    onSelectedItemsChange: ({type, selectedItems: currentSelectedItems}) => {
```
shleewhite, line 37: ❓ Why do we need both `onSelectedItemsChange` and `onSelectedItemChange`?

packages/paste-core/primitives/combobox/stories/mutliselect.stories.tsx
```diff
@@ -0,0 +1,136 @@
...
+      if (type === FunctionAddSelectedItem) {
+        const [toRemove] = difference(currentSelectedItems, selectedItems);
+        setFilteredItems((state) => state.filter((item) => item !== toRemove));
```
shleewhite, line 40: if toRemove is an array, will the item ever equal toRemove?

shleewhite, line 40: If `toRemove` is an array of multiple items, wouldn't `item` never equal `toRemove`?

tsconfig.json
```diff
@@ -15,12 +15,14 @@
     "noImplicitAny": true,
     "noResolve": false,
     "paths": {
-      "*": ["./@types/*"]
+      "*": ["./@types/*", "./node_modules/*/@types/*"],
```
shleewhite, line 18: I tried building this branch and these lines are causing it to fail

packages/paste-core/primitives/combobox/stories/mutliselect.stories.tsx
```diff
@@ -0,0 +1,136 @@
...
+
+  const dropdownProps = getDropdownProps({
+    preventKeyAction: isOpen,
+    ...getToggleButtonProps({tabIndex: 0}),
+  });
+
+  return (
+    <>
+      <Box marginBottom="space40" position="relative">
+        <Label htmlFor={seed('multi-select-primitive-input')} {...getLabelProps()}>
```
shleewhite, line 84: Would using `seed('multi-select-primitive-input')` for htmlFor on the label and for the id of the input generate 2 different ids?

packages/paste-core/primitives/combobox/stories/index.stories.tsx
```diff
@@ -80,14 +78,14 @@ export const AutocompleteCombobox = (): React.ReactNode => {
...
+  const seed = useUIDSeed();
   return (
     <>
-      <Label htmlFor={uid} {...getLabelProps()}>
+      <Label htmlFor={seed('autocomplete-combobox')} {...getLabelProps()}>
         Choose a component:
       </Label>
       <Box display="flex" {...getComboboxProps()}>
-        <Input id={uid} type="text" {...getInputProps()} />
+        <Input id={seed('autocomplete-combobox')} type="text" {...getInputProps()} />
```
shleewhite, line 88: Random thought: I think downshift might be overriding the ids because theyre set before we pass `{...getInputProps()}`

andipants12: Oh good catch! Yeah we can actually pass a few diff config options to the `useCombobox` hook to either prefix each generated ID with our custom string, or specifically set an id (e.g. input id versus menu id).

I'll check out both when I switch back to this branch and update. 👍

tsconfig.json
andipants12: interesting, I can check it out. Do you mind sharing the error output?

packages/paste-core/primitives/combobox/stories/mutliselect.stories.tsx
andipants12: the way this works currently it leverages `onSelectedItemChange` from `useCombobox` (which is a callback for the on-click handlers for the options in the dropdown) to trigger the appropriate action from `useMultipleSelection` to either add or remove the selected item to the collection of selected items.

This is a callback that runs when that selection changes. You can think of it as an alternative to adding `useEffect` to handle the side effect inside this method body, which is adjusting the list of filtered items (if a selected item is removed, it adds it back, if an item is selected, it is removed from the filtered items).

packages/paste-core/primitives/combobox/stories/mutliselect.stories.tsx
andipants12: Admittedly, this is a bit brittle, but since `addSelectedItem` and `removeSelectedItem` only handle a single item, I made the assumption that the diff count would always be one.

Since this is an example in our docs, I think it's Good Enough™️. It would be a bug in the hook if this were not triggered per change.

shleewhite (approved):
LGTM! Just need to update the website examples to match the latest changes in the storybook file.

packages/paste-website/src/component-examples/ComboboxPrimitiveExamples.ts
```diff
@@ -104,9 +103,9 @@ render(
 
 export const multiSelectExample = `const items = ['Alert', 'Anchor', 'Button', 'Card', 'Heading', 'List', 'Modal', 'Paragraph'];
 
-const BasicMultiCombobox = () => {
+const MultipleSelectCombobox = () => {
   const seed = useUIDSeed();
-  const [filteredItems, setFilteredItems] = React.useState([...items]);
+  const [filteredItems, setFilteredItems] = React.useState([...items].sort());
```
TheSisb, line 108:
```suggestion
  const [filteredItems, setFilteredItems] = React.useState(items);
```

Can add the sort to the tail end of line 104 and not need to do anything fancy here

tsconfig.json
```diff
@@ -29,6 +30,7 @@
     "resolveJsonModule": true,
     "esModuleInterop": true
   },
+  "files": ["node_modules/@storybook/react/dist/ts3.9/client/preview/types-6-0.d.ts"],
```
TheSisb, line 33: Hmm why did we add this line?

packages/paste-core/components/combobox/src/Combobox.tsx
```diff
@@ -153,9 +159,8 @@ const Combobox = React.forwardRef<HTMLInputElement, ComboboxProps>(
         >
           <ComboboxInputWrapper {...getComboboxProps({role: 'combobox'})}>
             <ComboboxInputSelect
-              {...getToggleButtonProps({tabIndex: 0})}
               // we spread props into `getInputProps` so that Downshift handles events correctly
-              {...getInputProps({disabled, required, ref, ...props})}
+              {...getInputProps({disabled, required, ...props, ...toggleButtonProps})}
```
TheSisb, line 163: Is it safe to pass toggleButtonProps into getInputProps? (Made a similar comment elsewhere, pointing out both locations)

tsconfig.json
```diff
@@ -17,10 +17,11 @@
     "paths": {
       "*": ["./@types/*"]
     },
+    "baseUrl": ".",
```
TheSisb, line 20: Also curious about this line addition

andipants12: Setting a base url in tsconfig files essentially [tells the compiler where the base or root is for relative path resolution](https://www.typescriptlang.org/tsconfig#baseUrl). So this allows us to use configuration options like  [`paths`](https://www.typescriptlang.org/tsconfig#paths).

packages/paste-website/src/component-examples/ComboboxPrimitiveExamples.ts
```diff
@@ -23,8 +23,7 @@ const BasicCombobox = () => {
         <Input
           id={uid}
           type="text"
-          {...getInputProps()}
-          {...getToggleButtonProps({tabIndex: 0})}
+          {...getInputProps(getToggleButtonProps({tabIndex: 0}))}
```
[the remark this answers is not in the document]
andipants12: Unfortunately the downshift docs are not complete and [there is content in the `useMultipleSelection` documentation about strategies to combine these prop getters](https://github.com/downshift-js/downshift/tree/master/src/hooks/useMultipleSelection#prop-getters) that is missing in `useCombobox`.

The toggle button prop getter doesn't actually return button-specific attributes, and since we are using the input box as the menu/dropdown trigger, this is as intended. It's actually the same thing we do in `Combobox`, with the extra getter for the multiple selection hook.

packages/paste-core/primitives/combobox/stories/mutliselect.stories.tsx
```diff
@@ -0,0 +1,134 @@
...
+    onSelectedItemChange: ({selectedItem: selected}) => {
+      if (selected != null) {
+        addSelectedItem(selected);
+      }
+      // ℹ️ This action from the dropdown library is strongly typed to accept only a string
+      //   to reset, we need to pass "null" so we will cast it as a string e.g. "null as string".
+
+      // ℹ️ Downshift does not clear the "selectedItem" state after a selection has changed
+      //   with the multiple selection implementation, we need to force this case, so that if
+      //   we remove the item, it isn't still marked as "Selected" in the combobox state.
```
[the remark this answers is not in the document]
andipants12: No, this is the intended behavior strangely. @shleewhite did research on this initially and can provide more context.

tsconfig.json
andipants12: There are strange things with our TS configuration 😅 I wanted to actually use the `StoryFn` type and this was a patch to make it accessible. I can back it out though and revert the story to just be a function that returns a react node.

andipants12: I really would love to take a look at our TS configs and see if we can improve some of the long standing by low-priority funniness. Def not needed for this patch tho.

packages/paste-core/components/combobox/src/Combobox.tsx
andipants12: I [left a response to this right here](https://github.com/twilio-labs/paste/pull/2155/files#r793115262) 👈

gloriliale:
oh my find about the website example might be bc of this that @shleewhite had said?

> LGTM! Just need to update the website examples to match the latest changes in the storybook file.

packages/paste-core/primitives/combobox/stories/mutliselect.stories.tsx
```diff
@@ -0,0 +1,134 @@
+import * as React from 'react';
+import difference from 'lodash/difference';
+import {useUIDSeed} from '@twilio-paste/uid-library';
+import {Box} from '@twilio-paste/box';
+import {Input} from '@twilio-paste/input';
+import {Label} from '@twilio-paste/label';
+import {FormPill, FormPillGroup, useFormPillState} from '@twilio-paste/form-pill-group';
+import {ComboboxListbox, ComboboxListboxGroup, ComboboxListboxOption} from '@twilio-paste/combobox';
+
+import type {StoryFn} from '@storybook/react/dist/ts3.9/client/preview/types-6-0';
```
SiTaggart, line 10: I suspect what you actually want is 

`import { ComponentStory, ComponentMeta } from '@storybook/react';`

as seen in the typescript examples https://storybook.js.org/docs/react/writing-stories/introduction

Hopefully that saves you from doing anything to the tsconfig files

tsconfig.json
SiTaggart: The TS configuration is fine, honestly. I don't think those are types you should be reaching for. There are better ones correctly exported from storybook react

packages/paste-core/primitives/combobox/stories/mutliselect.stories.tsx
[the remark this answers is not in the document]
shleewhite: 😬  its definitely confusing. It seems like behavior carried over from the single select combobox - when the user picks a value, it doesn't get un-selected until they pick a different one. 

For multiselect: from the comboboxes perspective, when the user selects the value and then removes it, the selectedItem is the same - only the list of combobox items has changed. selectedItem changes when they pick an entirely different value. So, to get around this we have to clear out the selectedItem every time the user picks something.

packages/paste-core/primitives/combobox/stories/mutliselect.stories.tsx
andipants12: Huh, interesting. The other day when I added this I _tried_ to import from that package but it didn't work for some reason. 🤔 

Will update.

packages/paste-website/src/component-examples/ComboboxPrimitiveExamples.ts
andipants12: ❓ Is there a technical concern with line 108?

FWIW I spread the items array into the state hook because I've seen bugs from not cloning objects (in this case it's just a shallow clone bc it's just an array of strings).

I just quickly [spun up this sandbox that demonstrates this](https://codesandbox.io/s/original-state-when-passed-an-array-zsps7?file=/src/App.tsx).

if you press the buttons with the sandbox terminal open, you can see that when the original array is passed, it's super easy to accidentally modify the original object. Worth noting, that this could be solved in the specific instance of updating the state by just spreading the old state in with the new.

tsconfig.json
andipants12: FWIW I was trying to access the `StoryFN` type because [storybook strongly recommends using CSF](https://storybook.js.org/docs/react/api/csf). 

In a comment above I mentioned that previously I was not able to import the expected types from the react package, but that is no longer an issue.

packages/paste-website/src/component-examples/ComboboxPrimitiveExamples.ts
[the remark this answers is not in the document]
TheSisb: The way I'm reading that, from what I understand it is saying to do:

```js
getDropdownProps(getToggleButtonProps({})
or
getDropdownProps(getInputProps({})

but not

getInputProps(getToggleButtonProps({})
```

andipants12:
@SiTaggart @TheSisb Here are some additional resources on "merging" props like refs and event handlers:
- [Article that goes in depth on how the getters work and how to merge props from downshift with dev supplied](https://www.middle-engine.com/blog/posts/2021/12/20/performance-issues-with-the-prop-getters-pattern-in-react#the-prop-getters-pattern)
- [Here is the implementation of `getInputProps` which shows an example of merging the refs as well as handing multiple event handlers](https://github.com/downshift-js/downshift/blob/master/src/hooks/useCombobox/index.js#L359)

packages/paste-core/primitives/combobox/stories/mutliselect.stories.tsx
```diff
@@ -32,102 +36,105 @@ export const BasicMultiCombobox: StoryFn = () => {
...
+
+        <Box {...getComboboxProps({role: 'combobox'})} display="flex">
+          <Box display="flex" flexGrow={1}>
+            <Input
+              {...getInputProps(getDropdownProps({preventKeyAction: isOpen, type: 'text', id: inputId}))}
+              flexGrow={1}
+            />
+          </Box>
+          <Box display="flex" flexGrow={0}>
+            <Button {...getToggleButtonProps({tabIndex: 0})} aria-label="toggle-button">
```
shleewhite, line 100: For a11y: aria-label should be a readable string that describes what the button does, like "Expand combobox" or something

packages/paste-core/primitives/combobox/stories/mutliselect.stories.tsx
```diff
@@ -32,102 +36,105 @@ export const BasicMultiCombobox: StoryFn = () => {
...
+              <ChevronDownIcon size="sizeIcon30" decorative={false} title="toggle menu" />
```
shleewhite, line 101: Also for a11y: this can actually be a decorative icon because the aria-label on the button describes what the icon is for.

packages/paste-core/primitives/combobox/stories/mutliselect.stories.tsx
andipants12: Could we move this to a separate ticket and address for all the combobox examples? I can do a jira-less PR too, just want to prevent scope creep for this one.

shleewhite: Yee - if its in other examples lets make another ticket.

gloriliale (approved):
Looking good!

packages/paste-core/components/combobox/src/Combobox.tsx
```diff
@@ -153,12 +159,18 @@ const Combobox = React.forwardRef<HTMLInputElement, ComboboxProps>(
...
+                disabled,
+                required,
+                // we can pass the toggle button ID as the "inputId" to ensure it won't get overwritten
+                inputId,
+                autoComplete: autocomplete,
+                ...props,
+                ...toggleButtonProps,
+                'aria-describedby': helpTextId,
+                ...(!autocomplete ? {onChange: (event: React.ChangeEvent) => event.preventDefault()} : undefined),
+              })}
```
SiTaggart, line 173: Ok, yes, this all makes sense to me 👍🏼

packages/paste-website/src/component-examples/ComboboxPrimitiveExamples.ts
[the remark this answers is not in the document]
TheSisb: If we revert this change, does anything break?

packages/paste-core/primitives/combobox/stories/mutliselect.stories.tsx
```diff
@@ -0,0 +1,140 @@
...
+            setInputValue('');
+            addSelectedItem(selectedItem);
+
+            // ℹ️ This action from the dropdown library is strongly typed to accept only a string
+            //   to reset, we need to pass "null" so we will cast it as a string e.g. "null as string".
+
+            // ℹ️ Downshift does not clear the "selectedItem" state after a selection has changed
+            //   with the multiple selection implementation, we need to force this case, so that if
+            //   we remove the item, it isn't still marked as "Selected" in the combobox state.
+            selectItem((null as unknown) as string);
```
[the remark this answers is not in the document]
andipants12: This is the actual "fix" for the reported issue in the example. This relates to @shleewhite's comment previously about the root cause for the bug. It's also described in the original pull request.

.changeset/flat-colts-learn.md
```diff
@@ -0,0 +1,6 @@
+---
+'@twilio-paste/combobox': patch
+'@twilio-paste/core': patch
+---
+
+[Combobox] Fix internal refs so primitive doesn't override them
```
SiTaggart, line 6: You might need a more descriptive changelog entry for this as it's not refs related anymore

packages/paste-core/components/combobox/src/extractPropsFromState.tsx
```diff
@@ -0,0 +1,65 @@
+import * as React from 'react';
+import {useComboboxPrimitive} from '@twilio-paste/combobox-primitive';
+import type {UseComboboxPrimitiveStateChange, UseComboboxPrimitiveReturnValue} from '@twilio-paste/combobox-primitive';
+import {isEmpty} from 'lodash';
```
SiTaggart, line 4: Be careful here. For component packages we tend to only use the individual lodash packages, otherwise you end up compiling all of lodash.

This is why the package size has blown up I suspect.

image

packages/paste-core/primitives/combobox/stories/index.stories.tsx
```diff
@@ -203,14 +203,27 @@ export const BasicMultiCombobox: React.FC = () => {
...
     },
   });
   const uid = useUID();
 
+  const dropdownProps = getDropdownProps({
+    preventKeyAction: isOpen,
+    ...getToggleButtonProps({tabIndex: 0}),
+  });
+
+  const toggleButtonProps = getToggleButtonProps({tabIndex: 0});
```
SiTaggart, line 225: Might be a nothing burger, but you call `getToggleButtonProps({tabIndex: 0})` here and in line 222. 

I only call it out as it's in the public example that we use in the docsite too.

SiTaggart (commented):
This seems to be in such a better space now. Just a couple of small things to tidy up

packages/paste-core/components/combobox/src/extractPropsFromState.tsx
andipants12: Yea this was an auto import thing. Thanks, I'll change to `import isEmpty from 'lodash/isEmpty';` 👍

packages/paste-core/primitives/combobox/stories/index.stories.tsx
andipants12: Yeah, that's not intentional and shouldn't be there. Will update, thanks.

.changeset/flat-colts-learn.md
```diff
@@ -0,0 +1,9 @@
+---
+'@twilio-paste/combobox': patch
+'@twilio-paste/core': patch
+---
```
andipants12, line 4: @SiTaggart How does this look?

SiTaggart: Much better!

SiTaggart (approved):

PR #2167, opened 2022-01-28
fix(checkbox): hide icon when not checked

shleewhite:
- [x] hide check icon with display prop when the checkbox is not checked
- [x] explicitly sets if the component is controlled or uncontrolled, so the icon renders correctly 

Fixes a bug found in: https://github.com/twilio-labs/paste/discussions/2165

**Contributing to Twilio**


packages/paste-core/components/checkbox/stories/checkbox.stories.tsx
```diff
@@ -13,8 +13,18 @@ export default {
 };
 
 export const DefaultCheckbox = (): React.ReactNode => {
```
shleewhite, line 15: Note: needed to make all the checkboxes in the stories all controlled so the icon displays correctly.

zahnster (approved):
Nice! Thanks for sprucing this up.

packages/paste-core/components/checkbox/src/Checkbox.tsx
```diff
@@ -64,7 +64,16 @@ const CheckboxIcon: React.FC<{
   if (indeterminate) {
     return <MinusIcon element={element} decorative color={color} size="sizeIcon10" />;
   }
-  return <CheckboxCheckIcon element={element} hidden={!checked} decorative color={color} size="sizeIcon10" />;
+  return (
+    <CheckboxCheckIcon
+      element={element}
+      hidden={!checked}
```
SiTaggart, line 70: So, the bug is on this line. `hidden` isn't being set on the svg element in the DOM. The logic is already there. Needs a test

shleewhite: Seems like no icons use the `hidden` prop... they all use `display`. I can just take out the `hidden` line?

SiTaggart: Sure, lets still have a test though, hey?

shleewhite: yeee adding them rn

packages/paste-core/components/checkbox/__tests__/checkbox.test.tsx
```diff
@@ -48,12 +51,15 @@ describe('Checkbox', () => {
...
   it('should render as checked', () => {
-    const {getByLabelText} = render(
+    const {getByLabelText, container} = render(
       <Checkbox {...defaultProps} checked onChange={NOOP}>
         foo
       </Checkbox>
     );
     expect((getByLabelText('foo') as HTMLInputElement).checked).toBeTruthy();
+
+    const checkIcon = container.querySelector('[data-paste-element="CHECKBOX_ICON"');
```
andipants12, line 61: Does the checkbox use a React Portal or something else that causes the target element to render outside of the main render return target?

packages/paste-core/components/checkbox/stories/checkbox.stories.tsx
andipants12: Interesting, does it not work with uncontrolled components? I know we had a regression a while back for folks who were using the checkbox component as an uncontrolled component. Did we ensure that functionality is not lost?

andipants12 (commented):
Generally LGTM! When we confirm that the checkbox icon renders correctly for uncontrolled components too, I'll go ahead and apporove. 🥳

packages/paste-core/components/checkbox/__tests__/checkbox.test.tsx
shleewhite: Nope, I just followed the format of how the customization tests target elements.

packages/paste-core/components/checkbox/stories/checkbox.stories.tsx
shleewhite: I don't have context on that... but in storybook when it was uncontrolled the check icon didn't appear - but the background color did change. If this is a regression, it should be fixed in a separate ticket.

SiTaggart: Probably does need to work in both controlled and uncontrolled situations

shleewhite: Ohhhhhh ok I see what was happening now. It seems like it was working in both situations kind of on accident.

Before, the checkbox icon was always visible because the hidden prop doesn't do anything. But it wasn't actually visible because it is the same color as the background when its unchecked. It was supposed to work where the icon is only visible if the checked prop is true.

The BaseRadioCheckboxControl has styles that change the backgroundColor when the sibling element (the hidden checkbox) is active.

Now, that the icon only shows is working as it was supposed to before, when the component is uncontrolled the background color changes as expected but the checked prop is undefined so the icon doesnt appear.

So now, I think I should add a new state or something that tracks the checked state of the hidden checkbox when the checked prop is undefined so the icon can show correctly.

packages/paste-core/components/checkbox/src/Checkbox.tsx
```diff
@@ -25,6 +25,7 @@ export interface CheckboxProps extends React.InputHTMLAttributes<HTMLInputElemen
   isSelectAll?: boolean;
   isSelectAllChild?: boolean;
   checked?: boolean;
+  defaultChecked?: boolean;
```
shleewhite, line 28: Adding the defaultChecked prop to help differentiate between a controlled component and an uncontrolled component. This follows what react recommends for uncontrolled inputs: https://reactjs.org/docs/uncontrolled-components.html#default-values

packages/paste-core/components/checkbox/src/Checkbox.tsx
```diff
@@ -90,15 +100,29 @@ const Checkbox = React.forwardRef<HTMLInputElement, CheckboxProps>(
     const disabled = props.disabled != null ? props.disabled : checkboxGroupContext.disabled;
     const name = props.name != null ? props.name : checkboxGroupContext.name;
     const hasError = props.hasError != null ? props.hasError : checkboxGroupContext.hasError;
-    const handleChange = props.onChange != null ? props.onChange : checkboxGroupContext.onChange;
+
+    // Manages what the value of checked is depending on if it is controlled or uncontrolled
+    const [checkedState, setCheckedState] = React.useState(!!defaultChecked);
```
shleewhite, line 105: This is a simplified version of what Chakra does (see [their useCheckbox hook](https://github.com/chakra-ui/chakra-ui/blob/e1057b085f47916fb148561e40fe0b4afbdd633f/packages/checkbox/src/use-checkbox.ts#L160))

If the user passes checked & onChange it is a controlled component, if they don't pass them it is uncontrolled.

packages/paste-core/components/checkbox/src/Checkbox.tsx
```diff
@@ -124,12 +148,12 @@ const Checkbox = React.forwardRef<HTMLInputElement, CheckboxProps>(
...
         <HiddenCheckbox
           {...safelySpreadBoxProps(props)}
-          checked={checked}
+          checked={mergedChecked}
           disabled={disabled}
           name={name}
           onChange={handleChange}
           aria-describedby={helpTextId}
-          aria-checked={indeterminate ? 'mixed' : checked}
+          aria-checked={indeterminate && 'mixed'}
```
SiTaggart, line 156: Careful here: there are 3 states for `aria-checked`. `null` | `"true"` | `"mixed"`

packages/paste-core/components/checkbox/src/Checkbox.tsx
```diff
@@ -90,15 +100,29 @@ const Checkbox = React.forwardRef<HTMLInputElement, CheckboxProps>(
     const disabled = props.disabled != null ? props.disabled : checkboxGroupContext.disabled;
     const name = props.name != null ? props.name : checkboxGroupContext.name;
     const hasError = props.hasError != null ? props.hasError : checkboxGroupContext.hasError;
-    const handleChange = props.onChange != null ? props.onChange : checkboxGroupContext.onChange;
+
+    // Manages what the value of checked is depending on if it is controlled or uncontrolled
+    const [checkedState, setCheckedState] = React.useState(!!defaultChecked);
+    const isControlled = checkedProp !== undefined;
+    const mergedChecked = isControlled ? checkedProp : checkedState;
```
TheSisb, line 107: If we replace this with 
```js
const checked =  checkedProp || defaultChecked;
```
Can we remove the internal state handling stuff?

packages/paste-core/components/checkbox/src/Checkbox.tsx
shleewhite: Oops! Yah fixed it.

packages/paste-core/components/checkbox/__tests__/checkbox.test.tsx
```diff
@@ -34,8 +34,11 @@ const defaultGroupProps = {
 
 describe('Checkbox', () => {
   it('should render', () => {
-    const {getByRole} = render(<Checkbox {...defaultProps}>foo</Checkbox>);
+    const {getByRole, container} = render(<Checkbox {...defaultProps}>foo</Checkbox>);
     expect(getByRole('checkbox')).not.toBeNull();
+
+    const checkIcon = container.querySelector('[data-paste-element="CHECKBOX_ICON"');
```
nkrantz, line 40: I guess it doesn't matter bc the tests are passing, but aren't these queries missing the closing bracket?
`[data-paste-element="CHECKBOX_ICON"'`
`[data-paste-element="CHECKBOX_ICON"]'`

packages/paste-website/src/pages/components/checkbox/index.mdx
```diff
@@ -101,6 +106,22 @@ Use a checkbox to present a user with a binary (e.g. on/off) choice that is part
 - When in an error state display an inline error message below the offending checkbox that clearly describes the error along with an icon that depicts the severity.
 - When displaying additional content based on checking a checkbox, be sure that the new content appears after the checkbox so that it is naturally discoverable by assistive technology users.
 
+## Controlled vs. uncontrolled checkboxes
+
+The checkbox can either be controlled, meaning there is an external state that determines if the checkbox is checked or not, or uncontrolled, meaning the checkbox manages its own state.
+
```
nkrantz, line 112: Super helpful to have this context in the docs, might be worth adding a sentence about why one would want an uncontrolled vs controlled checkbox

packages/paste-core/components/checkbox/src/Checkbox.tsx
nkrantz: because `defaultChecked` is a boolean, what's the difference between using the double bang operator and just saying `useState(defaultChecked)`? wouldn't that still give you either `true` or `false`?

nkrantz (approved):
noice🥇  left a few questions but nothing blocking

packages/paste-core/components/checkbox/__tests__/checkbox.test.tsx
shleewhite: Ah thanks, great catch!

TheSisb (approved):

PR #2168, opened 2022-01-29
feat: remove console theme

shleewhite:
- `paste/design-tokens`
   - [x] removes console theme tokens
   - [x] removes console snapshot test
- `paste/theme`
   - [x] adds test that passing `<Theme.Provider theme="console">` returns a default theme color token
   - [x] console.error message that console theme is deprecated
   - [x] updates the console theme and associated type to be the default theme
   - [x] removes console snapshot test

**Contributing to Twilio**


packages/paste-theme/src/themeProvider.tsx
```diff
@@ -22,10 +22,10 @@ function getProviderThemeProps(theme: ThemeVariants, customBreakpoints?: string[
...
         breakpoints: customBreakpoints || SendGridTheme.breakpoints,
       };
-    case ThemeVariants.CONSOLE:
+    case DeprecatedThemeVariants.CONSOLE:
       return {
-        ...ConsoleTheme,
-        breakpoints: customBreakpoints || ConsoleTheme.breakpoints,
+        ...DefaultTheme,
+        breakpoints: customBreakpoints || DefaultTheme.breakpoints,
       };
```
TheSisb, line 29: If you omit these lines, the case will fall into the DEFAULT block and _just work_

.changeset/cold-scissors-hunt.md
```diff
@@ -0,0 +1,6 @@
+---
+'@twilio-paste/design-tokens': major
+'@twilio-paste/theme': major
+---
+
+[Design Tokens, Theme] Deprecate the legacy console theme in favor of the default theme.
```
shleewhite, line 6: Do I need 2 separate changesets? Also, is this an adequate message or do I need to add more information?

.changeset/cold-scissors-hunt.md
```diff
@@ -0,0 +1,6 @@
+---
+'@twilio-paste/design-tokens': major
+'@twilio-paste/theme': major
+---
+
+[Design Tokens, Theme] Deprecate the legacy console theme. The default theme should be used instead.
```
shleewhite, line 6: Do I need 2 separate changesets? Also, is this an adequate message or do I need to add more information?

shleewhite: Also does core need to be included?

SiTaggart: Core must be included. 

Theme is technically a minor as it's not breaking. You don't need to do anything.

Design tokens is a major.

Also consider the message you would want to read from the changelog about this? What would you want to know, how would you want to be prepared for it as a consumer.

TheSisb: It looks like the theme package has a breaking change too, check out https://github.com/twilio-labs/paste/pull/2168/files#diff-086a3fa08d1651562b139d92ed171a8e1e9c5af00da5988e754eb3a7d4b7edc7R15-R17

I left a comment on how to resolve the breaking change

packages/paste-theme/src/themeProvider.tsx
```diff
@@ -22,10 +23,11 @@ function getProviderThemeProps(theme: ThemeVariants, customBreakpoints?: string[
...
         breakpoints: customBreakpoints || SendGridTheme.breakpoints,
       };
-    case ThemeVariants.CONSOLE:
+    case DeprecatedThemeVariants.CONSOLE:
+      isDeprecatedTheme(DeprecatedThemeVariants.CONSOLE);
       return {
-        ...ConsoleTheme,
-        breakpoints: customBreakpoints || ConsoleTheme.breakpoints,
+        ...DefaultTheme,
+        breakpoints: customBreakpoints || DefaultTheme.breakpoints,
```
TheSisb, line 30: If you remove these lines, it will call the function and fall into the default case (since there's no return or break)

packages/paste-theme/src/utils/isDeprecatedTheme.ts
```diff
@@ -0,0 +1,11 @@
+import type {DeprecatedThemeVariants} from '../constants';
+
+export const isDeprecatedTheme = (theme: DeprecatedThemeVariants): null => {
+  const themeToDeprecationMessage: Record<DeprecatedThemeVariants, string> = {
```
TheSisb, line 4: Nice!

packages/paste-theme/src/types/themes.ts
```diff
@@ -1,9 +1,7 @@
 import type {DefaultTheme} from '../themes/default';
 import type {SendGridTheme} from '../themes/sendgrid';
-import type {ConsoleTheme} from '../themes/console';
 import type {DarkTheme} from '../themes/dark';
 
 export type SendGridThemeShape = typeof SendGridTheme;
-export type ConsoleThemeShape = typeof ConsoleTheme;
```
TheSisb, line 7: So removing Console things from the Theme package _is_ a breaking change. But we want this to be seemless. Just as how we're pointing `theme="console"` to use the `default` theme, we can do things like
```
export type ThemeShape = typeof DefaultTheme;
export type ConsoleThemeShape = ThemeShape;
```

So we don't import any console stuff, but we continue to export it as default.

SiTaggart: This is a good point, and also affects the theme/inde.ts file too. https://github.com/twilio-labs/paste/pull/2168/files#diff-e12cea4607d991859b3334966fe268a7b9c542d9d643607f2422787dab61618bL3

We export the Console theme from the theme package, so we should also remap that export to the Default theme, so to the consumer, we still `export {DefaultTheme, ConsoleTheme, SendGridTheme, DarkTheme};`

packages/paste-theme/src/themeProvider.tsx
shleewhite: Yeah... I tried to do that but typescript said it expected a return. I think it is because I have  line 27 where I run `isDeprecatedTheme`

packages/paste-theme/src/types/themes.ts
shleewhite: Ok! I added the exports back in this file and the `paste-theme/src/themes/index.ts`

.changeset/cold-scissors-hunt.md
shleewhite: Fixed the breaking changes!

TheSisb (approved):
👍

packages/paste-theme/src/utils/isDeprecatedTheme.ts
nkrantz: This is a cool way of writing the error message

nkrantz (approved):
Nice deleting things 🔪 looks good and it's not a blocker, but should we add a brief message in the Theme Package docs page about this change and UDL? In case someone tries to use the console theme and goes to the theme docs for more info about _why_ it was removed

PR #2194, opened 2022-02-08
fix(link-button): updates default style for link buttons

zahnster:
* Updates default size prop for Link buttons to 'default' from 'reset', to ensure proper font & line height sizing.

### DIFF
--- a/.changeset/lucky-lies-travel.md
+++ b/.changeset/lucky-lies-travel.md
@@ -3,4 +3,4 @@
 '@twilio-paste/core': patch
 ---
 
-Updates default size for link buttons to 'default' from 'reset'
+Updates 'reset' styles to set inherited font size
--- a/.changeset/lucky-lies-travel.md
+++ b/.changeset/lucky-lies-travel.md
@@ -1,0 +1,6 @@
+---
+'@twilio-paste/button': patch
+'@twilio-paste/core': patch
+---
+
+Updates default size for link buttons to 'default' from 'reset'
--- a/packages/paste-core/components/button/src/index.tsx
+++ b/packages/paste-core/components/button/src/index.tsx
@@ -30,14 +30,14 @@ import {ResetButton} from './ResetButton';
 const AnimatedBox = animated(Box);
 
 // If size isn't passed, come up with a smart default:
-// - 'reset' for reset
+// - 'reset' for variant 'link'
 // - 'icon' if there's 1 child that's an icon
 // - 'default' otherwise
 const getButtonSize = (variant: ButtonVariants, children: React.ReactNode, size?: ButtonSizes): ButtonSizes => {
   let smartSize: ButtonSizes = 'default';
   if (size != null) {
     smartSize = size;
-  } else if (variant === 'reset') {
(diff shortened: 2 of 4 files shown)


packages/paste-core/components/button/src/index.tsx
```diff
@@ -30,14 +30,14 @@ import {ResetButton} from './ResetButton';
...
-// - 'reset' for variant 'link'
+// - 'reset' for reset
 // - 'icon' if there's 1 child that's an icon
 // - 'default' otherwise
 const getButtonSize = (variant: ButtonVariants, children: React.ReactNode, size?: ButtonSizes): ButtonSizes => {
   let smartSize: ButtonSizes = 'default';
   if (size != null) {
     smartSize = size;
-  } else if (variant === 'link' || variant === 'destructive_link' || variant === 'reset') {
+  } else if (variant === 'reset') {
```
SiTaggart, line 40: So the reason you can't use the default size for links is that default sizing comes with padding. 

image

As I mentioned in Slack, the real fix is to make sure reset, at the very least _sets a font size and line height_ by explicitly inheriting them from their surrounding text.

Which I suspect is exactly the behaviour you would get from an Anchor component too. It inherits the parent font and line height sizes

SiTaggart (changes requested):

packages/paste-core/components/button/src/index.tsx
zahnster: I see, I see. I did see an inherit font size property, but that was exclusively on the ResetButton component, and not in the broader base styles for the size prop. Adjusted the code, hopefully this is closer to what you had in mind.

.changeset/lucky-lies-travel.md
```diff
@@ -0,0 +1,6 @@
+---
+'@twilio-paste/button': patch
+'@twilio-paste/core': patch
+---
+
+Updates default size for link buttons to 'default' from 'reset'
```
andipants12, line 6: Could you please update this to reflect the change we pivoted to?

andipants12 (commented):
Looks great, just need to update the changeset message to reflect the direction we decided to go for a fix. 💯

.changeset/lucky-lies-travel.md
zahnster: _great_ point, yes

SiTaggart (approved):
Note worth mentioning. Our button stories are not super useful for link variant. 

I don't think we really intend on using link style button with actual sizes, so the "it's just a link button" case isn't very well captured in the VRT. But it does what it should do

zahnster:
> Note worth mentioning. Our button stories are not super useful for link variant.
> 
> I don't think we really intend on using link style button with actual sizes, so the "it's just a link button" case isn't very well captured in the VRT. But it does what it should do

I noticed this 😅 . Locally, I verified through the website, as the docs examples are a bit more authentic to the expected implementations. I have _thoughts_ about our Button stories, but that's for another day...

andipants12 (approved):
:shipit:

PR #2200, opened 2022-02-09
feat(data-visualization): add package

TheSisb:
This PR adds a new package called `@twilio-paste/data-visualization-library`.  This package will serve as a container for future data viz utilities. In its current form, we export a helpful React hook which exports a Highcharts option object with Paste theme values applied. This will help developers build themed and accessible charts in Highcharts more easily. I've added a couple stories showing how it works.


packages/paste-libraries/data-visualization/package.json
```diff
@@ -0,0 +1,53 @@
...
+  },
+  "devDependencies": {
+    "@twilio-paste/animation-library": "^0.3.1",
+    "@twilio-paste/color-contrast-utils": "^2.0.0",
+    "@twilio-paste/design-tokens": "^7.0.0",
+    "@twilio-paste/styling-library": "^0.3.5",
+    "@twilio-paste/theme": "^6.0.0",
+    "@twilio-paste/types": "^3.1.3",
+    "highcharts": "^9.3.3",
+    "highcharts-react-official": "^3.1.0",
```
TheSisb, line 49: I mark these two as devDeps because I want to use them in the story file. Highcharts alone is a peerDep because we expect consumers to only need Highcharts to use the hook.

packages/paste-libraries/data-visualization/src/usePasteHighchartsTheme.tsx
```diff
@@ -0,0 +1,100 @@
...
+export const usePasteHighchartsTheme = (options: Highcharts.Options): Highcharts.Options => {
+  const context = React.useContext(ThemeContext) as ThemeShape;
+  if (!context) {
+    throw new Error(
+      '[usePasteHighchartsTheme]: must be used within the @twilio-paste/theme provider. https://paste.twilio.design/introduction/for-engineers/manual-installation#setting-up-the-theme-provider'
+    );
+  }
+
+  return merge(options, {
+    colors: Object.values(context.dataVisualization),
```
TheSisb, line 23: We originally only designed for these sequential color values, but I took the liberty of applying our text and font styles elsewhere to make the charts look more branded. The nice thing about this package is we can update the styles as needed and ship it as a patch, which folks can get for free.

packages/paste-libraries/data-visualization/src/usePasteHighchartsTheme.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+const merge = require('deepmerge');
+
+/**
+ * A hook to apply the Paste theme to Highcharts options.
+ *
+ * @param {Highcharts.Options} options - options to be provided into Highcharts for rendering.
+ * @returns {Highcharts.Options} optionsWithPasteTheme - Passed options with deepmerged Paste Theme
+ */
+export const usePasteHighchartsTheme = (options: Highcharts.Options): Highcharts.Options => {
```
TheSisb, line 14: I needed to take an `options` object as argument here because the theme in Highcharts is applied over the same object that configures chart data. I was running into issues where my styles weren't applying because the options object defined different values for certain keys. So we need to take the options object, deepmerge our styles over them, and return a new object which is passed into Highcharts.  This works pretty well, as can be seen in the stories.

TheSisb, line 14: We need this to be a React hook so that the styles update upon toggling between light and dark theme. Can't just export a static object, unless we instruct developers to swap it themselves which is much less ergonomic and more error prone.

packages/paste-libraries/data-visualization/stories/index.stories.tsx
```diff
@@ -0,0 +1,116 @@
...
+    text: 'Solar Employment Growth by Sector, 2010-2016',
+  },
+  subtitle: {
+    text: 'Source: thesolarfoundation.com',
+  },
+  series: [
+    {
+      name: 'Installation',
+      // @ts-expect-error copied from official highcharts docs
+      data: [43934, 52503, 57177, 69658, 97031, 119931, 137133, 154175],
```
andipants12, line 27: Looking at the beta docs for TS and Highcharts, they[ suggest casting the data as the correct type](https://www.highcharts.com/docs/advanced-chart-features/highcharts-typescript-declarations#debugging). Did you try that before adding the expect error statement?

TheSisb: Yea I tried that first but kept getting a slew of other issues, even after doing `as unknown) as SeriesXrangeOptions` etc. 🤷‍♂️

SiTaggart (commented):
Is it worth adding tests for the hook? Even snap shotting the returned data? Just to preserve the expected outcome, given a certain highcharts config doesn't accidently regress over time.

SiTaggart (commented):
Overall, a pretty decent place to call this v1. Some tests would be nice and an open question on chart type coverage

packages/paste-libraries/data-visualization/stories/index.stories.tsx
```diff
@@ -0,0 +1,116 @@
...
+/* eslint-enable */
+
+import {usePasteHighchartsTheme} from '../src';
+
+// eslint-disable-next-line import/no-default-export
+export default {
+  title: 'Libraries/data-visualization',
+} as Meta;
+
+const lineChartOptions: Highcharts.Options = {
```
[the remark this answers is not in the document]
SiTaggart: Yeah, I get it, but I do still think it's worthwile expanding coverage.

For the hook, it might seem like a safe assumption that the chart implementation would cover it, but that assumes you're not doing something unexpected with it before it's fed to the chart. we'd just be ensuring the contract between the input and ouput of the hook stays the same, which might become super important later if we tweak anything in the guts of the function. They are peer dependencies to each other after all, so they should be treated with isolation.

For the charts, I think even though they are tested, it is probably worth while to ensure the changes we make to chart styles have the expected results on chart types that we aren't using in stories. Just to be extra cautious.

SiTaggart (approved):
Code looks fine. I'm more than fine with copy pasta examples from highcharts too with ts-ignores.

One final thing, which I didn't spot, we will need to add highcharts as a root dependency because Storybook is a root application. Right now it works when import highcharts into a story because yarn is hoisting the dep from the data visualisation package to the root of the repo.

gloriliale (approved):
this makes sense, looks really, really awesome and the stories are fantastic. I'm glad you included what it looks like w/o the hook as well for easy comparison. People are going to be happyyyyyyy 🎉

serifluous:
A few things:

1. Tooltip and pie chart font need to use Inter
2. The border in between each pie chart slice should be thicker (closer to what's default in HighCharts) so there's clear differentiation between each slice.
3. We could consider shifting the alias values so we start with some of the more vibrant colors first. For example, I was thinking we could get the alias for `color-data-visualization-4` to be the new `color-data-visualization-1`, and then shift the sequence accordingly.

@katieporter I chatted with Shadi, and since this PR is really just to add the tokens, with a few fixes to fix really obvious things, he's going to create a ticket for followup work to polish up the theme. We can address these comments ^ when that polish work happens.

zahnster (approved):
Praise: Thank you for the PR notes and the tests! This looks good to me.

📊 📈 👍🏼 🎉

serifluous (approved):

PR #2208, opened 2022-02-11
chore: loosen eng lead review requirement to only core

SiTaggart:
This PR modifies the Code owner requirements such that Eng Leads, @TheSisb, @zahnster and myself, only require a review for things that are core library specific.

Rules are now:

- any file must have 2 reviews from different engineers.
- anything file in design-tokens, core, customization, style-props and theme must also have an eng lead
- icons must also have a design review
- design tokens must also have a design review

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,7 +1,16 @@
-* @twilio-labs/design-system-eng-leads
+* @twilio-labs/design-system-eng
 
-# Paste core - round robin someone from the team and a lead for review
-/packages/ @twilio-labs/design-systems-eng @twilio-labs/design-system-eng-leads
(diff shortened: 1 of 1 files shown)


TheSisb (approved):

.github/CODEOWNERS
```diff
@@ -1,7 +1,16 @@
-* @twilio-labs/design-system-eng-leads
+* @twilio-labs/design-system-eng
```
TheSisb, line 1: Did you mean to change this?

SiTaggart: Yup, flipping it round. 

Every code change must have a engineering review. But only specific things must _also_ have a leads review.

shleewhite (approved):

PR #2210, opened 2022-02-14
chore: enable kodiak merge queue

SiTaggart:
Enable the kodiak merge queue.

Add a `🚀 merge it!` label to add a pr to the queue once you are happy with it.

Add `Status: WIP` or `Status: Do Not Merge` to block / override.

For the future, I've added a rule for dependencies via Renovate

### DIFF
--- a/.kodiak.toml
+++ b/.kodiak.toml
@@ -1,0 +1,13 @@
+version = 1
+
+[merge]
+automerge_label = "🚀 merge it!"
+blocking_labels = ["Status: WIP", "Status: Do Not Merge"]
+method = ["rebase_fast_forward", "rebase", "squash"]
+
+[merge.automerge_dependencies]
+# only auto merge "minor" and "patch" version upgrades.
+# do not automerge "major" version upgrades.
+versions = ["minor", "patch"]
+usernames = ["renovate"]
+
--- a/.github/opened-pr-labeler.yml
+++ b/.github/opened-pr-labeler.yml
@@ -1,0 +1,5 @@
+# https://github.com/actions/labeler
+
+# Add 'Status: Do Not Merge' label to any newly opened PR
+"Status: Do Not Merge":
+  - "**/*"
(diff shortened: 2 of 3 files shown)


.kodiak.toml
```diff
@@ -0,0 +1,13 @@
+version = 1
+
+[merge]
+automerge_label = "🚀 merge it!"
+blocking_labels = ["Status: WIP", "Status: Do Not Merge"]
```
gloriliale, line 5: What would be a situation in which to use `WIP` and not `Do Not Merge`?

SiTaggart: 🤷🏼 Just giving you options I guess.

shleewhite (approved):
STOKED

.kodiak.toml
```diff
@@ -0,0 +1,13 @@
...
+method = ["rebase_fast_forward", "rebase", "squash"]
+
+[merge.automerge_dependencies]
```
TheSisb, line 8: praise: this will come in clutch with the new renovate bot's 40+ PRs

.kodiak.toml
```diff
@@ -0,0 +1,13 @@
+version = 1
+
+[merge]
```
TheSisb, line 3: thought (not-blocking): have you given any thought to this config toggle: https://kodiakhq.com/docs/config-reference#mergeprioritize_ready_to_merge

While this does some "unfairness", this likely saves build cycles by merging in the logical fastest way and not blocking a chain of PRs on a fail for whatever is ahead in the queue

TheSisb (approved):

gloriliale (approved):

.kodiak.toml
SiTaggart: I don't necessarily have a good sense of how they actually deal with the queue yet. It sounds like they don't just keep rebasing and rebuilding everything, they only do it one at a time based on whats next. So hopefully they shouldn't be spinning up CI all the time for no reason.

If that's true, skipping the queue might actually introduce more CI usage becuase something in the queue that is currently building, might then have to be rebuilt if something jumps ahead of it.

Lets keep it in our back pocket. There are likely a lot of config options that could help optimise our queue once we get a good picture on how it actually works for us.

PR #2229, opened 2022-02-18
fix(website): fixes sidebar page selection on page load

TheSisb:
Currently, loading the Paste docs on a nested route doesn't expand the sidebar or select the current active path (try: https://paste.twilio.design/components/alert-dialog/). I investigated the root cause to be that we were relying on the `window.location` object which isn't available when we SSR our pages. I migrated our functions to now rely on `@reach/router` which is being used by Gatsby to get the correct location information. This fixes the issue of the route not expanding the sidebar in production builds.

Update: window.location is still being used opportunistically to natively support client side behaviors, but will fall back to use Reach Router in cases where the window object is unavailable (such as Gatsby SSG). Resolves client-side environments that lack Reach (such as Storybook).


packages/paste-website/src/utils/RouteUtils.ts
```diff
@@ -33,7 +27,7 @@ export function getNameFromPackageName(packageName: string): string {
 
 // Returns "Aspect Ratio" from "@twilio-paste/aspect-ratio"
 export function getHumanizedNameFromPackageName(packageName: string): string {
-  return startCase(toLower(packageName.replace(PASTE_PACKAGE_PREFIX, '')));
+  return startCase(toLower(getNameFromPackageName(packageName)));
```
gloriliale, line 30: praise: good switch here to keep things nice and DRY 🎉

packages/paste-website/stories/SidebarNavigation.stories.tsx
```diff
@@ -8,79 +9,87 @@ import {SidebarDisclosureButton} from '../src/components/site-wrapper/sidebar/si
...
-            </SidebarAnchor>
-            <SidebarAnchor level={2} to="https://www.google.com">
-              Level 2 anchor
-            </SidebarAnchor>
-          </SidebarDisclosureContent>
-        </SidebarDisclosure>
-      </SidebarDisclosureContent>
-    </SidebarDisclosure>
-  </Box>
+  <Router>
```
gloriliale, line 12: issue: I tried to open storybook from my browser and got an error for these stories that looked like this:
image
I'm confused by the error since it seems to be saying that there's another, bigger error?

gloriliale (commented):
hooray for no more 🐛 sidebar! I had a few q's and noticed something strange in the online storybook. Gonna look at it locally now tho.

cypress/integration/sidebar-navigation/index.spec.ts
```diff
@@ -56,6 +56,19 @@ describe('Sidebar navigation', () => {
   });
 });
 
+describe('Sidebar opens correct section on first load', () => {
+  it('opens the sidebar category and selects the page correctly on first load', () => {
```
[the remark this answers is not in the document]
zahnster: answer: it will happen on first load on any deep link, which you can see here: https://paste.twilio.design/primitives/box

The nav will work as you navigate through the website, but I've consistently observed deep links / refreshes dropping the nav context. Test seems good from my perspective.

packages/paste-website/src/components/shortcodes/feedback-popover/index.tsx
```diff
@@ -13,42 +13,45 @@ import {ThumbsDownIcon} from '@twilio-paste/icons/esm/ThumbsDownIcon';
 import {SupportIcon} from '@twilio-paste/icons/esm/SupportIcon';
 import {trackCustomEvent} from 'gatsby-plugin-google-analytics';
 import {SimpleStorage} from '../../../utils/SimpleStorage';
-import {getCurrentPathname} from '../../../utils/RouteUtils';
+import {useLocationPathname} from '../../../utils/RouteUtils';
```
zahnster, line 16: suggestion: This component directly interfacing with the route utils is causing complexities when it comes to trying to render in isolation (i.e. in Storybook). Would it make more sense to pass in the current location from a prop on the component call?

shleewhite (approved):
WooO!!

cypress/integration/sidebar-navigation/index.spec.ts
```diff
@@ -56,6 +56,19 @@ describe('Sidebar navigation', () => {
...
+
+    const componentsCategoryButton = cy.get(`[data-cy="sidebar-disclosure-button-components"]`);
+    componentsCategoryButton.should('have.attr', 'aria-expanded', 'true');
+
+    const componentsCategoryList = cy.get(`[data-cy="sidebar-disclosure-content-components"]`);
+    const alertDialogLink = componentsCategoryList.contains('Alert Dialog');
+    alertDialogLink.should('have.attr', 'aria-current', 'page');
+  });
+});
+
```
SiTaggart, line 71: praise: zing!

packages/paste-website/src/components/site-wrapper/sidebar/SidebarAnchor.tsx
```diff
@@ -67,12 +68,17 @@ const StyledSidebarAnchor = styled(Link, {
...
-    </StyledSidebarAnchor>
-  </SidebarItem>
-);
+const SidebarAnchor: React.FC<SidebarAnchorProps> = ({children, level, to}) => {
+  const pathname = useLocationPathname();
+  const pathnameWithoutTrailingSlash = pathname.endsWith('/') ? pathname.slice(0, -1) : pathname;
+
+  return (
+    <SidebarItem>
+      <StyledSidebarAnchor level={level} to={to} aria-current={pathnameWithoutTrailingSlash === to ? 'page' : false}>
```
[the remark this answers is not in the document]
SiTaggart: Oh sweet!

SiTaggart (approved):
love this

PR #2234, opened 2022-02-24
feat(toast): add i18n props

gloriliale:
- [x]  add `i18nDismissLabel` prop for dismiss button
- [x]  add `i18nErrorLabel`, `i18nNeutralLabel`, `i18nSuccessLabel`, `i18nWarningLabel` props for icons of variants
- [x]  add i18n story
- [x]  add tests for default and custom i18n props
- [x]  add example to website
- [x]  add prop to Popover prop list on website

### DIFF
--- a/packages/paste-core/components/toast/__tests__/index.spec.tsx
+++ b/packages/paste-core/components/toast/__tests__/index.spec.tsx
@@ -56,6 +56,92 @@ describe('Toast', () => {
     });
   });
 
+  describe('i18n', () => {
+    it('should have default dismiss button text', () => {
+      render(
+        <Toast variant="neutral" onDismiss={onDismissMock}>
+          This is a toast
+        </Toast>
+      );
+      const dismissButton = screen.getByRole('button', {name: 'Dismiss toast'});
+      expect(dismissButton).toBeDefined();
+    });
+
+    it('should use i18nDismissLabel for dismiss button text', () => {
+      render(
+        <Toast variant="neutral" i18nDismissLabel="Cerrar notificacion" onDismiss={onDismissMock}>
+          Esta es una notificacion
+        </Toast>
+      );
+      const dismissButton = screen.getByRole('button', {name: 'Cerrar notificacion'});
+      expect(dismissButton).toBeDefined();
+    });
+
+    it('should have default error variant icon text', () => {
+      render(<Toast variant="error">This is a toast</Toast>);
+      const iconText = screen.getByText('(error)');
+      expect(iconText).toBeDefined();
+    });
+
+    it('should have default neutral variant icon text', () => {
+      render(<Toast variant="neutral">This is a toast</Toast>);
+      const iconText = screen.getByText('(information)');
+      expect(iconText).toBeDefined();
+    });
+
+    it('should have default success variant icon text', () => {
+      render(<Toast variant="success">This is a toast</Toast>);
+      const iconText = screen.getByText('(success)');
+      expect(iconText).toBeDefined();
+    });
+
+    it('should have default warning variant icon text', () => {
+      render(<Toast variant="warning">This is a toast</Toast>);
+      const iconText = screen.getByText('(warning)');
+      expect(iconText).toBeDefined();
+    });
+
+    it('should use i18n prop for error variant icon text', () => {
+      render(
+        <Toast variant="error" i18nErrorLabel="(erreur)">
+          This is a toast
+        </Toast>
+      );
+      const iconText = screen.getByText('(erreur)');
+      expect(iconText).toBeDefined();
+    });
+
+    it('should use i18n prop for neutral variant icon text', () => {
+      render(
+        <Toast variant="neutral" i18nNeutralLabel="(informacion)">
+          This is a toast
+        </Toast>
+      );
+      const iconText = screen.getByText('(informacion)');
+      expect(iconText).toBeDefined();
+    });
+
+    it('should use i18n prop for success variant icon text', () => {
+      render(
+        <Toast variant="success" i18nSuccessLabel="(éxito)">
+          This is a toast
+        </Toast>
+      );
+      const iconText = screen.getByText('(éxito)');
+      expect(iconText).toBeDefined();
+    });
+
+    it('should use i18n prop for warning variant icon text', () => {
+      render(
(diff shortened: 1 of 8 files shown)


packages/paste-core/components/toast/__tests__/index.spec.tsx
```diff
@@ -56,6 +56,92 @@ describe('Toast', () => {
...
+      expect(iconText).toBeDefined();
+    });
+
+    it('should have default warning variant icon text', () => {
+      render(<Toast variant="warning">This is a toast</Toast>);
+      const iconText = screen.getByText('(warning)');
+      expect(iconText).toBeDefined();
+    });
+
+    it('should use i18n prop for error variant icon text', () => {
```
shleewhite, line 104: praise: great test coverage!!

shleewhite (approved):
🎉 Looks great!

packages/paste-website/src/pages/components/toast/index.mdx
```diff
@@ -457,11 +475,16 @@ If you choose to handle state entirely yourself, you can use `<Toaster />` on it
...
+| children         | `ReactNode`  |                                                                                           | null    |
+| onDismiss?       | `() => void` | Create dismissable toast by providing an onDismiss callback                               | null    |
+| variant          | `string`     | `error` / `neutral` / `sucess` / `warning`                                                | null    |
+| i18nDismissLabel | `string`     | Label for the dismiss button in a dismissable toast. The default value is 'Dismiss toast' |
+| i18nErrorLabel   | `string`     | Icon label text for the `error` variant. The default value is '(error)'                   |
+| i18nNeutralLabel | `string`     | Icon label text for the `neutral` variant. The default value is '(information)'           |
+| i18nSuccessLabel | `string`     | Icon label text for the `success` variant. The default value is '(success)'               |
+| i18nWarningLabel | `string`     | Icon label text for the `warning` variant. The default value is '(warning)'               |
```
zahnster, line 487: Suggestion: Can we update this so the default field is populated for the i18n props? It'll be easier to scan that way.

gloriliale: Done 🎉

zahnster (approved):
Great work! 🚢

packages/paste-core/components/toast/__tests__/index.spec.tsx
```diff
@@ -56,6 +56,92 @@ describe('Toast', () => {
...
+    });
+
+    it('should use i18n prop for neutral variant icon text', () => {
+      render(
+        <Toast variant="neutral" i18nNeutralLabel="(informacion)">
+          This is a toast
+        </Toast>
+      );
+      const iconText = screen.getByText('(informacion)');
+      expect(iconText).toBeDefined();
```
SiTaggart, line 121: thought(non-blocking): I actually have no good solution for this, it's just an observation. 

It terms of a test I wonder how effective this will be long term. We essentially test, "does _an_ element with this string exist?" Which proves the prop works, but might not prove the prop does what it's semantically meant to do.  To explain that further, I could theoretically move the screen reader only text to anywhere in the toast composition and the test would still pass, but the icon would no longer with described with a translated string.

SiTaggart: Further expansion, the dismiss button tests are really good examples. It tests that there is a button, it is a dismiss button and it is translated.

There really isn't a good "icon" specific alternative

packages/paste-core/components/toast/src/Toast.tsx
```diff
@@ -78,11 +62,20 @@ const Toast = React.forwardRef<HTMLDivElement, ToastProps>(
       }
     }, [setFocus]);
 
+    // uses variant to return the correct i18n label prop for that variant
+    const i18nVariants = {
+      error: i18nErrorLabel,
+      neutral: i18nNeutralLabel,
+      success: i18nSuccessLabel,
+      warning: i18nWarningLabel,
+    };
```
SiTaggart, line 71: praise: smart, I like

packages/paste-core/components/toast/src/types.ts
```diff
@@ -18,6 +18,14 @@ export interface ToastProps extends React.HTMLAttributes<HTMLDivElement>, Pick<B
...
    */
   setFocus?: boolean;
+  /**
+   * These labels support i18n
+   */
+  i18nDismissLabel?: string;
+  i18nErrorLabel?: string;
+  i18nNeutralLabel?: string;
+  i18nSuccessLabel?: string;
+  i18nWarningLabel?: string;
```
SiTaggart, line 28: suggestion: it might seem redundant, but each one of these having their own tsdoc comment would be beneficial. These provide comments in VS code when you hover over the prop to let you know what the prop is for. Right now the last 4 wouldn't get a VS Code hint, but `i18nDismissLabel` would say "These labels support i18n", which in isolation might not feel very useful to the developer using it.

For example, hovering over `onDismiss` 
image

packages/paste-core/components/toast/__tests__/index.spec.tsx
shleewhite: For sure, I think the only way we can check that specifically is if we add a `data` attribute somewhere and then checking if the text exists within the element with that attribute.

It could go on the `MediaFigure` that wraps the icon & text... or we could go back to using the `title` attribute on the icon and use the `data-paste-element` attribute on the icon... or add a data attribute on the `ScreenReaderOnly` element..,

Is there a reason why we decided to use `ScreenReaderOnly` instead of the `title` attribute on the icon for i18n?

packages/paste-core/components/toast/src/types.ts
gloriliale: ooo thanks for the catch on this; we don't employ it very often that I've seen in our types so I'm glad I know more about it now. It's been implemented! I used the descriptions from the anatomy table in the docs.

packages/paste-core/components/toast/__tests__/index.spec.tsx
gloriliale: I'll be merging this as the conversation continues, but I appreciate everyone's ideas and input, especially since it's bound to come up again. Thank you!

PR #2237, opened 2022-02-24
Chore: esbuild  upgrade and remove form package

SiTaggart:
- Update esbuild from https://github.com/twilio-labs/paste/pull/2166
- Remove form package to resolve some weird 'I can't find textarea' build failure, because it was trying to get it via the form package. We deprecated a long time ago. Time to remove it entirely. Issue resolved.


.changeset/angry-ants-reply.md
```diff
@@ -0,0 +1,82 @@
...
+'@twilio-paste/nextjs-template': patch
+'@twilio-paste/style-props': patch
+'@twilio-paste/theme': patch
+'@twilio-paste/theme-designer': patch
+'@twilio-paste/token-contrast-checker': patch
+'@twilio-paste/types': patch
+'@twilio-paste/utils': patch
+---
+
+esbuild dependency update for compiled source code
```
shleewhite, line 82: question: what does this mean?

shleewhite (approved):

shleewhite:
Looks good, but why are all the packages larger now?

.changeset/angry-ants-reply.md
TheSisb: suggestion: "Updated a build dependency (esbuild) which changes the output of our builds slightly, without materially changing anything about the code."

packages/paste-core/components/form/stories/form.stories.tsx
```diff
@@ -4,29 +4,38 @@ import {Text} from '@twilio-paste/text';
 import {Box} from '@twilio-paste/box';
 import {Grid, Column} from '@twilio-paste/grid';
 import {Anchor} from '@twilio-paste/anchor';
-import {Checkbox, FormLabel, FormInput, Select, Option, Radio, FormTextArea, CheckboxGroup, RadioGroup} from '../src';
-import type {FormInputProps, SelectProps, FormTextAreaProps, CheckboxGroupProps, RadioGroupProps} from '../src';
+import {Checkbox, CheckboxGroup} from '@twilio-paste/checkbox';
```
TheSisb, line 7: nitpick (non-blocking, out of scope): Keeping the `components/form` folder for a story file can be misleading to future developers of Paste. How would we feel about moving these to the respective package story files?

tools/build/esbuild.js
```diff
@@ -96,6 +97,8 @@ function build(packageJson) {
       minify: true,
       format: 'esm',
       outfile: outFileESM,
+      // Needed to bug with replacing require with import statements https://github.com/evanw/esbuild/issues/566
```
TheSisb, line 100: suggestion: typo in the comment? "Needed to fix a bug..."

TheSisb (approved):
Finally. And thank you.

TheSisb:
> Looks good, but why are all the packages larger now?

The way esbuild compiles the files has changed so that they need a little more boilerplate. It kinda sucks, but isn't too big a deal after gzipping itll all get compressed away.

zahnster:
Worthy of one of these

eee

.changeset/angry-ants-reply.md
SiTaggart: Yeah, what he said. Becuase we change the compiled code _ever so slightly_ with this esbuild upgrade, we need to force publish all packages even though nothing actually changes

packages/paste-core/components/form/stories/form.stories.tsx
SiTaggart: I struggled with this. The reason we have this big story is more to capture the spacing between the different type of controls, vertically and horizontally. 

We had a collapsing margin bug once, so we need to ensure the vertical rythmn between all the form controls doesn't degrade. 

But that doesn't give it a good home as it uses all form components... 😬

PR #2265, opened 2022-03-04
chore(chromatic): add chromatic delay to data-viz to allow load

zahnster:
* Adds Chromatic Delay for Data Visualization stories, to allow them to animate in before taking a snapshot
  * Should resolve false positive VRT diffs with Data Viz stories in Chromatic


SiTaggart (approved):
FINALLY! 🎉

shleewhite (approved):
Heck yeah!

PR #2269, opened 2022-03-04
fix: custom plugin to fix esbuild esm imports

SiTaggart:
The plugin we used to fix the esbuild issue with require statements in esm compiled packages reintroduced an issue with incorrectly importing defaults from json files, using `import * as x from 'x'`.

This is because it blanketly transformed all import statements, even though the bug we were trying to fix was just require statements.

This customized version of the plugin only modifies imports that are require statements and leaves already correct import statements alone.


tools/build/build-plugins/EsmExternals.js
```diff
@@ -0,0 +1,60 @@
...
+
+  // Escape characters with special meaning either inside or outside character sets.
+  // Use a simple backslash escape when it’s always valid, and a `\xnn` escape when the simpler form would be disallowed by Unicode patterns’ stricter grammar.
+  return string.replace(/[|\\{}()[\]^$+*?.]/g, '\\$&').replace(/-/g, '\\x2d');
+}
+
+const NAME = 'esm-externals';
+const NAMESPACE = NAME;
+
+// this is a modified version of https://github.com/remorses/esbuild-plugins/tree/master/esm-externals
```
zahnster, line 16: praise: nice commenting, much appreciated

tools/build/esbuild.js
```diff
@@ -1,25 +1,6 @@
...
-    });
-
-    // Change all .es6 design-token imports to .common
-    build.onResolve({filter: /\/tokens.es6$/}, ({path}) => {
-      return {path: path.replace('.es6', '.common'), external: true};
-    });
-  },
-};
+const {EsmExternalsPlugin} = require('./build-plugins/EsmExternals');
+const {PasteCJSResolverPlugin} = require('./build-plugins/PasteCJSResolver');
```
zahnster, line 3: praise: nice code organization

zahnster (approved):
Obama Not Bad meme

shleewhite (approved):

zahnster:
Great additions, and look at all those green checks! ✅

PR #2274, opened 2022-03-07
feat(label): remove hardcoded strings for i18n

shleewhite:
Upon investigation, the original ticket didn't account for when there is a RequiredDot in a legend.

When the RequiredDot is in a label for an an input, it should not have text because the input has the required attribute (hence why the default value of i18nLabel prop on RequiredDot is ''.

When RequiredDot is in a legend for a fieldset, it should have the text 'required' because legends and fieldsets do not have the required attribute.

This PR includes:
**Label**
- [x] i18nLabel prop to RequiredDot
- [x] i18nRequiredLabel prop to Label
- [x] adds an i18n story
- [x] updates existing required dot test to just check the element is in the dom
- [x] tests that the default text is empty and that i18nRequiredLabel changes the text 
- [x] updates the prop table on the website
- [ ] adds an i18n example to the website 

**Checkbox**
- [x] i18nRequiredLabel prop to CheckboxGroup
- [x] tests that the default text is '(required)' and that i18nRequiredLabel changes the text 
- [x] updates the CheckboxGroup prop table on the website
- [x] adds an i18n example to the website 

**RadioGroup**
- [x] i18nRequiredLabel prop to RadioGroup
- [x] tests that the default text is '(required)' and that i18nRequiredLabel changes the text 
- [x] updates the RadioGroup prop table on the website
- [x] adds an i18n example to the website 

**InlineControlGroup**
- [x] i18nRequiredLabel prop
- [ ] tests that the default text is '' and that i18nRequiredLabel changes the text 

**Contributing to Twilio**


packages/paste-core/components/checkbox/__tests__/checkbox.test.tsx
```diff
@@ -75,12 +75,15 @@ describe('Checkbox', () => {
   });
 
   it('should render a required dot', () => {
-    const {getByText} = render(
+    render(
```
SiTaggart, line 78: praise: nice little tidy up to best practives here

packages/paste-core/components/label/src/Label.tsx
```diff
@@ -15,10 +15,24 @@ export interface LabelProps extends React.LabelHTMLAttributes<HTMLLabelElement>,
...
+  (
+    {
+      as = 'label',
+      marginBottom,
+      required,
+      disabled,
+      children,
+      variant,
+      element = 'LABEL',
+      i18nRequiredLabel = '',
```
SiTaggart, line 31: question: is it safe to say that if the label is rendered as a label `i18nRequiredLabel` should be empty by default, but if it is rendered as legend `i18nRequiredLabel` should be `(required)` by default. Is that an over simplification?

SiTaggart (approved):

packages/paste-core/components/label/src/Label.tsx
shleewhite: Oh! thats exactly correct.

I just had it default to nothing and then checkboxGroup and radioGroup set it to '(required)'.

zahnster (approved):
Looks good 👍🏼

PR #2275, opened 2022-03-08
chore(applitools-landing): add vrt to landing page; update config

zahnster:
* Adds Applitools VRT validation to Docs Site Landing page
* Updates Applitools Config to not fail Cypress when a diff is detected. We do not need to fail the overall Cypress check - we have a distinct check for Applitools, which will fail on a diff, and resolve once the baseline is accepted. Cypress won't re-run on a baseline acceptance, so Applitools should not cause it to fail.


SiTaggart (approved):

shleewhite (approved):
LGTM

PR #2282, opened 2022-03-09
fix(react-spring-raf): update react-spring deps to resolve raf overload

zahnster:
* Updates dependencies for React Spring to resolve [this observed issue](https://github.com/twilio-labs/paste/discussions/2281) with RequestAnimationFrame call overload:
  * react-spring/shared
  * react-spring/web
* Updates Skeleton Loader to use a CSS animation instead of React Spring, to resolve issue with an infinite looped animation freezing its Jest test.
  * Updates Skeleton Loader to use tokens for colors, to better adapt to theming & dark mode.

When tested before/after on any component using @twilio-paste/animation-library (such as Alert Dialog, Badge, Toast, etc) I see the unnecessary calls to RAF no longer happening.


shleewhite (approved):

askel4dd:
Is there any blocker to merge this PR?

nkrantz (approved):

netlify.toml
```diff
@@ -4,8 +4,7 @@
   framework = "#custom"
 
 [context.deploy-preview]
-  # only do a preview deply when there is a website change
-  ignore = "git diff --quiet main...HEAD -- packages/paste-website packages/paste-theme-designer ':!packages/**/__tests__/*' ':!packages/paste-website/stories'"
+  ignore = "git diff --quiet main...HEAD -- packages/paste-core packages/paste-website packages/paste-theme-designer ':!packages/**/__tests__/*' ':!packages/paste-website/stories'"
```
TheSisb, line 7: question: what does adding paste-core do here?

packages/paste-core/components/skeleton-loader/__tests__/index.spec.tsx
```diff
@@ -118,8 +118,10 @@ describe('SkeletonLoader', () => {
   describe('Accessibility', () => {
     it('Should have no accessibility violations', async () => {
       const {container} = render(<SkeletonLoader />);
+
       const results = await axe(container);
-      expect(results).toHaveNoViolations();
+
+      await waitFor(async () => expect(await results).toHaveNoViolations());
```
TheSisb, line 124: thought: `await results` might not be needed, since results is already an awaited result on line 122. 

question: why do we need the rest of the wrapping async code?

packages/paste-core/components/skeleton-loader/src/index.tsx
```diff
@@ -1,32 +1,23 @@
...
-
+const SkeletonLoaderInner = styled.div({
+  position: 'absolute',
+  top: 0,
+  bottom: 0,
+  left: 0,
+  right: 0,
+  background:
+    'linear-gradient(90deg, transparent, rgba(255, 255, 255, 0.4) 40%, rgba(255, 255, 255, 0.4) 60%, transparent)',
+  animationTimingFunction: 'cubic-bezier(.06,.42,.57,.89)',
```
TheSisb, line 16: issue (blocking): we need to keep the initial `transform: ...` line because otherwise on computers with prefers-reduced-motion the halted state is with the animation stopped in the center of the skeletonLoader:

Without it:
image

With it: 
image

packages/paste-core/components/skeleton-loader/src/index.tsx
```diff
@@ -1,32 +1,23 @@
...
-  // on the bottom right of the loader at the end state.
-  to: {translateX: '105%', skew: '155deg'},
-  config: {
-    mass: 0.1,
-    tension: 80,
-    friction: 50,
-  },
-};
-
+const SkeletonLoaderInner = styled.div({
```
TheSisb, line 8: polish (non-blocking): since we don't need `styled` to wrap the animated element, can we replace this with `Box` now?

netlify.toml
zahnster: This includes changes under paste-core as things that will trigger a preview website build - the builds were being skipped, but I do believe we would want core changes to trigger preview builds.

packages/paste-core/components/skeleton-loader/src/index.tsx
zahnster: @TheSisb how would I handle the css animation if this were a <Box> ?

TheSisb (approved):
I'm not sure if we (can) support keyframe animations in Box currently, so this looks perfectly fine.

PR #2067, opened 2021-11-22
feat(styled-library): remove props from html element attributes using shouldForwardProp

abereghici:
Added `shouldForwardProp` function to `styled-library` that's used to remove the props from html element attributes.

Before:
Screenshot 2021-11-22 at 13 42 15

After:
Screenshot 2021-11-22 at 13 42 24

**Contributing to Twilio**

> All third-party contributors acknowledge that any contributions they provide will be made under the same open-source license that the open-source project is provided under.

- [x] I acknowledge that all my contributions will be made under the project's license.

PR #2285, opened 2022-03-09
feat: add should forward props to box and text

SiTaggart:
Taking over from https://github.com/twilio-labs/paste/pull/2067 with some extra additions for custom style props that we add to the system.


zahnster (approved):
👍🏼  This looks great, much cleaner html on the implementation side.

packages/paste-core/components/modal/__tests__/index.spec.tsx
```diff
@@ -156,7 +156,7 @@ describe('Modal', () => {
     expect(document.activeElement).toEqual(getByTestId('modal-header').querySelector('button'));
   });
 
-  it('should focus on the element provided as the initialFocus element in the modal', () => {
+  it.skip('should focus on the element provided as the initialFocus element in the modal', () => {
```
SiTaggart, line 159: This test causes a blow up for some reason now. totally can't figure out what is up with it.

shleewhite: will you investigate that in another PR?

shleewhite (approved):

packages/paste-website/src/components/shortcodes/table-of-contents/TableOfContentsListItem.tsx
```diff
@@ -1,21 +1,21 @@
 import {styled, themeGet} from '@twilio-paste/styling-library';
 import type {TableOfContentsListItemProps} from './types';
 
-export const marginTopPartial = ({depth}: TableOfContentsListItemProps): string => {
+export const marginTopPartial = ({depth}: TableOfContentsListItemProps): any => {
```
SiTaggart, line 4: By adding the types package themeGet is retyped as returning `any` which is kind of annoying, but I get it.

packages/paste-core/components/modal/__tests__/index.spec.tsx
SiTaggart: Whoa, actual bug, terrible message. Good catch Jest

SiTaggart: Fixed

zahnster (approved):
👍🏼

shleewhite (approved):

PR #2298, opened 2022-03-15
feat: add localized fonts to support i18n

shleewhite:
- [x] new font family aliases for non-latin characters: `font-family-30`,` font-family-40`, `font-family-50`, `font-family-60`
- [x] new font family tokens: `font-family-text-japanese`, `font-family-text-korean`, `font-family-text-chinese-traditional`, `font-family-text-chinese-simplified`
- [x] add `class='paste-theme-provider'` to `themeProvider`
- [x] add global styles that change the font depending on the html:lang()
- [x] replace fontFamily in components with `'inherit'` (if not monospace)
- [x] add stories with English and Japanese
- [x] `react-helmet` as a devDependency
 

**Contributing to Twilio**


packages/paste-design-tokens/tokens/aliases/font-family.yml
```diff
@@ -1,3 +1,7 @@
 aliases:
   font-family-10: "'Inter var experimental', 'Inter var', -apple-system, BlinkMacSystemFont, 'Segoe UI', Roboto, Oxygen-Sans, Ubuntu, Cantarell, 'Helvetica Neue', sans-serif"
   font-family-20: "'Fira Mono', 'Courier New', Courier, monospace"
+  font-family-ja: "'Inter var experimental', 'Inter var', Hiragino Sans, 'ヒラギノ角ゴ ProN W3', 'Hiragino Kaku Gothic ProN', 'メイリオ', Meiryo, Osaka, 'MS PGothic', sans-serif"
```
shleewhite, line 4: Question: the other font family aliases use a number, but it felt kind of weird for this case. What should the alias name be?

`font-family-ja`, `font-family-30`, or `font-family-ja-10`?

packages/paste-design-tokens/tokens/aliases/font-family.yml
```diff
@@ -1,3 +1,7 @@
...
+  font-family-ko: "'Inter var experimental', 'Inter var', 'Microsoft JhengHei', 微軟正黑體, 'Microsoft JhengHei UI', 'Microsoft YaHei', 微軟雅黑, 宋体, SimSun, sans-serif"
+  font-family-zh-traditional: "'Inter var experimental', 'Inter var', 'Microsoft JhengHei', 微軟正黑體, 'Microsoft JhengHei UI', 'Microsoft YaHei', 微軟雅黑, 宋体, SimSun, sans-serif"
```
shleewhite, line 6: Question: the selector where this is used is `html:lang(zh-TW), html:lang(zh-HK)`. Is this an ok name? Or should it be something else?

packages/paste-design-tokens/tokens/aliases/font-family.yml
```diff
@@ -1,3 +1,7 @@
...
+  font-family-zh-simplified: "'Inter var experimental', 'Inter var', 'Microsoft YaHei New', 微软雅黑, 'Microsoft Yahei', 'Microsoft JhengHei', 宋体, SimSun, sans-serif"
```
shleewhite, line 7: Question: The selector for this one is `html:lang(zh-CN)`. Should it be `font-family-zh-cn`? This also depends on the answer to the 2 above questions.

packages/paste-design-tokens/tokens/global/font-family.yml
```diff
@@ -8,3 +8,11 @@ props:
     value: "{!font-family-10}"
   font-family-code:
     value: "{!font-family-20}"
+  font-family-text-ja:
```
shleewhite, line 11: Question: in the ticket, Shadi wanted the token to be `fontFamilyTextJA`, but when it is compiled it is `fontFamilyTextJa`. Is this ok? or do I need to update how the tokens are compiled?

packages/paste-theme/stories/themeProvider.stories.tsx
```diff
@@ -0,0 +1,70 @@
...
+    <Box>
+      <Paragraph>
+        「ヒラギノ角ゴシック体」は、「ヒラギノ明朝体」との混植を意識してデザインした角ゴシック体です。モダンでカウンターが広く明るいだけではなく、少しフトコロを締め気味にして、現代的な明るさを残しつつもオーソドックスなデザインになっています。文字のすき間を均等に配置することと画線両端のアクセントにより、可読性と存在感を両立させた角ゴシック体です。極小サイズの本文から、特大サイズの見出しまで、広い範囲をカバーできます。
+      </Paragraph>
+      <Paragraph>
+        あ い う え お ア イ ウ エ オ 安 以 宇 衣 於 か き く け こ カ キ ク ケ コ 加 幾 久 計 己 さ し す せ そ サ シ
+        ス セ ソ 左 之 寸 世 曽 た ち つ て と タ チ ツ テ ト 太 知 川 天 止 A B C D a b c d 1 2 3 4 、 。 ！ ？ 「 」
+      </Paragraph>
+      <Stack orientation="vertical" spacing="space50">
+        <Button variant="primary" onClick={() => {}}>
```
shleewhite, line 56: Question: I thought it would be a good idea to add example components to check that the font inherit is working correctly. Should I keep this?

packages/paste-design-tokens/tokens/aliases/font-family.yml
SiTaggart: General naming convention for alias is that they are number steps. They are generic, buckets of values that have no meaning, and we don't need to think about naming them when adding more.

I'd go with ont-family-30 personally

packages/paste-design-tokens/tokens/aliases/font-family.yml
SiTaggart: This is more to be applied to the tokens file, but we tend to favour verbosity where possible. It's just easier for everyone to understand when coming into the code base, even if you have no prior experience, when things are spelt out in full.

I wouldn't be ashame of 

```
font-family-japanese: foo;
font-family-korean: bar;
font-family-traditional-chinese: boo;
font-family-simplified-chinese: bam;
```

As the token names

packages/paste-design-tokens/tokens/aliases/font-family.yml
SiTaggart: see above ^

packages/paste-design-tokens/tokens/global/font-family.yml
SiTaggart: `fontFamilyTextJa` is fine, there's only so much you can control when you are automating things.

packages/paste-theme/src/styles/global.ts
```diff
@@ -3,6 +3,27 @@ import {css} from '@twilio-paste/styling-library';
 export const pasteGlobalStyles = css({
   html: {
     fontSize: '100%',
+    fontFamily: 'fontFamilyText',
+  },
+  'html:lang(ja), html:lang(ja-JP)': {
+    '.paste-theme-provider, .paste-theme-provider input, .paste-theme-provider button, .paste-theme-provider textarea, .paste-theme-provider select, .paste-theme-provider ::placeholder': {
```
SiTaggart, line 9: question: Why do we need the element qualified selectors? It looks like if you set font-family to inherit on those components, it should just work, even for placeholder text? https://codepen.io/SiTaggart/pen/zYpGXMv

packages/paste-theme/stories/themeProvider.stories.tsx
SiTaggart: Great idea

packages/paste-theme/stories/themeProvider.stories.tsx
```diff
@@ -0,0 +1,70 @@
...
+      <TextArea aria-label="Feedback" value="Lorem ipsum dolor sit amet, consectetur adipiscing elit" />
+    </Stack>
+  </Box>
+);
+
+export const JapaneseFontFamily = (): React.ReactNode => (
+  <>
+    <Helmet>
+      <html lang="ja" />
+    </Helmet>
```
SiTaggart, line 46: suggestion: We do have this Storybook plugin installed https://storybook.js.org/addons/storybook-addon-i18n. If it supports it, it might be a better solution to using Helmet.

packages/paste-theme/src/styles/global.ts
shleewhite: Oh! If that's the case, I can remove them. I think this was included in the ticket just to be extra sure they had the right font.

shleewhite: Oh! Do we want everything that is a child of the `ThemeProvider` to get the font, even if its not a Paste component? Thats the only reason I can think of to keep the selectors.

SiTaggart: Kinda, the way we do it now gives us a little wriggle room for migration from "old stuff". In that, we try to not set global css styles so that we don't affect random components thrown into the mix. Because we don't own them, we don't really have a right to change their styles, out side of the very minimum amount of container styles we must set on the provider.

If it doesn't work for you, you should move to Paste components to get all the theming and goodness for free.

For that reason, avoiding element selectors would be preferable.

SiTaggart (commented):
Before I give final approval I'd like to see the storybook and chromatic builds, but everything looks nice and neat. Really like where it landed

packages/paste-design-tokens/tokens/global/font-family.yml
```diff
@@ -8,3 +8,11 @@ props:
...
   font-family-code:
     value: "{!font-family-20}"
+  font-family-text-japanese:
+    value: "{!font-family-30}"
+  font-family-text-korean:
+    value: "{!font-family-40}"
+  font-family-text-chinese-traditional:
+    value: "{!font-family-50}"
+  font-family-text-chinese-simplified:
+    value: "{!font-family-60}"
```
gloriliale, line 18: praise: I really like this approach and think it's going to be useful if we need flexibility in the future. So good!

packages/paste-theme/stories/themeProvider.stories.tsx
```diff
@@ -0,0 +1,70 @@
...
+        あ い う え お ア イ ウ エ オ 安 以 宇 衣 於 か き く け こ カ キ ク ケ コ 加 幾 久 計 己 さ し す せ そ サ シ
+        ス セ ソ 左 之 寸 世 曽 た ち つ て と タ チ ツ テ ト 太 知 川 天 止 A B C D a b c d 1 2 3 4 、 。 ！ ？ 「 」
+      </Paragraph>
+      <Stack orientation="vertical" spacing="space50">
+        <Button variant="primary" onClick={() => {}}>
+          ダミーテキスト
+        </Button>
+        <Input aria-label="Search" placeholder="ダミーテキスト" type="text" />
+        <Select aria-label="Options">
+          <Option value="option1">ダミーテキスト</Option>
```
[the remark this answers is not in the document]
gloriliale: praise: good research!

gloriliale (approved):
This is awesome, well done Lee!

PR #2304, opened 2022-03-19
feat(button): add i18n props

gloriliale:
- [x]  add i18nExternalLinkLabel for external link icon
- [x]  add i18n stories
- [x]  add tests for default and custom i18n props
- [x]  add example to website
- [x]  add prop to Button prop list on website
- [x] make title not required for Spinner

Note: don't need to update spinner page on website because it incorrectly said title isn't required.


packages/paste-core/components/button/src/index.tsx
```diff
@@ -146,7 +152,7 @@ const ButtonContents: React.FC<ButtonContentsProps> = ({buttonState, children, s
           alignItems="center"
           lineHeight="lineHeight30"
         >
-          <Spinner decorative={false} title="Loading, please wait." delay={0} />
+          <Spinner decorative={false} title={i18nLoadingLabel} delay={0} />
```
shleewhite, line 155: Question: @SiTaggart title is required on spinner, even when its decorative. Could we make it not required? Otherwise we would need an i18nLoadingLabel prop.

shleewhite: We could do some manual prop validation where if they dont pass a title when its not decorative it throws an error?

shleewhite: Talked in slack: Going to make spinner like icon, title is not required but will throw an error if decorative={false} and there is no title

SiTaggart (approved):

zahnster (approved):
Looks set! 👍🏼

PR #2307, opened 2022-03-21
fix: data grid layout backwards compat with older tables. Support responsive tables

SiTaggart:
By default the datagrid is responsive, and horizontally scrolls. This is unlike past Twilio tables and causes data dense grids to have a lot of content scrolled off screen.

In this PR we implement responsive table support in the base Table component. We support noWrap for not line breaking cell content. We make the default behavior for a data grid, non-responsive to mimic better the default table behaviour and inherit the responsive options from the base Table.

### DIFF
--- a/.changeset/giant-pillows-lick.md
+++ b/.changeset/giant-pillows-lick.md
@@ -1,0 +1,6 @@
+---
+'@twilio-paste/table': patch
+'@twilio-paste/core': patch
+---
+
+[Table] Adds the ability to safely render tables for mobile screens using the `isResponsive` prop. Also adds the `noWrap` prop to control cell density by preventing line wrapping.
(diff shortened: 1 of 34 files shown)


.changeset/giant-pillows-lick.md
```diff
@@ -0,0 +1,6 @@
+---
+'@twilio-paste/table': patch
```
shleewhite, line 2: Suggestion: going off of the i18n tickets, should these changesets be a minor bc youre adding props?

shleewhite (approved):
🚢

.changeset/giant-pillows-lick.md
SiTaggart: Proooooooooooooooooooooooooobably 🙃

katieporter:
While you're in these files, would you mind making "About Data Grid" on line 102 in `data-grid/index.mdx` an H3 instead of an H2?

katieporter:
Do we want to add the new prop to the props table on the Data Grid page?

zahnster (approved):
+1 to Katie and Lee's comments (non-blocking but highly encouraged), otherwise looks good to go.

PR #2313, opened 2022-03-22
chore: Jest improvements part 1

SiTaggart:
- [x] add jest-swc as the faster transformer
- [x] remove enzyme
- [x] remove jest trasnforms and the need to run `pre-test`
- [x] remove silent running of jest to spot issues
- [x] fix existing issues that made jest super noisy when not run in silent mode
- [x] fix a couple of async test bugs in disclosure, breadcrumb and tooltip that were causing the `createEvent` bug


jest.config.js
```diff
@@ -4,8 +4,7 @@ module.exports = {
   globals: {
     TestTheme,
   },
-  verbose: true,
-  testRegex: '(/__tests__/.*|(\\.|/)(test|spec))\\.tsx?$',
+  testRegex: '(/__tests__/.*|(\\.|/)(test|spec))\\.(js|jsx|ts|tsx)?$',
```
SiTaggart, line 7: Single jest config that runs all tests in every file format

jest.config.js
```diff
@@ -30,11 +29,10 @@ module.exports = {
     },
   },
   transformIgnorePatterns: ['node_modules/'],
-  resolver: '<rootDir>/.jest/resolver.js',
   transform: {
-    '^.+\\.tsx?$': '<rootDir>/.jest/transform-typescript.js',
-    '^.+\\.js?$': '<rootDir>/.jest/transform-esm.js',
+    '^.+\\.(js|jsx|ts|tsx)?$': '@swc/jest',
```
SiTaggart, line 33: Whereas before we needed different transforms for js and ts, because they needed different babel configs, just using swc as the compiler not only means it's faster to run jest than babel, but it'll transform both source file types with no config switching. Much simpler.

package.json
```diff
@@ -63,13 +63,9 @@
...
-    "test": "yarn pre-test && concurrently \"yarn:test:packages\" \"yarn:test:tools\"",
-    "test:react-16": "yarn pre-test && concurrently \"yarn:test:packages-16\" \"yarn:test:tools-16\"",
-    "test:packages": "jest --silent ./packages/",
-    "test:packages-16": "USE_REACT_16=true jest --silent ./packages/",
-    "test:tools": "jest ./tools/ -c ./tools/jest.config.js",
-    "test:tools-16": "USE_REACT_16=true jest ./tools/ -c ./tools/jest.config.js",
-    "test:coverage": "yarn pre-test && jest --coverage",
+    "test": "jest",
+    "test:react-16": "USE_REACT_16=true jest",
+    "test:coverage": "jest --coverage",
```
SiTaggart, line 68: No need to run pre-test with yarn 3, and no need to have two concurrent jest just to test packages which are tsx and tools which are js.

packages/paste-codemods/tools/__tests__/tools.spec.js
```diff
@@ -39,12 +39,15 @@ describe('generatePackageExportsMap', () => {
...
       Stack: '@twilio-paste/core/stack',
+      StyledBox: '@twilio-paste/core/box',
+      getCustomElementStyles: '@twilio-paste/core/box',
       getStackChildMargins: '@twilio-paste/core/stack',
       getStackDisplay: '@twilio-paste/core/stack',
       getStackStyles: '@twilio-paste/core/stack',
       safelySpreadBoxProps: '@twilio-paste/core/box',
       useComboboxPrimitive: '@twilio-paste/core/combobox-primitive',
       useDisclosurePrimitiveState: '@twilio-paste/core/disclosure-primitive',
+      useMultiSelectPrimitive: '@twilio-paste/core/combobox-primitive',
```
SiTaggart, line 50: This test wasn't being run in the old config. It does now that we have a single jest config that covers all test source formats, across the entire repo.

packages/paste-core/components/avatar/__tests__/avatar.test.tsx
```diff
@@ -84,9 +84,9 @@ describe('Avatar', () => {
         expect(getCorrespondingIconSizeFromSizeToken('sizeIcon110')).toEqual('sizeIcon80');
       });
       it('should throw an error when non IconSize values are passed in', () => {
-        // @ts-expect-error
+        // @ts-expect-error non-icon token
         expect(() => getCorrespondingIconSizeFromSizeToken('size50')).toThrow();
-        // @ts-expect-error
+        // @ts-expect-error doesn't accept booleans
```
SiTaggart, line 89: Just trying to get rid of the squiggly lines

packages/paste-core/components/breadcrumb/__tests__/breadcrumb.spec.tsx
```diff
@@ -45,7 +45,7 @@ describe('Breadcrumb', () => {
         <BreadcrumbItem href="#">foo</BreadcrumbItem>
       </Breadcrumb>
     );
-    const renderedAnchor = screen.findByRole('link');
+    const renderedAnchor = screen.getAllByRole('link');
```
SiTaggart, line 48: `findByRole` is asynchronous. This is likely the main culprit of the `createEvent` error on react 17 tests.

`getBy` is not, and arguably does a better job.

packages/paste-core/components/button/__tests__/button.test.tsx
```diff
@@ -150,132 +149,169 @@ describe('Button', () => {
...
-      expect(() => shallow(<Button variant="primary" tabIndex="-2" />)).toThrow();
-      expect(console.error).toHaveBeenCalled();
-      console.error = originalError;
+      // hide console errors from terminal when throwing expected errors
+      const spy = jest.spyOn(console, 'error');
+      spy.mockImplementation(() => {});
+      // @ts-expect-error invalid tabindex
+      expect(() => render(<Button variant="primary" tabIndex="-2" />)).toThrow();
+      expect(spy).toHaveBeenCalled();
+      spy.mockRestore();
```
SiTaggart, line 308: Goodbye enzyme and extremely noisy terminal when throwing expected errors. 

Part of the problem we had is that we silenced jest because it was so noisy we could't figure out what was a real error. To be a little more intentional, and enable us to stop silencing jest, we can mock the console errors, meaning we can actually see when jest is telling us something useful.

packages/paste-core/components/checkbox/__tests__/checkboxdisclaimer.test.tsx
```diff
@@ -119,16 +117,13 @@ describe('Customization', () => {
...
-    render(
+    const {container} = render(
       <CheckboxDisclaimer {...defaultProps} errorText={errorText}>
         foo
-      </CheckboxDisclaimer>,
-      container
+      </CheckboxDisclaimer>
     );
-    const results = await axe(document.body);
+    const results = await axe(container);
```
SiTaggart, line 126: Just consistency here. We use RTL to run axe tests everywhere else now, using react DOM was an old way.

packages/paste-core/components/disclosure/__tests__/disclosure.test.tsx
```diff
@@ -184,12 +186,18 @@ describe('Disclosure', () => {
...
+      const toggleButton = screen.getByTestId('external-toggle');
+      const disclosureButton = screen.getByTestId('disclosure-button');
+      const disclosureContent = screen.getByTestId('disclosure-content');
+      expect(disclosureButton.getAttribute('aria-expanded')).toEqual('false');
+      expect(disclosureContent).not.toBeVisible();
+      await waitFor(() => {
+        userEvent.click(toggleButton);
+      });
+      expect(disclosureButton.getAttribute('aria-expanded')).toEqual('true');
+      expect(disclosureContent).toBeVisible();
```
SiTaggart, line 200: This is a much better test for the state hook. The previous one just mimicked the default behaviour, it tested nothing.

This now at least, tests that you can use the state hook to affect the behaviour of the disclosure by controlling is externally using one of the returned hook methods.

packages/paste-core/components/disclosure/__tests__/disclosure.test.tsx
```diff
@@ -161,15 +159,19 @@ describe('Disclosure', () => {
       expect(renderedDisclosureButton.getAttribute('tabindex')).toEqual('0');
       expect(screen.getByTestId('disclosure').id).toEqual('disclosure');
     });
-    it('should render a disclosure open', () => {
-      render(<MockDisclosure visible />);
+    it('should render a disclosure open', async () => {
+      await waitFor(() => {
+        render(<MockDisclosure visible />);
+      });
```
SiTaggart, line 165: This also seemed to be tripping up jest with an incomplete async test. Setting visible causes a state change which wasn't wrapped in `act`. Waiting for it, fixes that, and the potential for the `createEvent` bug to appear.

packages/paste-core/components/help-text/__tests__/helptext.test.tsx
```diff
@@ -1,27 +1,14 @@
...
-
-  it('should have an error icon', () => {
-    expect(container.find('ErrorIcon').length).toEqual(1);
-  });
-
-  it('should have colorTextError', () => {
-    expect(container.prop('color')).toEqual('colorTextError');
-  });
-});
-
```
SiTaggart, line 19: covered by VRT

packages/paste-core/components/menu/__tests__/index.spec.tsx
```diff
@@ -13,7 +13,7 @@ const PreferencesMenu = React.forwardRef<HTMLButtonElement, MenuButtonProps>((pr
 
   return (
     <>
-      <MenuButton ref={ref} {...menu} {...props} data-testid="example-submenu-trigger">
+      <MenuButton ref={ref} {...menu} {...props} variant="secondary" data-testid="example-submenu-trigger">
```
SiTaggart, line 16: Console error ing, but never saw it as we silenced jest.

packages/paste-core/components/popover/__tests__/index.spec.tsx
```diff
@@ -94,9 +101,11 @@ describe('Popover', () => {
...
       expect(renderedPopoverButton.getAttribute('aria-expanded')).toEqual('false');
       expect(renderedPopover).not.toBeVisible();
-      renderedPopoverButton.click();
-      await waitFor(() => expect(renderedPopoverButton.getAttribute('aria-expanded')).toEqual('true'));
-      await waitFor(() => expect(renderedPopover).toBeVisible());
+      await waitFor(() => {
+        userEvent.click(renderedPopoverButton);
+      });
+      expect(renderedPopoverButton.getAttribute('aria-expanded')).toEqual('true');
+      expect(renderedPopover).toBeVisible();
```
SiTaggart, line 108: These might have been causing async test problems due to state updates, but I'm not sure they are necessary changes. 

`ButtonTwo.click();` will cause a state update but the update isn't wrapped in `act` which is what you need to prevent a async test problem. by being strict about use of `userEvent`, we at least guarantee the state update is wrapped by `act`, as `userEvent` does this for you.

Not convinced you need to `waitFor` subsequently, but it's hard to debug locally

packages/paste-core/components/popover/__tests__/index.spec.tsx
```diff
@@ -245,6 +259,11 @@ describe('Popover', () => {
         </Theme.Provider>
       );
 
+      const popoverButton = screen.getByRole('button');
+      await waitFor(() => {
+        userEvent.click(popoverButton);
+      });
+
```
SiTaggart, line 266: Again, preset visible causes an unwrapped state update. We can achieve the same test, by just clicking the thing and using the test utils to handle it all safely.

packages/paste-core/components/toast/__tests__/useToaster.spec.tsx
```diff
@@ -119,7 +122,7 @@ describe('useToaster', () => {
       act(() => {
         result.current.pop('custom_timeout');
       });
-      expect(clearTimeout).toHaveBeenCalled();
+      expect(clearTimeoutSpy).toHaveBeenCalled();
```
SiTaggart, line 125: This is a required update on how to spy on window methods in the new version of Jest

packages/paste-core/primitives/sibling-box/__tests__/siblingBox.test.tsx
```diff
@@ -1,66 +1,62 @@
...
+          }}
+          _checkedAndHoverSibling={{
+            borderRadius: 'borderRadius10',
+          }}
+        >
+          child
+        </SiblingBox>
+      </CustomizationProvider>
+    );
+    expect(asFragment()).toMatchSnapshot();
```
SiTaggart, line 59: Consistency in render snapshots using rtl

packages/paste-core/primitives/text/__tests__/StyleFunctions.spec.tsx
```diff
@@ -1,194 +1,196 @@
...
+            foo: {
+              borderColor: 'colorBorderError',
+            },
           },
         },
       },
     },
-  },
-};
+  };
```
SiTaggart, line 48: This is a little memory leak. I'm pretty sure that because mockTheme is outside the describe block and is a global, for what ever reason it's not torn down after the test, so it keeps a reference of it in memory. 

Regardless if you run `yarn jest --detectLeaks` before and after, this test stops being flagged as a memory leak.

packages/paste-core/primitives/text/__tests__/text.spec.tsx
```diff
@@ -1,620 +1,576 @@
...
+          _before={{content: `"Before text"`, position: 'absolute', bottom: 0, left: 0}}
+          _after={{content: `"After text"`, position: 'absolute', bottom: 0, left: 0}}
+          _focusWithin={{color: 'colorTextWeak'}}
+        >
+          PseudoBox
+        </Text>
+      </CustomizationProvider>
+    );
+
+    expect(asFragment()).toMatchSnapshot();
```
SiTaggart, line 573: Consistency around a single tool set.

packages/paste-theme/__tests__/themeProvider.spec.tsx
```diff
@@ -20,17 +20,6 @@ describe('Theme.Provider', () => {
...
-    const {getByText} = render(
-      <Theme.Provider theme="console">
-        <ThemeConsumerExampleTextColor />
-      </Theme.Provider>
-    );
-
-    // console theme link color: rgb(4, 60, 181)
-    expect(getByText('Color: rgb(2, 99, 224)')).toBeDefined();
-  });
-
```
SiTaggart, line 33: Deprecated theme in the theme provider. We don't even import it so there's no point in testing it if it's just the same as default theme.

packages/paste-website/src/pages/introduction/about-paste/index.mdx
```diff
@@ -253,9 +253,7 @@ Systems should never have single points of failure. Good documentation, reliable
     </DisclosureHeading>
     <DisclosureContent>
       <Paragraph>
-        We use Jest coupled with Enzyme for older components and React-Testing-Library for newer components to do
-        component unit testing. We're transitioning away from Enzyme to React-Testing-Library in the long term as it is
-        now the recommended testing library of the React team. We aim for 85-90% test coverage.
+        We use Jest coupled with React-Testing-Library to do component unit testing. We aim for 85-90% test coverage.
```
SiTaggart, line 256: Little website update as we now solely use rtl

zahnster (approved):
Praise: This all looks great and works really well locally. Thank you for the comments in the PR, really helped to understand the context for these changes.

package.json
shleewhite: LOVE

shleewhite: Also love that silent is removed.. I delete that on my local all the time and then have to put it back lol

shleewhite (approved):

package.json
andipants12: Totally! 🪙 🪙 Silent _can_ be nice in CI, helps to make output cleaner, but can def obfuscate issues when they arise. So glad we made these improvements. 🏅

packages/paste-core/components/disclosure/__tests__/disclosure.test.tsx
andipants12: 🪙 🪙 [@testing-library/react-hooks](https://github.com/testing-library/react-hooks-testing-library) is super useful for unit testing hooks and make things simpler/less opaque.

packages/paste-core/components/breadcrumb/__tests__/breadcrumb.spec.tsx
andipants12: 🪙 🪙 The find methods are useful for testing, but without an await statement will cause async errors. [KCD summarizes it here](https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#using-waitfor-to-wait-for-elements-that-can-be-queried-with-find) succinctly.

I wish `.find...` methods threw a better error when used without some kind of async logic.

PR #2315, opened 2022-03-24
fix(spinner): match title prop types to types

SiTaggart:
When up dating the spinner title requirements we forgot to update the prop types to go with it.

It is hard to tell when prop types don't match types as we don't explicitly check them anywhere.

This became visible in the newer jest setup.


nkrantz (approved):

zahnster (approved):
👍🏼

shleewhite (approved):

PR #2323, opened 2022-03-25
fix(website): fix buggy tokens list filter

nkrantz:
- [X] Query runs on backspace as well as letter key down (and same results are displayed in either direction)
- [X] Query params in url are updated in real time when the input value changes
- [X] Enter button / submitting form does nothing (removed the form altogether)
- [X] Correctly sets input value on page load when present in url
- [x] Correctly filters tokens and displays queried results even when input value doesn't change (on page load with url params, on "clear search" button, etc. - fix `filterTokenList` to run for this to work)
- [X] Add empty state for no matches (using current matching)
- [X] Add e2e tests to validate expected functionality
- [X] Add unit tests for the functions
- [X] Refactor index.js - pull out types and helper functions into their own files

**Contributing to Twilio**

### DIFF
--- a/cypress/integration/token-list/index.spec.ts
+++ b/cypress/integration/token-list/index.spec.ts
@@ -6,7 +6,7 @@ import {
 } from '../../support/utils/applitools';
 
 // ℹ️ If we use anon functions instead of arrow functions, we can leverage Mocha's context and pull the test name directly from this
-describe('Token list filter', function () {
+describe('Token list filter with no existing search params', function () {
   // ℹ️ We are able to reference `this.title` because we have bound this describe block to the Cypress context.
   const testSuiteName = this.title;
 
@@ -18,6 +18,12 @@ describe('Token list filter', function () {
     cy.get('input[name="tokens-filter"]').type('background').should('have.value', 'background');
     cy.get('#background-colors').should('exist');
     cy.get('#border-colors').should('not.exist');
+  });
+
+  it('shows empty state', () => {
+    cy.get('input[name="tokens-filter"]').type('abc');
+    cy.get('[data-cy="tokens-empty-state"]').should('exist');
+    cy.get('img[src="/images/patterns/empty-no-results-found.png"]').should('exist');
   });
 
   describe('Visual regression tests', () => {
--- a/packages/paste-website/src/components/tokens-list/helpers.tsx
+++ b/packages/paste-website/src/components/tokens-list/helpers.tsx
@@ -1,0 +1,71 @@
+import * as React from 'react';
+import {trackCustomEvent} from 'gatsby-plugin-google-analytics';
+import type {ThemeVariants} from '@twilio-paste/theme';
+import type {TokenCategory, TokensListProps} from './types';
+import {Callout, CalloutTitle, CalloutText} from '../callout';
+
+const debounce = require('lodash/debounce');
+
+export const trackTokenFilterString = debounce((filter: string): void => {
+  if (filter !== '') {
+    trackCustomEvent({
+      category: 'Design Tokens',
+      action: 'filter',
+      label: filter,
+    });
+  }
+}, 500);
+
+const filterDeprecatedTokens = (tokens: TokenCategory[]): TokenCategory[] => {
+  return tokens.map((category) => {
+    return {...category, tokens: [...category.tokens.filter((token) => !token.deprecated)]};
+  });
+};
+
+export const getTokensByTheme = (props: TokensListProps, theme: ThemeVariants): TokenCategory[] => {
+  const unfilteredTokensBasedOnTheme =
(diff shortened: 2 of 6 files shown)


packages/paste-website/src/components/tokens-list/index.tsx
```diff
@@ -125,30 +132,35 @@ export const TokensList: React.FC<TokensListProps> = (props) => {
...
+    if (filter === '') navigate(window.location.pathname, {replace: true});
+    else navigate(`?tokens-filter=${filter}`, {replace: true});
   };
 
   const uid = useUID();
 
   return (
     <>
-      <Box as="form" marginTop="space100" marginBottom="space100" maxWidth="size40">
+      <Box as="div" marginTop="space100" marginBottom="space100" maxWidth="size40">
```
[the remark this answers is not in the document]
nkrantz: The form has default behavior that was making the page reload and the input value clear on hitting the enter key, and the AC says that the enter key shouldn't be necessary to a search (or perform any actions really, since they should happen live on each keystroke, which they now are). I was under the impression that it's ok to have one search filter Input field outside of a form field (a11y-wise), is that true?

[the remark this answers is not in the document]
SiTaggart: Technically, it's still better to have it contained in a form for accessibility reason. You can navigate by forms with AT. 

You can prevent the default behaviour of the form element though, by adding 

`onSubmit={ error => error.preventDefault() }`

to the form. So when you press enter, you submit the form, but the forms default behaviour of reloading the page is prevented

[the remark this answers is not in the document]
nkrantz: That was the first approach before changing it to a div, so I can definitely change it back to prevent default 👍

cypress/integration/token-list/index.spec.ts
```diff
@@ -20,6 +20,12 @@ describe('Token list filter', function () {
     cy.get('#border-colors').should('not.exist');
   });
 
+  it('shows empty state', () => {
+    cy.get('input[name="tokens-filter"]').type('abc');
+    cy.get('[data-paste-element="CARD"]').should('exist');
```
SiTaggart, line 25: suggestion(non-blocking): use of `data-testid` as a test identifier.

By targeting a generic selector like `[data-paste-element="CARD"]` you're in effect targeting any card on the page. If someone were to add another card to the page, this test might break. To be more explicit, adding a test id to your target will guarantee your test is looking at only the thing you are trying to test

nkrantz: not sure how to use `data-testid` in these cypress tests since I'm not rendering anything in the test, but would adding a `data-cy` to the card and checking for that be just as good?

`cy.get('[data-cy="tokens-empty-state"]').should('exist');`

SiTaggart: Yup, that's probably the cypress equivalent 👍

shleewhite (approved):
Awesome!

packages/paste-website/src/components/tokens-list/index.tsx
```diff
@@ -194,7 +121,36 @@ export const TokensList: React.FC<TokensListProps> = (props) => {
...
           );
-        })}
+        })
+      ) : (
+        <Card data-cy="tokens-empty-state">
+          <Stack orientation="horizontal" spacing="space70">
+            <Box
+              as="img"
+              src="/images/patterns/empty-no-results-found.png"
+              alt="No results found illustration"
```
shleewhite, line 131: Nit (non-blocking): This image is decorative because it doesn't add anything that the text on the page doesn't already say. You can just have alt="".

packages/paste-website/src/components/tokens-list/helpers.tsx
```diff
@@ -0,0 +1,71 @@
+import * as React from 'react';
+import {trackCustomEvent} from 'gatsby-plugin-google-analytics';
+import type {ThemeVariants} from '@twilio-paste/theme';
+import type {TokenCategory, TokensListProps} from './types';
+import {Callout, CalloutTitle, CalloutText} from '../callout';
+
+const debounce = require('lodash/debounce');
```
nkrantz, line 7: @SiTaggart should I update this to the import format like you have in your PR, or should I merge this one before yours and then you can add it to yours?

SiTaggart: I can handle it if you like. I'll need to do some of the images anyway for the empty state

katieporter:
Could we update the spacing and copy on the empty state to match [the design](https://www.figma.com/file/gHnoNWHbAKjMSjM02T8R3f/Design-Tokens?node-id=274%3A35622)?

It's `$space-150` padding around the card content, and `$space-110` between the illustration and the text. I'm realizing that our codesandboxes for the empty state pattern do not match the designs...we can fix that later haha.

PR #2328, opened 2022-03-29
fix(theme): animations no longer run when they shouldn't

TheSisb:
@zahnster noticed that there was a loud warning in her console due to `useLayoutEffect` in SSR environments:
image

While debugging I went through a bit of a rabbit hole trying to understand the best way to solve this. You see, the official [React-Spring docs](https://react-spring.io/guides/accessibility) recommend `useEffect`:
image

The problem with `useEffect` is that it fires after the first render (noted in this previous PR https://github.com/twilio-labs/paste/pull/2021) so we would have an issue where the animation would loop once before stopping. In that PR I swap to using `useLayoutEffect` which correctly prevents the first animation in client-side environments. However, the warning @zahnster shared above concerned me because if it does nothing in SSR environments, that means the animation loops for SSR and then stops after the client JS executes. I've attached a short video showing the issue on our built website:

https://user-images.githubusercontent.com/1592327/160682337-882103fc-5ed7-4d6f-9067-f5cbde5d9694.mov

---

I tried a few approaches such as the ones outlined by [Gaeron here](https://reactjs.org/link/uselayouteffect-ssr) or by checking for `typeof window !== undefined` and conditionally NOOPing on SSR, but that didn't improve the problem. After more thought, it finally hit me that I shouldn't be calling the animation `assign` function in an effect, but rather immediately.  There's no further spot up the DOM tree where I can move the call, as it depends on a prop from the `Theme.Provider`, so I execute it immediately in the render body.  I ended up wrapping it in a memo to avoid too many calls and it now works perfectly. 🎉


shleewhite (approved):
Coooooooool!!! I'm not too familiar with useMemo but I think it makes sense

zahnster (approved):
👏🏼 This is epic, thanks for the excellent writeup.

.storybook/preview.js
```diff
@@ -2,12 +2,20 @@ import React from 'react';
...
 import {Stack} from '@twilio-paste/stack';
 import {Grid, Column} from '@twilio-paste/grid';
 import {RenderPerformanceProfiler} from './RenderPerformanceProfiler';
 
+const disableAnimations = isChromatic();
+if (disableAnimations) {
+  AnimatedGlobals.assign({
+    skipAnimation: disableAnimations,
+  });
+}
```
SiTaggart, line 17: question: If theme provider calls skipAnimation internally, but you are also calling it here in the story global context, does that mean `disableAnimations` doesn't actually fully disabled animations?

TheSisb: Yeah so the first commit fixed the issue on the Gatsby website (dev and prod), but for some reason the issue still occurs in Chromatic where it loops once before halting.  I'm still debugging.

PR #2333, opened 2022-03-30
fix(live-preview): set LivePreview overflow as configurable

zahnster:
A recent change to the overflow behavior on the LivePreview component has caused some implementations (notedly examples on the Combobox and Menu docs pages) to not render as intended.

This change will set the overflow prop as configurable for the LivePreview component, so we can change it in instances that need visible overflow, such as the Combobox and Menu docs pages.


SiTaggart:
@zahnster thanks for tidying up my brute force fix.

Can you make sure this applies to the 

- Status pattern page
- Badge component page
- Flex component page
- Grid component page
- Pagination component page
- Skeleton loader page 
- Tabs page 
- and Toasts page

Please? 🙏

zahnster:
@SiTaggart thanks for sharing your deeper knowledge about the affected pages. Since this list is decently long, it makes me wonder if we should reverse the logic on the overflow behavior - should the default instead be to scroll overflow, and only turn it to visible on certain elements? Combobox and Menu are the only two components I've seen that need to break through their container. Your thoughts here are appreciated.

SiTaggart:
@zahnster yup, that was my initial thought too, which is why I went with the brute force approach. That coupled with the fact the site is responsive, by default we should make it harder to make code examples look bad on small screens.

I think reversing the logic would be a better idea too

shleewhite (approved):

andipants12 (approved):

SiTaggart (approved):

PR #2334, opened 2022-03-31
feat(ChatLog): add ChatMessageMeta component

shleewhite:
- [x] create ChatMessageMeta component
- [x] create ChatMessageMetaProps type
- [x] inbound and outbound ChatMessageMeta stories
- [x] new kitchen sink story
- [x] add unit tests for ChatMessageMeta (incl. customization tests)

**Contributing to Twilio**


packages/paste-core/components/chat-log/__tests__/customization.spec.tsx
```diff
@@ -28,175 +29,295 @@ const customBubbleStyles: PasteCustomCSS = {
...
+      render(
+        <CustomizationProvider baseTheme="default" theme={TestTheme} elements={customizedElements}>
+          <ChatMessage variant="inbound">
+            <ChatBubble data-testid="inbound-bubble">test</ChatBubble>
+          </ChatMessage>
+          <ChatMessage variant="outbound">
+            <ChatBubble data-testid="outbound-bubble">test</ChatBubble>
+          </ChatMessage>
+        </CustomizationProvider>
+      );
```
andipants12, line 164: ℹ️  Since these customization wrappers get repeated often, we started using the second optional argument of `render` to define them. It's kind of nice because you can just define it once, and use it in each render.

Not a required change, just a fun fact. 😸 

```suggestion
      render(
          <ChatMessage variant="inbound">
            <ChatBubble data-testid="inbound-bubble">test</ChatBubble>
          </ChatMessage>
          <ChatMessage variant="outbound">
            <ChatBubble data-testid="outbound-bubble">test</ChatBubble>
          </ChatMessage>,
          { wrapper: YourWrapperComponent }
      );
```

packages/paste-core/components/chat-log/src/types.ts
```diff
@@ -10,3 +10,9 @@ export interface ChatBubbleProps {
   children?: React.ReactNode;
   element?: string;
 }
+
+export interface ChatMessageMetaProps {
+  ['aria-label']?: string;
+  children: React.ReactNode;
```
andipants12, line 16: Are there any types within `React.ReactNode` that would not be valid for this component? 
More specifically, can `children` be `null` or `undefined` for this component?

I see that on line 35 above we are using `isRequired`, which makes it seem like this should exclude those values.

Happy to chat about this synchronously if you have questions, this stuff can be tricky.

packages/paste-core/components/chat-log/stories/index.stories.tsx
```diff
@@ -28,31 +33,172 @@ export const OutboundChatMessage = (): React.ReactNode => {
...
+              <Avatar name="Gibby Radki" size="sizeIcon20" />
+            </MediaFigure>
+            <MediaBody>Gibby Radki</MediaBody>
+          </MediaObject>
+          4 minutes ago
+        </ChatMessageMeta>
+      </ChatMessage>
+    </>
+  );
+};
```
andipants12, line 55: ℹ️ No required change here, just wanted to point this out.
I like to omit the curly braces when the arrow function only has a return statement, can save some lines in the long run.

```suggestion
export const InboundMessageWithMeta = (): React.ReactNode => (
  <>
    <ChatMessage variant="inbound">
      <ChatBubble>
        Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.
      </ChatBubble>
      <ChatMessageMeta aria-label="said by Gibby Radki 4 minutes ago">
        <MediaObject verticalAlign="center">
          <MediaFigure spacing="space30">
            <Avatar name="Gibby Radki" size="sizeIcon20" />
          </MediaFigure>
          <MediaBody>Gibby Radki</MediaBody>
        </MediaObject>
        4 minutes ago
      </ChatMessageMeta>
    </ChatMessage>
  </>
);
```

packages/paste-core/components/chat-log/stories/index.stories.tsx
```diff
@@ -28,31 +33,172 @@ export const OutboundChatMessage = (): React.ReactNode => {
   );
 };
 
+export const InboundMessageWithMeta = (): React.ReactNode => {
```
andipants12, line 36: ℹ️  You can import the following type from storybook and use it to type functional components used for stories. I recently started doing this and really like it because it provides more information specific to stories.

Something like this for the import, then the suggestion below 👇 
`import type {StoryFn} from '@storybook/react';`

```suggestion
export const InboundMessageWithMeta: StoryFn = () => {
```

packages/paste-core/components/chat-log/stories/index.stories.tsx
```diff
@@ -28,31 +33,172 @@ export const OutboundChatMessage = (): React.ReactNode => {
...
+            <Box>Read</Box>
+          </Box>
+        </ChatMessageMeta>
+      </ChatMessage>
+      <ChatMessage variant="outbound">
+        <ChatBubble>
+          Praesent in lectus commodo orci faucibus ullamcorper. Nulla sit amet sapien consectetur, tempor ante eget,
+          sagittis orci. Curabitur enim lorem, cursus et massa non, pretium faucibus lacus.
+        </ChatBubble>
+        <ChatMessageMeta>
```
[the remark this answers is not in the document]
andipants12: Will come back to this, will give it a think 🧠.

andipants12 (commented):
This looks great so far! I left quite a few comments/suggestions, nothing blocking though. I'll spend some time tomorrow thinking about the required label question and get back to it. 🙌

packages/paste-core/components/chat-log/__tests__/customization.spec.tsx
shleewhite: Nice! I didn't realize that was a thing. Will update.

packages/paste-core/components/chat-log/stories/index.stories.tsx
shleewhite: 👍🏽

packages/paste-core/components/chat-log/stories/index.stories.tsx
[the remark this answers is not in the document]
shleewhite: Discussed with Simon at a focus time, and decided that the use case where aria-label isn't necessary is way less common so might as well just keep it required.

packages/paste-core/components/chat-log/src/ChatMessageMeta.tsx
```diff
@@ -0,0 +1,38 @@
...
+  ({children, element = 'CHAT_MESSAGE_META', ...props}, ref) => {
+    const variant = React.useContext(MessageVariantContext);
+
+    return (
+      <Box
+        ref={ref}
+        element={element}
+        display="flex"
+        justifyContent={Array.isArray(children) ? 'space-between' : 'flex-end'}
+        textAlign={!Array.isArray(children) ? 'right' : undefined}
```
SiTaggart, line 18: Nit: Minor optimisation, you could call Array.isArray(children) once outside of the render block to save doing it twice

SiTaggart (approved):
Loving this consumer API 🎉

packages/paste-core/components/chat-log/src/ChatMessageMeta.tsx
TheSisb: Minor improvement on `Array.isArray` is to use the React top-level API `React.Children.count(children)` https://reactjs.org/docs/react-api.html#reactchildrencount

andipants12 (approved):

Issue #2308, opened 2022-03-22
Combobox starts behaving erratically in grouped options mode when passed options are not in sorted order


josamit:
Thanks for helping make Paste better! Note: For urgent requests, reach out at **#help-design-system**

**Describe the bug**
Combobox starts behaving erratically in grouped options mode when passed options are not in sorted order

**To Reproduce**
1. Go to 'https://paste.twilio.design/components/combobox/#combobox-with-option-groups'
2. Click on view code and **change combobox options in an unsorted order** similar to following screenshot.
<img width="788" alt="image" src="https://user-images.githubusercontent.com/62208722/159483525-ced75e86-4a1c-404a-a982-7898b7a3d02b.png">
3. Click on combobox now and scroll through options. **The options will start auto scrolling erratically** when you hover on the option menu from the bottom. More details in the screen-capture below:

![combobox_issue](https://user-images.githubusercontent.com/62208722/159483863-f1fa24aa-6ea5-4539-9cb5-07b1885f0396.gif)

**Expected behavior**
The scrolling behaviour should be consistent in both grouped and non group options. This bug request is coming from investigation of few of our customer escalations related to this. 

**Desktop (please complete the following information):**
 - OS: MacOs
 - Browser Chrome
 - Version 8.1.1

**Additional context**

shleewhite:
@josamit Just following up on this again, could you please provide more information on the customer escalation? Is it happening in their code or just on our website? Also, could you please send the Zendesk tickets? Thank you!

josamit:
@shleewhite  Ticket : https://issues.corp.twilio.com/browse/COMPUI-126 Slack Thread : https://twilio.slack.com/archives/CSDLH7HG9/p1647326225977409

andipants12:
@josamit Hello there 👋 We have a fix that is ready to be merged very soon. You can read more about the issue by checking out the linked PR above. I will reach out here and let you know when the fix is officially released. Please let us know if you have any questions or feedback in the meantime.

Thank you for taking the time to help up improve Paste! 🏅

josamit:
@andipants12 thank you so very much, I will keep an eye out for the update on release 🙇

TheSisb:
Hi @josamit,

The fix has been merged any will be deployed and available in the `@twilio-paste/core@11.5.0` version in a few minutes.

PR #2335, opened 2022-03-31
fix(combobox): grouped combobox scroll

andipants12:
## Description
Issue reported: [Issue 2308](https://github.com/twilio-labs/paste/issues/2308)

**tl;dr** Adds a conditional guard to a side effect hook that syncs virtual scroll position with current highlighted index when the combobox has grouped items.

### Summary of changes
Grouped comboboxes are not yet virtualized because we cannot yet do so accessibly. The main component ([Combobox.tsx](https://github.com/twilio-labs/paste/pull/2335/files#diff-22b2d509777ec919245d83593056eb5d0c5ac23808f3b543b288425b985cb360L106)) was running a side effect (`useEffect`) that syncs the virtualized list position according to the highlighted index. This caused increasingly erratic scrolling behavior (with continued scrolling) because the scrolling function is not "calibrated" correctly for the non-virtualized grouped list.

To resolve this we:
- check if the combobox is grouped by determining if a key was provided by which to group items (`groupItemsBy` prop)
- If grouped, the effect logic is skipped; if not, it runs
- added unit tests for grouped, controlled, and basic comboboxes to confirm that the scroll function runs or not as is appropriate

**Contributing to Twilio**

### DIFF
--- a/.changeset/loud-bags-jog.md
+++ b/.changeset/loud-bags-jog.md
@@ -1,5 +1,6 @@
 ---
 '@twilio-paste/combobox': minor
+'@twilio-paste/core': minor
 ---
 
 [Fix] Remediate erratic scrolling behavior for grouped combobox. Scroll function for the virtualization library was being called even though grouped comboboxes are not virtualized. The jumping behavior on scroll was observed because the indices in the virtual list and the actual list did not match. Added guard to prevent the scroll function from the virtualization library from being called.
--- a/.changeset/loud-bags-jog.md
+++ b/.changeset/loud-bags-jog.md
@@ -1,0 +1,5 @@
+---
+'@twilio-paste/combobox': minor
+---
+
+[Fix] Remediate erratic scrolling behavior for grouped combobox. Scroll function for the virtualization library was being called even though grouped comboboxes are not virtualized. The jumping behavior on scroll was observed because the indices in the virtual list and the actual list did not match. Added guard to prevent the scroll function from the virtualization library from being called.
--- a/packages/paste-codemods/tools/.cache/mappings.json
+++ b/packages/paste-codemods/tools/.cache/mappings.json
@@ -20,7 +20,6 @@
   "Card": "@twilio-paste/core/card",
   "ChatBubble": "@twilio-paste/core/chat-log",
   "ChatMessage": "@twilio-paste/core/chat-log",
-  "MessageVariantContext": "@twilio-paste/core/chat-log",
   "Checkbox": "@twilio-paste/core/checkbox",
   "CheckboxDisclaimer": "@twilio-paste/core/checkbox",
   "CheckboxGroup": "@twilio-paste/core/checkbox",
--- a/packages/paste-codemods/tools/.cache/mappings.json
+++ b/packages/paste-codemods/tools/.cache/mappings.json
@@ -20,6 +20,7 @@
   "Card": "@twilio-paste/core/card",
   "ChatBubble": "@twilio-paste/core/chat-log",
   "ChatMessage": "@twilio-paste/core/chat-log",
+  "MessageVariantContext": "@twilio-paste/core/chat-log",
   "Checkbox": "@twilio-paste/core/checkbox",
   "CheckboxDisclaimer": "@twilio-paste/core/checkbox",
   "CheckboxGroup": "@twilio-paste/core/checkbox",
--- a/packages/paste-core/components/combobox/__tests__/index.spec.tsx
+++ b/packages/paste-core/components/combobox/__tests__/index.spec.tsx
@@ -1,16 +1,37 @@
 import * as React from 'react';
 import * as _ from 'lodash';
 import {render, screen, fireEvent} from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
 import type {RenderOptions} from '@testing-library/react';
 import {Theme} from '@twilio-paste/theme';
 import {Button} from '@twilio-paste/button';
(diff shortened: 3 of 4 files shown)


packages/paste-codemods/tools/.cache/mappings.json
```diff
@@ -20,6 +20,7 @@
   "Card": "@twilio-paste/core/card",
   "ChatBubble": "@twilio-paste/core/chat-log",
   "ChatMessage": "@twilio-paste/core/chat-log",
+  "MessageVariantContext": "@twilio-paste/core/chat-log",
```
andipants12, line 23: This is odd, I didn't modify anything in codemods, but I saw this when I went to commit changes. 
I think this component was recently added, but shouldn't this diff be in a different pull request? 🤔

packages/paste-core/components/combobox/__tests__/index.spec.tsx
```diff
@@ -277,6 +317,22 @@ describe('Combobox', () => {
...
     });
+
+    it('should not run react virtual\'s "scroll to" function when "groupItemsBy" is defined as a string', () => {
+      // bc grouped comboboxes are not yet virtualized
+      render(<GroupedMockCombobox />, {wrapper: ThemeWrapper});
+
+      const targetIndex = 1;
+      const target = groupedItems[targetIndex];
+
+      userEvent.hover(screen.getByText(target.label));
```
andipants12, line 328: Hover to trigger state change `highlightedIndex`, which controls the effect.

packages/paste-core/components/combobox/__tests__/index.spec.tsx
```diff
@@ -277,6 +317,22 @@ describe('Combobox', () => {
...
+
+      expect(screen.getByRole('textbox').getAttribute('aria-activedescendant')).toMatch(
```
andipants12, line 330: Indirect confirmation that the correct item is active (which indirectly means it is highlighted).

packages/paste-core/components/combobox/__tests__/index.spec.tsx
```diff
@@ -277,6 +317,22 @@ describe('Combobox', () => {
...
+        /downshift-([1-9]\d\d|[1-9]\d|\d)-item-1/g
```
andipants12, line 331: Just want to make sure we have the correct item index, don't really care about the uuid prefix from downshift.

packages/paste-core/components/combobox/__tests__/index.spec.tsx
```diff
@@ -277,6 +317,22 @@ describe('Combobox', () => {
...
+      );
+
+      expect(mockScrollToIndex).not.toHaveBeenCalled();
```
andipants12, line 334: ✅ I verified that this expect statement fails when that guard is removed (meaning the scroll function is in fact called).

packages/paste-core/components/combobox/__tests__/index.spec.tsx
shleewhite: Oh, this answers one of my questions above - so even if the list doesnt actually scroll it still triggers the state change?

shleewhite:
Looks way better! I tried it out in a CodeSandbox: https://codesandbox.io/s/twilio-paste-nextjs-template-forked-9sze64?file=/pages/index.tsx

The only thing I noticed is that if an option is partially in view, it doesn't snap to be in view when it is highlighted... but I guess that is expected now because virtual is what handles that.

.changeset/loud-bags-jog.md
```diff
@@ -0,0 +1,6 @@
+---
+'@twilio-paste/combobox': minor
+'@twilio-paste/core': minor
+---
+
+[Fix] Remediate erratic scrolling behavior for grouped combobox. Scroll function for the virtualization library was being called even though grouped comboboxes are not virtualized. The jumping behavior on scroll was observed because the indices in the virtual list and the actual list did not match. Added guard to prevent the scroll function from the virtualization library from being called.
```
SiTaggart, line 6: suggestion: we tend to do the component name in the square brackets so that we can identify the package in the Core bundle changelog

SiTaggart (approved):
just a minor tweak the changeset, but the change and test coverage seem pretty great to me

packages/paste-core/components/combobox/__tests__/index.spec.tsx
```diff
@@ -208,6 +233,21 @@ describe('Combobox', () => {
...
       expect(renderedTextbox.getAttribute('required')).toEqual('');
     });
+
+    it('should call scroll function when highlighted index changes', () => {
+      render(<ComboboxMock />, {wrapper: ThemeWrapper});
+
+      const targetIndex = 1;
+      const target = items[targetIndex];
+
+      userEvent.hover(screen.getByText(target));
```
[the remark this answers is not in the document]
andipants12: Initial highlighted index is -1. We need to simulate a user event because the index change is coupled to the mouseover event.

These are things internal to the primitive from dropdown.

packages/paste-core/components/combobox/__tests__/index.spec.tsx
andipants12: Yeah the highlighted index is coupled to mouseover position, not scroll position.

shleewhite (approved):

.changeset/loud-bags-jog.md
SiTaggart: Bump, still need to Component name in the square brackets [ ]

PR #2347, opened 2022-04-06
fix: correct close icon hover state

gloriliale:
Hover states for the close icons on our components weren't showing up; as conjectured in the ticket, this is because they were being overwritten by color props in the CloseIcon components that were left behind from the previous implementation (before the `secondary_icon` variant in button). Taking those props out fixes this.

### DIFF
--- a/.changeset/eleven-mice-repair.md
+++ b/.changeset/eleven-mice-repair.md
@@ -6,4 +6,4 @@
 '@twilio-paste/core': patch
 ---
 
-Adjust close icon's hover state color, which was previously being overwritten.
+[Alert, Modal, Popover, Toast] Adjust close icon's hover state color, which was previously being overwritten.
--- a/.changeset/eleven-mice-repair.md
+++ b/.changeset/eleven-mice-repair.md
@@ -1,0 +1,9 @@
+---
+'@twilio-paste/alert': patch
+'@twilio-paste/modal': patch
+'@twilio-paste/popover': patch
+'@twilio-paste/toast': patch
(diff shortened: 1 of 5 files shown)


.changeset/eleven-mice-repair.md
```diff
@@ -0,0 +1,9 @@
+---
+'@twilio-paste/alert': patch
+'@twilio-paste/modal': patch
+'@twilio-paste/popover': patch
+'@twilio-paste/toast': patch
+'@twilio-paste/core': patch
+---
+
+Adjust close icon's hover state color, which was previously being overwritten.
```
gloriliale, line 9: I'm not sure if this is the best way to format the changeset for this type of change; let me know if there's a more appropriate way!

TheSisb: suggestion: It's helpful to prefix the message here with the package names in brackets so it's easier to parse the affected packages when reviewing the core changelog digest here https://paste.twilio.design/core/changelog/

TheSisb (approved):
Sweet!

shleewhite (approved):
Woo! Looks great! Just the one thing Shadi mentioned.

.changeset/eleven-mice-repair.md
nkrantz: suggestion: I also think we might want to pull these into separate changesets (one per package, core in each)

TheSisb: Since it is the same change message, it can be combined into this one changeset file.

PR #2351, opened 2022-04-08
fix: add react/jsx-no-literals rule

shleewhite:
- [x] update `eslint-plugin-react` to get all features of the `react/jsx-no-literals` rule
- [x] enable `react/jsx-no-literals` rule for paste-core
- [x] disable `react/jsx-no-bind` and `react/no-unused-prop-types` because it is out of scope to fix those issues
- [x] fix small lint errors that came up when upgraded the plugin
- [ ] disable `react/jsx-no-literals`  for allowed strings

**Contributing to Twilio**


packages/paste-core/components/toast/src/ErrorToast.tsx
```diff
@@ -8,17 +8,27 @@ const ErrorToast = React.forwardRef<HTMLDivElement, ToastProps>(({element = 'TOA
     <Box
```
shleewhite, line 8: The new lint rule checks these for strings:
* Children
* Props
* Attributes (rare for us, we don't usually use html elements directly - we usually use Box or another component instead)

When we run the rule with all 3 checks in Core files that aren't stories or tests (we expect there to be hardcoded strings in those files), it has ~2000 errors and all of them are with props. This is because it flags _all_ strings passed as props. There are a lot of strings that we never need to translate (element, style props, variant names, etc.)

The only ways to customize the rule for props are to:

1. Turn off the rule for props
2. Disable the rule for entire files
3. Disable the rule for individual lines
4. Add allowed strings to a whitelist

There are pros/cons for each of these options:

1. Turn off the rule for props: we probably want the rule on because most of the strings found in @TheSisb  's audit were in props.
2. Disable the rule for entire files: if you disable the rule for the entire file - nothing stops someone from accidentally adding an english string later
3. Disable the rule for individual lines: Very repetitive to add individual comments, makes code look ugly
4. Add allowed strings to a whitelist: Hard to maintain, would need to be a really really long list

It seems like option 3 (disable the rule for individual lines) is the best option if we really want this rule. However, before I add a whole bunch of eslint-disable comments, I wanted to get the team's thoughts. Do y'all like this? If you look at the `chore: disable react/jsx-no-literals for allowed strings` commit, you can see a few files updated with the ignores.

PR #2353, opened 2022-04-08
test: site map VRT

SiTaggart:
Quick a dirty site VRT for use when doing up and coming large system upgrades such as Gatsby v4 and Emotion 11.

N.B. This isn't designed to be a long term solution, but we need something in the near term.

- Creates a simple command to VRT a particular url
- Creates tests for all the urls listed in the sitemap. This is manual for now. You might be able to automate it so new pages are automatically caught but that seems out of scope for now.
- runs cypress on the preview deploy Netlify, significantly reducing wait time.
- runs cypress in parallel across 3 boxes.

This roughly adds 100 new tests to cypress but halves the average completion time.

Waiting for ~30 minutes

image

Waiting for ~15 minutes

image

### 1. Applitools Github Check

The Applitools Github Check has a limitation where it only listens for "a completed ci check" before it reports back what the applitools status is. It doesn't care which one, it assumes there is only one.

What that means is, they cut that batch when ever the next CI check reports back as finished, even if it's not the one running the Cypress check. So you might see a status of "2 out of 43" when you know there are over 100 VRTs in that batch. It's just another CI checked finished, and Applitools went and reported back the current status of what was running.


package.json
```diff
@@ -66,7 +66,7 @@
     "test": "jest",
     "test:react-16": "USE_REACT_16=true jest",
     "test:coverage": "jest --coverage",
-    "test:website": "start-server-and-test 'yarn serve:website' http://localhost:9000 'yarn cypress run --record'",
+    "test:website": "yarn cypress run --record",
```
zahnster, line 69: Why are you keeping the --record flag?

SiTaggart: That's how Cypress records the test run for you to watch in the dashboard, no?

cypress/integration/sitemap-vrt/index.spec.ts
```diff
@@ -0,0 +1,83 @@
+// ℹ️ If we use anon functions instead of arrow functions, we can leverage Mocha's context and pull the test name directly from this
+describe('Full Site VRT', function () {
+  describe('Visual regression tests', () => {
+    it('vrt homepage', () => {
```
TheSisb, line 4: question: did you manually generate these or did you use some kind of script? We will have to add to these if we add/remove/change any page routes right?

SiTaggart: OK, my experiment failed. Sticking to this for now

shleewhite (approved):
This looks good to me... My one open question is when do the rest of the cypress tests run? Is it only when the label is added?

cypress/integration/sitemap-vrt/index.spec.ts
shleewhite: Thought: Can we make an object of pages/urls and then loop through that? The sidebar navigation tests do this so I didn't have to write really repetitive tests.

SiTaggart:
> This looks good to me... My one open question is when do the rest of the cypress tests run? Is it only when the label is added?

So, all cypress tests run on every push. The only difference being, when you add the label we set the environment variable `USE_CYPRESS_EYES` to be `true`, which makes applitools take the snapshots. The cypress tests are never _not_ run.

TheSisb (approved):

PR #2360, opened 2022-04-12
chore: add custom svg-needs-viewBox eslint rule

gloriliale:
Added new eslint rule to check for svgs having a viewBox attribute.

- [x] Add rule
~~- [ ] Restrict rule to paste-icons package (right?)~~
- [x] Fix instances of errors from rule
- [x] Add test


.eslint/rules/icon-has-viewBox.js
```diff
@@ -0,0 +1,62 @@
+/**
+ * @fileoverview All Paste icon svgs should have a viewBox attribute
+ * @author Glorilí Alejandro
+ */
+
+// TODO: have this only run on our icons package in its src directory
```
gloriliale, line 6: This is something we want, right? or do we want this rule to apply everywhere?

.eslint/rules/icon-has-viewBox.test.js
```diff
@@ -0,0 +1,9 @@
+// import noUnusedOwnProps from './no-stevenpetryk';
```
gloriliale, line 1: right now this is just commented out code to have handy as a basic test structure, pls ignore

package.json
```diff
@@ -161,6 +161,7 @@
     "eslint": "7.23.0",
     "eslint-config-airbnb-typescript": "12.0.0",
     "eslint-config-prettier": "6.11.0",
+    "eslint-config-svg-needs-viewBox": "link:./.eslint/rules",
```
zahnster, line 164: Since you made a plugin and not a config, I do not think this is needed?

.eslint/rules/icon-has-viewBox.js
zahnster: 💭 Thought: I could go either way here. It looks like we do have a handful of failures in svgs outside of our icons library, so if we keep the scope global we'll want to fix those. From local testing, filtering to just the icons package doesn't have a lot of perf advantages, so might be fine to just have it work globally.

zahnster:
Idea: Might be nice to try your hand at a fix function? https://eslint.org/docs/developer-guide/working-with-rules#applying-fixes

.eslint/rules/icon-has-viewBox.js
gloriliale: sweet, I made it global and fixed those errors so we should be good to go!

.eslint/rules/icon-has-viewBox.js
```diff
@@ -0,0 +1,60 @@
...
+              // if any of them are 'viewBox', we are done here.
+              return;
+            }
+          }
+          // if we are in an svg node (which in here, we know we are) and
+          // we haven't exited out already from viewBox being present,
+          // then we need to report this node. Bad node!
+          context.report({
+            node,
+            message: 'Every icon needs to have a viewbox attribute',
```
TheSisb, line 54: thought: looking at the code, this is checking that any `svg` has a `viewBox` attribute. SVGs can be used for more than icons, such as for illustrations. It may be confusing that the wording here is "icon" if this reports for a non-icon SVG.

packages/paste-website/src/components/icons/TwilioLogo.tsx
```diff
@@ -13,7 +13,7 @@ const TwilioLogo = React.memo(({title = 'Twilio Logo', decorative = true, classN
   const titleId = useUID();
   return (
     <span style={{color, width: size, height: size, display: 'inline-block'}} className={className}>
-      <svg role="img" aria-hidden={decorative} width="100%" height="100%" aria-labelledby={titleId}>
+      <svg role="img" aria-hidden={decorative} width="100%" height="100%" viewBox="0 0 20 10" aria-labelledby={titleId}>
```
TheSisb, line 16: thought: oof `0 0 20 10`. That's a little odd.

.eslintrc.js
```diff
@@ -19,6 +19,7 @@ module.exports = {
     'prettier',
     'prettier/react',
     'prettier/@typescript-eslint',
+    'plugin:svg-needs-viewBox/all',
```
TheSisb, line 22: suggestion: I kind of think this should be called `plugin:paste` so it can contain all our custom rules in the future.

.eslint/rules/index.js
```diff
@@ -0,0 +1,19 @@
...
+const ruleFiles = fs.readdirSync('.eslint/rules').filter((file) => file !== 'index.js' && !file.endsWith('test.js'));
+
+const configs = {
+  all: {
+    plugins: [projectName],
+    rules: Object.fromEntries(ruleFiles.map((file) => [`${projectName}/${path.basename(file, '.js')}`, 'error'])),
+  },
+};
+
+const rules = Object.fromEntries(ruleFiles.map((file) => [path.basename(file, '.js'), require('./' + file)]));
```
TheSisb, line 17: chore: can we add comments above this line and above line 13 that explains what these one-liners are returning?  A link at the top of the file to documentation source would be great too.

package.json
gloriliale: I took it out and installed again and all looks well. Good catch, thank you!

.eslint/rules/icon-has-viewBox.js
```diff
@@ -0,0 +1,60 @@
...
+      // this lets us work with all the JSXElement nodes
+      JSXElement(node) {
+        // I want to get the openingElement node of each, since that's
+        // where the details about both it being an svg and the
+        // attributes live.
+        const elementDetails = node.openingElement;
+
+        // To get find the name, we need to go into the openingElement
+        // we just got, get its JSXIdentifier, and then get that node's name
+        if ((getJSXIdentifier(elementDetails) || {}).name === 'svg') {
```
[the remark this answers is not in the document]
gloriliale: Great question! This rule is going to look at every node that's a JSXElement in the whole repo, and for various reasons (some of those nodes might not have a JSXIdentifier bc they don't have an `openingElement` or `name`) the `getJSXIdentifier` call could be returning `undefined`. We can't do `.name` on undefined since it breaks, but we can do it on this empty object so that it will be falsey in comparison to the 'svg' string, and the code can keep running.

Lemme know if that makes sense or not, happy to keep talkin' about it!

packages/paste-website/src/components/icons/TwilioLogo.tsx
gloriliale: There was a fair bit of trial and error to get it to look like it does currently and these values were the ones I found that work 🤷  . Maybe I'll get better intel from design when they have a look.

.eslint/rules/icon-has-viewBox.js
gloriliale: This is a great point, I definitely want to make this message better. I'm thinking of changing it to "An svg must have a viewBox". Thoughts?

.eslintrc.js
gloriliale: I was considering that as well, and I'm into it, especially if we think we'll have more of these. And the repetition was driving me nuts anyway 🤣 
:paste:

gloriliale: I would like it to be more specific tho. how about `plugin:paste-eslint`?

packages/paste-website/src/components/icons/TwilioLogo.tsx
serifluous: Where is this being used? Should we use the Paste component instead?

serifluous: Nvm it's the logo + logomark. It's cut off on the website

Screen Shot 2022-04-13 at 1 06 47 PM

.eslintrc.js
gloriliale: from standup: since we'll probably be publishing an external npm package of this plugin, we'll be naming this `paste-internal`

packages/paste-website/src/components/icons/TwilioLogo.tsx
gloriliale: @serifluous do you have the size of this logo anywhere? I'm just haphazardly plugging numbers into viewBox at this point and could use some guidance!

.eslint/rules/index.js
gloriliale: Just added some comments here! Lemme know what ya think!

.eslint/rules/icon-has-viewBox.js
TheSisb: I like it!

packages/paste-website/src/components/icons/TwilioLogo.tsx
TheSisb: You may get a quicker response if you send a designer the asset in Slack directly.

gloriliale: This has been addressed!

.eslint/rules/index.js
```diff
@@ -0,0 +1,39 @@
...
+
+*/
+
+const fs = require('fs');
+const path = require('path');
+
+const projectName = 'paste-internal';
+
+// This allRuleFiles constant will include ALL the rules in our rules directory.
+// We create an array of these rule files by filtering out this index.js file and any test files.
```
nkrantz, line 15: praise: helpful comments

.eslint/__tests__/svg-needs-viewBox.test.js
```diff
@@ -0,0 +1,78 @@
...
+      jsx: true,
+    },
+  },
+});
+
+//------------------------------------------------------------------------------
+// Tests
+//------------------------------------------------------------------------------
+
+var ruleTester = new RuleTester();
```
nkrantz, line 28: question: is there a reason this uses `var` instead of `const`?

.eslint/__tests__/svg-needs-viewBox.test.js
```diff
@@ -0,0 +1,78 @@
...
+ */
+'use strict';
+
+//------------------------------------------------------------------------------
+// Requirements
+//------------------------------------------------------------------------------
+
+const rule = require('../rules/svg-needs-viewBox.js');
+
+const RuleTester = require('eslint').RuleTester;
```
nkrantz, line 13: question (non-blocking): why did you go with the `require` instead of the `import` statement you have below (commented out)?

.eslint/__tests__/svg-needs-viewBox.test.js
```diff
@@ -0,0 +1,78 @@
...
+        {
+          message: 'An svg must have a viewbox attribute',
+          type: 'JSXElement',
+        },
+      ],
+    },
+  ],
+});
+
+// import {RuleTester} from 'eslint';
```
nkrantz, line 45: nit: reminder to delete this commented test

TheSisb, line 45: TODO: remove?

TheSisb (approved):

nkrantz (commented):
This is so nice to have in our linter! Making the most of the HAP 🥇 

I've never read through an eslint rule or tests before but you made it super clear with the comments, thank you!! I left a few tiny questions/comments but overall it looks really good.

.eslint/__tests__/svg-needs-viewBox.test.js
gloriliale: 😬 yup! done

gloriliale: 💯 done!

nkrantz (approved):

.eslint/__tests__/svg-needs-viewBox.test.js
gloriliale: ah this was left in from a source, good catch! Just changed it.

.eslint/__tests__/svg-needs-viewBox.test.js
gloriliale: It won't work with `import`; I believe it's because we're at the root? I talked about it w/ @zahnster but can't remember the exact reason. Any insight here @zahnster?

PR #2363, opened 2022-04-13
fix(icons): use React.forwardRef in icons

shleewhite:
- [x] Update the IconWrapper and reactIconTemplate to use React.forwardRef
- [x] Add prettierrc in /icons to configure how svg-to-react prettifies the icons
- [x] Upgrade svg-to-react package
- [x] Update Avatar to make the icon prop work with the new forwardRef icons

**Things of note to review:**
Please look over the updated icon template (`packages/paste-icons/tools/templates/reactIconTemplate.js`), because it is a string its kinda hard to read so it may be easier to review a couple of the icons tsx files. No need to review every single icon file tho because they are all generated from the template.

**Contributing to Twilio**


packages/paste-icons/.prettierrc
```diff
@@ -0,0 +1,9 @@
+{
+  "parser": "babel-ts",
```
shleewhite, line 2: Note: We need a specific .prettierrc file just for icons because svg-to-react queries it and it needs the parser.

I tried adding just the parser to the global prettierrc, but that caused problems because not all files in paste are typescript (ex. js, json).

shleewhite: I wanted to add this info as a comment in the file itself but we don't allow comments in JSON :///

packages/paste-core/components/avatar/package.json
```diff
@@ -24,6 +24,10 @@
     "clean": "rm -rf ./dist",
     "tsc": "tsc"
   },
+  "dependencies": {
+    "@types/react-is": "^17.0.3",
```
shleewhite, line 28: Note: I ended up needing to include the `@types/react-is` package as well as `react-is`

packages/paste-icons/tools/templates/reactIconTemplate.js
```diff
@@ -13,20 +13,22 @@ export interface ${componentName}Props extends IconWrapperProps {
   decorative: boolean;
 }
 
-const ${componentName}: React.FC<${componentName}Props> = ({as, display, element = 'ICON', size, color, title, decorative}) => {
+const ${componentName} = React.forwardRef(function ${componentName}(
```
shleewhite, line 16: Note: looking at [the React docs](https://reactjs.org/docs/forwarding-refs.html#displaying-a-custom-name-in-devtools), it seems like the way to add a display name to forwardRef components is to use a named function like this:

```tsx
 const MyIcon = React.forwardRef(function MyIcon() {})
```

but when I did that, Avatar still wasn't able to get the displayName correctly. So, I ended up leaving line 33 in too.

packages/paste-icons/src/helpers/IconWrapper.tsx
```diff
@@ -7,15 +7,8 @@ export interface IconWrapperProps extends BoxElementProps, Pick<BoxStyleProps, '
...
-const IconWrapper: React.FC<IconWrapperProps> = ({
-  as = 'span',
-  color = 'currentColor',
-  display = 'block',
-  element = 'ICON',
-  size = 'sizeIcon30',
-  ...props
-}) => {
-  return (
+const IconWrapper = React.forwardRef<HTMLElement, IconWrapperProps>(
```
TheSisb, line 10: issue: we're missing the named function here `function IconWrapper({as ...`

packages/paste-icons/tools/actions/convertNewAction.js
```diff
@@ -14,15 +14,21 @@ const {reactIconTemplate} = require('../templates/reactIconTemplate');
 const {writeToFile} = require('../../../../tools/utils/writeToFile');
 
 // Converts raw svg to react component
-function performFileConversion(fileName, outputPath, options) {
+async function performFileConversion(fileName, outputPath, options) {
```
TheSisb, line 17: question: does this need to be async? line 21's async function should be enough to wrap the await on line 27 maybe

packages/paste-icons/tools/templates/reactIconTemplate.js
TheSisb: > but when I did that, Avatar still wasn't able to get the displayName correctly. So, I ended up leaving line 33 in too.

question: does this mean line 33 still works? I thought it wasn't working and then started to work when we made this change.

packages/paste-icons/src/helpers/IconWrapper.tsx
shleewhite: I dont think we actually need to do this. Looking at the avatar storybook - IconWrapper's display name shows up as expected without naming the function.

packages/paste-icons/tools/actions/convertNewAction.js
shleewhite: I think you're right! I changed this when I was still trying to debug the svg-to-react issues.

gloriliale (approved):
lgtm!

TheSisb:
thought: I think we should still add the "correct" way of naming the functions in forwardRef so it works as intended in devTools.

TheSisb (approved):

SiTaggart (approved):

PR #2364, opened 2022-04-13
chore: add ADR template for future ADRs

SiTaggart:
Create a place for internal docs, and a template for creating and sharing architecture decision records


01_internal_docs/engineering/ADR/0000-template.md
```diff
@@ -0,0 +1,64 @@
+# [short title of solved problem and solution]
```
TheSisb, line 1: question: what does the `01` signify in the root folder name?

SiTaggart: Sticks it up at the top of the directory list so internal docs aren't fully buried? 

Kinda still workshopping that idea, but that was the current thinking

shleewhite: Could we use the github wiki for these docs?

SiTaggart: So, we discussed that a little bit in [Slack here](https://twilio.slack.com/archives/C02LKSS1H41/p1648580519288259). But essentially highlights include:

- The wiki doesn't support nested pages, which is super not helpful. IA will be useful to organise our content
- If we use markdown pages in the repo we can get peer review on our documentation via PRs
- ADRs require a link to the PR so you can review the discussion that was had about it. See line 5 in the template
- If we use markdown pages in the repo we can write danger rules that prompt us to update the docs when we make changes to specific things.
- If we use markdown pages we can leverage this [new diagram feature](https://github.blog/2022-02-14-include-diagrams-markdown-files-mermaid/), which doesn't seem to work in the Wiki https://github.blog/2022-02-14-include-diagrams-markdown-files-mermaid/

SiTaggart: > Sticks it up at the top of the directory list so internal docs aren't fully buried?

Which, in reality it kind of doesn't do due to `.` folders.

I need better name suggestions... go:

gloriliale: `!_internal_docs`? brings it to the top 🤷

gloriliale (commented):
Could there be an example ADR in this directory as well?

01_internal_docs/engineering/ADR/0000-template.md
TheSisb: nitpick/question: does it being at the top matter in a significant way?

shleewhite: yeah... i feel like `__internal_docs/` is fine?

SiTaggart: > nitpick/question: does it being at the top matter in a significant way?

Not attached at all, just felt like "a thing". Dunno why honestly

TheSisb (approved):

01_internal_docs/engineering/ADR/0000-template.md
SiTaggart: Went with `INTERNAL_DOCS/` All caps to match markdown file naming convention we already have

shleewhite (approved):
i like the new folder name!

PR #2373, opened 2022-04-28
chore(remove-knobs): Remove Storybook Knobs

zahnster:
* Removes deprecated Storybook knobs from dependencies, config, and all stories that had been using them.
  * In certain cases (eg: Alert, Heading, Input), we removed stories which were built to be primarily used with Knobs controls.
  * In other cases (eg: Anchor, Card), we simply removed the Knobs controlled variables and replaced the values with the declared defaults.
* Removes Knobs references from plop template.
* PR also removes the Knobs addon from the Storybook config, and the Knobs dependency in our package.json.

### DIFF
--- a/packages/paste-core/components/anchor/stories/index.stories.tsx
+++ b/packages/paste-core/components/anchor/stories/index.stories.tsx
@@ -1,6 +1,5 @@
 import * as React from 'react';
 import {action} from '@storybook/addon-actions';
-import {withKnobs, select, text} from '@storybook/addon-knobs';
 import {Box} from '@twilio-paste/box';
 import {Stack} from '@twilio-paste/stack';
 import {Heading} from '@twilio-paste/heading';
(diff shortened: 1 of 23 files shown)


packages/paste-core/components/anchor/stories/index.stories.tsx
```diff
@@ -1,57 +1,52 @@
...
 export default {
   title: 'Components/Anchor',
-  decorators: [withKnobs],
   component: Anchor,
 };
 
 export const Default = (): React.ReactNode => {
-  const tabIndexOptions = select('tabIndex', AnchorTabIndexOptions, 0) as AnchorTabIndexes;
-  const targetOptions = select('target', AnchorTargetOptions, '_self') as AnchorTargets;
+  const tabIndexOptions = 0 as AnchorTabIndexes;
```
shleewhite, line 17: Question: if we're not using knobs, could we just skip passing tabIndexOptions to Anchor? It has tabindex 0 by default.

shleewhite: Same question for targetOptions

shleewhite: Talked with Jade: since we might be using controls instead of knobs, we will have to do similar stuff with having all the props as variables - so we might as well just leave it as is for now to reduce duplicated work.

SiTaggart (approved):

PR #2375, opened 2022-04-28
fix(radio-group): allow uncontrolled radio and radiogroup

TheSisb:
Radio and RadioGroup weren't behaving correctly in controlled and uncontrolled scenarios where varying props could be passed around. For example, if we used a RadioGroup with a `value` and `onChange`, what would happen if the child Radio had either `checked` or `defaultChecked`?  This PR addresses those issues, adds plenty of scenario stories, and updates our tests to catch future pitfalls.


packages/paste-core/components/radio-group/stories/radiogroup-uncontrolled.stories.tsx
```diff
@@ -0,0 +1,111 @@
...
+    </RadioGroup>
+  );
+};
+
+UncontrolledRadioGroupB.story = {
+  name: 'Uncontrolled RadioGroup with controlled Radios',
+  parameters: {
+    chromatic: {disableSnapshot: true},
+  },
+};
```
SiTaggart, line 111: question: I see that we have dedicated VRT stories now, but given that we identified gaps in our functionality, do we think we have sufficient VRT coverage for all the controlled, uncontrolled initial render scenarios?

I just noticed we weren't adding many VRT stories considering the changes we've had to make.

TheSisb: Since VRT only tests the visuals, I made sure we VRT'd all the possible visual states. I added unit tests to cover the controlled and uncontrolled radio behavior.

SiTaggart (approved):

zahnster (approved):
Looks great, appreciate the helpful comments in the code.

PR #2379, opened 2022-04-29
feat: add custom date/time filter group example

shleewhite:
Creates a live version of the custom date/time popover in a filter group on the filter group pattern page. I also added `react-hook-form` and `date-fn` to make the filtering easier.

**Contributing to Twilio**


packages/paste-website/src/pages/patterns/filter-group/index.mdx
```diff
@@ -109,20 +110,6 @@ export const pageQuery = graphql`
       </Heading>
     </Card>
   </Column>
-  <Column>
```
shleewhite, line 112: Note: ended up removing Stack and Flex from the list of ingredients because I didn't use them in the examples 👀

packages/paste-website/src/component-examples/filter-group/components/DateTimePopover.tsx
```diff
@@ -0,0 +1,123 @@
...
+        <Box display="flex" flexDirection="column" marginBottom="space50" rowGap="space50">
+          <Box display="flex" flexDirection="row" columnGap="space50">
+            <Box width="100%">
+              <Label htmlFor={`${startId}-date`} required>
+                Start date
+              </Label>
+              <DatePicker
+                id={`${startId}-date`}
+                aria-describedby={`${startId}-date-error`}
+                required
```
TheSisb, line 42: thought (non-blocking): Does this accept min/max props? If so we can try passing max to startDate and min to endDate as better UX

TheSisb (commented):
Really solid code here, nice work!

packages/paste-website/src/component-examples/filter-group/components/DateTimePopover.tsx
shleewhite: I'm gonna leave this as out of scope... I can add a ticket in the backlog to implement that.

TheSisb (approved):

packages/paste-website/src/pages/patterns/filter-group/index.mdx
```diff
@@ -194,11 +179,7 @@ While we don’t yet have an official custom date/time component, in the interim
 
 When the Custom option is selected, the Popover containing the Date Inputs and Time Inputs necessary would be displayed to allow customers to enter their desired range. Once the Apply button is clicked, the Popover would hide. If a customer wants to change their original selections, they must click on the Calendar link style button to open the Popover again.
```
nkrantz, line 180: question: the popover doesn't open for me when the custom option is selected, only when the calendar icon is clicked. this makes it a kind of confusing interaction because I can select custom without knowing how to actually specify the custom dates. similarly, if I click the cal icon to open the date range popover, and click "apply", it doesn't switch the menu selection back to "custom" so it looks like the previous selection is still being applied. lmk if this is confusing and I can explain on a zoom/huddle

nkrantz: just saw your comment about how shifting focus on click of "custom" wouldn't be accessible... I wonder if we can add some text to clarify the interaction? like "Custom (click cal icon to specify dates)" or something less wordy but to the same point. maybe just me but I was definitely confused when using this filter

PR #2383, opened 2022-04-30
chore(style-props): upgrade csstype to latest

andipants12:
This handles an upgrade to the latest version of `csstype`, which is required by our Gatsby and Emotion upgrades.

- [x] upgrade `csstype` at root of monorepo
- [x] upgrade `csstype` in `@twilio-paste/style-props` package
- [x] move additional type declarations from `@twilio-paste/box` and `@twilio-paste/text` to `@twilio-paste/style-props`
- [x] export newly added types from the style props package, and remove `csstype` as a dependency.

**Contributing to Twilio**


.changeset/clever-needles-count.md
```diff
@@ -0,0 +1,8 @@
+---
+'@twilio-paste/box': minor
+'@twilio-paste/text': minor
+'@twilio-paste/style-props': minor
```
SiTaggart, line 4: I actually buy this now, I think you're right, it's technically a minor even though the package is a major bump. It doesn't change the type signatures to the Paste consumer as we don't really directly expose it.

packages/paste-core/primitives/text/src/types.ts
```diff
@@ -14,14 +7,19 @@ import type {
   SpaceProps,
   TypographyProps,
   VerticalAlign,
+  CursorProperty,
+  OutlineProperty,
+  TextTransformProperty,
+  TransitionProperty,
+  FontVariantNumericProperty,
```
SiTaggart, line 14: Praise: Both this and Box consuming types from style-props rather than directly depending on csstype makes a tonne of sense

packages/paste-style-props/src/types/background.ts
```diff
@@ -9,11 +9,11 @@ export type BackgroundColorOptions = keyof ThemeShape['backgroundColors'];
 export type BackgroundColor = ResponsiveValue<BackgroundColorOptions | 'none' | 'transparent' | 'inherit'>;
 
 // CSS native
-export type BackgroundImageOptions = CSS.BackgroundImageProperty;
-export type BackgroundSizeOptions = CSS.BackgroundSizeProperty<TLengthStyledSystem>;
-export type BackgroundPositionOptions = CSS.BackgroundPositionProperty<TLengthStyledSystem>;
-export type BackgroundRepeatOptions = CSS.BackgroundRepeatProperty;
-export type BackgroundAttachmentOptions = CSS.BackgroundAttachmentProperty;
+export type BackgroundImageOptions = Properties<TLengthStyledSystem>['backgroundImage'];
```
SiTaggart, line 12: Question: Is the reuse of the `<TLengthStyledSystem>` type argument / annotation on things that are not length or width type properties intentional? It was only ever used on widths, heights and other box like length dimensions to cast them to styled system.

package.json
```diff
@@ -223,7 +223,8 @@
   },
   "resolutions": {
     "@emotion/core": "10.0.28",
-    "react-hot-loader": "^4.13.0"
+    "react-hot-loader": "^4.13.0",
+    "csstype": "npm:csstype@3.0.11"
```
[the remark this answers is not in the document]
andipants12: This instructs yarn to resolve the alias from the npm registry. Check out the [yarn docs on this and other available protocols](https://yarnpkg.com/features/protocols#table).

I ended up using this protocol for the legacy resolutions (e.g. `emotion-theming`) to prevent installation of the legacy packages, essentially to replace them with the v11 equivalence.

In this case, it's not needed, since the resolution is for the version only. I did it this way purely for consistency.

packages/paste-style-props/src/types/background.ts
andipants12: No, this was totally me making a best guess 😅 I wanted to ask about that in follow up. I'm guessing they are now removed, but if not, go forth.

package.json
```diff
@@ -224,7 +224,7 @@
   "resolutions": {
     "@emotion/core": "10.0.28",
     "react-hot-loader": "^4.13.0",
-    "csstype": "npm:csstype@3.0.11"
+    "csstype": "^3.0.11"
```
andipants12, line 227: nice, thanks!

packages/paste-style-props/src/types/css-props.ts
```diff
@@ -0,0 +1,21 @@
+import type {Properties} from 'csstype';
+
+export type AnimationProperty = Properties['animation'];
+export type AppearanceProperty = Properties['appearance'];
+export type BoxSizingProperty = Properties['boxSizing'];
+export type ClipProperty = Properties['clip']; // 'clip' is deprecated
```
andipants12, line 6: I can remove this comment, looks like technically it's deprecated, but there is a special "bucket" on the interface for deprecated types so... 🤷‍♀️

andipants12:
@SiTaggart Thanks for cleaning this up! Looks good to me ✅

packages/paste-core/primitives/box/src/types.ts
```diff
@@ -7,21 +7,12 @@ import type {
   ShadowProps,
   SpaceProps,
   TypographyProps,
-} from '@twilio-paste/style-props';
-import type {CustomTheme} from '@twilio-paste/customization';
-import type {
   AnimationProperty,
   AppearanceProperty,
-  BorderCollapseProperty,
-  BorderSpacingProperty,
```
SiTaggart, line 16: Moved border props to `BorderProps`

packages/paste-core/primitives/box/src/types.ts
```diff
@@ -7,21 +7,12 @@ import type {
...
   BoxSizingProperty,
   ClipProperty,
   CursorProperty,
   FloatProperty,
-  FontVariantNumericProperty,
-  ListStyleImageProperty,
-  ListStylePositionProperty,
-  ListStyleTypeProperty,
```
SiTaggart, line 24: Moved font variant and list styles to `TypographyProps`

packages/paste-core/primitives/text/src/types.ts
```diff
@@ -14,19 +7,19 @@ import type {
...
   content?: string;
   cursor?: CursorProperty;
   display?: Display;
-  outline?: OutlineProperty<string>;
-  // Do not document, we prefer if folks do not use this property for i18n.
-  textTransform?: TextTransformProperty;
+  outline?: OutlineProperty;
   transition?: TransitionProperty;
   verticalAlign?: VerticalAlign;
-  fontVariantNumeric?: FontVariantNumericProperty;
```
SiTaggart, line 29: `textTransform` was a duplicate already declared on `TypographyProps`

`fontVariantNumber` moved to `TypographyProps`

packages/paste-style-props/src/types/border.ts
```diff
@@ -58,11 +59,19 @@ export interface BorderStyleProps {
   borderLeftStyle?: BorderLeftStyle;
 }
 
+export type BorderCollapseProperty = Properties['borderCollapse'];
+export type BorderCollapse = ResponsiveValue<BorderCollapseProperty>;
+
+export type BorderSpacingProperty = Properties['borderSpacing'];
+export type BorderSpacing = ResponsiveValue<BorderSpacingProperty>;
+
```
SiTaggart, line 67: Pulled in border props from generic css property types to keep border style props all in one place

packages/paste-style-props/src/types/css-props.ts
SiTaggart: I think it's fine 🤷🏻‍♂️

packages/paste-style-props/src/types/typography.ts
```diff
@@ -38,6 +42,10 @@ export type OverflowWrap = ResponsiveValue<OverflowWrapOptions>;
 export type TextDecoration = ResponsiveValue<TextDecorationOptions>;
 export type TextOverflow = ResponsiveValue<TextOverflowOptions>;
 export type TextTransform = ResponsiveValue<TextTransformOptions>;
+export type FontVariantNumeric = ResponsiveValue<FontVariantNumericProperty>;
+export type ListStyleImage = ResponsiveValue<ListStyleImageProperty>;
+export type ListStylePosition = ResponsiveValue<ListStylePositionProperty>;
+export type ListStyleType = ResponsiveValue<ListStyleTypeProperty>;
```
SiTaggart, line 48: Pulled in a few generic typography related css property types so type can be control from one location

package.json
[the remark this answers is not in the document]
SiTaggart: I think because we only ever pull from the npm registry, we probably shouldn't be using this. This line from that page is also worth noting (emphasis my own)

> While they work regardless of the context *we strongly recommend you to only use semver ranges* on published packages *as they are the one common protocol* whose semantic is clearly defined across *all package managers*

Useful context with regards to portability to pnpm later on.

andipants12:
@SiTaggart I just realized that we don't have `clipPath` exported here and it's used in some of our components (working on the emotion upgrade and I saw it). Should we add it before merge, or address later?

SiTaggart:
> I just realized that we don't have clipPath exported here and it's used in some of our components (working on the emotion upgrade and I saw it). Should we add it before merge, or address later?

I think generally speaking the style props could do with a little cosmetic work, and during a version bump of a dependency probably isn't the time to do it.

zahnster (approved):
Praise: Appreciate aligning csstypes imports through style-props. Nice work overall. 👍🏼

PR #2390, opened 2022-05-04
feat(website): add TokenCard component and rgbToHex util

TheSisb:
- [x] Adds a new TokenCard component that will render out token previews on the WIP tokens list page
- [x] Adds a preview for just BackgroundColor tokens. More to come in a later PR
- [x] Adds a RGB to HEX color utility with tests


packages/paste-website/src/components/tokens-list/token-card/index.tsx
```diff
@@ -0,0 +1,111 @@
+import * as React from 'react';
+import camelCase from 'lodash.camelcase';
```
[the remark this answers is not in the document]
TheSisb: I think it depends where you're importing it.  On dev tooling/scripts we don't ship anywhere, `import * as _ from 'lodash'` is perfectly fine and easier Dev Experience.

Everywhere else, probably as we do here.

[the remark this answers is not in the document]
SiTaggart: I think I managed to remove all the `from 'lodash` imports from the website at least in favour of individual packages in the gatsby upgrade to help keep things nice and intentional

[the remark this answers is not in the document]
shleewhite: 👀 i will update the filter group then to match this!

packages/paste-website/src/components/tokens-list/token-card/index.tsx
```diff
@@ -0,0 +1,132 @@
...
+      borderRadius="borderRadius30"
+      minHeight="sizeSquare170"
+    >
+      <Box margin="space40" width="sizeSquare200" justifyContent="center" alignItems="center" display="flex">
+        <PreviewComponent category={category} name={camelCaseName} value={value} />
+      </Box>
+      <Box
+        paddingY="space60"
+        paddingLeft="space70"
+        paddingRight="space70"
```
SiTaggart, line 80: nit: you can use `paddingX` here too

katieporter (approved):

packages/paste-website/src/components/tokens-list/token-card/index.tsx
```diff
@@ -0,0 +1,132 @@
...
+              >
+                {camelCaseName}
+              </Text>
+              <Tooltip text={tooltipText} state={tooltipState}>
+                <Button variant="secondary_icon" size="icon_small" onClick={handleCopyName}>
+                  <CopyIcon decorative />
+                  <span aria-live="polite">
+                    <ScreenReaderOnly>{clipboard.copied ? 'Copied token name!' : 'Copy token name'}</ScreenReaderOnly>
+                  </span>
+                </Button>
```
SiTaggart, line 106: issue(maybe): this is the screenshot from my chromatic comment. Notice the outline isn't truly a square, because we're applying column gap to the right on the icon due to the presence of screen reader text. Sure this is a known bug

image

packages/paste-website/src/components/tokens-list/token-card/index.tsx
```diff
@@ -0,0 +1,132 @@
...
+    setTooltipText(clipboard.copied ? 'Copied!' : 'Copy token name');
+  }, [clipboard.copied]);
+
+  // Reakit has a bug where the tooltip doesn't recalc position on content changes
+  // This is a workaround until we upgrade to Ariakit with Floating UI fixes
+  // https://github.com/twilio-labs/paste/discussions/2037
+  React.useEffect(() => {
+    // This prevents the tooltip from showing up on mount
+    if (isFirstRender.current) {
+      isFirstRender.current = false;
```
SiTaggart, line 55: thought(outloud): I really need to read a "what you can use refs for" blog post because I would never have thought to have used a ref for this.

TheSisb: I wouldn't worry about these use-cases too much. This is one of those "I didn't know I could do this but I figured it out because I needed to" situations.

PR #2402, opened 2022-05-10
chore: track github actions metrics

SiTaggart:
Track some event data in PRs about their workflow jobs and step in GH actions when:

- A workflow completes
- PR open or closed
- Commit pushed

It should be noted the data pushed to Datadog, differs per event type as they are tracking different things. https://github.com/int128/datadog-actions-metrics#metrics-for-workflow_run-event

**NOTE:** I'm not seeing workflow data being posted to DD yet, because of this caveat on the workflow_run event in the Github docs:

> Note: This event will only trigger a workflow run if the workflow file is on the default branch.

https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#workflow_run

So, until this is merged, no workflow data. Frustrating

### DIFF
--- a/.github/workflows/track-workflow-metrics.yml
+++ b/.github/workflows/track-workflow-metrics.yml
@@ -1,6 +1,6 @@
 name: Track workflow metrics
 
-# Controls when the workflow will run
+# Run when workflows complete, on pull request and on push
 on:
   workflow_run:
     workflows:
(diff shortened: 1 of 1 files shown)


TheSisb (approved):
Very clever!

nkrantz (approved):

andipants12 (approved):

PR #2413, opened 2022-05-17
fix: incorrect url to the designer role after reposting

SiTaggart:
Quick fix to the job add url, because we reposted it and the url changed.


shleewhite (approved):

zahnster (approved):
👍🏼

PR #2419, opened 2022-05-19
chore: remove extra lodash deps, update lodash

gloriliale:
We no longer need single package dependencies for lodash-  they are now removed, leaving only `lodash`


TheSisb:
- `plopfile.js` imports `lodash.camelCase` and `lodash.startCase` and may need migration to using just `lodash`.
- `.danger/missing-changesets-check.ts` uses `lodash.difference` and may need migration to using just `lodash`.

there may be more files.
pro tip: search the repo (cmd + shift + f) for `"lodash.` and migrate any monorepo scripts (things outside the `packages` folder) to using plain old lodash

TheSisb (changes requested):
- `plopfile.js` imports `lodash.camelCase` and `lodash.startCase` and may need migration to using just `lodash`.
- `.danger/missing-changesets-check.ts` uses `lodash.difference` and may need migration to using just `lodash`.

there may be more files.
pro tip: search the repo (cmd + shift + f) for `"lodash.` and migrate any monorepo scripts (things outside the `packages` folder) to using plain old lodash

packages/paste-icons/tools/utils.js
```diff
@@ -1,10 +1,13 @@
 const {join} = require('path');
 const fs = require('fs');
-const startcase = require('lodash.startcase');
+const startcase = require('lodash/startcase');
```
gloriliale, line 3: one thing I couldn't figure out: we use lodash here, but it's not in the package.json for the paste-icons package. When I tried to add it, I got a syntax error at the top of the package.json file:
image

I removed it to be able to push up and have the other changes reviewed. But any ideas on why that error would've come up for adding lodash as a dependency in the paste-icons package.json?

gloriliale: I'm adding it under a new "dependency" key, since that's the way it is in the other packages: 
image

packages/paste-core/components/combobox/src/helpers.tsx
```diff
@@ -2,7 +2,7 @@ import type {Item as ItemType} from './types';
 
 // This module can only be referenced with ECMAScript imports/exports by turning on the 'esModuleInterop' flag and referencing its default export
```
shleewhite, line 3: Question: is this comment still relevant?

packages/paste-style-props/src/constants.ts
```diff
@@ -1,6 +1,6 @@
 import RawJSON from '@twilio-paste/design-tokens/dist/tokens.raw.json';
 
-const camelCase = require('lodash/camelcase');
+const camelCase = require('lodash/camelCase');
```
shleewhite, line 3: suggestion: pretty sure this can be `import camelCase from 'lodash/camelCase'` (which is the preferred way to import stuff)

packages/paste-customization/package.json
```diff
@@ -23,8 +23,7 @@
     "tsc": "tsc"
   },
   "dependencies": {
-    "deepmerge": "4.2.2",
-    "lodash.merge": "^4.6.2"
+    "deepmerge": "4.2.2"
```
TheSisb, line 26: praise: good catch

packages/paste-types/package.json
```diff
@@ -23,7 +23,7 @@
     "tsc": "tsc"
   },
   "dependencies": {
-    "@types/lodash": "^4.17.19",
+    "@types/lodash": "^4.14.182",
```
shleewhite, line 26: Question: the I noticed the rest of the `package.json`s use `^4.7.21`. Is there a reason why this one is different?

packages/paste-design-tokens/formatters/common.ts
```diff
@@ -1,18 +1,16 @@
-import * as lodash from 'lodash';
+import camelCase from 'lodash/camelCase';
```
TheSisb, line 1: suggestion: let's add lodash as a devDependency in this package please

packages/paste-utils/src/hooks/useMutationObservable.ts
```diff
@@ -1,7 +1,7 @@
 import {useEffect, useState} from 'react';
 
 // This module can only be referenced with ECMAScript imports/exports by turning on the 'esModuleInterop' flag and referencing its default export
```
shleewhite, line 3: Question: is this comment still relevant?

shleewhite (commented):
Lookin great! just had a couple small questions.

packages/paste-types/package.json
```diff
@@ -23,10 +23,7 @@
     "tsc": "tsc"
   },
   "dependencies": {
-    "@types/lodash": "^4.14.136",
-    "@types/lodash.debounce": "^4.0.6",
-    "@types/lodash.groupby": "^4.6.6",
-    "@types/lodash.iselement": "^4.1.6",
+    "@types/lodash": "^4.14.182",
```
TheSisb, line 26: question: can we remove this too?

tools/build/__tests__/__snapshots__/compiledImports.test.ts.snap
```diff
@@ -632,63 +632,358 @@ var require_dist = __commonJS({
   }
 });
 
-// node_modules/lodash.camelcase/index.js
-var require_lodash = __commonJS({
-  \\"node_modules/lodash.camelcase/index.js\\"(exports, module) {
-    var INFINITY = 1 / 0;
+// node_modules/lodash/_freeGlobal.js
```
TheSisb, line 635: thought: looks like the new file is much larger with the new import method. I'm curious what the bundle-size CI check reports.

tools/build/__tests__/compiledImports.test.ts
```diff
@@ -19,7 +19,7 @@ describe('esbuild config', () => {
       module: filePaths.module,
       dependencies: {
         csstype: '^3.0.2',
-        'lodash.camelcase': '^4.3.0',
+        lodash: '^4.3.0',
```
TheSisb, line 22: issue: wrong version

tools/utils/updatePackageReferences.js
```diff
@@ -3,7 +3,7 @@
  * Updates devDependencies with paste peerDependencies
  */
 const chalk = require('chalk');
-const lodash = require('lodash');
+import difference from 'lodash/difference';
```
TheSisb, line 6: question: can we use `import` syntax in this JS file? `node tools/utils/updatePackageReference.js` in terminal works?

packages/paste-style-props/src/constants.ts
gloriliale: I don't exactly know _why_, but there are some spots where it won't build if I convert it to an `import` statement. For this latest set of changes, I converted all of the `require` lines to `import`, then systematically changed them back to `require` if they made the build fail. Any of them that have `require` right now have been checked and are like that (just checked this one again to be extra sure). I haven't figured out why and am very curious about it.

packages/paste-core/components/combobox/src/helpers.tsx
gloriliale: It's not!

tools/build/__tests__/__snapshots__/compiledImports.test.ts.snap
gloriliale: Not too much bigger, as discussed in focus time

tools/utils/updatePackageReferences.js
gloriliale: tested this and it does not work. Changing this back to `require`, along w/ other uses of lodash in `tools`

packages/paste-utils/src/hooks/useMutationObservable.ts
gloriliale: It is not!

shleewhite (approved):
🚢 🚢 🚢  Looks great!

TheSisb (approved):

Issue #2432, opened 2022-05-25
Spotted a documentation error on /components/flex with basis prop


thomaschaaf:
**Describe the bug**

The following usage 
```
    <Flex grow shrink basis>
```

results in invalid css. Because that results in `flex-basis: true;` as css.

**To Reproduce**

1. Go to https://paste.twilio.design/components/flex#flex-properties
2. Inspect "Left area"
3. See invalid css (also a warning from chrome)

**Expected behavior**

It should probably be

```
    <Flex grow shrink basis={1}>
```

per the API documentation at the end of the page `basis` may also not be boolean.

nkrantz:
Hey @thomaschaaf - I just opened a PR (#2433) for this. Thanks for letting us know!

thomaschaaf:
@nkrantz thanks by the way I went through the feedback and then report issue button. If I would have seen I could create a PR for the documentation I would have done that. Maybe an edit page link would be beneficial. See https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/translate at the bottom.

nkrantz:
@thomaschaaf thank you for this feedback! I've shared it with the DSYS team. Feel free to open a PR going forward if you spot any more bugs, or open another Issue to alert the team.

PR #2433, opened 2022-05-25
fix(docs): update to valid flex basis value

nkrantz:
Quick PR to fix a bug on our Flex docs that someone filed on Issues (#2432)

**Contributing to Twilio**

### DIFF
--- a/packages/paste-website/src/pages/components/flex/index.mdx
+++ b/packages/paste-website/src/pages/components/flex/index.mdx
@@ -123,7 +123,7 @@ you use the Flex as a custom element, it is up to you to manage the resulting ac
 
 <LivePreview scope={{Box, Flex}} language="jsx">
   {`<Flex>
-    <Flex grow shrink basis>
+    <Flex grow shrink basis="1px">
       <Box
         backgroundColor="colorBackgroundPrimaryWeak"
         padding="space40"
@@ -132,7 +132,7 @@ you use the Flex as a custom element, it is up to you to manage the resulting ac
         Left area
       </Box>
     </Flex>
-    <Flex grow shrink basis>
+    <Flex grow shrink basis="1px">
       <Box
         backgroundColor="colorBackgroundPrimaryWeaker"
         padding="space40"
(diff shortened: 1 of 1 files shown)


packages/paste-website/src/pages/components/flex/index.mdx
```diff
@@ -123,7 +123,7 @@ you use the Flex as a custom element, it is up to you to manage the resulting ac
 
 <LivePreview scope={{Box, Flex}} language="jsx">
   {`<Flex>
-    <Flex grow shrink basis>
+    <Flex grow shrink basis={0}>
```
andipants12, line 126: question: Did we consider setting this to a non-zero value? Previously, as the issue points out, the prop value for `basis` would have been `true`. Is there a benefit to using a truthy value here?

andipants12: question: Why is the `basis` prop on line 135 unchanged? Could we please check the docs page once more (and prob the stories) to ensure we catch other invalid prop values?

gloriliale (approved):
✅ question: why `{0}` as opposed to `{1}`?

andipants12 (commented):
Left two questions, with potential requested changes. Generally looks good!

Question: also, curious if the prop definitions for `grow`, `shrink`, and `flex` have been reviewed? I see [here](https://github.com/twilio-labs/paste/pull/2433/files#diff-ced09f1009b61b8fc05a577963bcb45e52e18e48cf4d880139cf2d8d41d39991L328) that grow and shrink don't have the same primitive types. I'm wondering if that is intentional.

packages/paste-website/src/pages/components/flex/index.mdx
nkrantz: update from convo: I'll update all the `basis` props on this page to be valid values and I updated the AC of [this ticket](https://issues.corp.twilio.com/browse/DSYS-2903) to include adding more helpful info for the basis, grow, and shrink properties to the docs.

andipants12 (approved):

PR #2453, opened 2022-06-07
fix(token-contrast-checker): fixing the build, also adding it to CI

TheSisb:
- [x] Fixes the build on the token-contrast-checker CRA app by setting that folder to the root for eslint, as it was conflicting with CRA's lint configs and our monorepo eslint config.
- [x] Added a script to build that package from the main package.json
- [x] Added it to codesandbox CI


shleewhite (approved):
Nice!!

PR #2454, opened 2022-06-07
feat(textarea): allow vertical resize

shleewhite:
To allow for vertical resizing, I needed to make a new paste library called `@twilio-paste/react-textarea-autosize`. This library is a version of [react-textarea-autosize](https://www.npmjs.com/package/react-textarea-autosize) (a more popular and actively maintained library than [the one we use now](https://www.npmjs.com/package/react-autosize-textarea)).

### Textarea component
* switch out the TextareaAutosize component to use `@twilio-paste/react-textarea-autosize`
* new `resize` prop on Textarea which allows `'none' | 'vertical'` and defaults to `'none'`
* new storybook of a resizeable textarea

### `@twilio-paste/react-textarea-autosize`
Everything is copy/pasted over from [react-textarea-autosize](https://www.npmjs.com/package/react-textarea-autosize). The only change is in the handleChange function in the index.tsx file (there is also a comment pointing it out).

### Website
* update the props table on the website
* add an example of a resizeable textarea on the website

## Why do we need to switch libraries?
react-autosize-textarea uses another library called autosize (which also hasn't been updated in year) to manage the height. When autosize is initiated in the component, [it manually changes resize to none if it is set to vertical](https://github.com/jackmoore/autosize/blob/8fc6a6d6f934bdfa621eb3bac341d34ad9337882/src/autosize.js#L50). 

To avoid the problems created by these libraries, it is easier to use a very similar component that does allow resizing.

**Contributing to Twilio**


packages/paste-libraries/react-textarea-autosize/src/index.tsx
```diff
@@ -0,0 +1,90 @@
...
+      onHeightChange(height, {rowHeight});
+    }
+  };
+
+  const handleChange = (event: React.ChangeEvent<HTMLTextAreaElement>): void => {
+    // There isn't a user resize event on the textarea, so this is a low-cost way to check if the user manually resized it
+    // By default, when the user resizes a textarea, the new height is set in the style attribute
+    // The way the react-textarea-autosize works, it calculates the height and sets the style attribute with !important
+    // So, we can assume that if the height has important it is the autosize height and if it doesn't have important the user resized it themselves
+    const didUserResize = libRef.current?.style.getPropertyPriority('height') !== 'important';
```
shleewhite, line 73: This is the change I made from the react-textarea-autosize library.

TheSisb (approved):

katieporter:
question: is it possible to put a min-height on the textarea so it _can't_ get shorter than our regular input?

packages/paste-libraries/react-textarea-autosize/src/__tests__/index.test.tsx
```diff
@@ -0,0 +1,19 @@
...
+  // it('renders with initial height passed in style prop', () => {
+  //   const props = {
+  //     style: {
+  //       height: 55,
+  //     },
+  //   };
+  //   const {asFragment} = render(<TextareaAutosize {...props} />);
+  //   expect(asFragment()).toMatchSnapshot();
+  // });
+});
```
SiTaggart, line 19: question: is there a plan to add tests to cover the functionality of this package? There are some hooks and utils etc that could do with some tests

shleewhite:
> question: is it possible to put a min-height on the textarea so it _can't_ get shorter than our regular input?

@katieporter unfortunately we can't- user resize is entirely done by the browser and it doesn't respect min-height.

packages/paste-libraries/react-textarea-autosize/src/__tests__/index.test.tsx
shleewhite: I agree that it could! These are the only tests the library had and I just copied them over.

Is this enough of a baseline for now or should I add them in this PR?

packages/paste-libraries/react-textarea-autosize/src/utils.ts
```diff
@@ -0,0 +1,7 @@
+export const noop = (): void => {};
+
+export const pick = <Obj extends {[key: string]: any}, Key extends keyof Obj>(props: Key[], obj: Obj): Pick<Obj, Key> =>
+  props.reduce((acc, prop) => {
+    acc[prop] = obj[prop];
+    return acc;
+  }, {} as Pick<Obj, Key>);
```
SiTaggart, line 7: question: does lodash pick do the same things? https://lodash.com/docs/4.17.15#pick? Should we use it instead?

packages/paste-libraries/react-textarea-autosize/src/__tests__/index.test.tsx
SiTaggart: I think a good starting point would be at least the functions like `getSizingData`, `getHeight`, `calculateNodeHeight`, `forceHiddenStyles`, `useWindowResizeListener` etc

Just felt a bit weird pointing a bunch of "missing unit test" tickets whilst adding code without unit tests 😬

shleewhite: Lol fair point - I'll add them today!

katieporter:
question(purely for my own learning): can non-cursor users leverage the resize functionality?

katieporter:
question: I might have missed convos around this while I was out, but was it intentional to have the regular Textarea get shorter with this update?

Current:
image

Updated with resizable variant:
image

shleewhite:
> question: I might have missed convos around this while I was out, but was it intentional to have the regular Textarea get shorter with this update?
> 
> Current: image
> 
> Updated with resizable variant: image
> 
> My suggestion would be to maintain the current height if at all possible, so there's obvious differentiation between an Input and a Textarea.

Whoops! My bad, just pushed a fix.

shleewhite:
> question(purely for my own learning): can non-cursor users leverage the resize functionality?

I don't think so.... Its one of those things where because we are using the native HTML/CSS functionality, it is the browser's responsibility to make it accessible. For most use cases, I think the autosizing is really helpful in that it prevents the need for someone to resize the textarea.

shleewhite:
> question: sorry I missed this in my initial review, but I just noticed that the resizeable variant still respects some type of max-height on the Textarea - is that intentional? IMO it might be nice to not have a max-height once the Textarea has been resized, if possible.

Hmmm... I can't be that specific about it. I made it so if the prop `resize="vertical"` is there, it doesn't have a max-height at all. So if they dont resize it and keep typing it will continue to grow. This seems Good Enough™ but LMK if i should just leave the max height all the time.

packages/paste-libraries/react-textarea-autosize/src/hooks.ts
```diff
@@ -0,0 +1,20 @@
+import * as React from 'react';
+import useLatest from 'use-latest';
+
+export {default as useComposedRef} from 'use-composed-ref';
+
+export const useWindowResizeListener = (listener: (event: UIEvent) => any): any => {
```
TheSisb, line 6: maybe rename this file to `useWindowResizeListener` because we have another hook as its own file. Can move them to a hooks folder if you really want

tools/build/__fixtures__/random_package/src/index.tsx
```diff
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import TextareaAutosize from 'react-autosize-textarea';
+import {Tab} from 'reakit';
```
TheSisb, line 2: question: what's this change about?

packages/paste-libraries/react-textarea-autosize/src/__tests__/index.test.tsx
```diff
@@ -0,0 +1,133 @@
...
+    // need to force sizingData type to not be null
+    const calculatedHeight = calculateNodeHeight(element, sizingData as SizingData, '');
+    expect(calculatedHeight).toEqual([102, 96]);
+  });
+});
+
+describe('useHiddenTextarea', () => {
+  it('creates a visually hidden textarea', () => {
+    render(<SampleComponent />);
+    const hiddenTextarea = screen.getByRole('textbox', {hidden: true});
```
TheSisb, line 122: Doh, so simple!

packages/paste-libraries/react-textarea-autosize/src/__tests__/index.test.tsx
```diff
@@ -0,0 +1,133 @@
...
+  });
+});
+
+describe('calculateNodeHeight', () => {
+  it('returns the height and rowHeight of an empty textarea', () => {
+    const element = document.createElement('textarea');
+    // need to set the box sizing property because the function returns null if it is not set
+    element.style.setProperty('box-sizing', 'border-box');
+    // Jest/JSDOM doesn't set scrollHeight, so we need to fake it
+    jest.spyOn(element, 'scrollHeight', 'get').mockImplementation(() => 100);
```
TheSisb, line 111: praise: excellent

packages/paste-core/components/textarea/stories/textarea.stories.tsx
```diff
@@ -59,6 +63,31 @@ TextareaInverse.story = {
   name: 'Textarea - inverse',
 };
 
+export const TextareaResizeVertical = (): React.ReactNode => {
```
TheSisb, line 66: thought(blocking): can you add a story similar to our codesandbox, where we add and remove multiple textareas? This will help us verify resize continues to work with several of them mounted

packages/paste-libraries/react-textarea-autosize/src/__tests__/index.test.tsx
```diff
@@ -0,0 +1,133 @@
...
+    const calculatedHeight = calculateNodeHeight(element, sizingData as SizingData, '');
+    expect(calculatedHeight).toEqual([102, 96]);
+  });
+});
+
+describe('useHiddenTextarea', () => {
+  it('creates a visually hidden textarea', () => {
+    render(<SampleComponent />);
+    const hiddenTextarea = screen.getByRole('textbox', {hidden: true});
+    expect(hiddenTextarea).toBeDefined();
```
TheSisb, line 123: thought(blocker): can you add a test that checks that the hiddenTextarea is removed on unmount please?

```
const { unmount } = render(<Component />);
unmount();
```

andipants12 (commented):
Looks great! Two requests around maintaining naming consistency with our other library packages, and polish on the stories.

I had some other comments, but they got lost when changes were pushed. Will try to add them again. 😄

packages/paste-libraries/react-textarea-autosize/src/index.tsx
andipants12: todo: Could you please leave a comment about what the change was?

.changeset/late-rings-poke.md
```diff
@@ -0,0 +1,6 @@
+---
+'@twilio-paste/textarea': minor
```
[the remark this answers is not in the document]
SiTaggart: It's a good question. If this was a peer dependency, maybe, because the dependency management kind of falls on the consumer and you need a way to tell them to do something.

As it's a dependency and not re-exported, it's bundled into the compiled code and distributed, so you shouldn't have any idea about the change. You can't directly get at it to use it, it's an internal implementation detail.

If we were treating it as a library and re-exporting it, then yup, that'd be a major.

packages/paste-libraries/react-textarea-autosize/src/useWindowResizeListener.tsx
```diff
@@ -0,0 +1,18 @@
+import * as React from 'react';
+import useLatest from 'use-latest';
```
andipants12, line 2: question: Is this being used to check the value of the ref?

suggestion(conditional): if that is the case, this is a pretty simple hook to implement. I suggest implementing it under our utilities rather than adding a new dependency.

packages/paste-website/src/pages/components/textarea/index.mdx
```diff
@@ -196,6 +196,18 @@ The textarea should include the base textarea, along with supporting elements to
 </>`}
 </LivePreview>
 
+### Resizeable Textarea
+
+By default, the textarea is not resizeable. To change this, add the prop `resize='vertical'`.
+
+<LivePreview scope={{Label, HelpText, TextArea}} language="jsx">
+  {`<>
```
andipants12, line 204: todo: missing id on the label and associated aria-labelledby

tools/build/__fixtures__/random_package/src/index.tsx
```diff
@@ -32,5 +32,5 @@ export const getContrastRatingsOfTokensWithTextContrastRequirements = (): void =
 // eslint-disable-next-line @typescript-eslint/explicit-module-boundary-types
 // eslint-disable-next-line @typescript-eslint/explicit-function-return-type
 export const RandomComponent = () => {
-  return <TextareaAutosize />;
+  return <Tab />;
```
andipants12, line 35: todo: I suggest using [jest module mocks](https://jestjs.io/docs/manual-mocks) to handle this instead. Happy to chat in person if you have questions.

andipants12 (commented):
Left some clarifying questions as well as conditional suggestions/todos.

packages/paste-libraries/react-textarea-autosize/package.json
```diff
@@ -0,0 +1,39 @@
+{
+  "name": "@twilio-paste/react-textarea-autosize",
+  "version": "0.0.1",
```
SiTaggart, line 3: suggestion: I think what I've learnt is that starting at v0 should be reserved for alpha status, mainly due to what it does as a knock on effect when bumping package versions. This is essentially GA, so we should start at v1

packages/paste-core/components/textarea/__tests__/textarea.test.tsx
```diff
@@ -73,6 +78,48 @@ describe('TextArea render', () => {
...
+    const pushButton = screen.getByRole('button', {name: 'Push textarea'});
+    const toggleButton = screen.getByRole('button', {name: 'Toggle textarea visibility'});
+
+    expect(screen.getAllByRole('textbox', {hidden: true})).toHaveLength(2);
+
+    fireEvent.click(pushButton);
+    expect(screen.getAllByRole('textbox', {hidden: true})).toHaveLength(4);
+
+    fireEvent.click(toggleButton);
+    expect(screen.queryAllByRole('textbox', {hidden: true})).toHaveLength(0);
```
TheSisb, line 119: quibble: we can merge these all into one `it` test to avoid having to re-render the same `<MultipleTextareas>` several times, since these tests follow a natural order. Test can be re-titled to `handles adding and removing multiple text areas`

TheSisb (approved):

packages/paste-core/components/textarea/stories/textarea.stories.tsx
```diff
@@ -25,12 +25,13 @@ export const Textarea = (): React.ReactNode => {
       <Label htmlFor={uid}>Label</Label>
       <TextArea
         id={uid}
+        aria-describedby={`help-text-${uid}`}
```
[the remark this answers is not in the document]
SiTaggart: `useUID()` is the generator, `uid` is the output of the generation, so I think this is showing best practices.

packages/paste-core/components/textarea/stories/textarea.stories.tsx
```diff
@@ -59,6 +63,31 @@ TextareaInverse.story = {
   name: 'Textarea - inverse',
 };
 
+export const TextareaResizeVertical = (): React.ReactNode => {
+  const uid = useUID();
+  return (
+    <>
+      <Label htmlFor={uid} required>
```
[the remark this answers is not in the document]
SiTaggart: There is no requirement for `aria-labelledby` here

.changeset/slow-buttons-film.md
```diff
@@ -0,0 +1,6 @@
+---
+'@twilio-paste/core': minor
+'@twilio-paste/react-textarea-autosize': minor
```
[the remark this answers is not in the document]
shleewhite: Will update!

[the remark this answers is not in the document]
shleewhite: Also, because this is starting at version 1.0.0, should the changeset be major?

cc: @andipants12 @SiTaggart

.changeset/late-rings-poke.md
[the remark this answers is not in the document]
shleewhite: wait. so what should it be?

packages/paste-core/components/textarea/stories/textarea.stories.tsx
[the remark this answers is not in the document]
shleewhite: Its not the generator, its a variable - `const uid = useUID();` is on line 23.

The ids match correctly in storybook.

.changeset/late-rings-poke.md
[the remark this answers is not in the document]
andipants12: @shleewhite Is `@twilio-paste/textarea-resize-library` a peer dependency of `@twilio-paste/textarea`? 
If yes I think the answer to the question is yes, it should be major.

@SiTaggart lmk if that isn't correct ☝️

packages/paste-libraries/react-textarea-autosize/src/useWindowResizeListener.tsx
SiTaggart: Counterpoint, I'd rather the team utilize well tested third parties, than create more code to maintain.

packages/paste-libraries/react-textarea-autosize/package.json
shleewhite: Ok! Will change it to 1.0.0

packages/paste-core/components/textarea/stories/textarea.stories.tsx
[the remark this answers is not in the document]
andipants12: Clarification: I am pointing out that the id for the text area and the label are the same for all of the stories. If there are multiple in the same page, this would be a problem.

It looked to me like there are multiple per page, if not disregard.

.changeset/late-rings-poke.md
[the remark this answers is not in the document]
SiTaggart: Correct, you are adding a peer dependency to the package, which won't be installed for the consumer automatically, and as such you should mark it as a major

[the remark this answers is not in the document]
SiTaggart: (I realise my wording was all jumbled up in my original explanation. I tried not to give the answer and made it more confusing)

packages/paste-libraries/react-textarea-autosize/src/calculateNodeHeight.ts
```diff
@@ -0,0 +1,50 @@
...
+): CalculatedNodeHeights {
+  const {paddingSize, borderSize, sizingStyle} = sizingData;
+  const {boxSizing} = sizingStyle;
+
+  if (hiddenTextarea == null) return [0, 0];
+
+  Object.keys(sizingStyle).forEach((_key) => {
+    const key = _key as keyof typeof sizingStyle;
+    // eslint-disable-next-line @typescript-eslint/no-non-null-assertion
+    hiddenTextarea!.style[key] = sizingStyle[key] as any;
```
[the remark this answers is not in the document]
shleewhite: Ok! I can add a check that style is defined. 

I'm not sure I need to do the second part because don't objects work where when you have an object and you set the value of a key, it creates the key if it doesn't already exist?

packages/paste-website/src/pages/components/textarea/index.mdx
shleewhite: This is not needed because the label has htmlFor

packages/paste-core/components/textarea/stories/textarea.stories.tsx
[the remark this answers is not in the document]
SiTaggart: The way `useUID` works as a hook in the component, is that if you do have multiple instances of the component on the same page it will generate a different ID per instance.

packages/paste-libraries/react-textarea-autosize/src/index.tsx
shleewhite: I thought the comment above explains the change?

.changeset/slow-buttons-film.md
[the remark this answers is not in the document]
shleewhite: OK - i will make the package json say "0.0.0" and the changeset major

packages/paste-libraries/react-textarea-autosize/src/useWindowResizeListener.tsx
shleewhite: I'm going to skip doing this for now, just because its Good Enough™ and I copied it over from the original library.

packages/paste-website/src/pages/components/textarea/index.mdx
shleewhite: This is not needed because the label has htmlFor

tools/build/__fixtures__/random_package/src/index.tsx
shleewhite: I'm going to leave that as out of scope, I just swapped the test for another 3rd party component because we don't have that dependency anymore.

packages/paste-libraries/react-textarea-autosize/src/index.tsx
SiTaggart: counterpoint is, because you created our own version, you only really need to explain what _this_ does going forward. No need to explain what is does differently to a package we no longer use

packages/paste-libraries/react-textarea-autosize/src/calculateNodeHeight.ts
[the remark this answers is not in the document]
shleewhite: I'm going to leave this as is because it is copied from the original library.

packages/paste-core/components/textarea/__tests__/textarea.test.tsx
shleewhite: Updated.

packages/paste-libraries/react-textarea-autosize/src/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+  if (process.env.NODE_ENV !== 'production' && props.style) {
+    if ('maxHeight' in props.style) {
+      throw new Error('Using `style.maxHeight` for <TextareaAutosize/> is not supported. Please use `maxRows`.');
+    }
+    if ('minHeight' in props.style) {
+      throw new Error('Using `style.minHeight` for <TextareaAutosize/> is not supported. Please use `minRows`.');
+    }
+  }
+  const isControlled = props.value !== undefined;
+  const libRef = React.useRef<HTMLTextAreaElement | null>(null);
```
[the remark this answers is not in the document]
shleewhite: Updated.

packages/paste-libraries/react-textarea-autosize/src/index.tsx
```diff
@@ -0,0 +1,94 @@
...
+) => {
+  if (process.env.NODE_ENV !== 'production' && props.style) {
+    if ('maxHeight' in props.style) {
+      throw new Error('Using `style.maxHeight` for <TextareaAutosize/> is not supported. Please use `maxRows`.');
+    }
+    if ('minHeight' in props.style) {
+      throw new Error('Using `style.minHeight` for <TextareaAutosize/> is not supported. Please use `minRows`.');
+    }
+  }
+  const isControlled = props.value !== undefined;
```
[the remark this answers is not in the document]
shleewhite: Updated.

tools/build/__fixtures__/random_package/src/index.tsx
shleewhite: Leaving as out of scope because I just swapped the component for another 3rd party component.

packages/paste-libraries/react-textarea-autosize/src/index.tsx
shleewhite: Ok - to clarify, the change is adding `didUserResize`. Because it is a hack, I added the comment

```
// There isn't a user resize event on the textarea, so this is a low-cost way to check if the user manually resized it
// By default, when the user resizes a textarea, the new height is set in the style attribute
// The way the react-textarea-autosize works, it calculates the height and sets the style attribute with !important
// So, we can assume that if the height has important it is the autosize height and if it doesn't have important the user resized it themselves
```

to explain why the hack works to distinguish between autosizing and user resizing.

katieporter (approved):

PR #2452, opened 2022-06-06
chore(avatar): allow both src and icon prop to be passed

shleewhite:
Made for: https://github.com/twilio-labs/paste/discussions/2437

We want to allow users to pass both src and icon so that they can display one or the other depending on if the image loads correctly.

This PR includes:
* updated AvatarProps
* updated Avatar prop types
* note on the docs site that if both src and icon are passed, the avatar will display the icon

**Contributing to Twilio**

> All third-party contributors acknowledge that any contributions they provide will be made under the same open-source license that the open-source project is provided under.

- [x] I acknowledge that all my contributions will be made under the project's license.

PR #2456, opened 2022-06-07
fix(avatar): prefer image over icon

shleewhite:
Fast fix for: https://github.com/twilio-labs/paste/pull/2452

Image should be preferred to icon.

**Contributing to Twilio**

### DIFF
--- a/.changeset/wicked-carrots-greet.md
+++ b/.changeset/wicked-carrots-greet.md
@@ -1,0 +1,6 @@
+---
+'@twilio-paste/avatar': patch
+'@twilio-paste/core': patch
+---
+
+[Avatar] prefer image over icon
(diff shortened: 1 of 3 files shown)


serifluous (approved):

.changeset/wicked-carrots-greet.md
```diff
@@ -0,0 +1,6 @@
+---
+'@twilio-paste/avatar': patch
```
TheSisb, line 2: Could this be a breaking change? If so, is it worth it? Can check reporting too

TheSisb (approved):

TheSisb:
Can we add a unit test to verify now and in the future that if both a src and icon are passed, the image is rendered please?

TheSisb (changes requested):
Requesting changes to add that test before we merge, to prevent more issues in the future

.changeset/wicked-carrots-greet.md
shleewhite: Its not a breaking change, I just misread the code initially. 

The logic was happening in 2 places before: 

In the Avatar component, it checked if the src prop was there and rendered an image. If it wasn't there, it rendered an icon or initials

Then, in the AvatarContents component it did the opposite,  if the icon prop was present, it rendered that. Then if the src was there it would render an image, then the initials.

I caused the confusion because i didn't see that the src attribute was being checked in the Avatar component first. So, it was always image, icon, then initials.

To also clear this up in the future, I simplified the logic. You can see now that all the logic is in the AvatarContents component.

TheSisb (approved):
thanks

PR #2457, opened 2022-06-08
chore: small improvements to turborepo locally

TheSisb:
I clean up the commands in this PR to make turborepo useful for local dev.  I tried making it work on CI, but it was failing and I didn't want to spend too much time debugging.  For now this should help us in local-dev and as we make more improvements to the monorepo we can reconsider adding it to CI.


packages/paste-design-tokens/package.json
```diff
@@ -18,9 +18,18 @@
     "build:watch": "gulp dev",
     "clean": "rm -rf ./dist",
     "tokens": "gulp default",
-    "test": "jest"
+    "test": "jest",
+    "tsc": "yarn build"
```
TheSisb, line 22: Since this uses `gulp` to generate both the dist files and the types, I re-use the build command for the `tsc` command rather than trying to split up internals.

packages/paste-design-tokens/package.json
```diff
@@ -18,9 +18,18 @@
...
   },
   "devDependencies": {
+    "@types/gulp": "^4.0.6",
```
TheSisb, line 25: I pulled these deps out from the monorepo `package.json` file into this package specific one because they're only used as devDeps in this one package.

package.json
```diff
@@ -77,18 +77,13 @@
...
-    "type-check": "lerna run tsc",
     "tsc": "echo 'Did you mean to run yarn type-check?'",
     "chromatic": "chromatic",
     "packages:check": "manypkg check",
     "packages:fix": "manypkg fix",
     "changeset": "changeset",
     "create:package": "plop create-package",
-    "token-usage": "ts-node tools/build/token-usage-detector.ts",
-    "turbo-build": "turbo run build",
-    "turbo-test": "turbo run test",
```
TheSisb, line 89: I wanted to add turborepo to other commands such as `test` `lint` `prettier` and `type-check`, but couldn't get it to work after a couple hours, so I decided to work with the main pain point for now (builds) and approach this again later.

.github/workflows/on_merge_to_main.yml
```diff
@@ -11,6 +11,9 @@ jobs:
   prerelease:
     name: Prerelease test and build
     runs-on: ubuntu-latest
+    env:
+      TURBO_TOKEN: ${{ secrets.TURBO_TOKEN }}
```
TheSisb, line 15: These were added to Github per [the docs](https://turborepo.org/docs/ci/github-actions)

SiTaggart (approved):

shleewhite (approved):

PR #2409, opened 2022-05-13
Chore/nextjs template update

zahnster:
* Updates nextjs-template dependencies, including major version upgrades for:
  * next: v10 -> v12
  * prop-types: v14 -> v17
* Adds resolution for types/react to v17 in order to prevent multiple major versions from being installed
* Adds lint as a dependency as well as eslint-config-next
* Updates eslint config to use next's lint config
* Updates Document from class to function
* Adds NextPage type to index page

PR #2466, opened 2022-06-13
fix(nextjs-template): upgrade nextjs to v12

TheSisb:
This PR does a few important things in my journey to upgrade NextJS for our nextjs-template package:

- [x] I upgrade nextjs to v12 in both `nextjs-template` and `theme-designer`, because you can't update one without the other within our monorepo structure sadly.
- [x] I localize the eslint config to the `nextjs-template` and `theme-designer` packages, because nextjs 12's eslint config conflicts with our own. I do this with `"root": true,  "parser": "@typescript-eslint/parser",` in those folders' respective eslintrc files. This fixes build issues around failed linting.
- [x] I upgrade Typescript and ESbuild to see if the build issue is in our builds. It isn't but these changes catch an error with the Table component's typing and improve our bundle size, so I keep them.
- [x] I copy over some of the changes @zahnster made from [her PR](https://github.com/twilio-labs/paste/pull/2409) because they're helpful and use the latest nextjs imports.

At this point I'm still getting build failures with theme-designer. I can't decypher what the error is from the compiled code so I decide to rerun `yarn build` on all Paste components but after removing `minify: true` in our esbuild.js config file. This makes all our dist files uncompressed and improves my ability to read the errors.  Turns out, theme-designer is failing on one TextArea. I'm not sure why, but it's not very needed so I remove it and the build passes.

I remove my changes to the build config and re-build Paste with minification on. Now I build theme-designer to verify and it fails. For some reason, theme-designer was building without minification and fails to build with minification. I update our esbuild config to minify everything but identifiers and it works again. This is probably actually a good change because it will help us/downstream debug more easily.  Looks like the builds grew a bit :/ but at least our code is safer


packages/paste-nextjs-template/pages/index.tsx
```diff
@@ -15,51 +13,13 @@ const Home: React.FC = () => {
...
-        </ListItem>
-        <ListItem>
-          <Heading as="h3" variant="heading30">
-            <Anchor href="https://vercel.com/import?filter=next.js" showExternal>
-              Deploy
-            </Anchor>
-          </Heading>
-          <Paragraph>Instantly deploy your Next.js site to a public URL with Vercel.</Paragraph>
-        </ListItem>
-      </UnorderedList>
```
SiTaggart, line 62: With the fixed esbuild config, is there an opportunity to put this stuff back and it still build and work?

I think there's an opportunity, just like with the token contrast checker, to add a verification step in our build for these templates so we also know they are working.

packages/paste-theme-designer/components/preview/FormPreview.tsx
```diff
@@ -36,11 +35,6 @@ const FormPreview: React.FC = () => {
           />
           <HelpText id={inputId('input-help-text')}>This is help text.</HelpText>
         </Box>
-        <Box>
-          <Label htmlFor={inputId('textarea')}>Textarea input</Label>
-          <TextArea aria-describedby={inputId('textarea-help-text')} id={inputId('textarea')} onChange={() => {}} />
-          <HelpText id={inputId('textarea-help-text')}>This is help text.</HelpText>
-        </Box>
```
SiTaggart, line 43: Same here. Was the identifier a fix for this too? Or is TextArea just a bit weird?

We've struggled in the past with textarea compilation FWIW so it might be an actual problem with textarea we need to investigate

SiTaggart (commented):
There's definitely a bump in package size, but in total of 25kb across then entire system isn't so drastic. Definitely worth keeping an eye on.

I'm still curious why it will build just fine in a next project outside of the monorepo, and not inside. It doesn't seem to be customer impacting which is good, but probably worth a ticket to investigate after

packages/paste-theme-designer/components/preview/FormPreview.tsx
TheSisb: Yeah it may be related to the library the textarea depends on.  Adding this back broke the builds again.

packages/paste-nextjs-template/pages/index.tsx
TheSisb: Adding this back in worked, good call.

packages/paste-theme-designer/components/preview/FormPreview.tsx
SiTaggart: I wonder if that will go away with this https://github.com/twilio-labs/paste/pull/2454

I've seen it struggle with that library in the past with builds / compilation

TheSisb: I've tried removing it from Textarea but the error persists. I'm completely at a loss.

TheSisb: https://issues.corp.twilio.com/browse/DSYS-3418

shleewhite (approved):

zahnster (approved):
👍🏼

TheSisb:
There still seems to be a typescript error in codesandbox, but we chatted offline that this was probably a codesandbox issue, since the template works when installed onto our local machines without any issues.

PR #2475, opened 2022-06-17
Feat: Custom Breakpoint toggle for Storybook

zahnster:
When building website components in Storybook, it's difficult to accurately preview how they will behave responsively on the Docs site, as the breakpoints on the Docs site are different from the default breakpoints in Paste's ThemeProvider.

This PR introduces a new control to Storybook which will allow us to set Storybook's ThemeProvider to use the custom responsive breakpoints from the Docs site, and will allow us to accurately preview how a component will respond with those custom breakpoint values.


SiTaggart:
We have this thing too, right? 

image

zahnster:
@SiTaggart I updated the PR description to better reflect the code change. This is to provide a way to load the Docs site's custom breakpoints to the ThemeProvider in Storybook, so we can accurately preview how components will respond when in the Docs site.

SiTaggart:
> @SiTaggart I updated the PR description to better reflect the code change. This is to provide a way to load the Docs site's custom breakpoints to the ThemeProvider in Storybook, so we can accurately preview how components will respond when in the Docs site. 

Do we need both? Is the other one useful? Should we remove it?

zahnster:
@SiTaggart they're meant to be used together. This new toggle only sets the docs site's custom breakpoint values on the ThemeProvider. The existing 'viewport resize' control is what actually changes the viewport.

SiTaggart:
> @SiTaggart they're meant to be used together. This new toggle only sets the docs site's custom breakpoint values on Storybook's ThemeProvider. The existing 'viewport resize' control is what actually changes the viewport. Both are useful and needed so we can accurately develop responsive docs site components w/in Storybook.

Oh yeeeeeeeeeeah, that makes sense... whoops  🤦

SiTaggart (approved):

shleewhite (approved):
🚢

gloriliale (approved):

PR #2476, opened 2022-06-17
chore(truncate): fix alignment

shleewhite:
Currently, there is a bug where if you put a Truncate component where it pushes the container its in down and it makes it look like there's extra padding. 

Before:

After:

## Changes in this PR
* sets `verticalAlign: 'bottom'` to fix how the Truncate contents are aligned
* sets `textDecoration: 'inherit'` because I noticed that the Anchor text isn't underlined when it is truncated
* adds a `TruncateInAnchor` story

## Why does this fix it?
When you set `display: 'inline-block'`, `verticalAlign` defaults to `'baseline'` [which behaves differently between HTML elements and browsers](https://developer.mozilla.org/en-US/docs/Web/CSS/vertical-align#values_for_inline_elements) and can end up looking weird.

Instead, I set it to `'bottom'`, which aligns the content to the bottom of the entire line. To confirm this was correct, I checked that the focus state on Anchor looked the same with and without the text in a Truncate component.

**Contributing to Twilio**

### DIFF
--- a/packages/paste-core/components/truncate/src/index.tsx
+++ b/packages/paste-core/components/truncate/src/index.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
 import * as PropTypes from 'prop-types';
-import {Box, BoxProps} from '@twilio-paste/box';
-
+import {Box} from '@twilio-paste/box';
+import type {BoxProps} from '@twilio-paste/box';
 export interface TruncateProps extends Omit<React.HtmlHTMLAttributes<HTMLSpanElement>, 'color'>, Pick<BoxProps, 'as'> {
   children: NonNullable<React.ReactNode>;
   title: string;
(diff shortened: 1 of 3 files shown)


serifluous (approved):

packages/paste-core/components/truncate/src/index.tsx
```diff
@@ -13,6 +13,8 @@ const Truncate = React.forwardRef<HTMLSpanElement, TruncateProps>(({children, ..
       {...props}
       as="span"
       display="inline-block"
+      verticalAlign="bottom"
+      textDecoration="inherit"
```
SiTaggart, line 17: I still don't know if this is technically the  "fix" because the problem really is, that inline-block is adding some vertical space to the parent.

So if I place this in a heading for example, and change the color of the background so you can see the physical space the heading takes up, inline-block makes the whole thing bigger.

display block or flex corrects this, but that seems like it could be a risk change.

image

shleewhite: Yeah... I thought about adding a `display` prop but I didn't want to change the default from `'inline-block'`. Even if we go that route, I still recommend keeping the vertical align because it does help the alignment when its inline block.

SiTaggart (approved):

PR #2480, opened 2022-06-21
chore: add tests for isDeprecatedTheme function

shleewhite:
Currently, the isDeprecatedTheme function is untested. In an effort to be resilient, we need to add tests to ensure it functions correctly.

## Changes in this PR
* creates `isDeprecatedTheme.spec.ts` file which tests:
  * it makes a console error with a specific error message when available
  * it makes a console error with a generic error message when a specific one isn't available
* fixes the error message formatting in `isDeprecatedTheme`

## Why does this fix it?
**Testing**
These test address both possible use cases for this function.

**Formatting** 
Before, when there wasn't a specific error message it would output this to the console: `'Paste: {themeName} is a deprecated theme. undefined'`

Now it works as expected and only outputs `'Paste: {themeName} is a deprecated theme.'`

**Contributing to Twilio**


nkrantz (approved):
nice!

PR #2494, opened 2022-06-24
chore: follow upgrade guide for gatsby netlify plugin

SiTaggart:
Essentially, follow this

image


shleewhite (approved):

gloriliale (approved):
I feel like this is a safe assumption but, the 'install it from the Netlify Plugins'. part is something that doesn't need to be in our repo, right? it's on netlify's side of things?

TheSisb (approved):

PR #2501, opened 2022-06-28
feature(chat-log): add Chat Attachment and related components

gloriliale:
**Contributing to Twilio**


shleewhite (approved):

packages/paste-core/components/chat-log/src/ChatAttachmentDescription.tsx
```diff
@@ -0,0 +1,37 @@
...
+  ({children, element = 'CHAT_ATTACHMENT_DESCRIPTION', ...props}, ref) => {
+    return (
+      <Text
+        element={element}
+        ref={ref}
+        as="div"
+        fontSize="fontSize20"
+        color="colorTextWeak"
+        lineHeight="lineHeight10"
+        {...props}
```
TheSisb, line 22: We need to safelySpreadTextProps here. Also, do note that you're spreading after applying your styles. So folks can override fontSize / color / etc.  Is this intended?

gloriliale: Ah, it is not! just updated it to use safelySpreadTextProps and changed the order. I also changed the order for props spreading in `ChatAttachment` and now use safeSpreadBoxProps in `ComposerAttachmentCard`

TheSisb (approved):

loreina (approved):
🎉 🎉 🎉

PR #2506, opened 2022-06-29
test: jest axe to storybook test runner

SiTaggart:
# Remove Jest aXe

When we run axe in jest we do so on a very limited set of examples and states, plus it can be sometimes without a theme. Ideally each component variant would have it's own aXe test, and we would test both dark and light modes for color contrast.

With Storybook test runner and test hooks, we can automatically test each story we write for VRT against aXe and fail CI checks if we are not meeting accessibility standards in our stories.

## In this PR

- We add storybook test runner and set up automatically aXe tests using Playwright.

- We remove jext-axe and all the jest-axe tests.

- fix modals, alert dialogs to not introduce multiple landmark regions such as banner and contentinfo, which will happen if you are using `header` and `footer` elements outside of sectioning elements.
- Fix breadcrumb to allow overriding the aria-label.

- Update internal docs detailing slight change in direction for testing

## How it works (briefly)

It runs `yarn test:storybook` and `yarn start:storybook` in parrallel, using `yarn start:test:storybook`. It forces each story into the stacked view for testing all themes in one view.

## 🔴 Things to note when creating Customization Stories

1. You *must* set the current theme from the theme context
2. You *must* disable animations for test environments

Otherwise you will get test failures from aXe based on dark mode and timings on when the tests run.

## After this PR

Do not write another jest axe test. All new stories will automatically get tested by aXe


.changeset/spotty-terms-count.md
```diff
@@ -0,0 +1,6 @@
+---
+'@twilio-paste/breadcrumb': patch
+'@twilio-paste/core': patch
+---
+
+[Breadcrumb] allow aria-label to be overwritten on the breadcrumb navigation element. This is useful to remove duplicately labelled pagination elements and internationalisation
```
TheSisb, line 6:
```suggestion
[Breadcrumb] allow aria-label to be overwritten on the breadcrumb navigation element. This is useful to remove duplicate labels and for internationalization
```

.storybook/main.js
```diff
@@ -9,9 +9,14 @@ module.exports = {
     '@storybook/addon-a11y',
     '@storybook/addon-actions',
     '@storybook/addon-viewport',
+    '@storybook/addon-interactions',
```
TheSisb, line 12: neat, https://storybook.js.org/addons/@storybook/addon-interactions

.storybook/preview.js
```diff
@@ -9,7 +9,8 @@ import {Grid, Column} from '@twilio-paste/grid';
 import {RenderPerformanceProfiler} from './RenderPerformanceProfiler';
 import {SITE_BREAKPOINTS} from '../packages/paste-website/src/constants';
 
-const disableAnimations = isChromatic();
+// disable animations in test environments
+const isTestEnvironment = isChromatic() || process.env.NODE_ENV === 'test';
```
TheSisb, line 13: praise: good change

.storybook/test-runner.js
```diff
@@ -0,0 +1,43 @@
...
+          enabled: false,
+        },
+        {
+          id: 'region',
+          enabled: false,
+        },
+        {
+          id: 'scrollable-region-focusable',
+          enabled: false,
+        },
```
TheSisb, line 34: suggestion: can we get comments above each of these for why we decided to disable them please?

packages/paste-core/components/alert-dialog/stories/index.stories.tsx
```diff
@@ -228,6 +228,7 @@ export const CustomizedAlertDialog = (): React.ReactElement => {
         ALERT_DIALOG_BODY: {backgroundColor: 'colorBackgroundSuccessWeakest'},
         ALERT_DIALOG_FOOTER: {paddingTop: 'space50', backgroundColor: 'colorBackgroundWarningWeakest'},
       }}
+      disableAnimations={true}
```
TheSisb, line 231: question: does this still need to be passed with the new `isTestServer` check?

packages/paste-core/components/alert-dialog/stories/index.stories.tsx
```diff
@@ -256,6 +257,7 @@ export const CustomizedDestructiveAlertDialog = (): React.ReactElement => {
         FOO_BODY: {backgroundColor: 'colorBackgroundSuccessWeakest'},
         FOO_FOOTER: {paddingTop: 'space50', backgroundColor: 'colorBackgroundWarningWeakest'},
       }}
+      disableAnimations={true}
```
TheSisb, line 260: Same question for this

packages/paste-core/components/anchor/stories/index.stories.tsx
```diff
@@ -72,16 +73,13 @@ ShowExternalProp.story = {
 };
 
 export const CustomAnchor = (): React.ReactNode => {
+  const currentTheme = useTheme();
```
TheSisb, line 76: question: why this particular change? Wouldn't passing `baseTheme="default"` with no `theme` prop have the same effect?

packages/paste-core/components/chat-log/stories/customization.stories.tsx
```diff
@@ -14,140 +14,147 @@ import {
...
-    <Box as="ul" width="100%">
-      <ChatMessage variant="inbound">
-        <ChatBubble>test</ChatBubble>
-        <ChatMessageMeta aria-label="said by Gibby Radki 4 minutes ago">
-          <ChatMessageMetaItem>
+export const CustomizedMessages: StoryFn = () => {
+  const currentTheme = useTheme();
+  return (
+    <CustomizationProvider
+      theme={currentTheme}
```
TheSisb, line 28: Same question here about replacing this with baseTheme

packages/paste-core/components/chat-log/stories/index.stories.tsx
```diff
@@ -40,17 +41,19 @@ export const NewMessagesButton: StoryFn = () => (
 
 export const ScrollingChatLog: StoryFn = () => {
   const [showButton, setShowButton] = React.useState(true);
+  const chatBoxUniqueID = useUID();
   return (
-    <Box maxHeight="size40" overflowY="scroll" id="chat-box">
+    <Box maxHeight="size40" overflowY="scroll" id={chatBoxUniqueID}>
```
TheSisb, line 46: praise: good catch

.storybook/main.js
SiTaggart: Ya, this could be super useful long term

packages/paste-core/components/alert-dialog/stories/index.stories.tsx
SiTaggart: Yeah, unfortunately, because CustomizationProvider is basically a new Theme Provider, it resets all the animation stuff. It's a new theme context.

I think what I'm about to regret suggesting I do, is create a Stories Customization Provider which only tables theme and elements. Turns off animations in a single place and also sets getting the current theme in a single location too. Given that is what most of the changes are.

packages/paste-core/components/anchor/stories/index.stories.tsx
SiTaggart: This is to do with the "Stacked" or "Side by Side" view of our stories when we render them in test environments. One side it light, one side is dark, and all we're doing is rendering the same story in two views. 

So if you set a static baseTheme of default, when it's rendered in Dark theme in Stacked or Side by Side, the component is using light mode tokens, in a dark background. You have to basically dynamically inherit the theme you're rendering the story in

packages/paste-core/components/alert-dialog/stories/index.stories.tsx
SiTaggart: Backing out of this idea, it's unfortunate copy pasta from here

zahnster (approved):
Go for iiit.

Screen Shot 2022-07-19 at 4 13 26 PM

PR #2513, opened 2022-07-01
fix(site-tsconfig): update docs site tsconfig settings

zahnster:
This resolves an earlier encountered issue with test files throwing a TS error when including React but not directly using it.

### DIFF
--- a/packages/paste-website/tsconfig.json
+++ b/packages/paste-website/tsconfig.json
@@ -6,8 +6,7 @@
       "dom",
       "esnext"
     ] /* Specify a set of bundled library declaration files that describe the target runtime environment. */,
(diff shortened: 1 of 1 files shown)


gloriliale (approved):

nkrantz (approved):

SiTaggart:
Looked at this some more and I don't think we're going to be able to do it, as it provides the emotion css prop types. It's annoying as it does too much, but I don't see a way round it right now.

This might be a legitimate case for having a `tsconfig.test.json` and a `tsconfig.build.json` for the website to provide different types to jest and the website. That obviously fails if you need to test the css prop anyway, so that might not even be the solution.

Problem for another day I suspect

PR #2514, opened 2022-07-05
feat(tokens-page): add theme and format controls

nkrantz:
## Theme and format controls for filtering tokens

[Jira Ticket](https://issues.corp.twilio.com/browse/DSYS-3169)

### In this PR
This PR adds 2 select fields to the filtering section of the new Token List page. 

One is for controlling the theme of the displayed tokens. It should operate independently from the website theme switcher.

The other is for controlling the format of the token name. If the user selects "javascript", the token names are displayed in camelCase.

The `TokensList` component is refactored because I pulled out `TokensListFilter` into its own component. I did this so that we could do VRT on the `TokensListFilter` section of the page.

When a new theme or format is selected, it's added to local storage so that the selection persists next time someone revisits the page.

Cypress tests for the new controls.

### To do
- There's a ts error that is currently being ignored on line 46 of `tokens-list/index.tsx`, it's stumping me

### Notes
- The performance is very slow, but it's out of the scope of this ticket to fix performance issues. That will be addressed when we revisit the filtering of the page.
- The token values I'm using to test that the theme switcher works are hardcoded values. I tried pulling in the dist file to check against that, but it was undefined in cypress. Let me know if you have any thoughts on how to better test that the theme switcher is working.

--------------------------------------------------------------------------------

**Contributing to Twilio**

### DIFF
--- a/packages/paste-website/src/components/tokens-list/index.tsx
+++ b/packages/paste-website/src/components/tokens-list/index.tsx
@@ -78,7 +78,9 @@ export const TokensList: React.FC<TokensListProps> = (props) => {
                   useCamelCase={useJavascriptNames ? true : false}
                   value={value}
                   comment={comment}
-                  backgroundColor={selectedTheme === 'dark' ? 'rgba(15, 22, 33, 1)' : 'rgb(255, 255, 255)'}
+                  backgroundColor={
+                    tokens['background-colors'].find((token) => token.name === 'color-background-body')?.value
+                  }
                 />
               ))}
             </Box>
--- a/packages/paste-website/src/components/tokens-list/index.tsx
+++ b/packages/paste-website/src/components/tokens-list/index.tsx
@@ -1,7 +1,5 @@
 import * as React from 'react';
 import {Box} from '@twilio-paste/box';
-import {Label} from '@twilio-paste/label';
-import {Input} from '@twilio-paste/input';
 import Tokens from '@twilio-paste/design-tokens/dist/tokens.generic';
 import DarkModeTokens from '@twilio-paste/design-tokens/dist/themes/dark/tokens.generic';
 import {AnchoredHeading} from '../Heading';
@@ -11,9 +9,7 @@ import type {Token, TokenCategory, TokensListProps} from './types';
 import {PageAside} from '../shortcodes/PageAside';
 import {NoTokensFound} from './NoTokensFound';
 import {TokenCard} from './token-card';
-import {FilterIcon} from '@twilio-paste/icons/esm/FilterIcon';
-import {Grid, Column} from '@twilio-paste/grid';
-import {Select, Option} from '@twilio-paste/select';
+import {TokensListFilter} from './TokensListFilter';
 
 const sentenceCase = (catName: string): string => {
   return catName
@@ -41,11 +37,6 @@ export const TokensList: React.FC<TokensListProps> = (props) => {
     trackTokenFilterString(filterString);
   }, [filterString, props, theme]);
 
-  const handleInput = (e: React.FormEvent<HTMLInputElement>): void => {
-    const filter = e.currentTarget.value;
-    setFilterString(filter);
-  };
-
   if (tokens === null) {
     return <NoTokensFound onClearSearch={() => setFilterString('')} />;
   }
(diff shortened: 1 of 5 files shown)


nkrantz:
Screen recording of the slow filtered tokens rendering on my machine

https://user-images.githubusercontent.com/75342690/177397861-e54fb48d-a40b-4b28-ad06-2bdf341f8e84.mov

SiTaggart:
> Instead of unit tests (there's no function to test), I could add a cypress test with some filtering actions.

Perfect suggestion!

> Performance is very slow for rendering the filtered tokens

Agreed, but I'm not sure whether that's the scope of your ticket right now though. I'd be comfortable to merge with slow performance into the feature branch and address perf in the filtering ticket.

packages/paste-website/src/components/tokens-list/index.tsx
```diff
@@ -56,19 +61,60 @@ export const TokensList: React.FC<TokensListProps> = (props) => {
...
+              <Label htmlFor="theme-control" id="theme-control-label">
+                Theme
+              </Label>
+              <Select
+                id="theme-control"
+                defaultValue={theme === 'dark' ? 'dark' : 'default'}
+                onChange={(evt) => {
+                  if (evt.target.value === 'dark') {
+                    setSelectedTheme('dark');
+                    setTokens(DarkModeTokens.tokens);
```
SiTaggart, line 89: question: Should we be thinking about these as separate, independent pieces of state, or should we start thinking about them being tightly coupled?

packages/paste-website/src/components/tokens-list/index.tsx
```diff
@@ -56,19 +61,60 @@ export const TokensList: React.FC<TokensListProps> = (props) => {
...
-          />
+        <Box marginBottom="space80">
+          <Grid gutter="space40">
+            <Column span={6}>
+              <Label htmlFor="test" id="test-label">
+                Filter tokens
+              </Label>
+              <Input
+                type="text"
+                id="test"
```
SiTaggart, line 72: chore: when you move this to a component, don't forget to use the `useUID` hook to generate these ids for us.

packages/paste-website/src/components/tokens-list/index.tsx
```diff
@@ -80,10 +126,10 @@ export const TokensList: React.FC<TokensListProps> = (props) => {
                 <TokenCard
                   key={`token${name}`}
                   category={tokenCategory}
-                  name={name}
+                  name={useJavascriptNames ? camelCase(name) : name}
                   value={value}
                   comment={comment}
-                  backgroundColor="rgb(255, 255, 255)"
+                  backgroundColor={selectedTheme === 'dark' ? 'rgba(15, 22, 33, 1)' : 'rgb(255, 255, 255)'}
```
SiTaggart, line 132: question: does `backgroundColor` take a token name as a value? Rather than passing hardcoded rgb colors should we be passing tokens?

packages/paste-website/src/components/tokens-list/index.tsx
```diff
@@ -56,19 +61,60 @@ export const TokensList: React.FC<TokensListProps> = (props) => {
...
+            </Column>
+            <Column span={3}>
+              <Label htmlFor="format-control" id="format-control-label">
+                Format
+              </Label>
+              <Select
+                id="format-control"
+                onChange={(evt) => {
+                  if (evt.target.value === 'javascript') setUseJavascriptNames(true);
+                  else setUseJavascriptNames(false);
```
SiTaggart, line 109: question: should this selection be persistent? i.e. I select JavaScript format, and when I come back to the page in a different session it remembers that I selected that preferred format last time and automatically shows me what I want?

packages/paste-website/src/components/tokens-list/index.tsx
nkrantz: The `backgroundColor` needs to change based on the selected theme in the filter, rather than the current site-wide theme. If we give the `TokenCard` a token name as a value it'll display the color based on the site-wide theme.

I'll try to get the correct rgb value from the dist tokens file and pass it in that way (conditionally based on selected filter theme) so that it's more accurate if we ever change the token values.

SiTaggart: Ah-ha! Yeah that makes perfect sense and great suggestion to make this implementation better too

packages/paste-website/src/components/tokens-list/index.tsx
nkrantz: good point, I think I'm actually removing the `selectedTheme` state though, because it was just being used to determine the background color of the example in the `TokenCard` so not really necessary anymore

packages/paste-website/src/components/tokens-list/index.tsx
nkrantz: I think we do want this and the theme to persist, I'm adding it to local storage. changes pushed up shortly

packages/paste-website/src/components/tokens-list/index.tsx
```diff
@@ -42,6 +41,19 @@ export const TokensList: React.FC<TokensListProps> = (props) => {
     setFilterString(filter);
   };
 
+  const handleThemeChange = (e: React.ChangeEvent<HTMLSelectElement>): void => {
+    // @ts-ignore todo: figure out ts error with DarkModeTokens
+    if (e.currentTarget.value === 'dark') setTokens(DarkModeTokens.tokens);
```
nkrantz, line 46: need some help debugging this ts error on `setTokens(DarkModeTokens.tokens)`

Argument of type '{ "background-colors": [{ type: "color"; category: "background-color"; value: "rgb(18, 28, 45)"; comment: "Default background color for any container"; originalValue: "{!palette-gray-100}"; name: "color-background"; }, ... 49 more ..., { ...; }]; ... 12 more ...; "z-indices": [...]; }' is not assignable to parameter of type 'SetStateAction<{ "background-colors": [{ type: "color"; category: "background-color"; value: "rgb(244, 244, 246)"; comment: "Background color used for containers."; originalValue: "{!palette-gray-10}"; name: "color-background"; }, ... 70 more ..., { ...; }]; ... 12 more ...; "z-indices": [...]; }>'.
  Type '{ "background-colors": [{ type: "color"; category: "background-color"; value: "rgb(18, 28, 45)"; comment: "Default background color for any container"; originalValue: "{!palette-gray-100}"; name: "color-background"; }, ... 49 more ..., { ...; }]; ... 12 more ...; "z-indices": [...]; }' is not assignable to type '{ "background-colors": [{ type: "color"; category: "background-color"; value: "rgb(244, 244, 246)"; comment: "Background color used for containers."; originalValue: "{!palette-gray-10}"; name: "color-background"; }, ... 70 more ..., { ...; }]; ... 12 more ...; "z-indices": [...]; }'.
    Types of property '"background-colors"' are incompatible.
      Type '[{ type: "color"; category: "background-color"; value: "rgb(18, 28, 45)"; comment: "Default background color for any container"; originalValue: "{!palette-gray-100}"; name: "color-background"; }, ... 49 more ..., { ...; }]' is not assignable to type '[{ type: "color"; category: "background-color"; value: "rgb(244, 244, 246)"; comment: "Background color used for containers."; originalValue: "{!palette-gray-10}"; name: "color-background"; }, ... 70 more ..., { ...; }]'.
        Source has 51 element(s) but target requires 72.ts(2345)

PR #2535, opened 2022-07-15
chore: make tokens filter responsive

nkrantz:
## In this PR

[Jira ticket](https://issues.corp.twilio.com/browse/DSYS-3175)

The purpose of the ticket was to add a modal for the filtering UI at smaller screen sizes, but [based on this slack thread](https://twilio.slack.com/archives/C02LKSS1H41/p1657905916067179), we decided to display all of the filtering controls at all sizes, but stack them at the smallest breakpoint.

This PR adds breakpoints and an additional grid to the `TokensListFilter` component to implement that behavior. Screen recording shows the proposed solution.

### Open questions
- I added a cypress test but I'm having trouble actually getting cypress to run. I'll dig into it more on Monday, but here's the error I get when I try to run a test using the gui

https://user-images.githubusercontent.com/75342690/179315855-85dc8ef3-51c2-4f95-a59f-171a86341277.mov

### DIFF
--- a/cypress/integration/token-list/index.spec.ts
+++ b/cypress/integration/token-list/index.spec.ts
@@ -54,4 +54,10 @@ describe('Token list filter format control and theme control', function () {
       'rgb(244, 244, 246)'
     );
   });
+
+  it('has a responsive layout', () => {
+    cy.get('[data-cy="input-column"]').should('have.attr', 'min-width', '0px');
+    cy.viewport('iphone-x');
+    cy.get('[data-cy="input-column"]').should('have.attr', 'min-width', '100%');
+  });
 });
--- a/packages/paste-website/src/components/tokens-list/TokensListFilter.tsx
+++ b/packages/paste-website/src/components/tokens-list/TokensListFilter.tsx
@@ -24,8 +24,8 @@ export const TokensListFilter: React.FC<TokensListFilterProps> = ({
   return (
     <>
       <Box marginBottom="space80">
-        <Grid gutter="space40">
-          <Column span={6}>
+        <Grid gutter="space40" vertical={[true, true, false]}>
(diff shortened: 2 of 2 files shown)


cypress/integration/token-list/index.spec.ts
```diff
@@ -54,4 +54,10 @@ describe('Token list filter format control and theme control', function () {
       'rgb(244, 244, 246)'
     );
   });
+
+  it('has a responsive layout', () => {
+    cy.get('[data-cy="input-column"]').should('have.attr', 'min-width', '0px');
```
TheSisb, line 59: I think the bug is that you're checking for `have.attr` but you should be checking for `have.css`

zahnster:
@nkrantz regarding the Cypress issue you're seeing locally, it looks like you might be running Node 17? I don't see the issue locally w/ Node 16. I'm also not seeing that issue on the CI, and regarding the issue on CI, follow @TheSisb's advice 😄 .

zahnster:
Suggestion: I'm a lil meh on how compressed the filtering gets as things get sized down - was thinking we could do a small refinement so it will wrap on our smallest & third breakpoints, and then go to one line for our second (large mobile; no right hand ToC nav) and fourth (full desktop).

It's a little hard to describe so I made a video to show what that would look like:
https://user-images.githubusercontent.com/36438/179587499-8fb6a9b6-d359-45cb-96ad-3491b350409e.mov

This is the code change driving that example:
Screen Shot 2022-07-18 at 2 05 57 PM

FWIW, I think these are the breakpoints I'm going to use for the TokenCards as well.

TheSisb (approved):

zahnster (approved):
Looks great!

PR #2539, opened 2022-07-19
chore(website): token list performance improvements

TheSisb:
- [x] Moves the copy to clipboard functionality to the token list component and only passes required, unchanged items down to TokenCard to prevent frequent list re-renders
- [x] Uses `useCallback` and `React.memo` tastefully to prevent re-renders

### DIFF
--- a/packages/paste-website/src/components/tokens-list/token-card/index.tsx
+++ b/packages/paste-website/src/components/tokens-list/token-card/index.tsx
@@ -4,7 +4,6 @@ import {Box} from '@twilio-paste/box';
 import {Text} from '@twilio-paste/text';
 import {Button} from '@twilio-paste/button';
 import {Tooltip, useTooltipState} from '@twilio-paste/tooltip';
-import {useClipboard} from '@twilio-paste/clipboard-copy-library';
 import {ScreenReaderOnly} from '@twilio-paste/screen-reader-only';
 import {CopyIcon} from '@twilio-paste/icons/esm/CopyIcon';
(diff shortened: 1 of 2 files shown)


zahnster (approved):
Praise: Nice work! Token perf is much better (using the same test as before I'm seeing an average of 298ms to perform the change, where before I was getting 677ms).

shleewhite (approved):
Nice!!

packages/paste-website/src/components/tokens-list/token-card/index.tsx
```diff
@@ -77,102 +76,106 @@ export const TokenCard: React.FC<{
...
   useCamelCase?: boolean;
-}> = ({category, name, value, comment, backgroundColor, useCamelCase}) => {
-  const tooltipState = useTooltipState();
-  const [tooltipText, setTooltipText] = React.useState('Copy token name');
-  // Prevents tooltip being visible on first render due to reakit positioning bug code
-  const isFirstRender = React.useRef(true);
-  const clipboard = useClipboard({copiedTimeout: 2000});
-  const tokenName = useCamelCase ? camelCase(name) : `$${name}`;
+  isCopied?: boolean;
+  onCopyText?: (tokenName: string) => void;
```
nkrantz, line 80: suggestion: should these props be abstracted into a `TokenCardProps` type?

nkrantz: merge went in before I finished my review!

PR #2540, opened 2022-07-19
feat(tokens-page): add collator and properly order tokens

zahnster:
* Re-orders the design tokens in the formatters to be alphanumeric (uses collator function previously in the Design Tokens List code in the Docs site)


packages/paste-website/src/components/tokens-list/index.tsx
```diff
@@ -97,17 +99,26 @@ export const TokensList: React.FC<TokensListProps> = (props) => {
...
-              ))}
+              {tokens[tokenCategory]
+                .sort((a, b) => {
+                  if (tokenCategory === 'font-weights') {
+                    const aValue: string = typeof a.value === 'string' ? a.value : a.value.toString();
+                    const bValue: string = typeof b.value === 'string' ? b.value : b.value.toString();
+                    return collator.compare(aValue, bValue);
+                  }
+                  return collator.compare(a.name, b.name);
+                })
```
TheSisb, line 110: question: just wondering if we can run this once in an effect and save it to state or if this needs to run every re-render

zahnster: Alternate thought: what if we did this sorting on the design-token formatters level? I was tempted to do it there first but then saw the prior art.

TheSisb: You can try that for the gatsby formatter export. If it doesn't work or is difficult to implement, you can alternatively consider a constant file somewhere in the website package that imports the theme, sorts it, and exports it out for this page to use.

SiTaggart (approved):
Seems reasonable to me

PR #2557, opened 2022-07-25
chore(website): reimplement token list filters with improvements

TheSisb:
- [x] Updates the design tokens package 'generic' export to remove `originalValue` keys and deprecated tokens. Also precomputes the altValue and includes them as a key.
- [x] Implements search on the token list. Can search by name, value, or altValue.
- [x] Fix the none-found card styles
- [x] Fix font-family tokencard smushing bug.


zahnster:
Arguably scope creep but can we fix the broken image for no results? 

Screen Shot 2022-07-25 at 4 04 54 PM

packages/paste-website/src/components/tokens-list/NoTokensFound.tsx
```diff
@@ -5,27 +5,25 @@ import {Heading} from '@twilio-paste/heading';
...
   <Card data-cy="tokens-empty-state" padding="space150">
-    <Stack orientation="horizontal" spacing="space110">
-      <Box
-        as="img"
-        src="/images/patterns/empty-no-results-found.png"
-        alt="No results found illustration"
-        size="size20"
-      />
+    <Box display="flex" columnGap="space110">
+      <Box as="img" src="/images/no-tokens-found.png" alt="No results found illustration" size="size20" />
```
shleewhite, line 17: Nit: this image is decorative because the text below says the same thing. You can do alt="" instead.

packages/paste-design-tokens/formatters/generic.d.ts.ts
```diff
@@ -55,6 +55,11 @@ export const formatGroupTokensWithTemplate = (
         .filter((prop) => prop !== undefined && cat === prop.get('category'))
         .toJS();
 
+      // Filter out deprecated tokens in exported file
+      // Add alt values
+      // Remove originalValue
```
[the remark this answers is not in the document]
TheSisb: It points to the alias name like `originalValue={pallette-red-80}`

shleewhite (approved):

.changeset/twelve-bobcats-sort.md
```diff
@@ -0,0 +1,6 @@
+---
+'@twilio-paste/design-tokens': patch
+'@twilio-paste/core': patch
```
SiTaggart, line 3: suggestion: I think technically this should be a major if we're moving things. Might be pretty important to call out for Flex too _just in case_

packages/paste-website/src/components/tokens-list/helpers.tsx
```diff
@@ -16,55 +13,34 @@ export const trackTokenFilterString = debounce((filter: string): void => {
...
-  if (unfilteredTokensBasedOnTheme != null) {
-    tokens = filterDeprecatedTokens(unfilteredTokensBasedOnTheme);
-  }
+function sanitizeForSearch(filterString: string): string {
+  // Remove space and `-` characters
+  // Lowercases
+  return filterString.replace(/[\s-]+/g, '').toLowerCase();
+}
+
+export const filterTokenList = (filterString: string, tokens: Tokens): Partial<Tokens> | null => {
```
SiTaggart, line 22: praise: this is great as a testable approach to this list

SiTaggart (approved):

packages/paste-design-tokens/formatters/generic.ts
```diff
@@ -11,10 +13,44 @@ export const categoryTemplate = (categoryName: string, props: DesignToken[]): st
   `;
 };
 
+const getTokenAltValue = ({category, value}: {category: string; value: string}): string | null => {
```
zahnster, line 16: praise: nice optimization

zahnster (approved):
Looks great!

cypress/integration/token-list/index.spec.ts
```diff
@@ -33,13 +33,28 @@ describe('Token list filter format control and theme control', function () {
     cy.visit('/tokens/list');
   });
 
-  it('controls format of token name', () => {
+  it('controls format of token name (default = css)', () => {
+    const tokenNode = cy.get('[data-cy="tokens-table-container"] li:first dt [data-paste-element="TEXT"]');
+    cy.get('[data-cy="format-control"]').should('have.value', 'css');
```
nkrantz, line 38: question: does cypress ever keep track of localStorage, or does it use a clean environment each time? in other words, could this test ever fail because 'javascript' was the last value used for the format control in the test, and we pull it from local storage on load? If so, should we do some cleanup after the tests run to clear local storage?

TheSisb: I didn't check, but I presume it would clear it. That said, I made the test end on CSS to not have to worry about it.

PR #2561, opened 2022-07-26
chore(website): add breakpoints, maxWidth

gloriliale:
## Changes in this PR
[[DSYS-3447](https://issues.corp.twilio.com/browse/DSYS-3447)] This PR affects the layout within the Token Card in the Tokens List page. Changes include:
- Set up two more break points for `TokenCardValue` component. This change allows for the right hand side nav to be used in more window sizes. You can see the new behavior from this change [in this video](https://www.loom.com/share/8edca1b57ca7421d9dc3e2eb28e7b452).
- Added a `maxWidth` to the Box surrounding the token values. Without a maxWidth value, some token's values that were extremely long (mainly font-family tokens) pushed the token name text to be much too small. 

### Open question
The `maxWidth` value is currently set to `size20`. Originally I chose `size30`, but certain screen widths still made font-family token names too squished. Now at `size20`, they look better. The trade-off is that there are now other token values, namely the shadow-focus ones, that wrap to two lines. This seemed fine by me but wanted to call it out in case there was someone with a strong case against it.

**Contributing to Twilio**

### DIFF
--- a/cypress/integration/token-list/index.spec.ts
+++ b/cypress/integration/token-list/index.spec.ts
@@ -7,20 +7,6 @@ describe('Token list filter with no existing search params', function () {
     cy.visit('/tokens/list');
   });
 
-  // Todo: uncomment search filter tests when filter work is done
-
-  //   it('searches for a background color', () => {
-  //     cy.get('input[name="tokens-filter"]').type('background').should('have.value', 'background');
-  //     cy.get('#background-colors').should('exist');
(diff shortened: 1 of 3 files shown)


zahnster (approved):
🙌🏼 woot!

shleewhite (approved):
Looks great!

TheSisb (approved):

PR #2568, opened 2022-07-28
Ensure header stays on top of tooltips

patcon:
On the tooltip page, tooltips with controlled state can show on top of the header:
https://paste.twilio.design/components/tooltip/#using-state-hooks

Screen Shot 2022-07-28 at 5 35 26 PM

Wasn't sure if all props were strings, but this might need to be a number :)

**Contributing to Twilio**

### DIFF
--- a/packages/paste-website/src/components/site-wrapper/site-header/SiteHeaderDesktop.tsx
+++ b/packages/paste-website/src/components/site-wrapper/site-header/SiteHeaderDesktop.tsx
@@ -20,6 +20,7 @@ const SiteHeaderDesktop: React.FC = () => {
       borderBottomWidth="borderWidth10"
       borderBottomStyle="solid"
       flex="1 0 auto"
+      zIndex="99"
     >
       <SiteHeaderLogo title="Paste" subtitle="UX Platform" />
       <Box


TheSisb:
Hey @patcon,

Sorry for the late reply on this one and thanks for the contribution! 

I'm not sure I'm seeing the issue here, isn't it expected that the tooltip appear above the sticky header?

patcon:
No worriesnon delay :)

And ah ok, if it seems right to you, then it's probably right! Guess I was just thinking in a "material" design sense, where the tooltip-emitting item going under an element might imply the tooltip should too.

Pls do feel welcome to close out if this doesn't feel actionable

TheSisb:
Interesting thought re: "where the tooltip-emitting item going under an element might imply the tooltip should too."

Out of curiosity I checked on the Material doc site and it looks like [Tooltips appear over their sticky header](https://mui.com/material-ui/react-tooltip/):
image

and so do [Popovers](https://paste.twilio.design/components/popover):
image


But not the [Popper library](https://mui.com/material-ui/react-popper/):
image


So I'll close this out, but appreciate the PR nonetheless as I hadn't consciously thought about this before.

PR #2569, opened 2022-07-28
fix(select): disabled background color

SiTaggart:
**Contributing to Twilio**

### DIFF
--- a/packages/paste-core/components/select/src/Select.tsx
+++ b/packages/paste-core/components/select/src/Select.tsx
@@ -31,7 +31,7 @@ export const SelectElement = React.forwardRef<HTMLSelectElement, SelectProps>(
         // We want the size attribute on the HTML element to set the height, not the css
         height={undefined}
         appearance="none"
-        backgroundColor={variant === 'inverse' ? 'colorBackgroundInverse' : 'colorBackgroundBody'}
+        backgroundColor="transparent"
         border="none"
         borderRadius="borderRadius20"
         boxShadow="none"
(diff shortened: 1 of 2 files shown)


serifluous (approved):

packages/paste-core/components/select/src/Select.tsx
```diff
@@ -31,7 +31,7 @@ export const SelectElement = React.forwardRef<HTMLSelectElement, SelectProps>(
         // We want the size attribute on the HTML element to set the height, not the css
         height={undefined}
         appearance="none"
-        backgroundColor={variant === 'inverse' ? 'colorBackgroundInverse' : 'colorBackgroundBody'}
+        backgroundColor="transparent"
```
TheSisb, line 34: Is this change intended? I dont imagine it will be working as intended if the select is placed over a different color background element

TheSisb (approved):
If that change is intended, lgtm

packages/paste-core/components/select/src/Select.tsx
SiTaggart: Yeah it is indeed. The select element is wrapped in InputBox. InputBox is the thing that sets all the background colors, including disabled.

The reason this broke, is because we started setting a background color on the select element, so you can't see InputBx handling the states behind it.

TheSisb: Aha, understood. Thanks

shleewhite (approved):

PR #2380, opened 2022-04-29
feat: add a stickyHeader prop to the THead element

krnan9525:
add a stickyHeader prop to the THead element.
This will be used by Table component and the Datagrid component

**Contributing to Twilio**

> All third-party contributors acknowledge that any contributions they provide will be made under the same open-source license that the open-source project is provided under.

- [x] I acknowledge that all my contributions will be made under the project's license.

PR #2574, opened 2022-07-30
feat: add table and data-grid sticky header

SiTaggart:
Taking over from https://github.com/twilio-labs/paste/pull/2380

This PR takes the work @krnan9525 started, and ensures that the sticky header is styled as we intend it to be from the Paste system.

Worked needed to be done to enough the head borders travel with the sticker header.


TheSisb (approved):

TheSisb:
Fix for https://github.com/twilio-labs/paste/discussions/1517

shleewhite (commented):
One thing: It looks like the right border on the table header has disappeared

Screen Shot 2022-08-01 at 12 46 50 PM

SiTaggart:
> One thing: It looks like the right border on the table header has disappeared

Is that in the Storybook? And which browser are you using?

shleewhite:
> > One thing: It looks like the right border on the table header has disappeared
> 
> Is that in the Storybook? And which browser are you using?

In storybook on firefox: https://5e53448165911c0022e68c74-edsnrvrfer.chromatic.com/?path=/story/components-alert-dialog--alert-dialog-vrt

shleewhite (approved):
Ok turns out borders in FF are super weird, so this PR is actually fine!

packages/paste-core/components/table/stories/index.stories.tsx
```diff
@@ -944,6 +944,60 @@ RowHeaders.story = {
...
+        ))}
+      </TBody>
+    </Table>
+  );
+};
+
+BorderlessStickyHeaders.story = {
+  name: 'Borderless Sticky headers',
+};
+
```
gloriliale, line 1000: suggestion: could we add a story for sticky headers that also uses `top`?

packages/paste-website/src/pages/components/table/index.mdx
```diff
@@ -590,6 +590,76 @@ rather than using the `TFoot`.
...
+    <TFoot>
+      <Tr>
+        <Th scope="row">Total tasks</Th>
+        <Td/>
+        <Td textAlign="right">70</Td>
+      </Tr>
+    </TFoot>
+  </Table>
+</Box>`}
+</LivePreview>
```
gloriliale, line 661: suggestion: the website example would be more realistic for the use case with more rows. For reference, right now the example only scrolls down two rows.

gloriliale (approved):
This is gonna be GREAT! Left a few suggestions but nothing blocking.

packages/paste-core/components/table/stories/index.stories.tsx
SiTaggart: Can and probably should

SiTaggart: Actually we can't. Because you can't actually see the top property apply until you scroll the page, which chromatic doesn't support

gloriliale: Is this something cypress can do?

SiTaggart: Yes, but not in this situation. 

Cypress targets the website, and talks to Applitools. 

We have Stories and talk to Chromatic.

serifluous (commented):
suggestion: add a sticky header example to data grid docs. OR if you want to rely on people going to the "Layouts" section in Table, I'd recommend being a little more specific in the content like:

> The Data Grid inherits from the [Table component](https://deploy-preview-2574--paste-docs.netlify.app/components/table/), so you can take advantage of all the same layout techniques found in the [Table examples](https://deploy-preview-2574--paste-docs.netlify.app/components/table#layouts), such as setting column widths, text alignment, footers, and sticky headers, to control how your Data Grids render on the page.

Everything else looks good. Just saw @kelseycavitt approved in Chromatic, so I didn't go too deep into reviewing the design.

SiTaggart:
> suggestion: add a sticky header example to data grid docs. 

Yeah, I'm going to add it to the Kitchen sink example just under the layout section. (Once it's published to NPM)

PR #2577, opened 2022-08-01
feat(website): new tokens page

zahnster:
It's heere! The new and improved tokens page. 

This PR is a culmination of feature tasks that have been vetted and completed on the feature branch. Let's validate this all looks as expected and will cleanly merge into main.

Tan tabby cat putting on round sunglasses, caption of "I'm ready."


katieporter:
issue: I'm not sure when this happened, and I apologize if we missed it in an earlier design review, but it looks like the token values have a top margin on them when in the side-by-side layout that's creating vertical misalignment. Can we adjust so that that margin is only there when the cards swap to the stacked layout on smaller viewports?

packages/paste-website/src/components/tokens-list/NoTokensFound.tsx
```diff
@@ -0,0 +1,51 @@
...
+
+interface NoTokensFoundProps {
+  onClearSearch: () => void;
+}
+
+export const NoTokensFound: React.FC<NoTokensFoundProps> = ({onClearSearch}) => (
+  <Card data-cy="tokens-empty-state" padding="space150">
+    <Box display="flex" columnGap="space110">
+      <Box flexShrink={0} size="size20">
+        <StaticImage
```
nkrantz, line 19: this is still showing up blurry for me, not a big deal, but I wonder why

zahnster:
> issue: I'm not sure when this happened, and I apologize if we missed it in an earlier design review, but it looks like the token values have a top margin on them when in the side-by-side layout that's creating vertical misalignment. Can we adjust so that that margin is only there when the cards swap to the stacked layout on smaller viewports?

@katieporter - the default browser margin on UL's slipped in after a refactor - nice catch! I've reset it and we should be good to go now.

packages/paste-website/src/components/tokens-list/token-card/token-example/SpacingExample.tsx
```diff
@@ -1,16 +1,16 @@
...
+import {remToPx} from '@twilio-paste/theme';
 
 interface SpacingExampleProps {
   tokenName: string;
   spacing: string;
+  color: Properties['color'];
 }
 
-export const SpacingExample: React.FC<SpacingExampleProps> = ({tokenName, spacing}) => {
+export const SpacingExample: React.FC<SpacingExampleProps> = ({tokenName, spacing, color}) => {
```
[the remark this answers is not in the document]
zahnster: Sizing tokens would be really large in comparison, wouldn't fit into the preview area (the largest one is 1,232 px). We can really only show these representative examples w/ the spacing tokens.

packages/paste-website/src/components/tokens-list/NoTokensFound.tsx
zahnster: I just upped the source to 2x, looks better (I think 😅).

shleewhite: Lol I usually make it like 8x just to be safe

katieporter:
issue: looks like there's a bug with the theme filter. when refreshing the page, the theme filter does save its state, and the token values are correct, but other styling of the token swatches (background color, text color, etc.) is not retaining the correct theme: https://www.loom.com/share/bc8419124fcc4710ae79b9fe39fa4937

packages/paste-website/src/components/tokens-list/NoTokensFound.tsx
```diff
@@ -0,0 +1,51 @@
...
+            Try adjusting your criteria to filter by the token name or value. If you still can’t find what you’re
+            looking for, reach out by creating a{' '}
+            <Anchor href="https://github.com/twilio-labs/paste/discussions" showExternal>
+              GitHub Discussion.
+            </Anchor>
+          </Text>
+          <Button
+            variant="secondary"
+            onClick={() => {
+              // setFilterString('');
```
SiTaggart, line 41: chore: commented code removal

TheSisb (approved):

SiTaggart (approved):

nkrantz (approved):
🎉

shleewhite (approved):

katieporter:
nit pick: I think the background color of the inverse swatches might have changed at some point—they should be using color-background-body-inverse for the background of the swatches, instead of color-background-inverse

katieporter (approved):

PR #2578, opened 2022-08-01
feat(box, style-props): enable css grid style props

TheSisb:
Adds CSS Grid to the Style Props and Box packages.

Since we already have `rowGap` and `columnGap` defined in the `space` category for Flex usage, I didn't bother adding them or moving them into `Grid`. The old version of those CSS rules were `grid-row/column-gap` but modern browsers all omit the `grid` prefix so this works out anyways.


SiTaggart (approved):
Nice! 🎉

shleewhite (approved):
YESSSSS

PR #2582, opened 2022-08-02
feat(button): add toggle button

shleewhite:
## Changes in this PR
* add toggle button option for secondary and secondary_icon variants
* add a storybook example for toggle buttons
* add documentation to the docs site

**Contributing to Twilio**


packages/paste-core/components/button/stories/index.stories.tsx
```diff
@@ -230,6 +234,28 @@ export const PrimaryIconButton = (): React.ReactNode => <IconSizeOptions variant
...
+  return (
+    <Box display="flex" flexDirection="row" columnGap="space60">
+      <Button variant="secondary" pressed={followPressed} onClick={() => setFollowPressed(!followPressed)}>
+        {followPressed ? <CheckboxCheckIcon decorative /> : <PlusIcon decorative />}
+        Follow
+      </Button>
+      <Button variant="secondary_icon" size="icon" pressed={boldPressed} onClick={() => setBoldPressed(!boldPressed)}>
+        <BoldIcon decorative={false} title="Bold" />
+      </Button>
+      <Button variant="secondary" size="icon" pressed={pausePressed} onClick={() => setPausePressed(!pausePressed)}>
```
shleewhite, line 251: Note: once circle buttons merge, I will make this size="circle"

packages/paste-website/src/component-examples/ButtonExamples.ts
```diff
@@ -0,0 +1,22 @@
...
+return (
+  <Box display="flex" flexDirection="row" columnGap="space60">
+    <Button variant="secondary" pressed={followPressed} onClick={() => setFollowPressed(!followPressed)}>
+      {followPressed ? <CheckboxCheckIcon decorative /> : <PlusIcon decorative />}
+      Follow
+    </Button>
+    <Button variant="secondary_icon" size="icon" pressed={boldPressed} onClick={() => setBoldPressed(!boldPressed)}>
+      <BoldIcon decorative={false} title="Bold" />
+    </Button>
+    <Button variant="secondary" size="icon" pressed={pausePressed} onClick={() => setPausePressed(!pausePressed)}>
```
shleewhite, line 15: Note: once circle buttons merge, I will make this size="circle"

packages/paste-core/components/button/src/SecondaryIconButton.tsx
```diff
@@ -40,15 +40,19 @@ const ButtonStyleMapping = {
...
 const SecondaryIconButton = React.forwardRef<HTMLButtonElement, DirectButtonProps>(
-  ({size, buttonState, fullWidth, ...props}, ref) => {
+  ({size, buttonState, fullWidth, pressed, ...props}, ref) => {
+    const toggleStyles = pressed === undefined ? {} : ToggleIconButtonStyles;
+
     // Must spread size styles after button styles
     return (
       <Box
         ref={ref}
+        aria-pressed={pressed}
```
SiTaggart, line 50: suggestion: it is probably worth while throwing some tests around these to check that

1) aria-pressed is set on the element correctly
2) the pressed styles are applied in the pressed state. I notice that you only have one preset pressed button example for VRT too. We probably need test cases for all states.

Random thought, is there anything required about setting aria-pressed to false? It's a string attribute which means it needs to be `aria-pressed="true"` or `aria-pressed="false"`, but we're treating it as a boolean which means it's `aria-pressed="true"` or not present. Does AT do OK with removal of pressed when doing a state change?

packages/paste-core/components/button/src/styles.ts
```diff
@@ -83,3 +83,52 @@ export const SizeStyles: {[key in ButtonSizes]: BoxStyleProps} = {
...
+
+  _hover: {
+    borderColor: 'colorBorderPrimaryStronger',
+    backgroundColor: 'colorBackgroundPrimaryWeakest',
+  },
+  _focus: {
+    backgroundColor: 'colorBackgroundPrimaryWeakest',
+    boxShadow: 'shadowFocus',
+  },
+  _focus_hover: {
```
gloriliale, line 118: thought: I was initially confused here by these hover, focus and focus_hover styles, since they are not for the `pressed` state, but are still spread onto the button. I realize now that this is because toggle styles are added when `pressed` is not undefined; when `pressed=false`, it's still defined, so these styles still apply. 

I'm back and forth on whether we mention `pressed=false` or not in the docs when we explain how to use it. There are props that we have people use in our system that we don't do this for. For example in button, `disabled` and `loading` are binary in that they are added and are true or aren't present and are false-y. `pressed` isn't like that since it's both telling our button that it's a toggle and which state of the toggle it's in. If we want a toggle button, you have to explicitly say that it's `pressed=true` or `pressed=false`. 

This might not be _that_ big of a mental leap for consumers, but maybe something to think about.

gloriliale (approved):
Noice! Left a few comments but nothing blocking.

packages/paste-core/components/button/src/SecondaryIconButton.tsx
shleewhite: Oh! I guess the way its rendered, when I set `aria-pressed={false}` it shows up in the dom as `aria-pressed="false"`. I double checked in the website and storybook and it behaves correctly. 

The only time it doesn't add the attribute in the DOM is if it is `aria-pressed={undefined}`.

packages/paste-core/components/button/src/styles.ts
shleewhite: Ohhhh true, I added more info about that on the docs site.

packages/paste-core/components/button/src/SecondaryIconButton.tsx
shleewhite: I added tests and confirmed that it works as expected.

SiTaggart (approved):

loreina (approved):

PR #2589, opened 2022-08-04
fix(form-pill-group): small style bugs

shleewhite:
## In this PR
* add margin0 to the FormPillButton
  * what this fixes: the x being pushed over 2px on safari. Safari by default has 2px horizontal margin on buttons, which other browsers don't have
* set borderStyle to solid on the FormPillButton
  * what this fixes: the border on the pill looked 2 tone because it was using the default inset before
* set the error icon to sizeIcon10 in FormPillButton
  * what this fixes: the pill has a fixed height and before the error icon was too tall. this pushed the content down weirdly in Safari. 

**Fixing the colors**
If the pill is selectable+dismissable, we want the 'x' color to be the same as the text color. 
If the pill dismissable, we want the 'x' to only change color if you directly hover it.

To accomplish this, I sometimes make the color of the 'x' inherit from the wrapper and the rest of the time it is set on the 'x' itself.

**Contributing to Twilio**


packages/paste-core/components/form-pill-group/package.json
```diff
@@ -24,6 +24,9 @@
     "clean": "rm -rf ./dist",
     "tsc": "tsc"
   },
+  "dependencies": {
+    "deepmerge": "4.2.2"
```
shleewhite, line 28: Note: Soooooooooooooooooooo to get this to work and not have super repetitive CSS, I needed to add this as a dependency. Does this make it a major update?

TheSisb (approved):

SiTaggart (approved):

PR #2595, opened 2022-08-05
feat(input-box): improve styling of InputChevronWrapper

TheSisb:
This PR updates the InboxBox ChevronWrapper designs. It also exports a function to get the color for the child icon. This is used in the `Combobox` and `Select` components now.

I also sneak in a fix for overflowing ToC on the website (easily seen on the Select page).


packages/paste-core/components/input-box/src/InputChevronWrapper.tsx
```diff
@@ -10,23 +56,31 @@ export interface InputChevronWrapperProps {
...
-      element={element}
-      ref={ref}
-    >
-      {children}
-    </Box>
-  )
+const InputChevronWrapper = React.forwardRef<HTMLDivElement, InputChevronWrapperProps>(
+  ({children, element = 'CHEVRON_WRAPPER'}, ref) => {
+    const {disabled, readOnly, variant} = useInputBoxContext();
+    const state = disabled ? 'disabled' : readOnly ? 'readOnly' : 'default';
```
SiTaggart, line 62: Question: this logic in the component looks to be the same as the logic in the utility. Is there a chance they get out of sync?

shleewhite (approved):
LGTM!

SiTaggart (approved):

packages/paste-core/components/input-box/__tests__/input-chevron-wrapper.test.tsx
```diff
@@ -2,20 +2,47 @@ import * as React from 'react';
...
 import {CustomizationProvider} from '@twilio-paste/customization';
-import {InputChevronWrapper} from '../src';
+import {getInputChevronIconColor, InputChevronWrapper, InputBox} from '../src';
 
 describe('HTML attributes', () => {
   it('should set a element data attribute for InputChevronWrapper', () => {
-    render(<InputChevronWrapper element="INPUT_CHEVRON_WRAPPER">input-chevron</InputChevronWrapper>);
+    render(
+      <InputBox element="INPUT_BOX">
+        <InputChevronWrapper element="INPUT_CHEVRON_WRAPPER">input-chevron</InputChevronWrapper>
```
gloriliale, line 11: Issue: For testing the default element attribute, this shouldn't have an element set on it (the one for `InputBox` isn't necessary either, but it's now what you're testing so :shrug-emoji:). Also, it looks like `InputChevronWrapper`'s default element attribute is just `CHEVRON_WRAPPER`. I personally prefer `INPUT_CHEVRON_WRAPPER` tho.

packages/paste-core/components/input-box/__tests__/input-chevron-wrapper.test.tsx
```diff
@@ -25,7 +52,9 @@ describe('Customization', () => {
           INPUT_CHEVRON_WRAPPER: {backgroundColor: 'colorBackground'},
         }}
       >
-        <InputChevronWrapper element="INPUT_CHEVRON_WRAPPER">input-chevron</InputChevronWrapper>
+        <InputBox element="INPUT_BOX">
+          <InputChevronWrapper element="INPUT_CHEVRON_WRAPPER">input-chevron</InputChevronWrapper>
```
gloriliale, line 56: Issue: The element set here should be removed as well.

packages/paste-core/components/input-box/src/InputBoxContext.tsx
```diff
@@ -0,0 +1,17 @@
...
+  disabled: boolean;
+  readOnly: boolean;
+  variant: Variants;
+}
+export const InputBoxContext = React.createContext<InputBoxContextData | null>(null);
+
+export const useInputBoxContext = (): InputBoxContextData => {
+  const context = React.useContext(InputBoxContext);
+  if (!context) {
+    throw new Error('useInputBoxContext must be used within a <InputBox> component.');
```
gloriliale, line 14: suggestion: this error confused me bc `useInputBoxContext` is also used in `<InputCheveronWrapper>`. I think it'd be more clear if it read "... within an `InputBox` _package_ component."

packages/paste-core/components/input-box/stories/inputbox.stories.tsx
```diff
@@ -30,6 +31,14 @@ const CommonStories: React.FC<Omit<InputBoxProps, 'children'>> = ({variant, ...p
       <InputBox variant={variant} insertAfter={<Insert />} insertBefore={<Insert />} {...props}>
         <Input />
       </InputBox>
+      <InputBox variant={variant} insertAfter={<Insert />} insertBefore={<Insert />} {...props}>
+        <Box display="flex" width="100%" position="relative">
+          <Input />
+          <InputChevronWrapper element="test">
```
gloriliale, line 37: question: is this element attribute used anywhere? if it's not, I don't think it's necessary, especially since this isn't how we want people to use it.

packages/paste-website/src/components/shortcodes/table-of-contents/TableOfContentsList.tsx
```diff
@@ -11,4 +11,9 @@ export const TableOfContentsList = styled(Scrollspy)`
   border-left-width: 1px;
   border-left-style: solid;
   border-left-color: ${themeGet('borderColors.colorBorderWeak')};
+  /* Fix overflowing on long pages.
+   * 220px is how tall the page is above this nav once you scroll.
+   */
+  max-height: calc(100vh - 220px);
+  overflow: auto;
```
gloriliale, line 18: praise: Woohoo!! yay for this fix!

gloriliale (commented):
Yay for chevron! Left some comments about customization, it's an easy fix.

packages/paste-core/components/input-box/src/InputBoxContext.tsx
TheSisb: Interesting. This needs to specifically be used within the `InputBox` component, not any component in the package. It's being used in _our_ code in InputChevronWrapper, but the error is for consumers to realize they can't use `InputChevronWrapper` outside of the `InputBox`. Does that make sense?

loreina (approved):

packages/paste-core/components/input-box/__tests__/input-chevron-wrapper.test.tsx
TheSisb: There's a comment in the code about this, element is always required on this component:

```
  // Requiring element here instead of extending directly from BoxProps.
  // This ensures an element prop is always passed on these composite components.
```

packages/paste-core/components/input-box/stories/inputbox.stories.tsx
TheSisb: It's added because it's a required prop

packages/paste-core/components/input-box/src/InputBoxContext.tsx
gloriliale: Ah okay. So if they try to use `InputChevronWrapper` inside a Button for example, they'd get this error? If so, I don't think a consumer would understand where `useInputBoxContent` is coming into play. Maybe referencing `InputChevronWrapper` where `useInputBoxContent` is in this message would be more clear?

packages/paste-core/components/input-box/__tests__/input-chevron-wrapper.test.tsx
gloriliale: ah okay. I mean, in that case, this test and the following one are repeats. I would still suggest changing the wrapper's default element to `INPUT_CHEVRON_WRAPPER` though, in keeping w/ naming conventions of subcomponents.

PR #2597, opened 2022-08-08
chore: improve DataGrid colspan props, types and docs

jb-twilio:
**Contributing to Twilio**


.changeset/dull-singers-decide.md
```diff
@@ -0,0 +1,6 @@
+---
+'@twilio-paste/core': patch
+'@twilio-paste/data-grid': patch
+---
+
+include supporting types and documentation for existing colSpan prop
```
nkrantz, line 6:
```suggestion
[Data Grid] Include supporting types and documentation for existing colSpan prop
```
suggestion (non-blocking): we usually add the package name to the changeset message

packages/paste-core/components/data-grid/src/DataGridHeader.tsx
```diff
@@ -1,14 +1,18 @@
...
+
 /**
  * DataGrid header (th) component.
  * Just a wrapper around the DataGridCell.
  *
  * @param {string} element - customization element
+ * @param {number} colspan - how many columns the cell spans across
  */
-export const DataGridHeader: React.FC<ThProps> = ({element = 'DATA_GRID_HEADER', ...props}) => {
+export const DataGridHeader: React.FC<DataGridHeaderProps> = ({element = 'DATA_GRID_HEADER', ...props}) => {
```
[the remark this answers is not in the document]
nkrantz: don't think it would be a breaking change since you're still using ThProps. what do you mean by supporting a super set of types?

nkrantz (commented):
first PR! 🔥 could you add the `colSpan` prop to the [props tables on the bottom of the Data Grid docs page](https://paste.twilio.design/components/data-grid#datagridheader-props)?

packages/paste-core/components/data-grid/__tests__/index.spec.tsx
```diff
@@ -42,6 +45,24 @@ describe('Data Grid', () => {
...
   });
 
+  describe('Column Span', () => {
+    it('applies colSpan attribute as expected', () => {
+      const {getByTestId} = render(<ColumnSpanDataGrid />);
+      const th = getByTestId('data-grid-header');
+      expect(th).toHaveAttribute('colspan', '5');
+    });
+
+    it('enforces numeric colSpan prop type', () => {
```
shleewhite, line 55: Issue: we don't usually test if prop types works as expected.... I'm also not sure this is test is correct because there isn't an expect. I think it would be ok to remove this test.

shleewhite (commented):
🥳 WooO! Just had one comment about the tests.

packages/paste-core/components/data-grid/__tests__/index.spec.tsx
SiTaggart: It does raise the question, should we throw? If you're not using Typescript, theoretically you can pass a string to the html element, nothing stops you.

packages/paste-core/components/data-grid/src/DataGridCell.tsx
```diff
@@ -23,6 +23,7 @@ export interface DataGridCellProps extends Pick<TdProps, 'textAlign'> {
  * DataGrid cell component. Every visible box in a data grid is powered by the cell.
  *
  * @param {"th" | "td"} as - is it a header or a regular cell
+ * @param {number} colspan - how many columns the cell spans across
```
SiTaggart, line 26: suggestion: probably need to add an old school proptype to the bottom of the file too for the new type restriction, for all the JS users.

packages/paste-core/components/data-grid/src/DataGridHeader.tsx
```diff
@@ -1,14 +1,18 @@
 import * as React from 'react';
 import * as PropTypes from 'prop-types';
-import {DataGridCell} from './DataGridCell';
+import {DataGridCell, DataGridCellProps} from './DataGridCell';
 import type {ThProps} from './table/Th';
+
+type DataGridHeaderProps = ThProps & Omit<DataGridCellProps, 'as'>;
```
SiTaggart, line 6: question: Does `thProps` really not expose `colSpan` as a props. It is I believe a global HTML attribute for a `th` element. Maybe we need to fix the types there instead? Potentially `ThProps` might be wrong... 🤔

packages/paste-core/components/data-grid/__tests__/index.spec.tsx
shleewhite: Isn't that what the old school react prop types are for?

SiTaggart: Yeah, you can make it noisy (see my other comment about adding an old school proptype), but as a consumer you can just ignore them. It's not fail proof.

nkrantz: I think it makes sense to throw and follows our patterns for props in other components

packages/paste-core/components/data-grid/src/DataGridHeader.tsx
SiTaggart: Asked a different way, should the `ThProps` interface in that file extend the types from the native `th` HTML element?

```jsx
Export interface ThProps extends React.HTMLElement<ThElement> {
  ...
}
```

(Note, just random pseudo code I just wrote on my iPad, exact code might differ)

I think the reason this, and future issues with these will crop up, is we're not extended the types of the html element we export a component for. 

The other risk might be that extending the grid cell props as the grid header props, in terms of HTML elements, there might be incorrect overlaps between the two. I.e. a `td` might not share something with a `th` element and vice-a-versa

SiTaggart (approved):
Praise: super neat and tidy!

nkrantz (approved):
Good stuff!! If you have time to add the `colSpan` prop to the Data Grid props table and also another example showing the different column sizes, that'd be great just so that people know it's available to them.

packages/paste-core/components/data-grid/src/DataGridHeader.tsx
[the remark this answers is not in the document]
TheSisb: more of a feat (minor) since we're adding functionality. Nothing breaking here.

TheSisb (approved):

shleewhite (approved):
🚢

PR #2616, opened 2022-08-17
chore: minor vs code plugin package improvements

TheSisb:
- [x] Update the readme
- [x] Update the package.json
- [x] Add LICENSE file
- [x] Verify publishing works 👍


gkassym (approved):
LGTM

SiTaggart (commented):
Question: The version number we have in the package json here, are we intending to version the plugin via changesets? Or is that handled else where in the VS Code / Azure workflows?

TheSisb:
the version is handled elsewhere by the `vsce` npm binary as documented

SiTaggart (dismissed):
Maybe try private package in the package.json and we're good to go otherwise

PR #2619, opened 2022-08-19
chore: danger check to ensure external library packages are pinned

SiTaggart:
# Whats in this PR

Danger JS check to see if, when we update or add an external library to the system, we pin the version number got safer and more repeatable install's.

## How this works

- It grabs all the package.json files in the changeset for the PR
- From each package.json file it grabs dependencies (we don't care about dev or peer deps for this PR)
- It filters out all the internal workspace deps
- It checks each dep for `^` and returns a list of unpinned dep names
- Throws an error if there are any unpinned deps

It works too.

image


gloriliale (approved):
Looks great!

shleewhite (approved):
Nice nice

TheSisb (approved):

PR #2590, opened 2022-08-04
Feat/toggle

nkrantz:
# In this PR

[Jira ticket](https://issues.corp.twilio.com/browse/DSYS-1770)

Adds a Toggle package (name may change to `Switch` or `Toggle Switch`), as well as tests, stories, and a docs page. The component is built like this:
```
<label>
  <TOGGLE role="switch">
    <TOGGLE_KNOB>
      <TOGGLE_ICON />
    </TOGGLE_KNOB>
  </TOGGLE>
  <TOGGLE_LABEL_TEXT_WRAPPER />
</label>
<TOGGLE_HELP_TEXT_WRAPPER />
```
### Changes not related to Toggle
- Added a few additional dependencies to the package.json template for the `create:package` script
- Fixed the github and storybook links in the Chat Log docs page because I noticed they pointed to combobox
- Added a new `_checked_hover` pseudo style prop to the Box package for Toggle styles
- Added new icon size and line height tokens for the Toggle check mark
- Added a note to checkbox and radio docs about help text usage
- Added to internal docs

### To do
- Clean up docs page a bit
- Improve toggle customization story
- Change package name?

PR #2625, opened 2022-08-22
feat(switch): add new package

nkrantz:
Replacement PR for #2590 

**Contributing to Twilio**


packages/paste-core/components/switch/src/SwitchHelpText.tsx
```diff
@@ -0,0 +1,28 @@
...
+import * as PropTypes from 'prop-types';
+import type {SwitchHelpTextProps} from './types';
+import {HelpText} from '@twilio-paste/help-text';
+import {Box, safelySpreadBoxProps} from '@twilio-paste/box';
+
+const SwitchHelpText = React.forwardRef<HTMLSpanElement, SwitchHelpTextProps>(
+  ({element = 'SWITCH_HELP_TEXT_WRAPPER', children, helpTextId, ...props}, ref) => {
+    return (
+      // @ts-ignore TODO:no 50px tokens
+      <Box {...safelySpreadBoxProps(props)} element={element} marginLeft="50px" ref={ref}>
```
nkrantz, line 11: We don't have any tokens for this margin (it aligns the help text with the label text). Is the ts-ignore here ok?

packages/paste-design-tokens/tokens/global/sizing.yml
```diff
@@ -114,6 +114,9 @@ props:
     comment: Generic square sizing token scale for UI components.
 
   # icons
+  size-icon-05:
+    value: "{!line-height-05}"
+    comment: "Mapps to line-height tokens 1:1"
```
gloriliale, line 119:
```suggestion
    comment: "Maps to line-height tokens 1:1"
```

packages/paste-theme/__tests__/__snapshots__/generateThemeFromTokens.spec.ts.snap
```diff
@@ -80,6 +80,7 @@ Object {
     "sizeSquare90": "1px",
   },
   "iconSizes": Object {
+    "sizeIcon05": undefined,
```
gloriliale, line 83: question: I know this is a generated snapshot but is this `undefined` what's supposed to be here? it shouldn't be 1px like the rest?

gloriliale (commented):
yaasss Switch yaaaaassss 🎉 Left a few comments but it's spectacular. Lemme know when to look at the docs!

packages/paste-design-tokens/tokens/global/sizing.yml
nkrantz: found this at the same time 😆

packages/paste-core/components/switch/src/SwitchContainer.tsx
```diff
@@ -0,0 +1,71 @@
...
+                )}
+                <MediaBody>{label ?? <></>}</MediaBody>
+              </MediaObject>
+            </SwitchLabelText>
+          </MediaObject>
+        </Label>
+        {helpText ? (
+          <SwitchHelpText element={`${element}_HELP_TEXT_WRAPPER`} helpTextId={helpTextId}>
+            {helpText}
+          </SwitchHelpText>
```
TheSisb, line 50: suggestion (blocking): omit the SwitchHelpText and the SwitchLabelText files/components and define them inline. Reason being that we don't necessarily want our users to compose those elements as they can be imported from `@twilio-paste/core/switch/dist/SwitchHelpText` for example. Instead we want to manage just SwitchContainer which defines them internally so our API surface area is smaller.


```suggestion
          <Box {...safelySpreadBoxProps(props)} element={element} marginLeft="50px" ref={ref}>	
        <HelpText id={helpTextId} marginTop="space0">
          {children}
        </HelpText>
      </Box>
```

packages/paste-core/components/switch/src/SwitchHelpText.tsx
TheSisb: Not ideal. A question for design @twilio-labs/design-systems-pd if we have a "Good enough" token for here?

packages/paste-core/components/switch/src/SwitchKnob.tsx
```diff
@@ -0,0 +1,76 @@
...
+    const moveSwitchKnob = React.useCallback(() => {
+      if (!disabled) {
+        if (!switchIsOn && isHovering) return 'translateX(5%)';
+        if (switchIsOn && !isHovering) return 'translateX(100%)';
+        if (switchIsOn && isHovering) return 'translateX(95%)';
+      } else if (disabled) {
+        if (!switchIsOn) return 'translate(0%)';
+        if (switchIsOn) return 'translateX(100%)';
+      }
+    }, [switchIsOn, isHovering, disabled]);
```
TheSisb, line 22: suggestion: I would recommend converting this to a [pure function](https://medium.com/javascript-scene/master-the-javascript-interview-what-is-a-pure-function-d1c076bec976) as it becomes easier to test and reason about:

```suggestion
  function getSwitchKnobTransformValue(disabled: boolean, isOn: boolean, isHovering: boolean) {
    // the code you have
  }
  ```

And then define it outside the React component (line 7's closure).

 Then you can use it here as
```js
const transformValue = React.useMemo(() => getSwitchKnobTransformValue(disabled, isOn, isHovering), [disabled, isHovering, isOn]);
```

Now in your test you can just import the function and pass it several arguments to test it works as intended in a very elegant manner without needing a whole component mounting and css reading strategy.

packages/paste-core/components/switch/src/types.ts
```diff
@@ -0,0 +1,35 @@
...
+  isHovering: boolean;
+  height: string;
+}
+
+export interface SwitchHelpTextProps extends Pick<BoxProps, 'element'> {
+  children: NonNullable<React.ReactNode>;
+  helpTextId: string;
+}
+
+export interface SwitchLabelTextProps extends Omit<TextProps, 'as'>, Pick<BoxProps, 'element'> {
```
TheSisb, line 33: question (non-blocking): Can you simplify these by extending LabelProps and HelpTextProps instead?

packages/paste-core/components/switch/src/types.ts
```diff
@@ -0,0 +1,35 @@
...
+  helpText?: string | React.ReactNode;
+  label?: string | React.ReactNode;
+  required?: boolean;
+}
+
+export interface SwitchProps extends React.HTMLAttributes<HTMLDivElement>, Pick<BoxProps, 'element'> {
+  id?: string;
+  disabled?: boolean;
+  on?: boolean;
+  labelId?: string;
```
TheSisb, line 17: question (non-blocking): what's the difference in use-case between labelId and id? Can this work with just one id?

packages/paste-core/components/switch/stories/index.stories.tsx
```diff
@@ -0,0 +1,88 @@
...
+};
+
+export const On: React.FC<{element?: string}> = ({element}) => {
+  const id = useUID();
+  return (
+    <SwitchContainer id={id} element={element} helpText="This is help text" label="This is the label">
+      <Switch on />
+    </SwitchContainer>
+  );
+};
```
TheSisb, line 60: suggestion: it would be helpful here to create a story with a controlled state where clicking toggles the Switch on/off for future development and maintenance.

[edit] And for our docs examples too

TheSisb (commented):
Really solid work! 🎉

packages/paste-core/components/switch/src/Switch.tsx
```diff
@@ -0,0 +1,104 @@
...
+      if (!disabled) setSwitchIsOn(!switchIsOn);
+      if (onClick) onClick();
+    }, [onClick, disabled, switchIsOn]);
+
+    const handleKeyDown = React.useCallback(
+      (event: React.KeyboardEvent): void => {
+        if (event.key === ' ' || event.key === 'Enter') setSwitchIsOn(!switchIsOn);
+      },
+      [switchIsOn]
+    );
```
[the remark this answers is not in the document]
nkrantz: ooooo cool

packages/paste-theme/__tests__/__snapshots__/generateThemeFromTokens.spec.ts.snap
SiTaggart: I wonder if this is indicative of a bug? I feel like it should be something.

.changeset/large-nails-wave.md
```diff
@@ -0,0 +1,6 @@
+---
+'@twilio-paste/core': patch
+'@twilio-paste/box': patch
+---
+
+[Box] add a `_checked_hover` pseudo selector style prop for use in switch package
```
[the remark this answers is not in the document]
nkrantz: I marked it as a patch since it's not really a new feature for the consumer, but I have no preference. Anyone else have thoughts on this?

packages/paste-core/components/switch/src/SwitchHelpText.tsx
nkrantz: this is resolved by simon's commit

packages/paste-core/components/switch/src/types.ts
nkrantz: using a labelId allows us to use the required aria-labelledby prop

.changeset/large-nails-wave.md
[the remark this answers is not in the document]
SiTaggart: > I marked it as a patch since it's not really a new feature for the consumer

That seems reasonable to me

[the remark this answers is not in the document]
TheSisb: This is actually a minor but I'm not too concerned about making it a patch myself 🤷

TheSisb (approved):

loreina (approved):

Issue #2618, opened 2022-08-18
Controlled Combobox caret jumping to end when trying to update a previous part of the text 


mannycolon:
Thanks for helping make Paste better! For general questions or feature requests, please post [a discussion](https://github.com/twilio-labs/paste/discussions/new).

**Describe the bug**
When typing a help request query, and trying to update a previous part of the text, the caret jumps to the end after the first keystroke. This negatively impacts UX and makes text editing difficult. 

**To Reproduce**

1. Go to [Controlled Combobox](https://paste.twilio.design/components/combobox/#controlled-combobox).
2. Click on the combobox and type any text.
3. Click on the beginning of the text and type 123, see how the caret jumps to the end of input box.

**Expected behavior**
Caret should not jump to the end of the input box when updating a Combobox value. 

**Screenshots**
![image](https://user-images.githubusercontent.com/17072300/185493682-6026430b-092b-4468-8e62-9a2c33806800.png)

Desktop (please complete the following information):

OS: macOS
Browser: Chrome
Version 104.0.5112.79

shleewhite:
Hi @mannycolon!

Thank you for posting this, we are investigating the issue now.

SiTaggart:
Hi @mannycolon 

There is a small fix incoming in this PR #2647, along with some doc example updates.

If you are using the `useCombobox` hood from Combobox, you don't need to wait for the fix to merge. We had a bad example. When doing controlled comboboxes and the hook, rather than setting `inputValue`, allow Downshift to control the state internally. Just switch `inputValue` to setting `initialInputValue`, and keep everything else. This fixes the cursor jumping.

If you are not using the hook, then you'll have to wait for that PR to merge and we'll come back to let you know if was released.

mannycolon:
Hi @SiTaggart,

Awesome, thank you so much. I've already fix the problem on our implementation by following along your PR changes. We are using the useCombobox hook and it was as easy as just updating the variable names as you said. Thank you so much for your help.

Regards!

SiTaggart:
This is now released in Paste Core v15.2

PR #2647, opened 2022-08-27
fix(combobox): controlled autocomplete cursor placement onInputValueChange bug

SiTaggart:
This is a super annoying bug in Downshift that's been open for 2 years. https://github.com/downshift-js/downshift/issues/1108

They actually, kind of allude to it here in the Github docs https://github.com/downshift-js/downshift#oninputvaluechange, where they link to some old issue and some hints towarsds a fix.

Basically, we're doing it wrong. If you're trying to use the combobox in a controlled mannor, you should be using `defaultInputValue`, not `inputValue`, and let Downshift manage the internal state afterwards. We re-map `inputValue` to `defaultInputValue` to address this. This should have no detrimental impact to the consumer and should just "fix the problem" with no code changes.

For folks using the hook, we're doing it wrong again. You want to set the _initial_ value and selected item, and let Downshift take over from there. In this case I have just updated the Docs example.

### DIFF
--- a/packages/paste-core/components/combobox/src/types.ts
+++ b/packages/paste-core/components/combobox/src/types.ts
@@ -37,21 +37,27 @@ export type HighlightedIndexChanges = {
 export interface ComboboxProps extends Omit<InputProps, 'id' | 'type' | 'value'>, Pick<BoxProps, 'element'> {
   autocomplete?: boolean;
   helpText?: string | React.ReactNode;
+  labelText: string | NonNullable<React.ReactNode>;
+  optionTemplate?: OptionTemplateFn<any>;
+  groupLabelTemplate?: (groupName: string) => React.ReactNode;
+  groupItemsBy?: string;
+  variant?: InputVariants;
+
+  // Downshift useCombobox Hook Props. Thes are mainly covered in https://github.com/downshift-js/downshift/blob/master/src/hooks/useCombobox/README.md#advanced-props docs
   initialIsOpen?: UseComboboxPrimitiveProps<any>['initialIsOpen'];
   initialSelectedItem?: UseComboboxPrimitiveProps<any>['initialSelectedItem'];
   items: UseComboboxPrimitiveProps<any>['items'];
   itemToString?: UseComboboxPrimitiveProps<any>['itemToString'];
-  labelText: string | NonNullable<React.ReactNode>;
   onHighlightedIndexChange?: UseComboboxPrimitiveProps<any>['onHighlightedIndexChange'];
   onInputValueChange?: UseComboboxPrimitiveProps<any>['onInputValueChange'];
   onIsOpenChange?: UseComboboxPrimitiveProps<any>['onIsOpenChange'];
   onSelectedItemChange?: UseComboboxPrimitiveProps<any>['onSelectedItemChange'];
-  optionTemplate?: OptionTemplateFn<any>;
-  groupLabelTemplate?: (groupName: string) => React.ReactNode;
   selectedItem?: UseComboboxPrimitiveProps<any>['selectedItem'];
   inputValue?: UseComboboxPrimitiveProps<any>['inputValue'];
-  groupItemsBy?: string;
-  variant?: InputVariants;
+  getA11yStatusMessage?: UseComboboxPrimitiveProps<any>['getA11yStatusMessage'];
+  getA11ySelectionMessage?: UseComboboxPrimitiveProps<any>['getA11ySelectionMessage'];
+
+  // Downshift useCombobox Hook return props for when you are using the hook outside of the component
   state?: Partial<UseComboboxPrimitiveReturnValue<any>>;
   /**
    * Use `onInputValueChange` instead.
--- a/packages/paste-core/components/combobox/stories/index.stories.tsx
+++ b/packages/paste-core/components/combobox/stories/index.stories.tsx
@@ -508,21 +508,30 @@ ComboboxOverflowLongValue.story = {
 };
 
 export const ComboboxControlled = (): React.ReactNode => {
-  const [value, setValue] = React.useState('');
-  const [selectedItem, setSelectedItem] = React.useState({});
+  const [value, setValue] = React.useState('United Arab Emirates');
+  const [selectedItem, setSelectedItem] = React.useState({code: 'AE', label: 'United Arab Emirates', phone: '971'});
   const [inputItems, setInputItems] = React.useState(objectItems);
   return (
     <>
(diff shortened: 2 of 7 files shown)


packages/paste-core/components/combobox/stories/index.stories.tsx
```diff
@@ -523,7 +523,8 @@ export const ComboboxControlled = (): React.ReactNode => {
             {item.code} | {item.label} | {item.phone}
           </div>
         )}
-        onInputValueChange={({inputValue}) => {
+        __onControlledAutoCompleteInputValueChangeWorkaround={(e) => {
```
TheSisb, line 526: question: what if instead of this new prop / API, we remapped `onInputValueChange` to `onChange` always?

shleewhite:
> As I write this though, I am probably leaning more towards just advising against it, pulling the docs for controlled entirely and just pointing everyone to the state hook Downshift provides as the solution to controlled comboboxes and the cursor problem listed in discussions #2618

If we decide not to add the workaround prop, I think we should at least fix it on the docs site. Seems like people are looking at that example, so it'd be good to at least have it working there.

packages/paste-core/components/combobox/stories/index.stories.tsx
SiTaggart: `onInputValueChange` is the Downshift approved, you should do it this way, API.

It's only because there's a "bug" in Downshift that when it's controlled, you need to circumvent it to stop it taking over the input value.

So we _could_ have it called `onChange` and do a bunch of stuff underneath to make sure it's only `onChange` when it's autocomplete and controlled, otherwise use `onInputValueChange` under the hood.

SiTaggart: OK thrown alll this away, better approach instead

SiTaggart:
> If we decide not to add the workaround prop, I think we should at least fix it on the docs site. Seems like people are looking at that example, so it'd be good to at least have it working there.

Yeah, planning on just having "the correct docs" as the fix to the problem after further investigating this.

TheSisb (dismissed):
LGTM

packages/paste-core/components/combobox/src/types.ts
```diff
@@ -37,21 +37,27 @@ export type HighlightedIndexChanges = {
 export interface ComboboxProps extends Omit<InputProps, 'id' | 'type' | 'value'>, Pick<BoxProps, 'element'> {
   autocomplete?: boolean;
   helpText?: string | React.ReactNode;
+  labelText: string | NonNullable<React.ReactNode>;
+  optionTemplate?: OptionTemplateFn<any>;
+  groupLabelTemplate?: (groupName: string) => React.ReactNode;
+  groupItemsBy?: string;
+  variant?: InputVariants;
```
nkrantz, line 44: don't we already have `variant` from InputProps being extended above?

nkrantz (approved):
just the changeset

packages/paste-core/components/combobox/stories/index.stories.tsx
```diff
@@ -508,21 +508,30 @@ ComboboxOverflowLongValue.story = {
...
         autocomplete
         items={inputItems}
         labelText="Choose a country:"
         helpText="This is the help text"
-        optionTemplate={(item: ObjectItem) => (
-          <div>
-            {item.code} | {item.label} | {item.phone}
-          </div>
-        )}
+        optionTemplate={(item: ObjectItem) => {
```
gloriliale, line 527: I noticed something in these stories when clicking around, and then saw it in subsequent ones, just wanted to check if it was expected behavior. Loom here: https://www.loom.com/share/52cfc26cc5154f8b98a6254b48f0921a

packages/paste-core/components/combobox/src/types.ts
```diff
@@ -37,21 +37,27 @@ export type HighlightedIndexChanges = {
...
+
+  // Downshift useCombobox Hook Props
```
gloriliale, line 46: suggestion (non-blocking): it took me a while to find the A11y props in the downshift documentation bc it isn't actually in their site, it's on the ReadMe for their repo. Might be good to paste the link to that here for future reference.

gloriliale (approved):
woop woop!

packages/paste-core/components/combobox/stories/index.stories.tsx
SiTaggart: Yeah it can be a little funky, it is intentional but not clear.

This is just the way we've implemented our stories. You get to choose how to display the list option. So in autocomplete comboboxes we filter the list of options based on what you type or what you select. The persist it on reopen. 

No _wrong_, not _right_. Just "a way" to do it. Good spot though

packages/paste-core/components/combobox/src/types.ts
SiTaggart: Yeah, you're right! I wonder why it was doing that before? 🤔

packages/paste-core/components/combobox/src/types.ts
SiTaggart: That's not a bad call, they're pretty hard to find.

packages/paste-core/components/combobox/src/types.ts
SiTaggart: I'm going to leave it for now, as I'm respectfully fearful of this component and how often it breaks

SiTaggart:
Fixes #2618

packages/paste-website/src/pages/components/combobox/index.mdx
```diff
@@ -436,6 +436,14 @@ This function allows you to use your own `jsx` template for the items in the dro
 
 The variant of the Combobox. Available variants are `default` or `inverse`.
 
+##### `getA11yStatusMessage?: () => void`
+
+Useful to compose accessible status messages to assistive technology users, including translations.
+
+##### `getA11ySelectionMessage?: () => void`
+
+Useful to compose accessible status messages to assistive technology users, including translations.
```
serifluous, line 445: Should this say "selection" instead of "status"? It's the exact same description as the line above.

nkrantz (approved):

packages/paste-website/src/pages/components/combobox/index.mdx
SiTaggart: It should, yes!!

TheSisb (approved):

nkrantz (approved):

serifluous (approved):

gloriliale (approved):

PR #2651, opened 2022-08-30
feat(icons): add contact center icon set to the Paste icon library

SiTaggart:
**Contributing to Twilio**


TheSisb (approved):

netlify.toml
```diff
@@ -3,9 +3,6 @@
   targetPort = 8000
   framework = "#custom"
 
-[context.deploy-preview]
```
shleewhite, line 6: Question: why did you remove this?

packages/paste-icons/src/AgentIcon.tsx
```diff
@@ -24,6 +24,7 @@ const AgentIcon = React.forwardRef<HTMLElement, AgentIconProps>(
         <svg
           role="img"
           aria-hidden={decorative}
+          xmlns="http://www.w3.org/2000/svg"
```
shleewhite, line 27: Question:  why did you add this to all the icons?

shleewhite (approved):
Seems good! Just had a couple clarifying questions.

netlify.toml
SiTaggart: Ah-ha, I meant to comment on this.

Because there is a weird caching behaviour when you add an icon, if you haven't _also_ changed something on the website, when you redeploy with without cache it won't actually happen because none of the files this is looking for has changed. It's kind of weird, because you would expect that behaviour to also happen just on a normal push, but it doesn't. So I still don't think this works in the way we want it to. So, to save confusion in the future I just removed it.

packages/paste-icons/src/AgentIcon.tsx
SiTaggart: I just ran convert all, so that the updated svg's of existing icons are converted to new React components. `convert-new` only does missing react components. 

I actually assumed it was also adding it to the new icons, but it isn't. So that's weird... 🤷🏻‍♂️

serifluous: yeah, I had the same question bc it was happening in my branch.

I removed xmlns from the SVG export on the new icons because I thought they were breaking the command somehow. But these SVGs weren't changed at all so I don't know why it's doing it now.

shleewhite: ... weird, nothing changed in svg-to-react recently

TheSisb: It's not needed, but not a big deal either. Just adds extra bytes needlessly to the icons. Won't call it a blocker. I wonder if it crept up due to not pinning deps in SVG-to-React

packages/paste-core/core-bundle/code-block/package.json
```diff
@@ -5,4 +5,4 @@
   "sideEffects": false,
   "main": "../dist/code-block.js",
   "types": "../dist/code-block.d.ts"
-}
+}
```
shleewhite, line 8: Question: why is this in the PR?

packages/paste-icons/tools/templates/jsonTemplate.ts
```diff
@@ -1,10 +1,26 @@
...
+
+const createIconsFilesObject = (icons: string[] | undefined): IconFiles[] | undefined =>
+  icons?.map((iconName: string) => ({
+    name: iconName,
+    svg: `/svg/${iconName}.svg`,
+    cjs: `/cjs/${iconName}.js`,
+    esm: `/esm/${iconName}.js`,
+  }));
+
 export const jsonTemplate = (iconList: string[] | undefined): string => {
```
shleewhite, line 18: Question: should this function be unit tested?

shleewhite (approved):
Nice!

packages/paste-icons/tools/templates/jsonTemplate.ts
SiTaggart: Maybe. Its output is largely covered by the existing test, but if you think we could benefit from one I'd be happy to add it

packages/paste-core/core-bundle/code-block/package.json
SiTaggart: Unfortunately, if you build main, this is an uncommitted change. I think it slipped through somehow.

serifluous (approved):

shleewhite (approved):

TheSisb (approved):

TheSisb (approved):

TheSisb:
Double approve by accident, Github was lagging like crazy with all those file changes lol

PR #2654, opened 2022-08-31
feat(help-text): add success and warning variants

nkrantz:
[DSYS-3469](https://issues.corp.twilio.com/browse/DSYS-3469)

Adds `warning` and `success` variants to the Help Text component. Also updates the docs and stories.

Out of scope: update Switch, Radio, and Checkbox to accept help text variants other than default. Will address when requested.

### DIFF
--- a/.changeset/olive-peas-end.md
+++ b/.changeset/olive-peas-end.md
@@ -1,0 +1,6 @@
+---
+'@twilio-paste/help-text': minor
+'@twilio-paste/core': minor
+---
+
+[Help text] add success and warning variants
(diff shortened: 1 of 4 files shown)


kelseycavitt (commented):
With the addition of these two new variants, should we include information about the use of multiple help text instances within a form field? Do we allow for stacking of them? For further context, we don't allow this for alerts right now.

kelseycavitt (dismissed):

SiTaggart (approved):

shleewhite (commented):
Looks good! Only thing missing is adding customization unit tests for the new variants.

nkrantz:
> Only thing missing is adding customization unit tests for the new variants.

@shleewhite those are in the big customization story! same as simon's comment above it's not the best way of doing things to lump them all into one story, but this component is getting worked on for 1twilio anyway so maybe the stories can be split up during that sweep

gloriliale (approved):
Looks good to me! 🎉

kelseycavitt (approved):

PR #2660, opened 2022-09-06
feat(combobox): add MultiselectCombobox

TheSisb:
This PR introduces a MultiselectCombobox component that enables selecting multiple options from a dropdown list. This component is exported from the `@twilio-paste/combobox` package because it uses the same underlying styles and library, so there was a lot of opportunity for code re-use.

This PR makes several small changes to a handful of packages throughout Paste, listed in the changesets committed. These changes were needed to power the MultiselectCombobox.

The component has been highly requested: https://github.com/twilio-labs/paste/discussions/2279, https://github.com/twilio-labs/paste/discussions/2259, https://github.com/twilio-labs/paste/discussions/2291, https://github.com/twilio-labs/paste/discussions/2490

PR includes:
- [x] A designed & coded component
- [x] Stories
- [x] Tests

Documentation to come in a follow-up PR.


packages/paste-core/components/combobox/src/styles/ComboboxListboxOption.tsx
```diff
@@ -1,74 +1,95 @@
...
-        >
-          {children}
-        </Text>
+        <Box as="span" minHeight="1.25rem">
+          <Text as="span" color="inherit" element={`${element}_LIST_ITEM_TEXT`}>
+            {children}
+          </Text>
+        </Box>
+        <Box as="span" minHeight="1.25rem" marginLeft="auto" hidden={selected ? false : true}>
+          <CheckboxCheckIcon decorative={false} title="TODO: INJECT AS PROP FOR I18N" color="inherit" />
```
gloriliale, line 87: Todo: complete this i18n todo!

packages/paste-core/components/combobox/src/styles/ComboboxListboxOption.tsx
```diff
@@ -1,74 +1,95 @@
...
       >
-        <Text
-          as="span"
-          color="colorText"
-          element={`${element}_LIST_ITEM_TEXT`}
-          textDecoration={highlighted ? 'underline' : null}
-        >
-          {children}
-        </Text>
+        <Box as="span" minHeight="1.25rem">
```
gloriliale, line 81: question: why do we need to give this a `minHeight`? is it because we've added an icon?

packages/paste-core/components/combobox/src/multiselect/MultiselectCombobox.tsx
```diff
@@ -0,0 +1,314 @@
...
+            <GrowingInput
+              // we spread props into `getInputProps` so that Downshift handles events correctly
+              {...getInputProps({
+                ...getDropdownProps({ref, preventKeyAction: isOpen}),
+                disabled,
+                required,
+                ...props,
+              })}
+              aria-describedby={helpTextId}
+              element={`${element}_ELEMENT`}
```
shleewhite, line 275: Issue: the element should be `${element}_INPUT`

packages/paste-core/components/combobox/src/multiselect/MultiselectCombobox.tsx
```diff
@@ -0,0 +1,314 @@
...
+            />
+            <InputChevronWrapper element={`${element}_CHEVRON_WRAPPER`}>
+              <ChevronDownIcon
+                aria-hidden="true"
```
shleewhite, line 279: Nit: you don't need aria-hidden if it is marked as decorative.

packages/paste-core/components/combobox/__tests__/MultiselectCombobox.spec.tsx
```diff
@@ -0,0 +1,235 @@
...
+      expect(renderedTextbox.getAttribute('aria-labelledby')).toEqual(document.querySelector('label')!.id);
+
+      // unique option ids
+      const renderedOptions = getAllByRole('option');
+      const optionIDs = renderedOptions.map((option) => option.id);
+      const uniqueIDs = uniq(optionIDs);
+      expect(uniqueIDs.length).toEqual(optionIDs.length);
+    });
+
+    it('should render a label for that matches the input id', () => {
```
shleewhite, line 140: Question: should this be combined with the "should render a combobox with aria attributes" test?

packages/paste-core/components/combobox/src/helpers.tsx
```diff
@@ -1,12 +1,17 @@
-import type {Item as ItemType} from './types';
+import type {Item} from './types';
```
shleewhite, line 1: Issue: I think these helper functions should probably have tests.

packages/paste-core/components/combobox/__tests__/multiselect-combobox-customization.spec copy.tsx
```diff
@@ -0,0 +1,206 @@
+import * as React from 'react';
+import {render, screen} from '@testing-library/react';
+import type {RenderOptions} from '@testing-library/react';
+import {CustomizationProvider} from '@twilio-paste/customization';
+import type {PasteCustomCSS} from '@twilio-paste/customization';
+import {InformationIcon} from '@twilio-paste/icons/esm/InformationIcon';
+import {Text} from '@twilio-paste/text';
+import {MultiselectCombobox} from '../src';
+
+const getStyles = (element = 'MULTISELECT_COMBOBOX'): {[key: string]: PasteCustomCSS} => ({
```
shleewhite, line 10: Praise: I love this way of doing the customization tests!!!! Will follow this moving forward.

packages/paste-core/components/combobox/__tests__/multiselect-combobox-customization.spec copy.tsx
```diff
@@ -0,0 +1,206 @@
...
+      'MULTISELECT_COMBOBOX_GROUPNAME'
+    );
+
+    expect(container.querySelectorAll('[data-paste-element="MULTISELECT_COMBOBOX_LIST_ITEM_TEXT"]')).toHaveLength(3);
+    expect(container.querySelector('[data-paste-element="MULTISELECT_COMBOBOX_PREFIX"]')).toBeInTheDocument();
+    expect(container.querySelector('[data-paste-element="MULTISELECT_COMBOBOX_SUFFIX"]')).toBeInTheDocument();
+  });
+  it('should set the correct attributes on all MultiselectCombobox components with custom element prop', () => {
+    const {container} = render(<MultiselectComboboxToTest element="FOO" />, {wrapper: initCustomizationWrapper()});
+    expect(container.querySelector('[data-paste-element="FOO"]')).toBeInTheDocument();
```
shleewhite, line 91: Question: just wanted some clarity on this... I thought we don't usually just query the dom to see if an element with the data-paste-elemnet attribute exists because we want to make sure that it is on the correct element. What is the right way?

Ex. `expect(screen.getByRole('textbox').dataset.pasteElement).toEqual("FOO_INPUT")` instead of `expect(container.querySelector('[data-paste-element="FOO_INPUT"]').toBeInTheDocument()`

packages/paste-core/components/combobox/__tests__/multiselect-combobox-customization.spec copy.tsx
```diff
@@ -0,0 +1,206 @@
...
+    initialIsOpen
+    element={element}
+    optionTemplate={(item) => <div>{item.number}</div>}
+    itemToString={(item) => (item ? item.letter : '')}
+    insertBefore={<Text as="span">Z</Text>}
+    insertAfter={<InformationIcon decorative={false} size="sizeIcon20" title="Get more info" />}
+  />
+);
+
+describe('Combobox data-paste-element attributes', () => {
```
shleewhite, line 58: Issue: this should be renamed "Multiselect combobox data-paste-element attributes" for clarity.

packages/paste-core/components/combobox/src/multiselect/GrowingInput.tsx
```diff
@@ -0,0 +1,61 @@
+import * as React from 'react';
```
shleewhite, line 1: Question: should the growingInput have any unit tests? Just to make sure the width is right or something?

shleewhite (commented):
Great work!!!!!!!! I tried the combobox out with VoiceOver to verify that it works and I noticed a couple things:
* A nit, but the strings in the storybooks could be cleaned up. It is confusing trying to test this with a screenreader when the combobox is called "Choose a book" but the selected pills are called "Products" and the options are names of paste components.
* all of the pills should have 'aria-selected="true"' - this was what we did at salesforce to let people know that the users know that these were the options they already picked (@SiTaggart this is still best practice, right?)
* Currently, the active item in the dropdown automatically has `aria-selected="true"`. This makes it really confusing as a screen reader user because the item isnt actually selected until I press enter and because as I navigate through the list, every single item is announced as selected. The aria-selected attribute should instead be set depending on if the item is in the pill group.

.yarn/releases/yarn-3.0.2.cjs
```diff
@@ -61,9 +61,9 @@ Please report this to https://github.com/szmarczak/http2-wrapper/`);p(m)}return}
...
...._mutate;this._mutate=!0;let i=e(this);return this._mutate=r,i}concat(e){if(!e||e===this)return this;if(e.type!==this.type&&this.type!=="mixed")throw new TypeError(`You cannot \`concat()\` schema's of different types: ${this.type} and ${e.type}`);let r=this,i=e.clone(),n=wn({},r.spec,i.spec);return i.spec=n,i._typeError||(i._typeError=r._typeError),i._whitelistError||(i._whitelistError=r._whitelistError),i._blacklistError||(i._blacklistError=r._blacklistError),i._whitelist=r._whitelist.merge(e._whitelist,e._blacklist),i._blacklist=r._blacklist.merge(e._blacklist,e._whitelist),i.tests=r.tests,i.exclusiveTests=r.exclusiveTests,i.withMutation(s=>{e.tests.forEach(o=>{s.test(o.OPTIONS)})}),i}isType(e){return this.spec.nullable&&e===null?!0:this._typeCheck(e)}resolve(e){let r=this;if(r.conditions.length){let i=r.conditions;r=r.clone(),r.conditions=[],r=i.reduce((n,s)=>s.resolve(n,e),r),r=r.resolve(e)}return r}cast(e,r={}){let i=this.resolve(wn({value:e},r)),n=i._cast(e,r);if(e!==void 0&&r.assert!==!1&&i.isType(n)!==!0){let s=(0,qV.default)(e),o=(0,qV.default)(n);throw new TypeError(`The value of ${r.path||"field"} could not be cast to a value that satisfies the schema type: "${i._type}".
```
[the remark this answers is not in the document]
TheSisb: Bad find and replace for pinning lodash, will fix.

package.json
```diff
@@ -170,7 +170,7 @@
     "commander": "^2.20.0",
     "concurrently": "^4.1.1",
     "cross-env": "^5.2.0",
-    "csstype": "^3.0.11",
+    "csstype": "3.0.11",
```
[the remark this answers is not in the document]
TheSisb: Simon added a DangerJS check to verify `dependencies` are pinned. It only runs against `package.json` files with changes in commit history, so old `package.json` files that aren't pinned are ignored. I use lodash in my package.json, so I had to pin it. When pinning lodash, yarn complains that all lodashes need to be pinned. In doing so, I touch a package.json file with csstype so that starts to complain. I pin csstype and we're good to go.

packages/paste-utils/src/hooks/useWindowSize.ts
```diff
@@ -0,0 +1,64 @@
...
+ * This hook returns the window size and the current breakpoint index.
+ * @returns {WindowSizeState} an object containing the window size and the current breakpoint index
+ */
+export function useWindowSize(): WindowSizeState {
+  // Initialize state with undefined width/height so server and client renders match
+  // Learn more here: https://joshwcomeau.com/react/the-perils-of-rehydration/
+  const [windowSize, setWindowSize] = useState<WindowSizeState>({
+    width: undefined,
+    height: undefined,
+    breakpointIndex: undefined,
```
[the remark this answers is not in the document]
TheSisb: Note: This hook is used by the website. I moved it to our utils package so I can use it in Combobox.

BreakpointIndex is helpful so you can render in JS exactly like how CSS breakpoints work, see https://github.com/twilio-labs/paste/blob/multiselect-combobox/packages/paste-website/src/components/homepage/Experiment.tsx#L26

packages/paste-core/primitives/combobox/src/index.tsx
```diff
@@ -13,9 +13,12 @@ export type {
   UseComboboxState as UseComboboxPrimitiveState,
   UseComboboxReturnValue as UseComboboxPrimitiveReturnValue,
   UseComboboxStateChange as UseComboboxPrimitiveStateChange,
+  UseComboboxGetItemPropsOptions as UseComboboxPrimitiveGetItemPropsOptions,
+  UseComboboxGetMenuPropsOptions as UseComboboxPrimitiveGetMenuPropsOptions,
+  GetPropsCommonOptions as GetComboboxPrimitivePropsCommonOptions,
   UseMultipleSelectionInterface as UseMultiSelectPrimitiveInterface,
   UseMultipleSelectionProps as UseMultiSelectPrimitiveProps,
-  UseMultipleSelectionState as UseMultiSelectState,
+  UseMultipleSelectionState as UseMultiSelectPrimitiveState,
```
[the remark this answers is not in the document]
TheSisb: It's undocumented API so we don't need to mark it as breaking. Also reporting says no on is using it.

packages/paste-core/components/combobox/src/styles/ComboboxEmptyState.tsx
```diff
@@ -0,0 +1,27 @@
...
+import type {BoxProps} from '@twilio-paste/box';
+import {ComboboxListboxOption} from './ComboboxListboxOption';
+import {ComboboxListboxGroup} from './ComboboxListboxGroup';
+
+export interface ComboboxEmptyStateProps {
+  element?: BoxProps['element'];
+}
+
+const ComboboxEmptyState = React.memo(
+  React.forwardRef<HTMLUListElement, ComboboxEmptyStateProps>(({element = 'COMBOBOX', ...props}, ref) => {
```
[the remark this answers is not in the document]
TheSisb: Heh not quite. Combobox is a complicated component, where children render other children in an intermixed way. If you check the code, this component renders `ComboboxListboxGroup` and passes `element={element}` into it.  It's confusing but, basically `COMBOBOX` (or the "root" element value) is passed all the way down and components concatenate it internally.

packages/paste-core/components/combobox/src/multiselect/MultiselectCombobox.tsx
TheSisb: I agree, but left it as ELEMENT to match Combobox :/

packages/paste-core/components/combobox/src/helpers.tsx
TheSisb: They do, in Combobox.spec.tsx under Unit Tests

packages/paste-core/components/combobox/__tests__/multiselect-combobox-customization.spec copy.tsx
TheSisb: Your way is better. I was just copying Combobox without thinking about it too deeply.

packages/paste-core/components/combobox/src/multiselect/GrowingInput.tsx
TheSisb: Wasn't sure what I could test tbh. It's all CSS, so nothing to unit test. And visually it's invisible, so not much to VRT either.

.danger/pin-external-deps.ts
```diff
@@ -58,7 +58,9 @@ export default () => {
 
     if (unpinnedExternalDeps.length > 0) {
       fail(
-        'There are some package.json files in this PR that contain unpinned external package libraries. Please pin your external package libraries by removing the ^ from the beginning of the version number.'
+        `There are some package.json files in this PR that contain unpinned external package libraries. Please pin your external package libraries by removing the ^ from the beginning of the version number. See: ${JSON.stringify(
+          unpinnedExternalDeps
```
SiTaggart, line 62: praise: nice, that's a useful addition

package.json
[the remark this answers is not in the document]
SiTaggart: For extra context, even though we ignore the root package.json file in the dangerJS check, when we change it at the workspace level, manypkg check makes it required that the monorepo uses the same version across all packages and root. Hence you have to change it here too.

packages/paste-core/components/combobox/__tests__/Combobox.spec.tsx
```diff
@@ -336,42 +344,38 @@ describe('Combobox', () => {
...
-      fireEvent.click(screen.getAllByRole('option')[0]);
+      fireEvent.click(getAllByRole('option')[0]);
+
       // @ts-ignore Property 'value' does not exist on type 'HTMLElement' (I get it, but this is right)
-      expect(screen.getByRole('textbox').value).toEqual('Andorra');
-      expect(screen.getByTestId('input-value-span').textContent).toEqual('"Andorra"');
-      expect(screen.getByTestId('selected-item-span').textContent).toEqual(
-        '{"code":"AD","label":"Andorra","phone":"376"}'
-      );
+      expect(getByRole('textbox').value).toEqual('Andorra');
```
SiTaggart, line 355: Observation: Where or not I agree, KCD suggests using `screen` as "the thing to do" https://kentcdodds.com/blog/common-mistakes-with-react-testing-library#not-using-screen

package.json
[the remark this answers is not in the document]
shleewhite: Ohhhh ok that makes sense! Thanks for explaining!

packages/paste-core/components/combobox/src/styles/ComboboxEmptyState.tsx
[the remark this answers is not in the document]
shleewhite: -______- dang this is so complicated hahah

packages/paste-core/components/combobox/__tests__/multiselect-combobox-customization.spec copy.tsx
```diff
@@ -0,0 +1,206 @@
+import * as React from 'react';
```
SiTaggart, line 1: chore: file name needs updating

packages/paste-core/components/combobox/__tests__/virtualization.spec.tsx
```diff
@@ -39,17 +39,17 @@ describe('Combobox - Virtualization', () => {
   describe('Without option template', () => {
     const items = ['a', 'b', 'c', 'd', 'e'];
     it('should apply the correct height to the virtualized list box', () => {
-      render(<Combobox data-testid="test" items={items} labelText="Choose a letter:" />, {
+      render(<Combobox data-testid="test" items={items} labelText="Choose a letter:" initialIsOpen />, {
```
SiTaggart, line 42: fyi: I want to say I ran into this previously, but I'm not 100% sure. when debugging those jest memory leaks, it looked like setting initial open was causing an state change that wasn't getting wrapped in `act`. I could also be totally thinking of something else.

SiTaggart:
That increase in size in combobox seems alarming. Are you sure you've caught all the right peer deps listed and it's not bundling things it souldn't be?
image

packages/paste-core/components/combobox/src/styles/ComboboxListboxOption.tsx
```diff
@@ -1,74 +1,103 @@
 import * as React from 'react';
-import * as PropTypes from 'prop-types';
 import {Box, safelySpreadBoxProps} from '@twilio-paste/box';
+import {CheckboxCheckIcon} from '@twilio-paste/icons/esm/CheckboxCheckIcon';
```
SiTaggart, line 3: chore: needs to be swapped with the new icon you added for this very purpose.

packages/paste-core/components/combobox/src/styles/ComboboxListboxOption.tsx
```diff
@@ -1,74 +1,103 @@
...
-        >
-          {children}
-        </Text>
+        <Box as="span">
+          <Text as="span" color="inherit" element={`${element}_LIST_ITEM_TEXT`}>
+            {children}
+          </Text>
+        </Box>
+        <Box as="span" hidden={selected ? false : true}>
+          <CheckboxCheckIcon decorative color="inherit" />
```
SiTaggart, line 95: issue: we probably need something to represent this icon to AT users. When using this with a screen reader, I actually get no indication that the item is part of a collection of selected items from the overall combobox. It might be a limitation with Downshift but as you arrow through the list, Downshift sets aria-selected to true for the highlighted index. So every items says "selected". This isn't necessarily "wrong" either, for the purposes of that interaction, moving through the list of options, it is the selected item in that list for the purpose of navigation.

packages/paste-core/components/combobox/src/multiselect/MultiselectCombobox.tsx
```diff
@@ -0,0 +1,314 @@
...
+            {...comboboxProps}
+            aria-expanded={ariaExpanded}
+            position="relative"
+            width="100%"
+            paddingY="space30"
+            paddingX="space40"
+          >
+            <FormPillGroup
+              {...pillState}
+              aria-label="Products:"
```
SiTaggart, line 244: issue: hard coded string.

SiTaggart, line 244: issue: hard coded string

packages/paste-core/components/combobox/src/multiselect/MultiselectCombobox.tsx
```diff
@@ -0,0 +1,314 @@
...
+          <Box
+            {...getToggleButtonProps({disabled})}
+            {...comboboxProps}
+            aria-expanded={ariaExpanded}
+            position="relative"
+            width="100%"
+            paddingY="space30"
+            paddingX="space40"
+          >
+            <FormPillGroup
```
SiTaggart, line 242: observation: just as an observation, the hidden string of keyboard instructions make for a weird experience with the combobox. Nothing we can do right now, just calling it out as not ideal.

The keyboard instructions are announced as soon as you focus on the comobbox becuase it appears as inner text for the control.

packages/paste-core/components/combobox/src/styles/ComboboxListboxOption.tsx
SiTaggart: @shleewhite this is my commentary on your question

> Currently, the active item in the dropdown automatically has aria-selected="true". This makes it really confusing as a screen reader user because the item isnt actually selected until I press enter and because as I navigate through the list, every single item is announced as selected. The aria-selected attribute should instead be set depending on if the item is in the pill group.

SiTaggart:
> all of the pills should have 'aria-selected="true"' - this was what we did at salesforce to let people know that the users know that these were the options they already picked (@SiTaggart this is still best practice, right?)

@shleewhite I'm not sure any more, honestly. And it's made more complicated in that our set of pills has a color change for selected pills. It might not be necessary, if you label the group in such a way that it tells you they are the selected items for the combobox, then you achieve what we were trying to achieve at Salesforce without the aria-select true needing to be set on all the items

packages/paste-core/components/combobox/__tests__/virtualization.spec.tsx
TheSisb: I fixed the leak

packages/paste-core/components/combobox/src/styles/ComboboxListboxOption.tsx
```diff
@@ -1,74 +1,103 @@
...
+        justifyContent="space-between"
+        alignItems="center"
+        paddingY="space40"
+        cursor={disabled ? 'not-allowed' : 'pointer'}
         ref={ref}
+        transition="background-color 150ms ease, color 150ms ease"
+        borderBottomWidth="borderWidth10"
+        borderBottomColor="colorBorderDecorative10Weaker"
+        borderBottomStyle="solid"
+        _before={{
```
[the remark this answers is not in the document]
TheSisb: Woops broke that in my other quick fix. Just fixed it.

packages/paste-libraries/dropdown/stories/multiselect.stories.tsx
```diff
@@ -0,0 +1,234 @@
...
+  return books.filter(function filterBook(book) {
+    return (
+      !selectedItems.includes(book) &&
+      (book.title.toLowerCase().includes(lowerCasedInputValue) ||
+        book.author.toLowerCase().includes(lowerCasedInputValue))
+    );
+  });
+}
+
+export const MultiselectComboBox: Story = () => {
```
gloriliale, line 36: praise: really glad we've got some library stories now, for combobox but especially multiselect!

packages/paste-libraries/dropdown/stories/multiselect.stories.tsx
```diff
@@ -0,0 +1,234 @@
...
+  const [selectedItems, setSelectedItems] = React.useState<Book[]>(initialSelectedItems);
+  const items = React.useMemo(() => getFilteredBooks(selectedItems, inputValue), [selectedItems, inputValue]);
+
+  const {
+    // Call and destructure its returned object on the selected item element.
+    getSelectedItemProps,
+    // Call optionally with preventKeyAction and destructure its returned object on the toggle button (select) or input (combobox) element.
+    getDropdownProps,
+    // Action prop that adds the item to the selection. Best used in useSelect and useCombobox prop onStateChange or onSelectedItemChange
+    // addSelectedItem,
```
gloriliale, line 47: issue: this shouldn't be commented out, right?

gloriliale (commented):
MultiselectComboboxxxxx!!!! This was such a big undertaking and it looks FANTASTIC! Go Shadi!

The only thing I wanted to add is that some of these stories are not open initially and the content that we'd want to VRT is in the options inside the listbox (I'm thinking **with optionTemplate**, **with initialSelectedItems**, maybe even just a **Basic** with that starts off open). Could we have them those with options open initially for more useful testing? And could we also have an open options story showing the empty state?

package.json
```diff
@@ -241,5 +241,5 @@
       "pre-push": "if git-branch-is main; then yarn pre-push; fi"
     }
   },
-  "packageManager": "yarn@3.1.0"
+  "packageManager": "yarn@3.0.2"
```
SiTaggart, line 244: issue: This might be an unintended consequence of a bad find and replace for lodash, if I read that correctly. This is downgrading yarn by a minor version.

packages/paste-core/components/combobox/src/styles/ComboboxEmptyState.tsx
```diff
@@ -0,0 +1,27 @@
...
+    return (
+      <ComboboxListboxGroup element={element} ref={ref}>
+        <ComboboxListboxOption {...props} element={`${element}_EMPTY_STATE`} variant="default" disabled>
```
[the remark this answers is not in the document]
SiTaggart: I think the resolution here was that we're going for composition, over doing empty state for the consumer. Because in some cases it can be disabled, in others it can't.

These will be covered in examples on the doc page

packages/paste-core/components/combobox/src/multiselect/MultiselectCombobox.tsx
```diff
@@ -0,0 +1,309 @@
...
+      // For multiselect, we need `selectedItem` to always be null because
+      // we use the stateChange event to add/remove items to the multiselect hook
+      selectedItem: null,
+      onInputValueChange,
+      onHighlightedIndexChange,
+      onIsOpenChange,
+      initialIsOpen,
+      // https://www.downshift-js.com/use-combobox#state-reducer
+      // Handles how state in Downshift should change as a result of user action
+      stateReducer(_state, actionAndChanges) {
```
SiTaggart, line 117: question: When we talked about override downshifts control over aria-selected on the dropdown listbox options, is this where we'd override it to `false` unless it's a selected item that appears in the pill list?

packages/paste-core/components/combobox/src/multiselect/GrowingInput.tsx
```diff
@@ -0,0 +1,61 @@
...
+import {InputElement} from '@twilio-paste/input';
+
+interface GrowingInputProps {
+  initialValue?: string;
+  element?: BoxProps['element'];
+  onChange?: (event: React.ChangeEvent<HTMLInputElement>) => void;
+  id: string;
+}
+
+export const GrowingInput = React.forwardRef<HTMLInputElement, GrowingInputProps>(
```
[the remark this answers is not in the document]
TheSisb: We want an input element to be inline to the selected FormPills. Our Paste Input is 100% width so would never be inline, it wraps to the next line. A normal input has a prefixed width, say 100px, so may wrap too early considering the multiselect won't be super wide either. This component starts off with iirc 5-10px width and grows as you type, so the input element stays inline as long as possible and feels less janky.

package.json
TheSisb: Strange, we have 3.0.2 installed but this says to use 3.1.0.  I'm going to replace our 3.0.2 installation with the 3.1.0 version then to keep this setting correct.

packages/paste-core/components/combobox/src/multiselect/MultiselectCombobox.tsx
TheSisb: I found a simpler approach that wouldn't be prone to Downshift's API changing by manually controlling it after the destructuring of downshift props on the ComboboxListboxOption component.

packages/paste-core/components/combobox/src/styles/ComboboxListboxOption.tsx
```diff
@@ -1,74 +1,107 @@
...
+    ref
+  ) => {
+    const virtualItemStyles = startHeight != null ? getVirtualStyles(startHeight) : {};
+
     return (
       <Box
         {...safelySpreadBoxProps(props)}
+        // FIX: Manually control aria-selected state to be true when an item has been selected.
+        // Downshift sets aria-selected to true when an item is highlighted, which we don't want.
+        aria-selected={selected}
```
SiTaggart, line 58: issue: weirdly this is getting rendered as `[Object object]`

image

packages/paste-core/components/combobox/src/multiselect/MultiselectCombobox.tsx
SiTaggart: just came across a problem with our plan to override `i18nKeyboardControls` so it's blank, and place that hidden keyboard controls text _after_ the combobox element. The `aria-decribedby` prop that you would also need to override to point this to the new hidden keyboard controls, isn't overridable in the form pill group.

packages/paste-core/components/combobox/src/styles/ComboboxListboxOption.tsx
TheSisb: Good catch, fixed

packages/paste-core/components/combobox/src/styles/ComboboxEmptyState.tsx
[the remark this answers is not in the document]
TheSisb: Deleted this component. It is now a prop passed to the component as `EmptyState`

packages/paste-core/components/combobox/src/styles/ComboboxListboxOption.tsx
TheSisb: Fixed.

packages/paste-core/components/combobox/src/multiselect/MultiselectCombobox.tsx
TheSisb: Good catch. Fixed.

TheSisb:
Fixing peerDeps fixed the package size! 
image

shleewhite (dismissed):
🚢 🚢 🚢  WOOOOOOOO

gloriliale (dismissed):
SHHHHHHIP IT

nkrantz (dismissed):

SiTaggart (approved):

gloriliale (approved):

kelseycavitt (approved):

PR #2674, opened 2022-09-09
feat: add file picker package

nkrantz:
# In this PR

The native file picker is `<input type="file" />`. This File Picker hides the native element visually using css, but it still appears in the DOM. The reason we're able to get the functionality while it's hidden is because the styled button is wrapped in a label that's connected to the native file picker. That way, clicking on anything inside of the label (the button) will trigger the file browser to open.


packages/paste-core/components/file-picker/src/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+              clip="rect(0 0 0 0)"
+              onChange={handleChange}
+            />
+            {required ? <RequiredDot /> : <></>}
+            <Button variant="secondary" as="span" size="small" disabled={disabled}>
+              Choose a file
+            </Button>
+          </Label>
+        </MediaFigure>
+        <Box height="100%" display="flex" verticalAlign="center">
```
SiTaggart, line 57: Suggestion: This Box wrapping the Media Body, essentially breaks the MediaObject. Media Object has a prob for alignment already. Try to only add things _inside_ the body

packages/paste-core/components/file-picker/src/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+      const file = evt.target.value;
+      setFileName(file);
+      console.log('the event', evt);
+      console.log('file name', fileName);
+    };
+
+    return (
+      <MediaObject {...props}>
+        <MediaFigure spacing="space30">
+          <Label htmlFor={id} element={element}>
```
SiTaggart, line 32: Suggestion: Due to the nature of the component, and Label, I suspect we might be better off not using `Label` internally. I don't think you need any of the styles, and if we update the styles to Label, that might have unintended concequences on File Picker. It should be a label element, maybe just a Box as label.

```
<Label>File picker label</Label>
<FilePicker />
<HelpText>help text goes here underneath</HelpText>

File: packages/paste-core/components/file-picker/src/index.tsx [hunk 3]
```diff
@@ -0,0 +1,78 @@
...
+
+    const handleChange = (evt: any): void => {
+      const file = evt.target.value;
+      setFileName(file);
+      console.log('the event', evt);
+      console.log('file name', fileName);
+    };
+
+    return (
+      <MediaObject {...props}>
```
SiTaggart, line 30: Suggestion: as the component is a react replacement, for the native file input element, I suspect the intended target for additional props should be the same element that the ref goes on it. Safely spread, of course.

packages/paste-core/components/file-picker/src/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+              overflow="hidden"
+              padding="space0"
+              margin="space0"
+              whiteSpace="nowrap"
+              textTransform="none"
+              position="absolute"
+              clip="rect(0 0 0 0)"
+              onChange={handleChange}
+            />
+            {required ? <RequiredDot /> : <></>}
```
SiTaggart, line 51: suggestion: I'd be tempted to just leave the Label component to do the display of the required dot, as it does it for us. You do still want to be able to place the `required` HTML attribute on the input element though, so keep that in mind too.

```
<Label htmlFor="resume" required>Resume</Label>
<FilePicker aria-describedby="resume_help_text" id="resume" name="resume" onChange="" required/>
<HelpText id="resume_help_text">Use your work email address.</HelpText>
```

packages/paste-core/components/file-picker/src/index.tsx
nkrantz: This is from me pushing up at the wrong stage of trying to vertically center the label text 🥲

packages/paste-core/components/file-picker/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,62 @@
...
+    expect(screen.getByTestId('test-file-picker')).toHaveAttribute('aria-describedby', textId);
+  });
+});
+
+describe('FilePicker customization', () => {
+  it('should set an element data attribute on File Picker', () => {
+    const {container} = render(<Default />);
+    expect(container.querySelector('[data-paste-element="FILEPICKER"]')).toBeInTheDocument();
+    expect(container.querySelector('[data-paste-element="FILEPICKER_BUTTON"]')).toBeInTheDocument();
+    expect(container.querySelector('[data-paste-element="FILEPICKER_TEXT"]')).toBeInTheDocument();
```
nkrantz, line 30: I know we typically like to have more specific ways of ensuring these get applied correctly, but I couldn't think of any great ones in this case. Lmk if you have thoughts

packages/paste-core/components/file-picker/src/FilePicker.tsx
```diff
@@ -0,0 +1,110 @@
...
+          </SiblingBox>
+        </label>
+      </>
+    );
+  }
+);
+
+FilePicker.displayName = 'FilePicker';
+
+FilePicker.propTypes = {
```
shleewhite, line 102: Issue(blocking): this is missing the i18n prop

packages/paste-core/components/file-picker/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,62 @@
+import * as React from 'react';
+import {render, screen} from '@testing-library/react';
+import {Default, Disabled, Customized} from '../stories/index.stories';
+
+describe('FilePicker', () => {
+  it('should render', () => {
+    const {getByText} = render(<Default data-testid="test-file-picker" />);
+    expect(getByText('Choose a file')).toBeDefined();
+  });
+  it('should render the file input in the DOM', () => {
```
shleewhite, line 10: Suggestion(non-blocking): this test could be combined with the should render test. You'd just check if the text is there and that the input is there.

packages/paste-core/components/file-picker/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,62 @@
...
+    render(<Default data-testid="test-file-picker" />);
+    expect(screen.getByTestId('test-file-picker')).toHaveAttribute('type', 'file');
+  });
+  it('should render as disabled', () => {
```
shleewhite, line 14: Suggestion(non-blocking): you might want to add a test that the required prop works

packages/paste-core/components/file-picker/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,62 @@
...
+      "font-family: 'Fira Mono',Courier,monospace"
+    );
+    expect(container.querySelector('[data-paste-element="FILEPICKER_BUTTON"]')).toHaveStyle(
+      'background-color: rgba(242, 47, 70, 0.1)'
+    );
+    expect(container.querySelector('[data-paste-element="FILEPICKER_TEXT"]')).toHaveStyle(
+      'margin: 0px 0px 0px 0.125rem'
+    );
+  });
+  it('should add custom styling to a custom named File Picker', () => {
```
shleewhite, line 50: Suggestion(non-blocking): I think we also usually test that the styles are applied if you add a custom element name too

packages/paste-core/components/file-picker/notes
```diff
@@ -0,0 +1,17 @@
+# File Picker
```
shleewhite, line 1: Note: just a reminder to remove this file when you're done

shleewhite (commented):
Great stuff!!! Love how the API has turned out. Just had some smaller issues/suggestions.

.changeset/late-tables-shop.md
```diff
@@ -0,0 +1,6 @@
+---
+'@twilio-paste/core': minor
```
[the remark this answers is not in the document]
nkrantz: I have core as minor and file picker as major here, how should it be?

packages/paste-core/components/file-picker/notes
nkrantz: ty

packages/paste-core/core-bundle/CHANGELOG_ARCHIVE.md
```diff
@@ -1,3 +1,1213 @@
+## 10.19.0
```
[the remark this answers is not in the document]
nkrantz: The changelog we show on the website is too large and it's messing with gatsby so I moved the oldest few versions to the archived file

.changeset/late-tables-shop.md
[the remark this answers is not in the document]
shleewhite: Oh 🤦🏽 I completely misread it. Sorry!

packages/paste-core/components/file-picker/__tests__/index.spec.tsx
shleewhite: Oh, I think I had it backwards. You don't have a test for adding the styles when you don't add a custom element name

nkrantz: That's the previous one `'should add custom styling to File Picker'`

kelseycavitt (commented):
I left a comment in the thread in plz, but I think we could refine the design just a tad, I also am not sure if we wanted the ability to "remove" the file, I provided a design idea in Slack.

packages/paste-core/components/file-picker/src/FilePicker.tsx
```diff
@@ -0,0 +1,110 @@
...
+  children: React.ReactElement;
+  disabled?: boolean;
+  i18nNoSelectionText?: string;
+  required?: boolean;
+}
+
+const FilePicker = React.forwardRef<HTMLInputElement, FilePickerProps>(
+  (
+    {
+      element = 'FILEPICKER',
```
gloriliale, line 20: suggestion: should we make this FILE_PICKER, since it's two words?

packages/paste-website/src/pages/components/file-picker/index.mdx
```diff
@@ -0,0 +1,257 @@
...
+
+#### FilePicker Props
+
+| Prop                 | Type                                        | Description                                                                                                                                                                                                                                                               | Default             |
+| -------------------- | ------------------------------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------------- |
+| accept?              | <inlineCode>string</inlineCode>             | Specify the desired file type. Note: file type should still be validated server-side, as this prop is not a complete validation. <Anchor href="https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/accept" showExternal>Read more about the accept prop</Anchor> | null                |
```
gloriliale, line 245: issue: we need the `element` prop in this table

packages/paste-website/src/pages/components/file-picker/index.mdx
```diff
@@ -0,0 +1,257 @@
...
+<LivePreview scope={{FilePicker, FilePickerButton, Label, HelpText, useUID}} language="jsx" noInline>
+  {defaultFilePicker}
+</LivePreview>
+
+## Guidelines
+
+### About File Picker
+
+Use a File Picker when users need to upload a document in a form flow. The File Picker allows any type of file by default, but the desired file type can be specified using the <inlineCode>accept</inlineCode> prop. This is a native HTML file picker, `<input type="file" />`, so there's currently no drag-and-drop functionality. <Anchor href="https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/file" showExternal>Read more about the native file picker on MDN.</Anchor> The input is hidden visually, but still appears in the DOM, so that we're able to apply custom styles in an accessible way.
+
```
gloriliale, line 91: suggestion: I tried to pick multiple files and I couldn't, which I think is what we want. Should we point that out here, that this file picker is for one file at a time?

packages/paste-core/components/file-picker/src/FilePicker.tsx
```diff
@@ -0,0 +1,110 @@
...
+      accept,
+      id = useUID(),
+      children,
+      disabled = false,
+      i18nNoSelectionText = 'No files selected',
```
gloriliale, line 25: suggestion: since we can only choose one file at a time, I think this should match that by being singular
```suggestion
      i18nNoSelectionText = 'No file selected',
```

packages/paste-core/components/file-picker/src/FilePicker.tsx
nkrantz: totally can, but I followed the date picker and time picker patterns of making it one word (prob should've split those in hindsight)

packages/paste-core/components/file-picker/src/FilePicker.tsx
nkrantz: agreed :) talked to Aren about language and we're going with "No file uploaded" and "Upload a file"

packages/paste-website/src/pages/components/file-picker/index.mdx
nkrantz: I can add a note about that to the docs!

packages/paste-core/components/file-picker/src/FilePicker.tsx
gloriliale: ah gotcha gotcha

gloriliale (approved):
Looks goooood to me!

gloriliale (dismissed):

kelseycavitt (approved):

TheSisb (approved):

gloriliale (approved):

TheSisb (dismissed):

gloriliale (approved):

packages/paste-website/src/pages/components/file-picker/index.mdx
```diff
@@ -0,0 +1,259 @@
...
+  <ListItem>
+    Include inline error text with the error icon on any field that errors to make it visually clear that the field
+    changed. You can also use{' '}
+    <Anchor href="/components/help-text#error-help-text">the "error" variant of Help Text</Anchor>.
+  </ListItem>
+  <ListItem>
+    If the File Picker has associated help text or error text, include the <inlineCode>aria-describedby</inlineCode>{' '}
+    prop on the File Picker. This should match the <inlineCode>id</inlineCode> of the help or error text.
+  </ListItem>
+</UnorderedList>
```
SiTaggart, line 120: suggestion: just for future reference, you don't need to use components to achieve this. Pure markdown would achieve the same thing.

```
- Include a visible [Label](/components/label) on **all** form fields.
- Each label must use the `htmlFor` prop that equals the `id` of the associated input.
```

packages/paste-website/src/pages/components/file-picker/index.mdx
```diff
@@ -0,0 +1,259 @@
+---
+title: File Picker
+package: '@twilio-paste/file-picker'
+description: A File Picker is a form element used to upload files..
```
serifluous, line 4: sorry, one more nitpick here: It has an extra period.

serifluous (approved):

SiTaggart (approved):

gloriliale (approved):

PR #2676, opened 2022-09-12
feat: add new Side Modal component

shleewhite:
## This PR includes:
* new @twilio-paste/side-modal package, which includes
  * SideModal
  * SideModalHeader
  * SideModalHeading
  * SideModalBody
  * SideModalFooter
  * SideModalFooterActions
  * SideModalContainer
  * SideModalButton
  * useSideModalContext
* MinimizableDialog
  * fixes the generic HTMLDiv type extension
  * add missing prop types
  * export all of the component interfaces
  * use Box instead of a styled.div for the StyledMinimizableDialog
* Modal
  * spread props with safelySpreadBox props on ModalFooterActions
  * use column gap instead of conditional margin in ModalFooterActions
* Code Block
  * add missing peer/dev dependencies
* Box
  * add support for the inset CSS prop
* Style Props
  * add support for the inset CSS prop 

## To do
- [x]  docs page

## Open questions
* I'm not sure why the dangerJS check is failing for unpinned dependencies. Should I add React/PropTypes/React-dom to the allow list?

**Contributing to Twilio**


packages/paste-core/components/minimizable-dialog/package.json
```diff
@@ -26,10 +26,11 @@
   },
   "peerDependencies": {
     "@twilio-paste/box": "^7.0.0",
+    "@twilio-paste/button": "^11.1.1",
     "@twilio-paste/design-tokens": "^8.0.0",
     "@twilio-paste/icons": "^9.0.0",
     "@twilio-paste/non-modal-dialog-primitive": "^0.3.10",
-    "@twilio-paste/popover": "^10.0.0",
+    "@twilio-paste/screen-reader-only": "^10.0.0",
```
shleewhite, line 33: Note: popover isnt used in the Minimizable Dialog package, but screen reader only is.

packages/paste-core/components/modal/__tests__/customization.spec.tsx
```diff
@@ -54,11 +54,6 @@ describe('Modal Customization', () => {
 
       const modalFooterActions = modalFooter.firstChild as HTMLElement;
       expect(modalFooterActions.getAttribute('data-paste-element')).toEqual('MODAL_FOOTER_ACTIONS');
-
-      const modalFooterActionItemOne = modalFooterActions.firstChild as HTMLElement;
-      const modalFooterActionItemTwo = modalFooterActions.lastChild as HTMLElement;
-      expect(modalFooterActionItemOne.getAttribute('data-paste-element')).toEqual('MODAL_FOOTER_ACTIONS_ITEM');
-      expect(modalFooterActionItemTwo.getAttribute('data-paste-element')).toEqual('MODAL_FOOTER_ACTIONS_ITEM');
```
SiTaggart, line 61: issue: Unfortunately, the removal of this customization element target is a breaking change.

packages/paste-core/components/modal/src/ModalFooterActions.tsx
```diff
@@ -10,29 +10,22 @@ export interface ModalFooterActionsProps {
...
-        flexShrink={justify === 'start' ? null : 0}
-        flexWrap="wrap"
-        flexGrow={1}
-        element={element}
-        alignItems="center"
-        ref={ref}
-      >
-        {React.Children.map(children, (child, index) => {
-          return (
-            <Box marginRight={count !== index + 1 ? 'space50' : null} element={`${element}_ITEM`}>
```
SiTaggart, line 28: issue: unfortunately, the remove of this styling hook for a customer is a breaking change. If a customer is relying on that hook, we might break their app. We should be very conscious about whether we need to remove styling hooks from components.

packages/paste-core/components/side-modal/src/SideModal.tsx
```diff
@@ -0,0 +1,71 @@
...
+    transform: 'none!important',
+    inset: `0rem 0rem 0rem auto!important`,
+    ['&:focus']: {
+      outline: 'none',
+    },
+    ['div:first-child']: {
+      height: '100%',
+    },
+  })
+);
```
SiTaggart, line 25: question: I'm being somewhat lazy, what is not possible here with `Box`? Should it be something that `Box` supports?

packages/paste-core/components/side-modal/src/SideModalButton.tsx
```diff
@@ -0,0 +1,31 @@
+import * as React from 'react';
+import * as PropTypes from 'prop-types';
+import {Button} from '@twilio-paste/button';
+import {NonModalDialogDisclosurePrimitive} from '@twilio-paste/non-modal-dialog-primitive';
+import {SideModalContext} from './SideModalContext';
+import type {ButtonProps} from '@twilio-paste/button';
```
SiTaggart, line 6: nit: usually package imports come before relative imports

packages/paste-core/components/side-modal/src/SideModalFooterActions.tsx
```diff
@@ -0,0 +1,57 @@
+import * as React from 'react';
+import * as PropTypes from 'prop-types';
+import {Box, safelySpreadBoxProps} from '@twilio-paste/box';
+import type {BoxElementProps, BoxStyleProps} from '@twilio-paste/box';
+
+export interface SideModalFooterActionsProps extends React.HTMLAttributes<HTMLDivElement> {
+  children: NonNullable<React.ReactNode>;
+  element?: BoxElementProps['element'];
+  justify?: BoxStyleProps['justifyContent'];
```
SiTaggart, line 9: question: of this full list of all flex justify options, do you have opinions on whether some of these _shouldn't_ be used because they might "look bad"? Should we restrict to a collection of options that we as a design team think encourages good design?

packages/paste-core/components/side-modal/src/SideModalHeading.tsx
```diff
@@ -0,0 +1,29 @@
+import * as React from 'react';
+import * as PropTypes from 'prop-types';
+import {Heading} from '@twilio-paste/heading';
+import type {HeadingProps} from '@twilio-paste/heading';
+import type {BoxElementProps} from '@twilio-paste/box';
+
+export interface SideModalHeadingProps extends Omit<HeadingProps, 'marginBottom' | 'variant' | 'as'> {
+  children: NonNullable<React.ReactNode>;
+  as?: 'h1' | 'h2' | 'h3' | 'h4' | 'h5' | 'h6';
```
SiTaggart, line 9: praise: good continuation of API design from other, similar components. Build familiarity with using the system

packages/paste-core/components/side-modal/src/SideModal.tsx
shleewhite: Box doesn't support inset yet - this came up with minimizable dialog too but I was also being lazy and didn't want to add it.

packages/paste-core/components/side-modal/src/SideModalFooterActions.tsx
shleewhite: This came up from a talk with @TheSisb. The way modal does it, theres only 2 options 'start' and 'end' which map to 'flex-start' and 'flex-end'. This felt weird because it is renaming the options so people dont actually know what theyre getting... Evergreen also has a use case for 'space-between' where there is a back button on the left and the save/cancel button on the right.

I'd be fine reducing the options to just flex-start/end and space-between.

shleewhite: Would it be weird to also rename the prop justifyContent? That would be a change from how modal does it but I think its a better practice.

packages/paste-core/components/modal/src/ModalFooterActions.tsx
shleewhite: @TheSisb and I talked about this too.... This is the wrong way to space the elements inside the FooterActions. 

Is there a happy medium where we keep the iterating over children part to add the element but also use columnGap instead of marginRight?

TheSisb: Oh, maybe I mispoke but let's leave Modal as is but do the right way in SideModal for now

packages/paste-core/components/side-modal/src/SideModal.tsx
SiTaggart: What's the "it"? Depending on the "it" it may or may not be worthwhile adding. They aren't super hard to add at this stage either, so if it's something that' pretty useful,  we should try and do it.

packages/paste-website/src/pages/components/side-modal/index.mdx
```diff
@@ -0,0 +1,364 @@
+---
+title: Side Modal
+package: '@twilio-paste/side-modal'
+description: An accessible modal dialog that sits on the right side of the page.
```
SiTaggart, line 4: suggestion: for descriptions here and on the package.json file we tend to follow the following pattern

> A(n) ${ComponentName} is a ...

"A Card is a styled container that groups related content and actions."
"A Popover is a page overlay triggered by a click that displays additional interactive content."
"A Modal is a page overlay that displays information and blocks interaction with the page until an action is taken or the Modal is dismissed."

packages/paste-website/src/pages/components/side-modal/index.mdx
```diff
@@ -0,0 +1,364 @@
...
+    SideModalHeading,
+  }}
+  noInline
+>
+  {defaultExample}
+</LivePreview>
+
+### Side Modal with a footer
+
+By default the footer actions are aligned to the right of the modal using the Flexbox property `justify-content`. This should be the most common configuration you should choose.
```
SiTaggart, line 156: suggestion: as it's a paired down API, you might consider just listing the alternative values and what they do here too.

packages/paste-core/components/side-modal/src/SideModal.tsx
shleewhite: Ok! ill add inset to box

packages/paste-website/src/pages/components/side-modal/index.mdx
shleewhite: How is this?
> A Side Modal is a page overlay that displays information that does not block the user from interacting with the rest of the page.

SiTaggart: I buy that 👍🏼

kelseycavitt (approved):

SiTaggart (approved):

PR #2677, opened 2022-09-13
chore: GH workflow for Percy baseline creation on main

SiTaggart:
## Create `main` baselines

Enables Percy to run via cypress on main, every time we push to main. 

Percy will auto approve the baselines create on main, so that we have up to date baselines to compare with on every PR.

## Stablize snapshots

- Create a utility in the website to detect running in Cypress
- Disable the animation in the one chart we display using the `inCypress` util
- Disable animations that use our animation library based on being in cypress or not
- set the clock for all cypress runs so that anything using `Date now()` never changes on different days


packages/paste-website/src/components/site-wrapper/site-header/SiteHeaderDesktop.tsx
```diff
@@ -34,6 +35,7 @@ const SiteHeaderDesktop: React.FC = () => {
         width="100%"
       >
         <DarkModeToggle />
+        <Box color="colorTextInverse">{inCypress() ? 'Cypress' : 'Not Cypress'}</Box>
```
TheSisb, line 38: question: what is this for?

cypress/support/commands/parent-commands.ts
```diff
@@ -108,6 +108,10 @@ Cypress.Commands.add('visualRegressionTestUrl', ({url, testName}) => {
       ...DEFAULT_VRT_OPTIONS,
     };
 
+    // Freeze the date for any date time diffs
+    const now = new Date(2018, 1, 1);
```
TheSisb, line 112: question: what does this fix? Mind adding that reference inline in a comment for future maintainers please?

.github/workflows/on_merge_to_main_cypress.yml
```diff
@@ -0,0 +1,94 @@
+name: Merge to main Cypress for VRT Baseline
+
+on:
+  push:
+    branches:
+      - main
+
+jobs:
+  # run this job first to create a single timestamped id for all percy batches on parallel machines
+  percy-batch-id:
```
TheSisb, line 10: question: why didn't we add all this to `on_merge_to_main.yml`? Is it because of this percy batch id job?

SiTaggart: Hmmm I think I naively followed the pattern from the existing cypress workflow, but in hindsight that's only done because of the events different from `on_pull_request`.

Probs will put it in `on_merge_to_main`

cypress/support/commands/parent-commands.ts
SiTaggart: Comment updated

packages/paste-website/src/components/homepage/HomeHeroIllustration.tsx
```diff
@@ -32,7 +33,7 @@ const IllustrationChildren: React.FC<{state: IllustrationStatesType}> = ({state}
  * - If prefers reduced motion, render static.
  */
 const HomeHeroIllustration: React.FC = () => {
-  const prefersReducedMotion = useReducedMotion();
+  const prefersReducedMotion = useReducedMotion() || inCypress();
```
SiTaggart, line 36: We use lottie and not our animation library, so disabling animations on the theme provider does nothing on the homepage

TheSisb: suggestion: mind leaving the above comment inline in the code? Annoying suggestion I know, but it will make a future maintainer happy.

TheSisb (approved):

packages/paste-website/src/components/homepage/HomeHeroIllustration.tsx
SiTaggart: done

TheSisb (approved):

shleewhite (approved):

PR #2681, opened 2022-09-19
feat(avatar): add new variant and color props

shleewhite:
In this PR:
* add variant prop which has the options: 'user', 'entity'. It defaults to 'user'
* add color prop which has the options: 'default', 'decorative10', 'decorative20', 'decorative30', 'decorative40'. It defaults to 'default'
* update snapshot tests
* updates the website content to include color and variant props

**Contributing to Twilio**

### DIFF
--- a/packages/paste-core/components/avatar/src/index.tsx
+++ b/packages/paste-core/components/avatar/src/index.tsx
@@ -6,7 +6,7 @@ import {Box, safelySpreadBoxProps} from '@twilio-paste/box';
 import type {BoxStyleProps} from '@twilio-paste/box';
 import {isIconSizeTokenProp} from '@twilio-paste/style-props';
 import {getComputedTokenNames, getInitialsFromName} from './utils';
-import type {AvatarProps, AvatarContentProps, ColorVariants, ShapeVariants} from './types';
+import type {AvatarProps, AvatarContentProps, ColorVariants, AvatarVariants} from './types';
 
 const DEFAULT_SIZE = 'sizeIcon70';
 
@@ -65,7 +65,7 @@ const colorVariants: Record<ColorVariants, BoxStyleProps> = {
   },
 };
 
-const shapeVariants: Record<ShapeVariants, BoxStyleProps> = {
+const variants: Record<AvatarVariants, BoxStyleProps> = {
   user: {
     borderRadius: 'borderRadiusCircle',
   },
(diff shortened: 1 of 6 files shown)


TheSisb (dismissed):

packages/paste-core/components/avatar/src/index.tsx
```diff
@@ -41,8 +42,43 @@ const AvatarContents: React.FC<AvatarContentProps> = ({name, size = DEFAULT_SIZE
...
+    backgroundColor: 'colorBackgroundDecorative30Weakest',
+    color: 'colorTextDecorative30',
+  },
+  decorative40: {
+    backgroundColor: 'colorBackgroundDecorative40Weakest',
+    color: 'colorTextDecorative40',
+  },
+};
+
+const shapeVariants: Record<ShapeVariants, BoxStyleProps> = {
```
nkrantz, line 68: suggestion (non-blocking): I know we want to be transparent about which styles the prop will affect, but could the prop be called 'variant'? Not blocking but it feels weird to me to have `shape="user"`

nkrantz (approved):

packages/paste-core/components/avatar/src/index.tsx
shleewhite: I don't disagree, I went with what was in the ticket - but I'll bring it up in standup tomorrow!

SiTaggart: mmmm, OK, yeah I can see that.

Essentially what happened is a few of us were discussing variants, and shoving too much into `variant` for Avatars. For a time is was just "add the square variant", then "add the decorative options like Badge". But square for entities and circles for users, could all be decorative colors and we didn't want 10 variants.

So we tried to abstract some overlapping stuff out of variant and _maaaaaaybe_ got carried away.

`user` and `entity` could probably be fine as variants, they could also be fine remaining as shapes.

Great call out though @nkrantz

packages/paste-core/components/avatar/stories/index.stories.tsx
```diff
@@ -84,8 +89,25 @@ export const Icon = (): React.ReactNode => {
...
+      <Avatar color="decorative40" name="avatar example" icon={UserIcon} />
+    </Stack>
+  );
+};
+
+export const ShapeVariants = (): React.ReactNode => {
+  return (
+    <Stack orientation="horizontal" spacing="space40">
+      <Avatar shape="user" name="avatar example" icon={UserIcon} />
+      <Avatar shape="entity" name="avatar example" icon={BusinessIcon} />
```
SiTaggart, line 108: suggestion: it might be a good addition to also have a User and Entity Image avatar to cover the shape when it's also an image background

SiTaggart (dismissed):

packages/paste-core/components/avatar/stories/index.stories.tsx
shleewhite: What would be an example of an entity image?

SiTaggart (dismissed):

packages/paste-core/components/avatar/stories/index.stories.tsx
SiTaggart: For the purpose of VRT, any old image would be just fine. You can even just reuse one of the human images if you didn't want to source one.

SiTaggart (approved):

kelseycavitt (approved):

PR #2682, opened 2022-09-20
build: add React18 Support

andreasonny83:
Update packages dependencies to include React v18

**Contributing to Twilio**


TheSisb:
Hey @andreasonny83, thanks for taking this initiative!

Heads up, our `@twilio-paste/uid-library` [may fail in React 18](https://github.com/thearnica/react-uid#readme): 

image


The good news is that we wrap `react-uid` in our own `uid-library` package, so we can fix this is a non-breaking patch change by [changing what we export](https://github.com/twilio-labs/paste/blob/main/packages/paste-libraries/uid/src/index.tsx) to something like this:

```js
const useUID = React.useId; // Use react 18's useId() hook
const useUIDSeed = (item: any) => React.useId(); // Function that calls React useId();
const UIDFork = ({children}) => <>{children}</>; // Do nothing
const uid = (item: any, index?: number) => React.useId(); // Just use React.useId() again
```

I _think_ this would just work, but we'd need to verify in tests and our website/storybook builds.

packages/paste-core/components/input-box/src/InputBox.tsx
```diff
@@ -21,7 +21,7 @@ export interface InputBoxProps {
   variant?: Variants;
 }
 
-const InputBox = React.forwardRef<HTMLDivElement, InputBoxProps>(
+const InputBox = React.forwardRef<HTMLDivElement, React.PropsWithChildren<InputBoxProps>>(
```
espipj, line 24: As we're using React.Props**WithChildren** InputBox will have already the children prop. No need to define children on the InputBoxProps L11

espipj (changes requested):
Prefix and Suffix files also need changing in L17 to be:
`const Prefix = React.forwardRef<HTMLDivElement, React.PropsWithChildren<PrefixProps>>(
`
path: `packages/paste-core/components/input-box/src/Prefix.tsx`

Still a problem in props like `insertAfter` in InputBox....

espipj (changes requested):
Btw running `yarn install` with a clean project has updated my `yarn.lock` file... which I guess is because it needs to be updated in the repo too... 😃

TheSisb:
We need to upgrade our underlying Modal library for React 18 support https://github.com/reach/reach-ui/issues/972

PR #2683, opened 2022-09-20
fix: react 16 types

andreasonny83:
Remove extra React-16 from type declaration files

After adding support for React 17, as part of #1965, a new `react-16` dependency has been introduced into the project for running specific test case scenarios against both React 16 and 17. Because of that, when building the distribution version of the package, an extra react-16 type declaration is added to the generated type declaration files. This is because we now have 2 React namespaces declared.

screenshot_2022-09-20_at_15 32 33

**Contributing to Twilio**


TheSisb:
This is close, but not quite right because [our jest.config.js file remaps `react` to `react-16`](https://github.com/twilio-labs/paste/blob/main/jest.config.js#L46) but this doesn't install a `react-16` package anymore.

I think a better approach, but similar to what you've done, is to have a specific `test:react-16` command in the package.json that does something like [how Ariakit does it.](https://github.com/ariakit/ariakit/blob/main/package.json#L6) Then we can also remove the remapping in the jest.config.json too.

TheSisb:
For tests to run successfully, I've taken over in https://github.com/twilio-labs/paste/pull/2684

PR #2684, opened 2022-09-20
fix: don't bundle react-16 types, only install for testing in CI

TheSisb:
Taking over https://github.com/twilio-labs/paste/pull/2683 so that Github actions run properly.


andreasonny83 (approved):

TheSisb:
Ignore the dangerjs check. I fix it in another PR to not worry about the root package.json file.

espipj (approved):
LGTM great effort @andreasonny83

.github/workflows/on_pull_request.yml
```diff
@@ -229,27 +229,24 @@ jobs:
...
-        uses: actions/cache@v2
-        id: node_modules_cache_id
-        with:
-          path: |
-            node_modules
-            */*/node_modules
-          key: ${{ runner.os }}-${{ hashFiles('**/yarn.lock') }}
-
       - name: Install Dependencies
-        if: steps.yarn_cache_id.outputs.cache-hit != 'true' || steps.node_modules_cache_id.outputs.cache-hit != 'true'
```
SiTaggart, line 242: suggestion: we might want to leave a little inline comment here reminding us why we don't want the cached node modules for this job. Otherwise we might forget and think it was a mistake.

andreasonny83: great point. Comment left

PR #2706, opened 2022-09-29
feat(sticky-token-filters): add support for sticky filters

nkrantz:
**This is a replacement PR for #2602 because the checks were failing**

## Update summary
* Updates Tokens Page filters to be sticky on the page.
* Uses IntersectionObserver to apply shadow on filters.
* Resolves issue with ToC nav on Tokens Page losing its position once the header is off screen.
* Adds support for route & breakpoint based scroll offset values.

### Additional notes
* Using a placeholder element ("filter-canary") to interact w/ IntersectionObserver since sticky position never triggers the events we need. 
* Setting a height on the IntersectionObserver canary allows us to fade in opacity of the shadow as the user scrolls down.
* Shadow is built using a pseudo element since applying a box-shadow to the whole element will bleed over to the sides. 
* Using negative margins and z-indicies to cover up the top and sides of the shadow pesudo.

### DIFF
--- a/packages/paste-website/src/components/shortcodes/PageAside.tsx
+++ b/packages/paste-website/src/components/shortcodes/PageAside.tsx
@@ -1,5 +1,6 @@
 import * as React from 'react';
 import {Box} from '@twilio-paste/box';
+import type {BoxProps} from '@twilio-paste/box';
 import {TableOfContents} from './table-of-contents';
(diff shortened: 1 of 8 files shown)


loreina (dismissed):
praise: lovin the sticky filter!!

TheSisb (approved):

PR #2728, opened 2022-10-11
chore: swap underlying Stack functionality to css gap, disabled wrap, add justify content control

TheSisb:
This PR refactors the Stack component so that it uses `columnGap` and `rowGap` instead of using `margin`. This approach is an improvement in terms of code cleanliness as well as fixing the margin issue on RTL locales.

This PR also removes the hardcoded `flexWrap: wrap` on Stack because that behavior is typically unintended and makes building responsive behaviors troublesome.


packages/paste-core/layout/stack/__tests__/stack.test.tsx
```diff
@@ -120,18 +125,6 @@ describe('Stack Customization', () => {
...
-  it('should set the data-paste-attribute attribute for Stack Children to "STACK_CHILD"', (): void => {
-    const {container} = render(
-      <CustomizationProvider baseTheme="default">
-        <Stack orientation="vertical" spacing="space0">
-          <p>One</p>
-          <p>Two</p>
-        </Stack>
-      </CustomizationProvider>
-    );
-    expect(container.querySelector('[data-paste-element="STACK_CHILD"]')).toBeInTheDocument();
```
SiTaggart, line 132: issue: This is technically a public styling API. We can't really take it away anymore, that contract is set for customer apps and their custom styling.

SiTaggart (commented):
I think the compromise here is going to be keeping the child div so that we:

- maintain the API contract for customers styling those div via customization
- Minimize effects on the styling or layout of children, think buttons suddenly becoming full with in vertical stacks as is caught in the VRT.

whilst also keeping the column and row gap implementation to solve for RTL etc.

PR #2737, opened 2022-10-14
chore: update create package template

gloriliale:
### Template
New components are created using the `create:package` script that uses `plop` templates. This PR updates the template to include:
- The `element` prop necessary for customization
- Giving the `element` prop its default value
- A new way to specify prop Types. 

### Button Group
Our system uses `props` as the convention for a component's props, this component is updated for consistency. ~~Since it's an entirely internal name change, we don't need a Changeset.~~ Changeset now included!


tools/plop-templates/component-index.hbs
```diff
@@ -1,11 +1,13 @@
 import * as React from 'react';
 import * as PropTypes from 'prop-types';
+import type {BoxProps} from '@twilio-paste/box';
```
gloriliale, line 3: Adding this means that every component has a dependency on `@twilio-paste/box`. However,[ in this PR](https://github.com/twilio-labs/paste/pull/2590#discussion_r942886603), we eliminated adding Paste dependencies to the package.json template. What's the reason that we don't want internal dependencies listed in the template?

tools/plop-templates/component-index.hbs
```diff
@@ -1,11 +1,13 @@
...
 
-export interface {{pascalCase component-name}}Props {
+export interface {{pascalCase component-name}}Props extends Omit<React.ComponentPropsWithRef<'div'>, 'children'> {
```
gloriliale, line 5: We noticed that we're not typing our components super specifically. We started using `extends Partial<Omit<HTMLDivElement, 'children'>>` in response. Recently, we used` extends Omit<React.ComponentPropsWithRef<'div'>, 'children'>` in the ButtonGroup package.

I would really like to get some clarity on:
- whether we even want to be this specific in our props(@TheSisb pointed out potential usability issues)
- which of these methods makes more sense for our system, if any
- which to include here in the template we're using for future components

SiTaggart: For consumers to make accessible applications, we have to extend the HTML element props that we expose to the consumer via the ref and where we spread our props onto. An Input has to reflect an input, a Select a select, etc. Our components really need to act like HTML elements, because underneath, that's all they are.

I really enjoyed the explanations in [this link](https://react-typescript-cheatsheet.netlify.app/docs/advanced/patterns_by_usecase/#wrappingmirroring-a-html-element) that @jb-twilio sent me on the subject, especially the "Why not ComponentProps or IntrinsicElements or [Element]HTMLAttributes or HTMLProps or HTMLAttributes" callout disclosure. Super good read.

tools/plop-templates/component-index.hbs
gloriliale: Since @SiTaggart said there was no reason not to, and compelling reasons to add them, I've added all the internal peer and dev dependencies Box has, as well as Box itself, to the template

packages/paste-core/primitives/box/package.json
```diff
@@ -29,6 +29,7 @@
   },
   "peerDependencies": {
     "@twilio-paste/animation-library": "^0.3.2",
+    "@twilio-paste/box": "^7.0.0",
```
TheSisb, line 32: Why does the box package depend on itself? lol Did you mean to add this to the hbs instead? https://github.com/twilio-labs/paste/pull/2737/files#diff-a68998e6714d0c1caeee57cc875a41564aff96679f3bee3a10b70aa1c596fc83R29

gloriliale: ugh yes 🙃 thanks for catching this!

tools/plop-templates/component-index.hbs
TheSisb: That's a good idea @jb-twilio - can you make a ticket for it pls?

SiTaggart (approved):

shleewhite (approved):

PR #2745, opened 2022-10-14
fix(theme): add missing Partial around iconSizes bucket

TheSisb:
Add missing `Partial<>` around iconSizes in theme typings

### DIFF
--- a/packages/paste-theme/src/types/GenericThemeShape.ts
+++ b/packages/paste-theme/src/types/GenericThemeShape.ts
@@ -77,7 +77,7 @@ export interface GenericThemeShape {
       [key in SizingKeys]: any;
     }
   >;
-  iconSizes: {
+  iconSizes: Partial<{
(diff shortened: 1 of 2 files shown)


packages/paste-theme/src/types/GenericThemeShape.ts
```diff
@@ -77,7 +77,7 @@ export interface GenericThemeShape {
       [key in SizingKeys]: any;
     }
   >;
-  iconSizes: {
+  iconSizes: Partial<{
```
SiTaggart, line 80: I really hope I didn't do this for an undocumented reason that I can no longer remember. 😟

SiTaggart (approved):

packages/paste-theme/src/types/GenericThemeShape.ts
TheSisb: Me too, but this change shouldn't be harmful anyways I don't think? I hope at least

PR #2746, opened 2022-10-14
add RadioButtonGroup package

jb-twilio:
**Contributing to Twilio**


packages/paste-core/components/radio-button-group/package.json
```diff
@@ -0,0 +1,69 @@
+{
+  "name": "@twilio-paste/radio-button-group",
+  "version": "0.0.0",
+  "category": "user input",
+  "status": "production",
+  "description": "A Radio Button Group is a form element that lets users select a single choice from a list of at least two options.",
```
nkrantz, line 6: this is the description in the spec: Radio Button Group is a set of related, mutually exclusive options where only one can be selected at a time.

gloriliale: updated!

packages/paste-core/components/radio-button-group/src/RadioButtonGroup.tsx
```diff
@@ -0,0 +1,129 @@
...
+//import type {BoxProps} from '@twilio-paste/box';
+import type {InlineControlGroupProps} from '@twilio-paste/inline-control-group';
+import {HelpText} from '@twilio-paste/help-text';
+import {Label} from '@twilio-paste/label';
+
+import {RadioButtonContext} from './RadioButtonContext';
+import {AttachedRadioButtonGroupWrapper, UnattachedRadioButtonGroupWrapper} from './StyledRadioButtonGroupWrapper';
+
+/*
+NOTE: not sure why this was added originally. seems we use the import on line 5 above?
```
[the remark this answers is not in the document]
shleewhite: This was temp, I initially copied this whole file from RadioGroup. When I realized I couldn't use the inline group component anymore, I copied the interface as a reference to know what props I should support! Sorry for the confusion. Should totally be deleted.

packages/paste-core/components/radio-button-group/src/RadioButtonGroup.tsx
```diff
@@ -0,0 +1,133 @@
...
+}
+*/
+
+export interface RadioButtonGroupProps extends InlineControlGroupProps {
+  name: string;
+  onChange?: (value: string) => void; // the context has this as taking in an event, is that ok?
+  value?: string;
+  i18nRequiredLabel?: string;
+  attached?: boolean;
+  // ref?: HTMLFieldSetElement;
```
shleewhite, line 38: Nit: can remove this prop

packages/paste-core/components/radio-button-group/src/RadioButtonGroup.tsx
```diff
@@ -0,0 +1,133 @@
...
+  orientation?: 'vertical' | 'horizontal';
+  ref?: HTMLFieldSetElement;
+  required?: boolean;
+  i18nRequiredLabel?: string;
+}
+*/
+
+export interface RadioButtonGroupProps extends InlineControlGroupProps {
+  name: string;
+  onChange?: (value: string) => void; // the context has this as taking in an event, is that ok?
```
shleewhite, line 34: ^ to your comment, yeah I think this should use event like the context.

packages/paste-core/components/radio-button-group/src/RadioButtonGroup.tsx
```diff
@@ -0,0 +1,133 @@
...
+      // Do we need to spread this contextValue object here?
+      <RadioButtonContext.Provider value={contextValue}>
+        <Box
+          as="fieldset"
+          element={element}
+          margin="space0"
+          padding="space0"
+          border="none"
+          display="flex"
+          flexDirection="column" // why is this column not row?
```
shleewhite, line 89: The column here is the label + help text + radio buttons + error text. Here is an example of the column layout:
image

packages/paste-core/components/radio-button-group/src/RadioButtonGroup.tsx
```diff
@@ -0,0 +1,133 @@
...
+          rowGap="space40"
+          ref={ref}
+          {...safelySpreadBoxProps(props)}
+        >
+          <Label
+            as="legend"
+            element="LEGEND"
+            htmlFor={undefined} // I know this is in inline control group but, why do we do this?
```
shleewhite, line 97: So, fieldsets don't use label the way inputs do. Fieldsets require there to be a legend tag that is the first child of the fieldset (see: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/fieldset)

It relies on the legend being the first child instead of the HTMLFor attribute. If the types don't yell about a missing htmlFor prop, you can remove this line.

packages/paste-core/components/radio-button-group/src/RadioButtonGroup.tsx
```diff
@@ -0,0 +1,133 @@
...
+        value: value || '',
+        disabled,
+        hasError: errorText != null,
+        onChange: onChangeHandler,
+      };
+    }, [name, value, disabled, errorText, onChangeHandler]);
+
+    return (
+      // Do we need to spread this contextValue object here?
+      <RadioButtonContext.Provider value={contextValue}>
```
shleewhite, line 81: Not necessarily... the difference between this: `value={contextValue}` and `value={{..contextValue}}` is that `value={{..contextValue}}` makes a new object with the same key/values as contextValue. This is copy pasta from RadioGroup so I think its fine.

packages/paste-core/components/radio-button-group/src/RadioButtonGroup.tsx
gloriliale: ahhh okay, thank you for pointing this out!

packages/paste-core/components/radio-button-group/src/RadioButtonGroup.tsx
gloriliale: Ha, yell it does 🙃 . Thanks for explaining tho.

packages/paste-core/components/radio-button-group/src/StyledRadioButtonLabel.tsx
```diff
@@ -0,0 +1,16 @@
...
+    '& > input[type=radio]:checked + span': {
+      ...ToggleButtonStyles._pressed,
+    },
+    '& > input[type=radio]:checked + span:hover': {
+      ...ToggleButtonStyles._pressed_hover,
+    },
+    '& > input[type=radio]:checked:focus + span': {
+      ...ToggleButtonStyles._pressed_focus,
+    },
+  })
```
TheSisb, line 15: question (blocking): rather than this sibling styling CSS, couldn't it be possible to pass `pressed` directly to the Button on line 112 (https://github.com/twilio-labs/paste/pull/2746/files#diff-a521b250aee50a0f703299853e41f573a04161259b607d498ebe3aec5e408cf4R112) based on the state we already have and are passing to the HiddenRadio??

packages/paste-core/components/radio-button-group/src/StyledRadioButtonGroupWrapper.tsx
```diff
@@ -0,0 +1,44 @@
+import {styled, css} from '@twilio-paste/styling-library';
+
+export const AttachedRadioButtonGroupWrapper = styled.span(
+  css({
+    // NOTE: allow focus indication to remain visible on all sides of a button despite
+    // sibling elements being rendered immediately adjacent under the attached styles.
+    // Not applied to hover for consistent styling with borders.
+
+    '& > label:focus-within': {
```
TheSisb, line 9: question (non-blocking): why can't we re-use/compose ButtonGroup in RadioButtonGroup to apply the styling - how is this different?

packages/paste-core/components/radio-button-group/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,324 @@
...
+import {Attached} from '../stories/index.stories';
+import {RadioButton, RadioButtonGroup} from '../src';
+import type {RadioButtonProps} from '../src';
+
+describe('RadioButton', () => {
+  const renderRadioButton = (props: Partial<RadioButtonProps> = {}): ReturnType<typeof render> =>
+    // eslint-disable-next-line react/no-children-prop
+    render(<RadioButton value="foo" name="foo" children="foo" {...props} />);
+
+  it('renders visibly into document', () => {
```
TheSisb, line 13: suggestion: we can delete this, it's redundant.

packages/paste-core/components/radio-button-group/src/StyledRadioButtonLabel.tsx
shleewhite: I don't think we can do that because the pressed state uses the `_pressed` pseudo selector which won't happen since this uses `checked` attribute on the hidden radio, not `pressed` attribute on the Button as span.

packages/paste-core/components/radio-button-group/src/StyledRadioButtonGroupWrapper.tsx
shleewhite: Oh! good point, this work started before ButtonGroup was finished. It can totally just re-export RadioGroup.

packages/paste-core/components/radio-button-group/src/StyledRadioButtonLabel.tsx
gloriliale: tldr; it's more accessible for it to be styled this way, but we'd like to avoid nested CSS in the future if at all possible.

packages/paste-core/components/radio-button-group/src/StyledRadioButtonGroupWrapper.tsx
gloriliale: these are different enough that they should be separate (span vs label), so we'll leave as is.

libertycarlin (approved):

PR #2748, opened 2022-10-17
chore(button-group): add toggle button story

nkrantz:
Button group package is missing a story for toggle buttons. Here it is.

### DIFF
--- a/packages/paste-core/components/button-group/stories/index.stories.tsx
+++ b/packages/paste-core/components/button-group/stories/index.stories.tsx
@@ -66,6 +66,47 @@ export const PopoverButtons: Story = () => {
   );
 };
 
+export const ToggleButtons: Story = () => {
+  const [boldPressed, setBoldPressed] = React.useState(false);
+  const [italicPressed, setItalicPressed] = React.useState(false);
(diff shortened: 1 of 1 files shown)


packages/paste-core/components/button-group/stories/index.stories.tsx
```diff
@@ -66,6 +66,25 @@ export const PopoverButtons: Story = () => {
   );
 };
 
+export const ToggleButtons: Story = () => {
+  const [boldPressed, setBoldPressed] = React.useState(false);
+  const [italicPressed, setItalicPressed] = React.useState(false);
```
shleewhite, line 71: suggestion(non-blocking): could be nice to have one of the buttons be pressed by default for VRT.

loreina: suggestion: could we also add disabled and selected disabled states?

shleewhite (approved):

SiTaggart (approved):

loreina (approved):
approved, and there'll be a fast follow fix for adding a box shadow to the pressed disabled button state

PR #2749, opened 2022-10-18
fix(ButtonGroup): customization tests 

shleewhite:
Currently, the customization tests use `toHaveStyle` instead of `toHaveStyleRule`. If you change the value, `toHaveStyle` always passes - only `toHaveStyleRule` actually works for our components. For example, there was currently a typo in the `toHaveStyle` rule and the test still passed.

**Contributing to Twilio**

### DIFF
--- a/packages/paste-core/components/button-group/__tests__/index.spec.tsx
+++ b/packages/paste-core/components/button-group/__tests__/index.spec.tsx
@@ -4,7 +4,7 @@ import {Button} from '@twilio-paste/button';
 import {CustomizationProvider} from '@twilio-paste/customization';
 
 import {ButtonGroup} from '../src';
-import {Attached, Unattached} from '../stories/index.stories';
+import {Unattached} from '../stories/index.stories';
 
 describe('ButtonGroup', () => {
   it('should render a group visibly into document', () => {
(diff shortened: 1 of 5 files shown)


TheSisb:
Do we need to change this in other packages too? I see it in alert-dialog, combobox, file-picker, and website tokencard tests

shleewhite:
> Do we need to change this in other packages too? I see it in alert-dialog, combobox, file-picker, and website tokencard tests

Oh! Yup we need to change it in all of them. @gloriliale found that if you change the value in `toHaveStyle` that is incorrect, the test still passes.

gloriliale (approved):
Woohoo, thanks Lee!

packages/paste-core/components/button-group/__tests__/index.spec.tsx
```diff
@@ -22,18 +22,6 @@ describe('ButtonGroup', () => {
     }
   });
 
-  it('should accept attached prop and apply styles', () => {
```
shleewhite, line 25: Note: I ended up removing these tests because they are tested by Chromatic and because jest was having issues with the nested styles for the middle button in the group.

gloriliale (dismissed):

SiTaggart (approved):

TheSisb (approved):

PR #2757, opened 2022-10-20
fix: radio button style import

SiTaggart:
- fix(button): export ToggleButtonStyles from index correctly
- fix(radio-button-group): import ToggleButtonStyles from compiled package source

**Contributing to Twilio**

### DIFF
--- a/packages/paste-codemods/tools/.cache/mappings.json
+++ b/packages/paste-codemods/tools/.cache/mappings.json
@@ -18,7 +18,7 @@
   "BreadcrumbItem": "@twilio-paste/core/breadcrumb",
   "ButtonGroup": "@twilio-paste/core/button-group",
   "Button": "@twilio-paste/core/button",
-  "ToggleButtonStyles": "@twilio-paste/core/button",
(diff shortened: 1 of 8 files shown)


shleewhite (approved):

nkrantz (dismissed):

gloriliale (approved):

packages/paste-codemods/tools/.cache/mappings.json
```diff
@@ -18,6 +18,7 @@
   "BreadcrumbItem": "@twilio-paste/core/breadcrumb",
   "ButtonGroup": "@twilio-paste/core/button-group",
   "Button": "@twilio-paste/core/button",
+  "ToggleButtonStyles": "@twilio-paste/core/button",
```
TheSisb, line 21: For the reason in which we create a mapping from package -> component exports, we have this loosely defined rule that exports from any package should be prefixed with the package name.

Can you instead export the styles as something like `ButtonToggleStyles` so there's no theoretical clash in the future and the mappings stay "obvious" please?

TheSisb (dismissed):

shleewhite (approved):

PR #2793, opened 2022-11-01
feat: add useChatLogger hook

jb-twilio:
This PR adds a new `useChatLogger` hook, `<ChatLogger />` component, associated stories, tests and docs site writeup. It also includes minor adjustments to `<ChatLog>` and other compound component markup to support the animations now occurring further upstream in ChatLogger.


packages/paste-core/components/chat-log/src/ChatLog.tsx
```diff
@@ -2,17 +2,60 @@ import * as React from 'react';
 import * as PropTypes from 'prop-types';
 import {Box, safelySpreadBoxProps} from '@twilio-paste/box';
 import type {BoxElementProps} from '@twilio-paste/box';
+import {useTransition, animated, useReducedMotion} from '@twilio-paste/animation-library';
+import {uid as buildID} from '@twilio-paste/uid-library';
```
shleewhite, line 6: Question: just out of curiosity, why did you rename this buildID?

packages/paste-core/components/chat-log/src/ChatLog.tsx
```diff
@@ -2,17 +2,60 @@ import * as React from 'react';
...
+
+const AnimatedChat = animated(Box);
+type StyleProps = React.ComponentProps<typeof AnimatedChat>['style'];
 
 export interface ChatLogProps {
   children?: React.ReactNode;
   element?: BoxElementProps['element'];
 }
 
+const isReactElement = (item: React.ReactNode): item is React.ReactElement => {
```
shleewhite, line 16: Suggestion(non-blocking): should isReactElement, isInboundChat, and isOutboundChat be moved into a utils file?

packages/paste-core/components/chat-log/src/useChatLog.ts
```diff
@@ -0,0 +1,61 @@
...
+      setChatDictionary((prev) => {
+        // eslint-disable-next-line @typescript-eslint/no-unused-vars
+        const {[id]: _, ...restPrev} = prev;
+        return restPrev;
+      });
+    },
+    [setChatDictionary]
+  );
+
+  const parsedChatData = React.useMemo(
```
shleewhite, line 48: Question: Could you explain how this function works? Sorry, I'm having some difficulty understanding.

shleewhite (commented):
The animation is looking ✨nice✨ 

Just had a couple clarifying questions. Once the interactive example is ready for the website this is perfect!!

packages/paste-core/components/chat-log/__tests__/useChatLog.spec.tsx
```diff
@@ -0,0 +1,78 @@
...
+        result.current.push(chat);
+      });
+
+      expect(result.current.chats).toHaveLength(1);
+      expect(result.current.chatIds).toHaveLength(1);
+    });
+  });
+
+  describe('pop', () => {
+    it('pops chats by id', async () => {
```
TheSisb, line 65: todo: can we get another pop test for without an id please? Would be fine to drop it into this `it` block and rename it to `pops chats correctly` so we don't need to create a new render

packages/paste-core/components/chat-log/src/ChatLog.tsx
```diff
@@ -2,17 +2,60 @@ import * as React from 'react';
...
```
TheSisb, line 16: suggestion (blocking): can we use the React top-level API here, [`React.isValidElement`](https://reactjs.org/docs/react-api.html#isvalidelement)?

packages/paste-core/components/chat-log/src/ChatLog.tsx
```diff
@@ -2,17 +2,60 @@ import * as React from 'react';
 import * as PropTypes from 'prop-types';
 import {Box, safelySpreadBoxProps} from '@twilio-paste/box';
 import type {BoxElementProps} from '@twilio-paste/box';
+import {useTransition, animated, useReducedMotion} from '@twilio-paste/animation-library';
+import {uid} from '@twilio-paste/uid-library';
+
+const AnimatedChat = animated(Box);
+type StyleProps = React.ComponentProps<typeof AnimatedChat>['style'];
```
TheSisb, line 9: praise: nice

packages/paste-core/components/chat-log/src/ChatMessage.tsx
```diff
@@ -31,8 +31,7 @@ export const ChatMessage = React.forwardRef<HTMLDivElement, ChatMessageProps>(
     return (
       <MessageVariantContext.Provider value={variant}>
         <Box
-          as="li"
-          listStyleType="none"
+          role="listitem"
```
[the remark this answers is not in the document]
TheSisb: Makes sense, thanks!

packages/paste-core/components/chat-log/src/ChatLog.tsx
```diff
@@ -2,17 +2,60 @@ import * as React from 'react';
...
   children?: React.ReactNode;
   element?: BoxElementProps['element'];
 }
 
+const isReactElement = (item: React.ReactNode): item is React.ReactElement => {
+  return (item as React.ReactElement)?.props !== undefined;
+};
+
+const isInboundChat = (item: React.ReactNode): boolean => {
+  return isReactElement(item) && item?.props?.variant === 'inbound';
```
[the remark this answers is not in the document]
SiTaggart: Yeah, this might start getting to why the toast hook is data driven and not jsx driven, because you don't need to necessarily inspect what is passed to you to figure out what to do. An example of that is toast variant, it's explicitly set on the object, so you don't need to look at passed in children, which you don't own technically, to figure out any specific knowledge

packages/paste-core/components/chat-log/src/useChatLog.ts
```diff
@@ -0,0 +1,61 @@
...
+    () =>
+      Object.entries(chatDictionary).reduce<{chats: Chats; chatIds: string[]}>(
+        (prev, [id, {content}]) => ({
+          chats: prev.chats.concat(React.cloneElement(content, {key: id})),
```
SiTaggart, line 52: Thought: again here, where we're cloning children is another example where the API using jsx might lend itself too much to overreaching into things we don't own. We're cloning and inspecting the children to make this work and it's not a pattern we tend to use due to component encapsulation rules. I.e no component should know about the existance of other components, but it relies on an implicit contract that could change if the consumer changes the child

packages/paste-core/components/chat-log/src/useChatLogger.ts
```diff
@@ -0,0 +1,45 @@
...
+type PushChat = (chat: PartialIDChat) => void;
+type PopChat = (id?: string) => void;
+
+export type Chat = {
+  id: string;
+  variant?: MessageVariants;
+  content: React.ReactElement;
+};
+
+export type PartialIDChat = Omit<Chat, 'id'> & Partial<Pick<Chat, 'id'>>;
```
shleewhite, line 15: Question: why do you have this partialIDChat type? could you just mark id as not required in the type above?

shleewhite (approved):
This is great!!!

packages/paste-website/src/component-examples/ChatLogExamples.ts
```diff
@@ -311,3 +311,66 @@ render(
...
+        <ChatBubble>{message}</ChatBubble>
+        <ChatMessageMeta aria-label={metaLabel + time}>
+          <ChatMessageMetaItem>{meta + time}</ChatMessageMetaItem>
+        </ChatMessageMeta>
+      </ChatMessage>
+    )
+  }
+};
+
+const chatTemplates = [
```
TheSisb, line 368: nit (not-blocking): could we move this and chatFactory to above the component "ChatLoggerExample"? More readable this way

TheSisb (approved):
I like how this turned out!

packages/paste-core/components/chat-log/src/useChatLogger.ts
shleewhite: Ohhhh.... I see. Thanks for explaining!

shleewhite (approved):

TheSisb (approved):

loreina (approved):
praise: docs site example looks great!

PR #2827, opened 2022-11-15
feat(in-page-navigation): add new component package

gloriliale:
# In this PR
This PR creates the new package In Page Navigation.

~~Right now, this is a draft with one story for preliminary design review.~~
This PR is ready for review!

- [x] Stories
- [x] Customization
- [x] Tests
- [x] Docs page 

**Contributing to Twilio**


packages/paste-core/components/in-page-navigation/src/InPageNavigationItem.tsx
```diff
@@ -0,0 +1,95 @@
...
+    },
+  };
+};
+
+export interface InPageNavigationItemProps extends React.AnchorHTMLAttributes<HTMLAnchorElement> {
+  currentPage?: boolean;
+  children: NonNullable<React.ReactNode>;
+  href: string;
+  ref?: any;
+  rel?: string; // do we need this? I don't understand how we're using it in Anchor
```
gloriliale, line 50: I added this prop bc it's in our Anchor component, but when I looked at Anchor, I didn't truly understand what we did with it there. The main question is, do we need this prop in InPageNavigationItem?

packages/paste-core/components/in-page-navigation/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,149 @@
...
+      <InPageNavigation aria-label="my-nav">
+        <InPageNavigationItem href="#">page 1</InPageNavigationItem>
+        <InPageNavigationItem href="#">page 2</InPageNavigationItem>
+      </InPageNavigation>
+    );
+
+    expect(getByRole('navigation').getAttribute('aria-label')).toBe('my-nav');
+  });
+
+  it('should render a list with list items', () => {
```
shleewhite, line 19: Suggestion(non-blocking): could this render test and the "should render anchors" test be combined? that would reduce renders

packages/paste-core/components/in-page-navigation/src/InPageNavigationItem.tsx
shleewhite: We probably should, its a HTML attribute: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/rel#:~:text=The%20rel%20attribute%20defines%20the,which%20the%20attribute%20is%20found.

Usually I see people set it to noreferrer.. 

However, I think it should be included in `React.AnchorHTMLAttributes`- not sure why it would need to be specified here too.

packages/paste-core/components/in-page-navigation/stories/index.stories.tsx
```diff
@@ -0,0 +1,100 @@
...
+
+import {InPageNavigation, InPageNavigationItem} from '../src';
+
+// eslint-disable-next-line import/no-default-export
+export default {
+  title: 'Components/In Page Navigation',
+  component: InPageNavigation,
+};
+
+export const Default: Story = () => {
```
shleewhite, line 18: Suggestion (non-blocking): do we need to have a comparison in the story?

packages/paste-website/src/component-examples/InPageNavigationExamples.ts
```diff
@@ -0,0 +1,45 @@
+export const basicExample = `
+const BasicInPageNavigation = () => {
+
+  return (
+    <InPageNavigation aria-label="my-nav">
```
shleewhite, line 5: Issue(blocking): aria label has to be a human readable string. So it could be "Primary navigation" or something like that. Same with the full width example.

kelseycavitt (commented):
@gloriliale Similar to what we did for the radio button group component ([tabs](https://paste.twilio.design/components/radio-button-group/#radio-button-groups-vs-tabs), [radio group](https://paste.twilio.design/components/radio-button-group/#button-group-vs-radio-button-group)) , could we have a lil paragraph/sentence about the difference between Tabs and In Page Navigation?

packages/paste-website/src/component-examples/InPageNavigationExamples.ts
nkrantz: the stories should also be updated to human readable strings

packages/paste-core/components/in-page-navigation/src/InPageNavigationItem.tsx
```diff
@@ -0,0 +1,95 @@
...
+        ref={ref}
+        element={element}
+        display="flex"
+        flexBasis={variant === 'fullWidth' ? '50%' : undefined}
+        flexGrow={variant === 'fullWidth' ? 1 : undefined}
+        flexShrink={variant === 'fullWidth' ? 1 : undefined}
+      >
+        <Box
+          {...safelySpreadBoxProps(props)}
+          {...getItemStyles(currentPage, variant)}
```
TheSisb, line 70: thought: looks like you're applying the fullWidth styles to this inner element and the outer `li`. Probably only need it on the outer `li`.

packages/paste-core/components/in-page-navigation/stories/index.stories.tsx
TheSisb: +1 to removing the comparison

packages/paste-core/components/in-page-navigation/stories/index.stories.tsx
```diff
@@ -0,0 +1,100 @@
...
+  return (
+    <>
+      <Heading as="h1" variant="heading30">
+        Comparison to Tabs style
+      </Heading>
+      <Stack orientation="vertical" spacing="space120">
+        <Box>
+          This is what <strong>In Page Navigation</strong> looks like (only for visual- clicking should not lead
+          anywhere)
+          <InPageNavigation aria-label="myNav">
```
TheSisb, line 28: todo (blocking): let's try and look on the web or in product for a real example with a human readable label. Let's use at least 2 different realistic examples on our storybook and website docs.

packages/paste-core/components/in-page-navigation/build.js
```diff
@@ -0,0 +1,3 @@
+const {build} = require('../../../../tools/build/esbuild');
```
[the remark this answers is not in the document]
TheSisb: In the future I plan to refactor it into a private `scripts` package, but haven't done so yet

packages/paste-core/components/in-page-navigation/__tests__/index.spec.tsx
gloriliale: yes yes, good idea

packages/paste-core/components/in-page-navigation/stories/index.stories.tsx
gloriliale: I left the comparisons in to make it easier for @twilio-labs/design-systems-pd to review. Maybe this is something we can remove after they've approved it?

packages/paste-core/components/in-page-navigation/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,149 @@
...
+        <InPageNavigationItem currentPage href="#">
+          page 2
+        </InPageNavigationItem>
+      </InPageNavigation>
+    );
+
+    expect(getByText('page 2').getAttribute('aria-current')).toBe('page');
+  });
+
+  it('should pass props given to InPageNavigationItem onto its Anchor child', () => {
```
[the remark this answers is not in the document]
gloriliale: great callout, it's the `a` element. Will update accordingly!

packages/paste-core/components/in-page-navigation/src/InPageNavigation.tsx
```diff
@@ -0,0 +1,51 @@
...
+import type {BoxProps} from '@twilio-paste/box';
+import {Box, safelySpreadBoxProps} from '@twilio-paste/box';
+
+import {InPageNavigationContext} from './InPageNavigationContext';
+
+export interface InPageNavigationProps extends Omit<React.ComponentPropsWithRef<'div'>, 'children'> {
+  children: React.ReactNode;
+  element?: BoxProps['element'];
+  'aria-label': string;
+  variant?: 'fullWidth';
```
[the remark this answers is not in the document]
gloriliale: ooo good call, adding this now!

packages/paste-core/components/in-page-navigation/src/InPageNavigationItem.tsx
gloriliale: I'm not sure exactly why, but it does need to be in both places for the inner text to be centered. Checked this out w/ @SiTaggart yesterday. Open to suggestions if there's anything that makes more logical sense though.

packages/paste-core/components/in-page-navigation/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,149 @@
...
+    const {getByText} = render(
+      <InPageNavigation aria-label="my-nav">
+        <InPageNavigationItem href="#">page 1</InPageNavigationItem>
+        <InPageNavigationItem currentPage href="#">
+          page 2
+        </InPageNavigationItem>
+      </InPageNavigation>
+    );
+
+    expect(getByText('page 2').getAttribute('aria-current')).toBe('page');
```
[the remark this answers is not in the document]
gloriliale: oh i definitely like this better, updating now!

packages/paste-core/components/in-page-navigation/src/InPageNavigationItem.tsx
```diff
@@ -0,0 +1,95 @@
+import * as React from 'react';
+import * as PropTypes from 'prop-types';
+import type {BoxStyleProps, BoxProps} from '@twilio-paste/box';
+import {Box, safelySpreadBoxProps} from '@twilio-paste/box';
+
+import {InPageNavigationContext} from './InPageNavigationContext';
+
+const getItemStyles = (currentPage: boolean, variant?: string): BoxStyleProps => {
```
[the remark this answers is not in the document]
gloriliale: @TheSisb I did this but I'd like to take a look at it together if you've got some time tomorrow (Friday).

packages/paste-core/components/in-page-navigation/src/InPageNavigationItem.tsx
gloriliale: update: 
- `rel` and `target` are included in `React.AnchorHTMLAttributes` as @shleewhite pointed out above, so it has been removed from the props
- `href`, though also included in `React.AnchorHTMLAttributes`, is optional there while we want it to be required here, so it is declared here to be required.
- `ref` is not necessary to add to our props, so it has been removed. @TheSisb and I tested this out by removing this line and adding a ref to a story. It worked.

packages/paste-core/components/in-page-navigation/src/InPageNavigationItem.tsx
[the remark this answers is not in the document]
gloriliale: update: refactored this even better, woohoo! thx @TheSisb!

packages/paste-core/components/in-page-navigation/stories/index.stories.tsx
kelseycavitt: I think having it initially was good, but post review, let's remove it!

kelseycavitt (commented):
@gloriliale 
Something that didn't occur to me to try with my first review was to stress test the # of items within the component, SO sorry about that! I did end up finding a little wonkiness when there's a large number or longer names of an item. I think the longer names is very edge case, but the number of items for smaller screen widths is a use case worth addressing I think.

1. Here's a [**Loom**](https://www.loom.com/share/7724fb2867a449179312cfab0b3a38f4) that shows the wonkiness in the In Page Nav component. I tried this out on the docs site page as well.

2. Here's another [**Loom**](https://www.loom.com/share/6feba25fbfe644749163c70930642760) that shows what we do for tabs

I think the solution here would be to actually not truncate (like tabs does) and allow a user to horizontally scroll within the component. Both things aren't super ideal from a design perspective (specifically if you're on a web browser, vs. a mobile web browser). Let me know what you think!

TheSisb (dismissed):

loreina (dismissed):

nkrantz (dismissed):

TheSisb (approved):

nkrantz (approved):

loreina (approved):

shleewhite (approved):

PR #2838, opened 2022-11-17
fix: eslint and peer deps

SiTaggart:
Changes to eslint and how it runs, so that we can correctly identify missing internal packages, and lint otherwise un-linted projects in the monorepo.

### Eslint Configs

Right now we are unable to detect missing peer dependencies via eslint and the `no-extraneous-dependencies` rule because of the way we give eslint a cache of where the packages live for the gloabl root linting of Storybook and Jest files.

Internal pacakges are not installed in the repo. So when linting a story which is global to the monorepo, it can't find button as an installed package. So we patch it and tell eslint where it lives.

Problem with that config was that when you are linting a package src, the internal packages are always found as they were globally patched to the file path. Hence we never caught any.

This config change splits the configs into packages and repo. We keep the patch cache for global or root files like Storybook, but remove it for packages, so that eslit looks at the local package json file when linting a component package.

### Linting

This change also opened up the ability to lint each type of project on their own. We now lint:

- core
- website
- remix
- contrast checker
- nextjs template 
- vscode extension
- all the rest as root files

This makes sure previously unlinted files, are now correctly linted via the Twilio Style Eslint Configs (hence all the files changes)

## Things to fix

- [x] Why are the perf stories suddenly erroring with errors that the repo config doesn't even have?

## Checklist


gloriliale (approved):

TheSisb (approved):
Plenty of good fixes in here.

PR #2863, opened 2022-12-06
chore(website): remove guideline of not using dark theme as only theme

serifluous:
Removes outdated Do / Don't about not using dark theme as the only theme.

See #2861


SiTaggart (approved):

loreina (approved):
suggestion(non-blocking): could we add some guidance on this page that folks can use either theme as their main theme?

PR #2895, opened 2022-12-14
chore: update Storybook to v7 and use Vite

Niznikr:
Update Storybook to v7 and use Vite in place of Webpack for faster development:

- Follow [migration guidelines](https://github.com/storybookjs/storybook/blob/next/MIGRATION.md)
- Update packages to be ESM-compatible
- Update usage of `require` in components to use import instead

**SB v6.5**
Screenshot 2022-12-13 at 9 33 33 AM

**SB v7 Webpack**
Screenshot 2022-12-13 at 10 12 16 AM

**SB v7 Vite**
Screenshot 2022-12-14 at 4 32 06 PM


.github/workflows/on_merge_to_main.yml
```diff
@@ -183,6 +183,7 @@ jobs:
           exitOnceUploaded: true
         env:
           STORYBOOK_GITHUB_SHA: ${{ github.sha }}
+          NODE_OPTIONS: --max-old-space-size=6144
```
SiTaggart, line 186: Suggestion: worth a comment here to explain why we need this? Just so no one deletes it later

packages/paste-core/components/form-pill-group/src/PillCloseIcon.tsx
```diff
@@ -57,7 +56,6 @@ export const PillCloseIcon: React.FC<PillCloseIconProps> = ({
       borderStyle="solid"
       borderColor="transparent"
       transition="color 150ms ease-in, border-color 150ms ease-in, background-color 150ms ease-in"
-      {...computedStyles}
```
SiTaggart, line 60: Question: is there any chance these computed styles are intentionally set after the default style props? Does the computation calculate a state that sometimes needs to override a style prop that is fine _most of the time_, but not always?

SiTaggart: Should get caught by VRT but I know we don't VRT hover and focus styles for example

packages/paste-customization/src/types/CreateCustomTheme.ts
```diff
@@ -5,6 +5,6 @@ import type {PasteCustomCSS} from './PasteCSS';
 export interface CreateCustomTheme {
   baseTheme: Partial<GenericThemeShape>;
   customBreakpoints?: string[];
-  elements?: PasteCustomCSS;
+  elements?: {[key: string]: PasteCustomCSS};
```
SiTaggart, line 8: Question: I wonder how this actually worked then, if I didn't type PasteCustomCSS as a keyed object?

packages/paste-customization/src/utils.ts
```diff
@@ -1,6 +1,6 @@
-import type {CustomTheme, CreateCustomTheme} from './types';
+import merge from 'deepmerge';
 
-const merge = require('deepmerge');
+import type {CustomTheme, CreateCustomTheme} from './types';
```
SiTaggart, line 3: Question: I guess my question is any idea why this wasn't getting caught by eslint? It is an eslint rule, so kinda scary it wasn't catching it here

packages/paste-design-tokens/gulpfile.ts
```diff
@@ -29,7 +29,7 @@ theo.registerFormat('es6.js', es6TokenFormat);
 theo.registerFormat('common.d.ts', dTSTokenFormat);
 theo.registerFormat('es6.d.ts', dTSTokenFormat);
 theo.registerFormat('sketchpalette', sketchpaletteTokenFormat);
-theo.registerFormat('generic.js', genericCommonJsTokenFormat);
+theo.registerFormat('generic.js', genericJsTokenFormat);
```
SiTaggart, line 32: Question: If anyone is using this import directly from design tokens package, should we be more explicit in our change log for design tokens? Is a consumer going to need to change the babel, webpack, whatever configs? Is that technically a breaking change?

.storybook/preview.tsx
```diff
@@ -169,3 +170,14 @@ export const parameters = {
...
+global.___loader = {
+  enqueue: () => {},
+  hovering: () => {},
+};
+
+global.__BASE_PATH__ = '/';
+
+window.___navigate = (pathname) => {
+  action('NavigateTo:')(pathname);
+};
```
SiTaggart, line 183: Question: assuming these are all Gatsby Link patching?

.storybook/main.ts
```diff
@@ -0,0 +1,47 @@
...
+  staticDirs: ['./static'],
+  typescript: {
+    // enable type checking
+    check: true,
+  },
+  docs: {
+    docsPage: 'automatic',
+    defaultName: 'Docs',
+  },
+  async viteFinal(config) {
```
SiTaggart, line 33: Question: does the profiler still work without the aliasing and subsequently forcing a dev build for production?

SiTaggart (commented):
Speedy improvements, just some general questions

packages/paste-core/components/form-pill-group/src/PillCloseIcon.tsx
Niznikr: Had to move those to address TS errors since importing instead of requiring deepmerge resulted in actual typing instead of `any`. Will see if it causes visual diffs otherwise we could ts-ignore if needed.

packages/paste-customization/src/utils.ts
Niznikr: Not sure tbh

.storybook/preview.tsx
Niznikr: Yup can add a comment in there.

.storybook/main.ts
Niznikr: Will verify that.

packages/paste-design-tokens/gulpfile.ts
Niznikr: Yeah can make an explicit note for that package in the changeset. Hard to say if this should be considered breaking. The Paste website for example uses that import and its Webpack configuration didn't need tweaking.

packages/paste-core/components/form-pill-group/src/PillCloseIcon.tsx
Niznikr: Verified hover/focus styles are not impacted by this change.

.storybook/main.ts
Niznikr: Had to change `process.env` to `import.meta.env` (Vite's version). Also needed to [include the aliases](https://github.com/twilio-labs/paste/pull/2895/commits/d3bce166f50b04a1c067931978c20e56455801b2) plus one for `react-dom/client` which may help inform the React 18 upgrade. Verified this gets profiling to work as expected.

packages/paste-customization/src/types/CreateCustomTheme.ts
SiTaggart: Unrelated, we need to make deepmerge a system library and share it. We bundle it in a number of components now

.changeset/twelve-parents-glow.md
```diff
@@ -0,0 +1,72 @@
...
+'@twilio-paste/design-tokens': patch
+'@twilio-paste/data-visualization-library': patch
+'@twilio-paste/style-props': patch
+'@twilio-paste/website': patch
+'@twilio-paste/core': patch
+---
+
+Update packages to be ESM-compatible
+
+[Design Tokens] Update `tokens.generic.js` export from commonjs to ESM
```
SiTaggart, line 72: suggestion: I'd pull design tokens out of this changeset, and create its own changeset file, with this line as it's entry. Otherwise, every changelog for the packages listed will get this line about design tokens added

shleewhite (dismissed):

.storybook/main.js
```diff
@@ -1,90 +0,0 @@
...
-      'react-dom': path.resolve(__dirname, '../node_modules/react-dom/profiling'),
-      'scheduler/tracing': 'scheduler/tracing-profiling',
-      '@emotion/core': path.resolve(__dirname, '../node_modules/@emotion/react'),
-      '@emotion/styled': path.resolve(__dirname, '../node_modules/@emotion/styled'),
-      'emotion-theming': path.resolve(__dirname, '../node_modules/@emotion/react'),
-    };
-    config.resolve.alias = config.resolve.alias == null ? customAlias : {...config.resolve.alias, ...customAlias};
-
-    // FIX: Tell Storybook to look at dev files if available
-    config.resolve.mainFields = ['main:dev', 'browser', 'module', 'main'];
```
TheSisb, line 83: question: this was needed in order to tell storybook to hot reload on src file changes. Do we not need this anymore? Does it know how to find src files?

TheSisb (commented):
Love the speed up! nice!

.storybook/main.js
Niznikr: Oh good catch! Seems we still need it to detect changes to peer dep src changes. Added it back in [this commit](https://github.com/twilio-labs/paste/pull/2895/commits/00214dab86ef79ce75b0b8f5979f3689c775f966).

packages/paste-core/components/alert-dialog/src/AlertDialogFooter.tsx
```diff
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import * as PropTypes from 'prop-types';
+import PropTypes from 'prop-types';
```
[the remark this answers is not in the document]
Niznikr: Great question! Vite uses Rollup for production builds and is very strict about using the correct import type. In this case, barrel imports are unnecessary and were causing errors. Using the es6 imports [per their docs](https://github.com/facebook/prop-types#importing) fixed things.

shleewhite (approved):

nkrantz (approved):

.storybook/main.ts
```diff
@@ -0,0 +1,51 @@
...
+  },
+  docs: {
+    autodocs: true,
+    defaultName: 'Docs',
+  },
+  async viteFinal(config, {configType}) {
+    return mergeConfig(config, {
+      resolve: {
+        alias: {
+          gatsby: path.resolve(__dirname, './gatsby'),
```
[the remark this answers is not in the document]
Niznikr: Not quite. The ESM-equivalent of `__dirname` is:
```
import { dirname } from 'path';
import { fileURLToPath } from 'url';
const __dirname = fileURLToPath(dirname(import.meta.url));
```

TheSisb (dismissed):
Solid!

shleewhite (approved):

TheSisb (approved):

PR #2899, opened 2022-12-16
ci: pin node version for codesandbox

Niznikr:
Pin the node version to match our `nvmrc` as codesandbox builds were failing when using `16.18.0`.


Niznikr:
Looks like codesandbox only accepts major versions. We'll have to fix issues with using `16.18.0`.

PR #2902, opened 2022-12-17
chore: swap lerna, almost entirely, for nx

SiTaggart:
Lerna > Nx

### Switch out Lerna as the repo running for Nx

This works entirely as a runner and is much faster.

### Build caching with Nx

Enable remote caching of the build with Nx cloud. This largely all works except for one edge case with the website build.

#### Edge case failure

- [x] In some cached builds in Netlify, the open graph serverless function will sometimes fail to build because the cached node modules make the function size too large. But it's not all the time.

## Checklist


.github/workflows/on_merge_to_main.yml
```diff
@@ -8,67 +8,17 @@ on:
       - main
 
 jobs:
-  prerelease:
-    name: Prerelease test and build
+  version_or_publish:
```
SiTaggart, line 11: Combining prerelease, and publish as it takes longer to upload the build assets than to build them with NX.

.github/workflows/on_merge_to_main.yml
```diff
@@ -167,11 +116,10 @@ jobs:
         if: steps.yarn_cache_id.outputs.cache-hit != 'true' || steps.node_modules_cache_id.outputs.cache-hit != 'true'
         run: yarn install --immutable
 
-      - name: Download build artifacts
-        uses: actions/download-artifact@v3
-        with:
-          name: compiled-js-and-types
-          path: packages/
```
SiTaggart, line 174: Again, takes longer to download assets than build then with NX

.github/workflows/on_pull_request.yml
```diff
@@ -6,66 +6,18 @@ on:
 
 jobs:
   build:
-    name: Build system packages
+    name: Build system packages and type check
```
SiTaggart, line 9: Type check needs to come with build because type checking website as an example requires built icon types.

I'm still not convinced that each package needs to be type checked after build, becuase to compile the types in build, it is type checked. 

But we do need to type check somethings, like website and theme designer

internal-docs/engineering/standard-operating-procedures.md
```diff
@@ -49,7 +49,7 @@ Then we think about the development and technical debt incurred by this product.
 ### Component Library specific
 
 - The `package.json` file contains
-  - [ ] the required yarn scripts: `[build, build:js, build:props, clean, tsc]`.
+  - [ ] the required yarn scripts: `[build, build:js, clean, tsc]`.
```
SiTaggart, line 52: deleting the `build:props` script as it doesn't get run and we don't need it

packages/paste-core/core-bundle/package.json
```diff
@@ -12,6 +12,53 @@
   "publishConfig": {
     "access": "public"
   },
+  "nx": {
```
SiTaggart, line 15: Core bundle needs it's own nx config overrides as some of the scripts are different and how we build it is different.

TheSisb (approved):
So so so so good!

.storybook/preview.tsx
```diff
@@ -135,7 +135,7 @@ export const decorators = [
         return (
           <RenderPerformanceProfiler id={context.id} kind={context.kind} view="stacked">
             <GlobalStyles />
-            <Stack orientation="vertical">
+            <Stack orientation="vertical" spacing="space0">
```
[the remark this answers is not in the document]
SiTaggart: No it's the same, it's just a required prop and this file is type checked 🙃

.github/workflows/on_pull_request.yml
```diff
@@ -98,22 +50,22 @@ jobs:
...
-        uses: actions/download-artifact@v3
-        with:
-          name: compiled-js-and-types
-          path: packages/
+      - name: Build packages
+        run: yarn build --base=origin/main
 
       - name: Run type checker
-        run: yarn type-check
+        run: yarn type-check --base=origin/main
```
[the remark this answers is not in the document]
SiTaggart: I think ideal would be:

- each build must type check. So website should type check. Storybook should type check etc.

But I might look to do that later on as a seperate fast follow

PR #2915, opened 2023-01-04
[AIONAPFMWK-170] Types Fix

espipj:
Re-uses commits from https://github.com/adamb-92/paste/pull/1
**Contributing to Twilio**

> All third-party contributors acknowledge that any contributions they provide will be made under the same open-source license that the open-source project is provided under.

- [ ] I acknowledge that all my contributions will be made under the project's license.

PR #2918, opened 2023-01-05
fix: react 18 update tests to pass with latest react version

adamb-92:
**Contributing to Twilio**

### DIFF
--- a/packages/paste-core/components/data-grid/__tests__/index.spec.tsx
+++ b/packages/paste-core/components/data-grid/__tests__/index.spec.tsx
@@ -56,18 +56,14 @@ describe('Data Grid', () => {
   describe('Composable Cells functionality', () => {
     it('has proper keyboard navigation behavior', async () => {
       const {getByTestId} = render(<ComposableCellsDataGrid />);
-      const wrapper = getByTestId('data-grid');
       const headerCell = getByTestId('header-1');
       const firstRowFirstInputCell = getByTestId('input-0-0');
-      const firstRowSecondInputCell = getByTestId('input-0-1');
-      const secondRowFirstInputCell = getByTestId('input-1-0');
       const firstInputCell = firstRowFirstInputCell?.parentElement?.parentElement;
 
       if (firstInputCell == null) {
         throw new Error('cannot find firstInputCell');
       }
 
-      // TEST: moves with arrow keys in navigational mode
       act(() => {
         headerCell.focus();
       });
@@ -87,8 +83,19 @@ describe('Data Grid', () => {
       expect(firstInputCell.getAttribute('tabindex')).toBe('-1');
       expect(headerCell.getAttribute('tabindex')).toBe('0');
       expect(headerCell).toHaveFocus();
+    });
 
-      // TEST: toggles actionable mode with [enter] and [escape] keys
+    it('toggles actionable mode with [enter] and [escape] keys', async () => {
+      const {getByTestId} = render(<ComposableCellsDataGrid />);
+      const wrapper = getByTestId('data-grid');
+      const headerCell = getByTestId('header-1');
+      const firstRowFirstInputCell = getByTestId('input-0-0');
+      const firstInputCell = firstRowFirstInputCell?.parentElement?.parentElement;
+
+      if (firstInputCell == null) {
+        throw new Error('cannot find firstInputCell');
+      }
+
       act(() => {
         headerCell.focus();
       });
@@ -103,8 +110,20 @@ describe('Data Grid', () => {
 
       await userEvent.click(headerCell);
       expect(wrapper.getAttribute('data-actionable')).toBe('true');
+    });
 
-      // TEST: should correctly tab through focusable elements in actionable mode
+    it('should correctly tab through focusable elements in actionable mode', async () => {
+      const {getByTestId} = render(<ComposableCellsDataGrid />);
+      const headerCell = getByTestId('header-1');
+      const firstRowFirstInputCell = getByTestId('input-0-0');
+      const firstRowSecondInputCell = getByTestId('input-0-1');
+      const secondRowFirstInputCell = getByTestId('input-1-0');
+      const firstInputCell = firstRowFirstInputCell?.parentElement?.parentElement;
+
+      if (firstInputCell == null) {
+        throw new Error('cannot find firstInputCell');
+      }
+
       act(() => {
         headerCell.focus();
       });
@@ -135,7 +154,7 @@ describe('Data Grid', () => {
       await userEvent.keyboard('{enter}');
       await userEvent.tab();
       expect(firstRowSecondInputCell).toHaveFocus();
-    }, 25000);
+    });
 
     it('has one tab stop in navigational mode and remembers the last focus', async () => {
       const {getByTestId} = render(


Niznikr:
I wonder if instead of wrapping individual instances of userEvent with act we instead [create a utility as seen here](https://github.com/testing-library/user-event/issues/938#issuecomment-1111976312) to use across all tests?

package.json
```diff
@@ -138,7 +138,7 @@
     "@testing-library/jest-dom": "^5.16.5",
     "@testing-library/react": "^13.4.0",
     "@testing-library/react-hooks": "^8.0.1",
-    "@testing-library/user-event": "^14.4.3",
+    "@testing-library/user-event": "^13.5.0",
```
TheSisb, line 141: question: why do we need to downgrade here?

packages/paste-customization/__tests__/CustomizationProvider.test.tsx
```diff
@@ -51,7 +51,7 @@ const BaseThemeConsumerExample: React.FC<React.PropsWithChildren<BaseThemeConsum
 
 describe('CustomizationProvider', () => {
   it('should render without crashing', (): void => {
-    const root = ReactDOM.createRoot(document.querySelector('div') as HTMLElement);
+    const root = ReactDOM.createRoot(document.createElement('div') as HTMLElement);
```
TheSisb, line 54: question: why this change? querySelector should be finding a div on the page but createElement creates an entirely new div, so the behavior is different

.github/workflows/on_pull_request.yml
```diff
@@ -168,27 +168,27 @@ jobs:
...
-
-      - name: Node modules cache
-        uses: actions/cache@v3
-        id: node_modules_cache_id
-        with:
-          path: |
-            node_modules
-            */*/node_modules
-          key: ${{ runner.os }}-${{ hashFiles('**/yarn.lock') }}
+      # - name: Get yarn cache directory path
```
TheSisb, line 171: question: is it  a todo to re-enable this caching?

jest.config.js
```diff
@@ -30,7 +30,11 @@ module.exports = {
       statements: 100,
     },
   },
-  moduleNameMapper: {'\\.css$': 'identity-obj-proxy'},
+  moduleNameMapper: {
+    '\\.css$': 'identity-obj-proxy',
+    '^@testing-library/user-event$': '<rootDir>/tools/test/act-user-event.ts',
```
TheSisb, line 35: question: why these changes? What are we trying to solve

packages/paste-core/components/chat-log/__tests__/chatLog.spec.tsx
```diff
@@ -44,8 +44,7 @@ describe('Customization', () => {
   });
 
   it('should add custom styles with a custom element data attribute', () => {
-    render(<ChatLog element="FOO_LOG" />, {wrapper: CustomizationFooWrapper});
-
+    render(<ChatLog element="foo_log" />, {wrapper: CustomizationFooWrapper});
```
TheSisb, line 47: quibble: we can probably undo this change since it doesn't really do anything for this pr

jest.config.js
TheSisb: Looks like this is a monkey patch. Let's add a comment with context for future maintainers please.

Include this link: https://github.com/adobe/react-spectrum/blob/main/packages/dev/test-utils/src/renderOverride.js#L21-L29

adamb-92: Yes! Not a problem, I'll add comments with the link included

.github/workflows/on_pull_request.yml
adamb-92: This was removed to debug some issues I noticed in the pipeline, I'll re-introduce these

packages/paste-core/components/chat-log/__tests__/chatLog.spec.tsx
adamb-92: This test fails because of an update that was done on a previous PR: https://github.com/twilio-labs/paste/pull/2915/files#diff-e0608fc8b28f10e9fdaf99f3a54d8aba1d44ae3bacc4862271255d2fd8e96eebL18

It addressed a linting error that came up (and caused the linting pipeline to fail) - we now need this change in this element value so the test passes

packages/paste-customization/__tests__/CustomizationProvider.test.tsx
adamb-92: We currently have createElement on main - https://github.com/twilio-labs/paste/blob/main/packages/paste-customization/__tests__/CustomizationProvider.test.tsx#L54

I'll double check if a div is on the page but they were failing when we ran the tests with the querySelector

packages/paste-website/tsconfig.json
```diff
@@ -16,7 +16,8 @@
     "skipLibCheck": true /* Skip type checking all .d.ts files. */,
     "outDir": "dist/",
     "emitDeclarationOnly": false,
-    "typeRoots": ["../../@types", "../../node_modules/@types", "./types"]
+    "typeRoots": ["../../@types", "../../node_modules/@types", "./types"],
+    "noUnusedLocals": false
```
adamb-92, line 20: This is required for paste-website as the tsconfig has "jsx":  "react-jsx". This is needed because we're using "jsxImportSource": "@emotion/react". This allows us to re-introduce the 'import * as React from 'react';' at the top of the test files without the type-check step failing in the pipeline

packages/paste-core/components/data-grid/__tests__/index.spec.tsx
```diff
@@ -68,64 +68,74 @@ describe('Data Grid', () => {
...
+      await userEvent.keyboard('{enter}');
+      await userEvent.keyboard('{arrowdown}');
       expect(firstInputCell).toHaveFocus();
-      userEvent.keyboard('{enter}');
-      userEvent.tab();
+      await userEvent.keyboard('{enter}');
+      await userEvent.tab();
       expect(firstRowSecondInputCell).toHaveFocus();
-    });
+    }, 25000);
```
adamb-92, line 138: I know this isn't ideal but this test kept timing out at around 20 seconds. I bumped it up to 25 and it's now passing.

Open to suggestions if someone can spot an alternative

packages/paste-customization/__tests__/CustomizationProvider.test.tsx
TheSisb: Oh if this is just reverting to something we're doing in Paste that works. I guess this was introduced by accident somewhere earlier in this upgrade?

packages/paste-core/components/chat-log/__tests__/chatLog.spec.tsx
TheSisb: I see so this is correct on main, but was broken along the way in the upgrade. Same as the other comment I had, np

packages/paste-core/components/data-grid/__tests__/index.spec.tsx
andreasonny83: it looks like this test case is quite big and we're testing several conditions. would it be possible to split this test case in smaller ones?

packages/paste-customization/__tests__/CustomizationProvider.test.tsx
adamb-92: Yes exactly, looks like this was an accident 🙂

packages/paste-website/tsconfig.json
TheSisb: thought: let's not add `"noUnusedLocals": false`. While it may solve the React import problem that was randomly introduced, it opens up a lot of room for leaving unused locals in all the other code.

packages/paste-core/components/data-grid/__tests__/index.spec.tsx
TheSisb: Same recommendation here, we should split this up imo

packages/paste-website/tsconfig.json
adamb-92: Okay understood, I'll trace back where the react import originally came from and I'll remove it.

packages/paste-core/components/data-grid/__tests__/index.spec.tsx
adamb-92: Agreed, I'll split this out into smaller test

adamb-92:
Closing this PR, I've had to make a change to how we're calling user-event which updates most of these files. I've opened a new PR that addresses this change as well as fixing the tests for 17 and 16: https://github.com/twilio-labs/paste/pull/2929

It's also a smaller PR so I feel this PR is no longer needed

PR #2923, opened 2023-01-05
feat: add Description List package

nkrantz:
Adds a new component package (Description List) for displaying small bits of data in term-detail format.

Also fixes a small eslint error in the stories and test plop template files for the `yarn create:package` script.


packages/paste-core/components/description-list/src/DescriptionListTerm.tsx
```diff
@@ -0,0 +1,28 @@
...
+
+export interface DescriptionListTermProps extends Omit<React.ComponentPropsWithRef<'div'>, 'children'> {
+  children?: React.ReactNode;
+  element?: BoxProps['element'];
+}
+
+const DescriptionListTerm = React.forwardRef<HTMLElement, DescriptionListTermProps>(
+  ({element = 'DESCRIPTION_LIST_TERM', ...props}, ref) => {
+    return (
+      <Box as="dt" ref={ref} element={element} fontWeight="fontWeightSemibold">
```
TheSisb, line 14: question: since we're using Box for all of these (dt,dl,dd) without passing any styles, they may render differently on different browsers. Worth trying to see how they look, spacing and font wise, on Safari Chrome FF and Edge if you have it.  

Alternatively we can hardcode our tokens as props here by passing default values.
`<Box as="dt" color="colorText" fontSize="fontSize30" lineHeight="lineHeight30" margin="space0" fontWeight="fontWeightSemibold">`

nkrantz: I tested on Safari, Chrome, and FF, and there are no differences I can see, but I'll add the styles just in case the default designs change in the future

gloriliale (commented):
Looks good so far! The only thing to add is getting the spacing between these things set. Lemme know when that's up!

packages/paste-core/components/description-list/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,130 @@
...
+  const details = screen.getAllByRole('definition');
+
+  it('should render as terms and definitions', () => {
+    expect(terms.length).toBe(3);
+    expect(details.length).toBe(4);
+  });
+
+  describe('DescriptionListTerm', () => {
+    it('should apply the correct styling', () => {
+      expect(terms[0]).toHaveStyleRule(FONT_WEIGHT, '600');
```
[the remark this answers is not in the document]
nkrantz: eslint suggested it 🤷

kelseycavitt (changes requested):
@nkrantz I left a comment in Slack and Chromatic, but just sharing here too.
Something I noticed when looking at smaller screen sizes is that if there's a long string, it keeps going past the viewport. @shleewhite  just encountered this too with the File Uploader and I think she used a word break?
Screen Shot 2023-01-09 at 8 03 45 PM
Screen Shot 2023-01-09 at 8 00 50 PM

kelseycavitt (approved):

tools/plop-templates/component-stories.hbs
```diff
@@ -1,4 +1,5 @@
 import * as React from 'react';
+
```
Niznikr, line 2: Thanks for correcting this!

Niznikr (approved):
Nice work!

PR #2925, opened 2023-01-06
feat: create new Form package

shleewhite:
Creates the new form package with tests, stories, and customization.

**Contributing to Twilio**


packages/paste-core/components/form/src/FormControl.tsx
```diff
@@ -0,0 +1,24 @@
...
+import type {BoxProps} from '@twilio-paste/box';
+
+export interface FormControlProps extends Omit<React.ComponentPropsWithRef<'div'>, 'children'> {
+  element?: BoxProps['element'];
+  children: React.ReactNode;
+}
+
+export const FormControl = React.forwardRef<HTMLDivElement, FormControlProps>(
+  ({children, element = 'FORM_CONTROL', ...props}, ref) => (
+    <Box ref={ref} flexGrow={1} element={element} {...safelySpreadBoxProps(props)}>
```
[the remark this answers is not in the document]
shleewhite: I added it so all of the two column layouts to take up the full width available.

Here is a screenshot without flexGrow:
Screen Shot 2023-01-06 at 2 16 43 PM


Here is a screenshot with the iPhone 5 screen size
Screen Shot 2023-01-06 at 2 15 23 PM

packages/paste-core/components/form/src/Form.tsx
```diff
@@ -0,0 +1,48 @@
...
+    </Box>
+  )
+);
+
+Form.displayName = 'Form';
+
+Form.propTypes = {
+  children: PropTypes.node,
+  element: PropTypes.string,
+  maxWidth: PropTypes.oneOf([
```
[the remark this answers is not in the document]
shleewhite: oooh I forgot about those! Ill update it.

packages/paste-core/components/form/stories/index.stories.tsx
```diff
@@ -0,0 +1,417 @@
...
+  FormControlTwoColumn,
+  FormSection,
+  FormSectionDescription,
+  FormSectionHeading,
+} from '../src';
+
+// eslint-disable-next-line import/no-default-export
+export default {
+  title: 'Components/Form',
+  component: Form,
```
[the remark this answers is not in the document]
shleewhite: Yah!

TheSisb (approved):
lgtm

Niznikr (dismissed):

packages/paste-core/components/form/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,179 @@
...
+    );
+
+    const form = screen.getByRole('form');
+    expect(form).toBeDefined();
+    expect(form).toHaveAttribute('aria-label', 'Settings');
+
+    const description = screen.getByTestId('section-desc');
+    expect(description.id).toBe('foo-section-description');
+
+    const section = screen.getByRole('group', {name: 'Settings'});
```
gloriliale, line 36: question: this was unclear to me, but maybe it's from a lack of familiarity w/ `group`- is the name of this group 'Settings' because of the content of `<FormSectionHeading>` or because of `aria-label="Settings"` on `Form`?

packages/paste-core/components/form/src/FormControlTwoColumn.tsx
```diff
@@ -0,0 +1,31 @@
+import * as React from 'react';
+import PropTypes from 'prop-types';
+import {Box, safelySpreadBoxProps} from '@twilio-paste/box';
+import type {BoxProps} from '@twilio-paste/box';
+
+export interface FormControlTwoColumnProps extends Omit<React.ComponentPropsWithRef<'legend'>, 'children'> {
```
gloriliale, line 6: question: why are we getting props from `legend` not `div`?

packages/paste-core/components/form/src/FormSection.tsx
```diff
@@ -0,0 +1,48 @@
...
+          padding="space0"
+          marginBottom="space60"
+          ref={ref}
+          element={element}
+          display="flex"
+          flexDirection="column"
+          rowGap="space80"
+          _last={{marginBottom: 'space0'}}
+          {...safelySpreadBoxProps(props)}
+          aria-describedby={`${id}-section-description`}
```
gloriliale, line 33: thought: would this break our i18n rules for no strings? Or since it's used more like a key here will it not be something actually seen by a user?

packages/paste-core/components/form/stories/index.stories.tsx
```diff
@@ -0,0 +1,425 @@
...
+  const [comboboxInputValue, setComboboxInputValue] = React.useState('');
+  const filteredItems = React.useMemo(
+    () => getMultiselectComboboxFilteredItems(comboboxInputValue),
+    [comboboxInputValue]
+  );
+
+  return (
+    <Form aria-labelledby={seed('form-heading')}>
+      <Heading id={seed('form-heading')} as="h2" variant="heading20" marginBottom="space0">
+        {/* Added UID to the heading because forms are HTML Landmarks. Axe was failing because in the stacked view, there were 4 forms with the same name and landmarks have to have unique names. */}
```
gloriliale, line 109: praise: thanks for leaving this comment!

packages/paste-core/components/form/stories/index.stories.tsx
```diff
@@ -0,0 +1,425 @@
...
+        <FormSectionDescription>Please add the details of your contact information.</FormSectionDescription>
+        <FormControl>
+          <Label htmlFor={seed('phone-number')}>Phone number</Label>
+          <Input id={seed('phone-number')} type="tel" insertAfter={<InformationIcon decorative />} />
+        </FormControl>
+        <FormControl>
+          <Label htmlFor={seed('email')}>Email address</Label>
+          <Input id={seed('email')} type="email" />
+        </FormControl>
+        <MultiselectCombobox
```
gloriliale, line 124: question: to clarify, you don't need to use `FormControl` with `MultiselectCombobox` bc it controls its own label, is that correct? What would happen if there _were_ a `FormControl` around it? (i'm thinking nothing, but wanted to check).

packages/paste-core/components/form/stories/index.stories.tsx
```diff
@@ -0,0 +1,425 @@
...
+    </Form>
+  );
+};
+
+export const ErrorForm: StoryFn = () => {
+  const seed = useUIDSeed();
+
+  return (
+    <Form aria-labelledby={seed('form-heading')}>
+      <Heading id={seed('form-heading')} variant="heading30" as="h3">
```
gloriliale, line 225: nitpick: could we do `marginBottom="space0"` here too?

packages/paste-core/components/form/stories/index.stories.tsx
```diff
@@ -0,0 +1,425 @@
...
+        </CalloutList>
+      </Callout>
+      <FormControl>
+        <Label htmlFor={seed('street-address')} required>
+          Street address
+        </Label>
+        <Input id={seed('street-address')} type="text" hasError required />
+        <HelpText variant="error">Please add a valid street address.</HelpText>
+      </FormControl>
+      <FormControlTwoColumn>
```
gloriliale, line 243: suggestion: it might be nice to show what an error would look like in a two column form control

packages/paste-core/components/form/stories/index.stories.tsx
```diff
@@ -0,0 +1,425 @@
...
+              ))}
+            </Select>
+          </FormControl>
+          <FormControl>
+            <Label htmlFor={seed('zip')}>Zip code</Label>
+            <Input id={seed('zip')} type="text" />
+          </FormControl>
+        </FormControlTwoColumn>
+      </FormSection>
+      <FormControl>
```
gloriliale, line 340: thought: I notice that having a dangling `FormControl` after a `FormSection` makes that control seem still related to the previous section spacing-wise, since the two are still the standard space80 apart (the section ends before **Select payment method.** 
image

Is that something we should guide people not to do in the docs?

gloriliale (commented):
This is gonna be revolutionaryyyy! Left a few comments here and there, nothing huge though

packages/paste-core/components/form/__tests__/index.spec.tsx
shleewhite: This is because the fieldset element (which is rendered by FormSection) has role="group" implicitly and then when a fieldset has a legend (which is rendered by FormSectionHeading) as the first child, that legend is automatically assumed to be the name of the group.

To clear up confusion, I changed the Form's aria-label attribute to be "My Form" instead of also being "settings".

packages/paste-core/components/form/src/FormControlTwoColumn.tsx
shleewhite: Oops, will fix

packages/paste-core/components/form/stories/index.stories.tsx
shleewhite: @kelseycavitt What do you think about this? Should we add guidance about this in the docs?

packages/paste-core/components/form/stories/index.stories.tsx
shleewhite: added!

gloriliale (approved):

Niznikr (approved):

gloriliale (approved):

Niznikr (dismissed):

TheSisb (approved):

gloriliale (approved):

PR #2926, opened 2023-01-06
fix: react 18 - fix reakit ssr issues with portals

adamb-92:
# Reakit SSR Fix

We realised that Tooltip component was failing when using Nextjs 13 (React 18).

We found these two GitHub issue:

- [https://github.com/ariakit/ariakit/issues/656](https://github.com/ariakit/ariakit/issues/656)
- [https://github.com/ariakit/ariakit/issues/844](https://github.com/ariakit/ariakit/issues/844) 

There's a linked PR to the first one but it is still a draft: [https://github.com/ariakit/ariakit/pull/802](https://github.com/ariakit/ariakit/pull/802)

We managed to reproduce that PR fix on [codesandbox](https://codesandbox.io/p/sandbox/heuristic-knuth-lk4m55) changing the dependency in the package.json to that commit:

```json
"reakit": "https://pkg.csb.dev/reakit/reakit/commit/b6285795/reakit"
```

This PR aims to integrate those changes into @twilio-paste/reakit-library, using the fix above wouldn't work well as the base of the branch of the `b6285795` commit is some releases older.

**Contributing to Twilio**

### DIFF
--- a/packages/paste-libraries/reakit/src/Tooltip/Tooltip.tsx
+++ b/packages/paste-libraries/reakit/src/Tooltip/Tooltip.tsx
@@ -1,0 +1,78 @@
+import * as React from 'react';
+import {createComponent} from 'reakit-system/createComponent';
+import {createHook} from 'reakit-system/createHook';
+import {useForkRef} from 'reakit-utils/useForkRef';
+import {getDocument} from 'reakit-utils/getDocument';
+
+import {useDisclosureContent} from '..';
+import type {TooltipStateReturn, DisclosureContentHTMLProps, DisclosureContentOptions} from '..';
+import {Portal} from '../Portal/Portal';
+import {TOOLTIP_KEYS} from './__keys';
+import globalState from './__globalState';
+
+export type TooltipOptions = DisclosureContentOptions &
+  Pick<Partial<TooltipStateReturn>, 'unstable_popoverRef' | 'unstable_popoverStyles'> & {
+    /**
+     * Whether or not the tooltip should be rendered within `Portal`.
+     */
+    unstable_portal?: boolean;
+  };
+
+export type TooltipHTMLProps = DisclosureContentHTMLProps;
+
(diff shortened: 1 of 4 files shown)


eyerean (approved):

packages/paste-libraries/reakit/src/Tooltip/Tooltip.tsx
```diff
@@ -0,0 +1,78 @@
+import * as React from 'react';
+import {createComponent} from 'reakit-system/createComponent';
+import {createHook} from 'reakit-system/createHook';
+import {useForkRef} from 'reakit-utils/useForkRef';
+import {getDocument} from 'reakit-utils/getDocument';
+
+import {useDisclosureContent} from '..';
```
TheSisb, line 7: issue (blocking): we should import these types from reakit directly, because right now there's a circular dependency.

Tooltip.tsx imports from `..` (index.tsx) and index.tsx imports from Tooltip.tsx.

This can cause issues

packages/paste-libraries/reakit/src/index.tsx
```diff
@@ -10,6 +11,7 @@ export {
   CompositeGroup,
   CompositeItem,
   // https://reakit.io/docs/disclosure/
+  useDisclosureContent,
```
TheSisb, line 14: This file exposes these types to consuming applications, so unless we have an explicit need for consumers to use these new disclosure exports, we should avoid doing it here.

packages/paste-libraries/reakit/src/Tooltip/Tooltip.tsx
adamb-92: Ahhh fair enough, I'll update this and all other occurrences where this is happening

packages/paste-libraries/reakit/src/index.tsx
adamb-92: Makes sense! I'll remove all the types that have been added here

Niznikr (approved):

PR #2933, opened 2023-01-11
fix: AIONAPFMWK-171: fix Hydration issues SSR on Paste Remix

espipj:
As reakit is not releasing a fix we've used patch-package to fix the reakit Portal build files. In the process changes from https://github.com/twilio-labs/paste/pull/2926 were reverted.

Also installed `postinstall-postinstall` package, see reason here: https://github.com/ds300/patch-package#why-use-postinstall-postinstall-with-yarn

Hopefully this fixes can be merged and released on the main reakit package in the following PR https://github.com/ariakit/ariakit/pull/2198 and then we can revert this config

Some smaller fixes were made:

- Storybook quiet mode
- Option/Select only client side **important** see: https://github.com/facebook/react/issues/24519#issuecomment-1122780621 this needs further investigation in the future
- Main modal (with video) only showing on the client 

How to test:
- Upgrade Paste Remix nextjs version to "^13"
- Change the @twilio-paste/core to be a link to paste-core-bundle folder
- build everything
- run Paste Remix Dev

**Contributing to Twilio**


eyerean (approved):

Niznikr (commented):
While this solution works for development I believe more work is needed to ensure `paste-libraries` has the patch applied when published. [This approach is worth exploring](https://github.com/ds300/patch-package/issues/198#issuecomment-873975267).

Niznikr (approved):

PR #2965, opened 2023-01-20
feat: add support for React 18

Niznikr:
- Update peer dependencies to include React 18
- Update dev dependencies to use React and React DOM 18
- Run unit tests for React 16, 17, and 18
- Update component types to [explicitly include children ](https://solverfox.dev/writing/no-implicit-children/)
- Update `@twilio-paste/uid-library` to detect and use React's `useId` hook if available
- Update to use [Reakit version with portal ssr fix](https://www.npmjs.com/package/@niznikr/reakit)
- Upgrade to [Next 13](https://nextjs.org/blog/next-13) to address hydration issues


.changeset/lucky-years-doubt.md
```diff
@@ -0,0 +1,96 @@
...
+'@twilio-paste/spinner': minor
+'@twilio-paste/switch': minor
+'@twilio-paste/table': minor
+'@twilio-paste/tabs': minor
+'@twilio-paste/textarea': minor
+'@twilio-paste/time-picker': minor
+'@twilio-paste/toast': minor
+'@twilio-paste/tooltip': minor
+'@twilio-paste/truncate': minor
+'@twilio-paste/core': minor
```
TheSisb, line 60: Let's make this a major for all

gloriliale (approved):

.changeset/lucky-years-doubt.md
SiTaggart: OK, I won't bother amending my changeset. Let's bundle in a few majors together

packages/paste-core/components/badge/stories/index.stories.tsx
```diff
@@ -26,7 +26,7 @@ const Wrapper = styled.div(
   })
 );
 
-export const AllBadges: React.FC = () => (
+export const AllBadges = (): JSX.Element => (
```
SiTaggart, line 29: suggest (non-blocking): If you were going to the effort of swapping all these, you _could_ swap them to 

```
export const AllBadges:StoryFn = () => (
```

Which returns a JSX.Element

packages/paste-core/components/data-grid/__tests__/index.spec.tsx
```diff
@@ -227,9 +260,13 @@ describe('Data Grid', () => {
...
       expect(header.getAttribute('aria-sort')).toBe('ascending');
-      headerSort.click();
+      act(() => {
+        headerSort.click();
+      });
       expect(header.getAttribute('aria-sort')).toBe('descending');
-      headerSort.click();
+      act(() => {
+        headerSort.click();
+      });
```
SiTaggart, line 269: praise: Nice. Wonder if we're seeing some speed improvements on these tests now?

packages/paste-core/components/select/src/Select.tsx
```diff
@@ -102,7 +107,7 @@ const Select = React.forwardRef<HTMLSelectElement, SelectProps>(
             size={multiple ? size : 0}
             variant={variant}
           >
-            {children}
+            {showOptions && children}
```
SiTaggart, line 110: question: should this have it's own changeset? I know its for hydration, but if someone upgrades would it be useful to us or them to have a call out specifically in this components changelog mentioning this change?

packages/paste-libraries/reakit/package.json
```diff
@@ -25,16 +25,16 @@
     "tsc": "tsc"
   },
   "dependencies": {
-    "reakit": "1.3.11",
+    "@niznikr/reakit": "1.3.12",
```
SiTaggart, line 28: question(non-blocking): I'm starting to wonder if when we need to fork things whether we should fork them to twilio-labs or a twilio namespace. This is the second personal fork we now rely on

SiTaggart (dismissed):

packages/paste-core/components/select/src/Select.tsx
Niznikr: Yeah a changeset could be helpful for this!

packages/paste-libraries/reakit/package.json
Niznikr: All for having them under `twilio-labs`.

Niznikr:
@SiTaggart Upgrading `@niznikr/reakit` to `1.3.18` should resolve the TS/build issue and this PR should be good to go.

cypress/support/commands/parent-commands.ts
```diff
@@ -98,7 +98,7 @@ Cypress.Commands.add('getInFixedContainer', (selector) => {
 
 Cypress.Commands.add('visualRegressionTestUrl', ({url, testName}) => {
   cy.visit(url);
-  cy.wait(1000);
+  cy.wait(2000);
```
Niznikr, line 101: @SiTaggart Percy was catching regressions as a result of [R18 stricter Strict Mode](https://reactjs.org/blog/2022/03/29/react-v18.html#new-strict-mode-behaviors) which does an extra unmount/remount. We want to keep it on to catch concurrency bugs for code resiliency. As such, we need to bump the wait time (might need more) to ensure all re-rendering is done.

SiTaggart (approved):

TheSisb (approved):

PR #3027, opened 2023-02-08
ci: enable TurboSnap for Chromatic

Niznikr:
To reduce the number of snapshots taken during Chromatic for faster checks and prevent overage costs, we can enable [TurboSnap](https://www.chromatic.com/docs/turbosnap). Going to keep it at `pull_request` for now for us to stress-test it and figure out how we might want to use it on `push` in a future iteration (since it requires packages to be built first). Verified `pull_request` works for this with the Chromatic team and based off [their example](https://github.com/chromaui/chromatic-cli/blob/main/.github/workflows/chromatic-main-and-prs.yml).

To test and see the generated dependency graph:

1. Run `yarn build:storybook`
2. Inspect `docs/preview-stats.json`


.github/workflows/on_pull_request.yml
```diff
@@ -419,6 +419,11 @@ jobs:
           buildScriptName: "build:storybook"
           autoAcceptChanges: "main"
           exitOnceUploaded: true
+          onlyChanged: true
+          externals: |
+            - 'packages/paste-design-tokens/tokens/**'
+            - 'packages/paste-icons/src/**'
```
Niznikr, line 425: These files are pre-processed/built and so require to be specified as [externals](https://www.chromatic.com/docs/turbosnap#specify-external-files-to-trigger-a-full-re-test-when-they-change).

nkrantz (dismissed):

gloriliale (dismissed):

gloriliale (approved):

nkrantz (dismissed):

SiTaggart (approved):

gloriliale (approved):

Issue #3036, opened 2023-02-14
Evergreen Theme's Modal backgroundColor is solid and covers page


brandongregoryscott:
### Description

When using the Evergreen theme, the background color of the overlay is completely solid and removes context of the page it's on. 

### Link to Reproduction

https://codesandbox.io/s/evergreen-theme-modal-background-xj2kdo

### Steps to reproduce

1. Click on the Open modal button
1. Observe the overlay background completely cover the page behind it

<img width="729" alt="image" src="https://user-images.githubusercontent.com/11774799/218757246-1f18db65-3d34-4c72-8f2f-756c8a9a65b1.png">

### Paste Core Version

17.7.0

### Browser

Google Chrome 109.0.5414.119 (Official Build) (x86_64) 

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

I don't think this is intended behavior based on how the modal works with the default or dark themes. I was able to work around it by wrapping the `Modal` in a `CustomizationProvider`:

```tsx
<CustomizationProvider
  baseTheme="evergreen"
  elements={{
    MODAL: {
      backgroundColor: "none",
    },
  }}
>
  <Modal>...</Modal>
```

nkrantz:
Thanks for bringing this to our attention @brandongregoryscott! I'm going to make a PR to fix this token so that you don't have to use the Customization Provider. I'll update this thread with the PR.

PR #3037, opened 2023-02-14
fix(CustomizationProvider): add missing theme types to baseTheme prop

brandongregoryscott:
While I was working around the Modal backgroundColor issue described in #3036, I noticed the type for the `CustomizationProvider.baseTheme` prop is out of date, limited to just `default | dark`, and requires a `@ts-ignore` comment to pass anything else. I've updated the type to include all of the values of `ThemeVariants`, but it's hard-coded. 

Something seems to be off with the TS config or version, because `ThemeVariants`, which I believe should resolve to a type union of the values of this constant object, just resolves to a string:

image

image

Happy to do some further digging to figure out what's going on here if we'd prefer to solve the root issue.

**Contributing to Twilio**

### DIFF
--- a/.changeset/eleven-deers-kiss.md
+++ b/.changeset/eleven-deers-kiss.md
@@ -1,0 +1,5 @@
+---
+'@twilio-paste/customization': patch
+---
+
+[CustomizationProvider] Add missing theme types to baseTheme prop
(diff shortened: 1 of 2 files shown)


brandongregoryscott:
I should've looked around more - according to #3026, this is intentional.

PR #3050, opened 2023-02-23
fix: discussions templates

nkrantz:
Fixes the rest of the Discussions templates that are also broken.


loreina (approved):

loreina:
question(non-blocking): just curious, why do the titles have to be an empty string for it to work?

nkrantz:
@loreina removing the `description` line was the fix. I made the titles empty because that refers to the title of the discussion, not the title of the category, so the discussions were pre-titled "Office Hours" etc. I assumed you were referring to the title of the category though. Let me know if that was the intended behavior and I'll replace them!

shleewhite (approved):

TheSisb (approved):

PR #3053, opened 2023-02-24
feat: Minimizable Dialog updates

shleewhite:
- [x] update styles to one twilio
- [x] fix border radius so background color doesn't show up in corners of the dialog
- [x] add toggleMinimized function to the useMinimizableDialogState return to match the toggle function that comes from the useNonModalDialogPrimitiveState hook
- [x] break out the styledMinimizableDialog component out so that it can be used in VRT (similar to how the modal VRT stories work)
- [x] update the default and default minimized stories so theyre not all stacked on top of each other

**Contributing to Twilio**

### DIFF
--- a/packages/paste-core/components/minimizable-dialog/src/StyledMinimizableDialog.tsx
+++ b/packages/paste-core/components/minimizable-dialog/src/StyledMinimizableDialog.tsx
@@ -1,7 +1,8 @@
 import * as React from 'react';
 import {Box} from '@twilio-paste/box';
+import type {BoxProps} from '@twilio-paste/box';
 
-export const StyledMinimizableDialog: React.FC<{element?: string; children: React.ReactNode}> = ({
+export const StyledMinimizableDialog: React.FC<React.PropsWithChildren<Pick<BoxProps, 'element'>>> = ({
   element,
   children,
 }) => (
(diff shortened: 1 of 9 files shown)


TheSisb (dismissed):

TheSisb (dismissed):

nkrantz (approved):

loreina (dismissed):

packages/paste-core/components/minimizable-dialog/src/StyledMinimizableDialog.tsx
```diff
@@ -0,0 +1,21 @@
+import * as React from 'react';
+import {Box} from '@twilio-paste/box';
+
+export const StyledMinimizableDialog: React.FC<{element?: string; children: React.ReactNode}> = ({
```
SiTaggart, line 4: issue(blocker): This needs to be changed for React18. We can no longer do `React.FC` typings without `React.PropsWithChildren`

```
React.FC<React.PropsWithChildren<AvatarContentProps>>
```

SiTaggart: Side note, worries me it doesn't fail type checking

shleewhite: ooohh is this a react 18 thing?

loreina (approved):

packages/paste-core/components/minimizable-dialog/src/StyledMinimizableDialog.tsx
SiTaggart: Yup Yup. I have updated the VS code snippet too, but I don't know if anyone uses them

PR #3059, opened 2023-02-28
chore: animate SideModal component

TheSisb:
Currently passing  an `animated` prop to the `useSideModalState` hook does not allow the modal to close. [See this bug reproduction](https://codesandbox.io/s/new-pine-83hx99?file=/src/index.tsx)

Originally reported via discussion here: https://github.com/twilio-labs/paste/discussions/2839

Let's animate the SideModal ourselves instead so it leverages prefers-reduced-motion properly.


loreina (approved):
looks amazing

PR #3119, opened 2023-03-23
chore(label): add title to required dot

nkrantz:
Adds title to label's required dot


loreina (approved):

packages/paste-core/components/label/src/RequiredDot.tsx
```diff
@@ -23,6 +23,7 @@ export const RequiredDot: React.FC<React.PropsWithChildren<RequiredDotProps>> =
       cursor="pointer"
       display="flex"
       height="sizeIcon30"
+      title={i18nLabel || 'Required'}
```
SiTaggart, line 26: suggestion: Instead of this, I might make 'Required' the default value of `i18nLabel` when it's being destructured. Right now there is also a little bit of a bug in that if you don't use i18nLabel, the required dot screen reader only text is blank. Adding `Required` as default value would do what you want and fix that bug

nkrantz: fixed

loreina (approved):

SiTaggart (dismissed):

SiTaggart (approved):

PR #3139, opened 2023-03-31
feat: update @twilio-paste/design-tokens dep for VS Code extension

brandongregoryscott:
Hey team! The VS Code extension is a little out of date with the most recent design tokens package, so this PR bumps it to the latest version. It might be nice to have a workflow that automatically pushes out a PR to update this dependency when it changes, so we don't need to manually keep it updated.

**Contributing to Twilio**


apps/vs-code-intellisense/package.json
```diff
@@ -71,7 +71,7 @@
     "test": "node ./out/test/runTest.js"
   },
   "dependencies": {
-    "@twilio-paste/design-tokens": "^8.5.0",
+    "@twilio-paste/design-tokens": "^9.0.2",
```
SiTaggart, line 74: Hmmmm internal dependency versioning should be handled by changesets. I wonder if this is set up correctly as workspace? If it's not detected as a workspace package it won't get version updates.

Publishing to the Microsoft extension registry is a whole other thing

brandongregoryscott: The VS Marketplace is definitely a little clunky.

Do you have an example of how this dependency should be setup? 🤔

PR #3154, opened 2023-04-12
feat(sidebar): add a Sidebar component with several variant styles

TheSisb:
This PR adds a new Paste package: `@twilio-paste/sidebar`. This package exports three components:
- Sidebar
- PushSidebarContentWrapper
- OverlaySidebarContentWrapper

The Sidebar comes prebuilt with two variants: `default` or `compact`.  `compact` keeps a sliver of the sidebar visible when collapsed, but otherwise is the same.

The `Sidebar` component is position fixed, locked to the left side of the screen and taking the full vertical height. By default, Sidebar is an overlay element. In order for the Sidebar to push the page contents, we wrap the page content with the corresponding `Push-` or `OverlaySidebarContentWrapper`. These wrappers handle animating the content position when the sidebar opens or closes. They accept the same props as Sidebar.

All the components are stateless. State is managed by the consuming application.

The Sidebar is responsive. It occupies the full 100% width, behaves as overlay, and is default variant always on the smallest breakpoint (mobile). This allows nested components to respond accordingly by also using responsive props and expecting the behavior on the smallest breakpoint too.

---

This PR also changes the `theme`'s `breakpoints` from `rem` to `px` values. MediaQueries are more easily grokked as PX, and it simplifies development. Some tests, like snapshot tests, may need to be updated.


.changeset/proud-bulldogs-share.md
```diff
@@ -0,0 +1,6 @@
+---
+'@twilio-paste/sidebar': minor
```
nkrantz, line 2:
```suggestion
'@twilio-paste/sidebar': major
```

.changeset/new-owls-refuse.md
```diff
@@ -0,0 +1,6 @@
+---
+'@twilio-paste/core': patch
+'@twilio-paste/theme': patch
+---
+
+[Theme] theme breakpoint values are now represented as `px` values instead of `rem` values
```
nkrantz, line 6: suggestion: include the tip about tests needing to be updated because of this change if it's relevant to consumers

packages/paste-core/components/sidebar/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,163 @@
...
+    </Theme.Provider>
+  );
+};
+
+describe('Sidebar', () => {
+  /**
+   * PUSH
+   */
+  describe('Push Sidebar', () => {
+    it('should render collapsed', async () => {
```
nkrantz, line 47: question: why are all of these `it` blocks async?

packages/paste-core/components/sidebar/package.json
```diff
@@ -0,0 +1,60 @@
+{
+  "name": "@twilio-paste/sidebar",
+  "version": "0.0.1",
+  "category": "layout",
+  "status": "alpha",
```
nkrantz, line 5: question: is this intentionally alpha instead of prod?

nkrantz (commented):
couple small things but looks great!

.changeset/proud-bulldogs-share.md
TheSisb: Keeping it a minor until we get all the pieces in it.

nkrantz (approved):

packages/paste-core/components/sidebar/src/Sidebar.tsx
```diff
@@ -0,0 +1,104 @@
...
+import type {BoxProps} from '@twilio-paste/box';
+import {useSpring, animated} from '@twilio-paste/animation-library';
+import {useTheme} from '@twilio-paste/theme';
+import {useWindowSize} from '@twilio-paste/utils';
+
+import type {Variants} from './types';
+
+const StyledSidebar = React.forwardRef<HTMLDivElement, BoxProps>((props, ref) => (
+  <Box
+    aria-label="Main Navigation"
```
SiTaggart, line 12: issue: hard coded string for i18n

packages/paste-core/components/sidebar/src/Sidebar.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+    return (
+      <AnimatedStyledSidebar
+        {...safelySpreadBoxProps(props)}
+        ref={ref}
+        element={element}
+        width={['100%', isCompact && collapsed ? 'sizeSidebarCompact' : 'sizeSidebar']}
+        style={styles}
+        aria-label={props['aria-label']}
+        aria-expanded={!collapsed}
```
SiTaggart, line 97: thought: this might end up being on the trigger not the container, but we can figure that out when we have all the bits together

SiTaggart (commented):
Praise: I really like the API and the way it's turned out 🎉

packages/paste-core/components/sidebar/src/Sidebar.tsx
TheSisb: So `'aria-label'?: string;` is a prop on the Sidebar which gets passed down to this to overwrite it. Should I remove the default label?

packages/paste-core/components/sidebar/src/Sidebar.tsx
TheSisb: yeah i was wondering about that too. tbd we have the rest, agreed

packages/paste-core/components/sidebar/src/Sidebar.tsx
SiTaggart: Depends on if it's a mandatory prop. I can imagine people not realizing they need to supply a label and not overriding it. Even if they internationalise it, they won't know necessarily that we buried a hard coded string in the component for them to override.

SiTaggart (approved):

loreina (approved):

PR #3173, opened 2023-04-19
chore: inject core version into the DOM

SiTaggart:
It's difficult to debug consuming applications, especially when folks find dependency management so hard. By injecting the version number of Core into the DOM we can see and prove what version of core the application is really resolving too.

It's static string replacement at compile time using an esbuild plugin.

image

## Checklist

### DIFF
--- a/tools/build/esbuild.js
+++ b/tools/build/esbuild.js
@@ -1,4 +1,7 @@
+const path = require('path');
 const esbuild = require('esbuild');
+const {esbuildPluginVersionInjector} = require('esbuild-plugin-version-injector');
+
 const {PasteCJSResolverPlugin} = require('./plugins/PasteCJSResolver');
 const {EsmExternalsPlugin} = require('./plugins/EsmExternals');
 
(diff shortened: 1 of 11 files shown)


tools/build/esbuild.js
```diff
@@ -49,15 +53,21 @@ async function build(packageJson) {
...
+     */
     target: ['chrome66', 'firefox58', 'safari11', 'edge79', 'node12.19.0'],
     define: {
       'process.env.NODE_ENV': `"${process.env.NODE_ENV}"`,
     },
     external,
   };
 
+  const versionInjectorConfig = {
+    packageJsonPath: path.join(__dirname, '../../packages/paste-core/core-bundle/package.json'),
```
TheSisb, line 68: question/suggestion: can we do this for all the individual packages too? In case people have core installed, but also use something like theme or button individually as well? Or is that a limitation right now?

TheSisb (approved):

tools/build/esbuild.js
SiTaggart: That's the dream. What I wanted initially was a DOM node that looked like:

```html
<div data-paste-component="@twilio-paste/button" data-paste-component-version="9.0.0" data-paste-core-version="19.0.1" />
```

but the only way I could think about doing that was going into each component and tagging the outer most dom node in each package... and that seemed like a lot of work...

I can't think of a way to do it centerally because Box source isn't bundled into each package, it doesn't know anything about button using it, for example

nkrantz (dismissed):

nkrantz (approved):

PR #3178, opened 2023-04-20
feat(sidebar): add SidebarHeader component

TheSisb:
- [x] Adds the SidebarHeader component
- [x] Tests
- [x] Stories
- [x] Customization


packages/paste-core/components/sidebar/src/SidebarHeader.tsx
```diff
@@ -0,0 +1,62 @@
...
+        display="flex"
+        alignItems="center"
+        justifyContent="flex-start"
+        columnGap="space50"
+        overflow="hidden"
+        element={element}
+        ref={ref}
+      >
+        <Button size="icon" variant="inverse" element={`${element}_BUTTON`} onClick={onClick}>
+          {productIcon ? productIcon : <ProductHomeIcon size="sizeIcon20" decorative element={`${element}_ICON`} />}
```
nkrantz, line 35: suggestion: specify icon color

packages/paste-core/components/sidebar/src/SidebarHeader.tsx
```diff
@@ -0,0 +1,62 @@
...
+        </Button>
+        <Box
+          opacity={collapsed ? '0' : '1'}
+          element={`${element}_TEXT`}
+          fontSize="fontSize30"
+          fontWeight="fontWeightSemibold"
+          lineHeight="lineHeight30"
+          letterSpacing="-.02em"
+          as="h5"
```
nkrantz, line 44: question: why did you go with h5

packages/paste-core/components/sidebar/src/SidebarCollapseButton.tsx
```diff
@@ -0,0 +1,58 @@
...
+
+    return (
+      <Button
+        ref={ref}
+        size="icon"
+        variant="inverse"
+        element={element}
+        {...props}
+        aria-expanded={!collapsed}
+        aria-controls={sidebarId}
```
SiTaggart, line 32: praise: excellent inclusion of accessibility

packages/paste-core/components/sidebar/src/SidebarPushContentWrapper.tsx
```diff
@@ -48,4 +48,4 @@ export const PushSidebarContentWrapper = React.forwardRef<HTMLDivElement, PushSi
     );
   }
 );
-PushSidebarContentWrapper.displayName = 'PushSidebarContentWrapper';
+SidebarPushContentWrapper.displayName = 'SidebarPushContentWrapper';
```
SiTaggart, line 51: praise: ah yeah, these seem like really obvious names now.

packages/paste-core/components/sidebar/src/SidebarHeader.tsx
SiTaggart: yeah, I might have some caution on the use of ay heading level given document hierarchy will be unknown. We might just get away with a div here

packages/paste-core/components/sidebar/src/SidebarHeader.tsx
TheSisb: I think `currentColor` as default should work fine. Any reason why you recommend specifying it?

nkrantz (approved):

packages/paste-core/components/sidebar/src/SidebarHeader.tsx
serifluous: Does that allow a product to change the icon color? Like, say in the future if we do want each product to use the brand colors for the icon color?

TheSisb: Yup. Teams can pass in their own icon entirely with the `productIcon` prop.

PR #3205, opened 2023-04-27
feat(topbar): add package

TheSisb:
Adding the Topbar component as part of the Navigation UI Kit


.changeset/kind-suns-cross.md
```diff
@@ -0,0 +1,5 @@
+---
+'@twilio-paste/codemods': patch
```
nkrantz, line 2: question: why do you have a changeset for codemods?

packages/paste-core/components/topbar/stories/topbar.stories.tsx
```diff
@@ -0,0 +1,124 @@
...
+
+const TonsOfContent: React.FC = () => {
+  const content = [];
+  for (let i = 0; i < 10; i++) {
+    content.push(<Paragraph key={i}>Dummy content</Paragraph>);
+  }
+  return <>{content}</>;
+};
+
+export const PushDefaultTopbar: StoryFn = () => {
```
nkrantz, line 30: suggestion: can you add a "open sidebar" button to this story like you have in the sidebar stories? once it's collapsed there's no way to re-open it

packages/paste-core/components/topbar/stories/topbar.stories.tsx
```diff
@@ -0,0 +1,124 @@
...
+      </Sidebar>
+      <SidebarPushContentWrapper collapsed={pushSidebarCollapsed} variant="compact">
+        <Topbar>Topbar</Topbar>
+        <TonsOfContent />
+      </SidebarPushContentWrapper>
+    </>
+  );
+};
+
+export const OverlayDefaultTopbar: StoryFn = () => {
```
nkrantz, line 78: same here re: open sidebar button

packages/paste-core/components/topbar/src/types.ts
```diff
@@ -0,0 +1,2 @@
+// TODO
```
nkrantz, line 1: reminder: remove comment

SiTaggart, line 1: question: Still `TODO`?

.changeset/kind-suns-cross.md
TheSisb: So the package can be shipped to npm and for it to upgrade with this new component

SiTaggart (dismissed):

nkrantz (dismissed):

loreina (commented):
issue(blocking): sidebar should be 76px in height so the bottom border matches sidebar header. in the breakdown doc, I suggested making a new token `size-topbar` but open to suggestions on that. 

suggestion(non-blocking): could we add multiple items next to each other for the left and right actions in the topbar? just to test what the spacing between each is.

figma reference: https://www.figma.com/file/kLinP2s645exXC2BEebC7p/feat(Navigation)%3A-create-components?node-id=27966-198627&t=HtxSPXYhccvo3vnF-4

breakdown doc: https://docs.google.com/document/d/1HRdIXaZIwQrtl5JDC5KkPWJm19MxVjEINXbDWpm8j8Y/edit#bookmark=id.5xpq821wbn3m

TheSisb:
> issue(blocking): sidebar should be 76px in height so the bottom border matches sidebar header. in the breakdown doc, I suggested making a new token size-topbar but open to suggestions on that.

Hmmm I don't think we should constrain the height, but I'll add the token and set it to `min-height` if that's cool

> suggestion(non-blocking): could we add multiple items next to each other for the left and right actions in the topbar? just to test what the spacing between each is.

Topbar doesn't have any opinions on what is inside it, it's just the container + inner padding, so this isn't necessary I don't think

loreina (approved):

SiTaggart (approved):

PR #3224, opened 2023-05-10
feat(sidebar): add SidebarNavigation elements

TheSisb:
Still a work in progress. 
- [x] Changes the Sidebar to be a `display="flex" flexDirection="column"` element so that the navigation portion fills up the middle area and doesn't overflow the old positioned fixed collapse button area.
- [x] Add Disclosure based navigation components which can automatically determine if they are nested to set the appropriate styles.
- [x] Add smoother animations for opening/closing with nav items.
- [x] Navigation area has custom scrollbar styles
- [x] Navigation area automagically adds a shadow to the top/bottom of its container when there's items scrolled past.


packages/paste-core/components/sidebar/src/nav/constants.ts
```diff
@@ -0,0 +1,8 @@
+export const IconSizeFromHeading = {
```
TheSisb, line 1: todo: remove file

packages/paste-core/components/button/src/index.tsx
```diff
@@ -153,6 +153,7 @@ const ButtonContents: React.FC<React.PropsWithChildren<ButtonContentsProps>> = (
         justifyContent={buttonVariantHasBoundingBox ? null : 'center'}
         columnGap="space20"
         alignItems="center"
+        width="100%"
```
SiTaggart, line 156: question: I wonder if this will have any unintended downstream conseuqneces?

packages/paste-core/components/sidebar/src/navigation/SidebarNavigation.tsx
```diff
@@ -0,0 +1,78 @@
+import * as React from 'react';
+import {Box, safelySpreadBoxProps} from '@twilio-paste/box';
+import type {BoxProps} from '@twilio-paste/box';
+import {styled, css} from '@twilio-paste/styling-library';
+import type {ThemeShape} from '@twilio-paste/theme';
+
+const SidebarNavigationWrapper = styled.div(({theme}: {theme: ThemeShape}) => {
+  const {colorBackgroundInverse, colorBackgroundInverseStronger} = theme.backgroundColors;
+
+  return css({
```
SiTaggart, line 10: suggestion: maybe worth a little comment here explaining this is the scrollbars and the overflow shadow effects.

packages/paste-core/components/sidebar/stories/navigation.stories.tsx
```diff
@@ -0,0 +1,266 @@
...
+      </SidebarPushContentWrapper>
+    </Box>
+  );
+  /* eslint-enable react/jsx-max-depth */
+};
+Default.parameters = {
+  padding: false,
+};
+
+export const Compact: StoryFn = () => {
```
SiTaggart, line 153: question: should we also have a story demo the state hook?

packages/paste-core/components/sidebar/__tests__/sidebarNavigation.spec.tsx
```diff
@@ -0,0 +1,315 @@
...
+
+    // Anchor works
+    expect(anchorIcon).toHaveAttribute('href');
+
+    // Disclosure works
+    expect(disclosure).toBeInTheDocument();
+    expect(disclosure).toHaveStyleRule('display', 'none');
+  });
+
+  it('renders expanded SidebarNavigationItems correctly', () => {
```
SiTaggart, line 168: question: are there any aria-attributes you want to be ensuring work correctly with the user interaction?

packages/paste-core/components/sidebar/src/index.tsx
```diff
@@ -1,10 +1,20 @@
...
+export * from './header/SidebarHeaderIconButton';
+/* Footer */
+export * from './footer/SidebarCollapseButtonWrapper';
+export * from './footer/SidebarCollapseButton';
+/* Navigation */
+export * from './navigation/SidebarNavigation';
+export * from './navigation/SidebarNavigationDisclosure';
+export * from './navigation/SidebarNavigationDisclosureContent';
+export * from './navigation/SidebarNavigationDisclosureHeading';
+export * from './navigation/SidebarNavigationItem';
```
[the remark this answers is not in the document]
nkrantz: +1

packages/paste-core/components/sidebar/src/navigation/SidebarNavigation.tsx
nkrantz: would be nice to have a story for showing the scrolling too - I don't think any are long enough

packages/paste-core/components/sidebar/src/navigation/types.ts
```diff
@@ -0,0 +1,7 @@
+import type {DisclosurePrimitiveInitialState, DisclosurePrimitveStateReturn} from '@twilio-paste/disclosure-primitive';
+
+export interface SidebarNavigationDisclosureStateReturn extends DisclosurePrimitveStateReturn {
+  [key: string]: any;
```
nkrantz, line 4: question: does this just mean that any prop can be added to the state return? why does it need to be extended this way

packages/paste-core/components/button/src/ResetButton.tsx
```diff
@@ -29,9 +29,9 @@ const ResetButton = React.forwardRef<HTMLButtonElement, DirectButtonProps>(
       <Box
         ref={ref}
         width={fullWidth ? '100%' : 'auto'}
-        {...safelySpreadBoxProps(props)}
         {...ButtonStyleMapping[buttonState]}
         {...SizeStyles[size]}
+        {...(props as any)}
```
nkrantz, line 34: question: don't we have to make `DirectButtonProps` extend `BoxProps` in order for reset buttons to accept other props in ts?

packages/paste-core/components/button/src/index.tsx
TheSisb: I don't believe it will, but Storybook/Website VRT should catch anything if it does.

packages/paste-core/components/sidebar/src/navigation/types.ts
TheSisb: Copied from the main Disclosure. I don't recall the prior art reasoning, but just kept it the same.

packages/paste-core/components/button/src/ResetButton.tsx
TheSisb: It's a little more complicated than that. First, the props passed to Button go through `ButtonProps` which we don't want to change. Those props then get passed down to the varianted button component internally which use DirectButtonProps. DirectButtonProps can't extend HTMLButtonElement and BoxProps at the same time to begin with, but even if we managed it fix that somehow, TS will never expect ResetButton to get any of the BoxProps anyways because the parent component has tighter typings.

This would need a larger rethink, but it's fine for now imo.

packages/paste-core/components/sidebar/src/navigation/types.ts
SiTaggart: I don't think we need it, Nora and I discussed this for another component, seems redundant in disclosure too

SiTaggart (approved):
I'm not sure what if anything is left, but it looks pretty solid from here

nkrantz (approved):

loreina (approved):
Super great work!! Looks sooooo nice 😍 🙌

PR #3231, opened 2023-05-16
feat(topbar): add user dialog

nkrantz:
Add User Dialog component to the Topbar package


loreina (approved):

packages/paste-core/components/topbar/src/UserDialog.tsx
```diff
@@ -0,0 +1,129 @@
...
+  children: PropTypes.node.isRequired,
+  'aria-label': PropTypes.string.isRequired,
+  element: PropTypes.string,
+};
+
+export const useUserDialogState = (
+  props: NonModalDialogPrimitivePopoverInitialState
+): NonModalDialogPrimitiveStateReturn => {
+  return useNonModalDialogPrimitiveState({...props, placement: 'bottom-start', gutter: 0});
+};
```
SiTaggart, line 129: suggestion: I'd potentially split some (if not all) of these out into their own files, under a UserDialog folder. Just so you can better understand what each thing is doing in isolation

packages/paste-core/components/topbar/src/UserDialogUserInfo.tsx
```diff
@@ -0,0 +1,100 @@
...
+      </>
+    );
+  }
+);
+UserDialogUserInfo.displayName = 'UserDialogUserInfo';
+
+UserDialogUserInfo.propTypes = {
+  children: PropTypes.node.isRequired,
+  element: PropTypes.string,
+};
```
SiTaggart, line 100: suggestion: same here, probably split all these into their own files.

SiTaggart (commented):
Along with splitting things up into separate files, you need some customization stories and tests. Looking pretty sharp though

packages/paste-core/components/topbar/stories/userDialog.stories.tsx
```diff
@@ -0,0 +1,120 @@
...
+    <UserDialogContainer name="User Name" icon={UserIcon} baseId="i-am-user-dialog">
+      <UserDialog aria-label="user menu" data-testid="basic-user-dialog">
+        <UserDialogUserInfo>
+          <UserDialogUserName>Name</UserDialogUserName>
+          <UserDialogUserEmail>email@email.com</UserDialogUserEmail>
+        </UserDialogUserInfo>
+      </UserDialog>
+    </UserDialogContainer>
+  );
+};
```
SiTaggart, line 34: question: as we're extending the avatar do we need an image or `src` version to prove it works?

nkrantz: yup good catch

SiTaggart (approved):

PR #3260, opened 2023-06-02
fix(combobox): prevent selection of disabled items

TheSisb:
This PR fixes the issue where disabled items in Combobox could still be selected if using keyboard navigation.

Newer versions of Downshift [will allow this more easily](https://github.com/downshift-js/downshift/issues/1176), but the fix isn't released yet. The idea for this approach comes from [one of their issue topics](https://github.com/downshift-js/downshift/issues/1295#issuecomment-1109797910).


nkrantz (approved):
Will this require any changes when Downshift releases the functionality?

TheSisb:
> Will this require any changes when Downshift releases the functionality?

Not necessarily, but it may be an opportunity to simplify our code if/when we upgrade

PR #3303, opened 2023-06-30
chore(sidebar): improve animation and scrollbar styles

TheSisb:
* Do a lil spike on janky animation when compact nav collapses and words wrap 
* Make scrollbar narrower, so collapsed Flex nav doesn’t have scrollbar covering items 
* Put 4-8px above scrollbar so it doesn’t touch the Sidebar Header border.

### DIFF
--- a/packages/paste-core/components/sidebar/src/navigation/SidebarNavigationDisclosureHeading.tsx
+++ b/packages/paste-core/components/sidebar/src/navigation/SidebarNavigationDisclosureHeading.tsx
@@ -6,6 +6,7 @@ import {DisclosurePrimitive} from '@twilio-paste/disclosure-primitive';
 import type {BoxProps} from '@twilio-paste/box';
 import {useTheme} from '@twilio-paste/theme';
 
+import {SidebarContext} from '../SidebarContext';
 import {SidebarNavigationDisclosureContext} from './SidebarNavigationDisclosureContext';
 import {
   sidebarNavigationLabelStyles,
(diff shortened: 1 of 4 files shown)


packages/paste-core/components/sidebar/src/navigation/SidebarNavigationDisclosureHeading.tsx
```diff
@@ -22,10 +23,26 @@ export interface SidebarNavigationDisclosureHeadingProps extends React.Component
 
 const StyledDisclosureHeading = React.forwardRef<HTMLDivElement, SidebarNavigationDisclosureHeadingProps>(
   ({children, element = 'SIDEBAR_NAVIGATION_DISCLOSURE_HEADING', selected, icon, ...props}, ref) => {
+    const {collapsed, variant} = React.useContext(SidebarContext);
     const [shouldIconMove, setShouldIconMove] = React.useState(false);
     const {nested} = React.useContext(SidebarNavigationDisclosureContext);
     const isExpanded = props['aria-expanded'];
     const theme = useTheme();
+    const isCompact = variant === 'compact';
+    const [visible, setVisible] = React.useState(!isCompact ? true : !isExpanded);
```
TheSisb, line 32: Note: I need a variable that copies `collapsed` but is set after a delay. This hook and timeout is how I achieve this. I use this to adjust whiteSpace and opacity AFTER the sidebar is expanded, so that we don't see word-wrapping occur throughout the expand process. It only triggers at the end of the animation once. This makes the animation end look like a single clean-up step rather than the entire animation to cause several height reflows.

packages/paste-core/components/sidebar/src/navigation/SidebarNavigationDisclosureHeading.tsx
```diff
@@ -52,8 +69,16 @@ const StyledDisclosureHeading = React.forwardRef<HTMLDivElement, SidebarNavigati
...
         </Box>
         {icon ? icon : null}
-        <Box as="span" display="block" marginLeft="space20">
-          {children}
+        <Box
+          as="span"
+          display="block"
+          marginLeft="space20"
+          transition="all 120ms ease"
+          float={visible ? 'none' : 'left'}
```
TheSisb, line 77: Note: since the wrapping Box uses column-gap and this Box isn't actually hidden, when in collapsed state we see too much paddingRight (due to columnGap). Setting float causes this element to be removed from the flex column-gap property while it is hidden

SiTaggart (approved):

serifluous (approved):

PR #3309, opened 2023-07-06
feat(code-editor): add package

TheSisb:
This PR adds two new packages.

## Code Editor Library

This new library is a wrapper around... a wrapper. That's right, we wrap https://github.com/suren-atoyan/monaco-react which wraps https://microsoft.github.io/monaco-editor/.
Why do this? Well we want to use the Monaco Editor in React, hence the first wrapper. And then we want to use that in Paste so our consumers can roll their own implementations as needed. One such consumer is even ourselves :O

## EditableCodeBlock

This new package wraps the above library package and applies the Paste theme. This is our "controlled" editable code block component we recommend consumers use. If for whatever reason our provided defaults don't work, folks can fallback to the library.


packages/paste-website/next.config.mjs
```diff
@@ -28,6 +28,20 @@ const nextConfig = {
   experimental: {
     legacyBrowsers: false,
   },
+  // https://nextjs.org/docs/pages/api-reference/next-config-js/headers
```
TheSisb, line 31: Note: I sneak fix in this PR so we can iframe the website from other websites. Normally this feature is defaulted off for security, because you dont want people embedding login flows into their websites and stealing credentials. However our docs doesn't have a login flow and is just information based and open source. So this is fine.

packages/paste-libraries/code-editor/package.json
```diff
@@ -0,0 +1,38 @@
...
+    "dist"
+  ],
+  "scripts": {
+    "build": "yarn clean && NODE_ENV=production node build.js && tsc",
+    "build:js": "NODE_ENV=development node build.js",
+    "clean": "rm -rf ./dist",
+    "tsc": "tsc"
+  },
+  "dependencies": {
+    "@monaco-editor/react": "4.5.1",
```
SiTaggart, line 27: question: what tests does this package have? Are they worth copying? Reason I ask, is what tests do we need to know an upgrade of the package is "fine"? Do we need tests?

packages/paste-core/components/editable-code-block/src/EditableCodeBlock.tsx
```diff
@@ -0,0 +1,41 @@
...
+     * Call provided onMount function.
+     * This can be used to add custom language support, or to grab a
+     * reference to the ref (monaco) object.
+     */
+    onMount?.(editor, monaco);
+  };
+
+  return (
+    <Box element={element} borderRadius="borderRadius10" overflow="hidden">
+      <CodeEditor {...props} onMount={handleEditorDidMount} />
```
SiTaggart, line 36: question: same question here too, what tests do we need to be confident "this works"? Are there even tests we can write? What is our boundary, the stuff we implement vs the stuff they implement?

packages/paste-libraries/code-editor/package.json
TheSisb: it still renders and looks like an editor, which is covered by storybook vrt, then that integration test is sufficient

packages/paste-core/components/editable-code-block/src/EditableCodeBlock.tsx
TheSisb: We don't implement anything other than the theme, so same answer here.

SiTaggart (approved):

PR #3327, opened 2023-07-13
fix(theme): update inter font to support number emojis

TheSisb:
- [x] Fixes https://github.com/twilio-labs/paste/discussions/3323
- [x] Adds internal-docs on where to find information on updating fonts in the future
- [x] (scope creep) added some clarity to Heading docs differentiating between `variant` and `as` props

### DIFF
--- a/.changeset/moody-impalas-rescue.md
+++ b/.changeset/moody-impalas-rescue.md
@@ -1,0 +1,6 @@
+---
+'@twilio-paste/core': patch
+'@twilio-paste/theme': patch
+---
+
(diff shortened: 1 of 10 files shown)


SiTaggart (approved):

nkrantz (approved):
non-blocking: would be better to link to the instructions on the paste-fonts repo rather than just mention it in the internal docs

TheSisb:
> non-blocking: would be better to link to the instructions on the paste-fonts repo rather than just mention it in the internal docs

I know but I didnt want to link to it.

PR #3378, opened 2023-08-03
feat(slider): add new component

TheSisb:
Adds the new Slider package based on React-Spectrum's headless UI hooks.  

[Spec document](https://docs.google.com/document/d/1IYdOLIueQkmq_lZomlFm71Ae3FaST3_kX0XCRIowQqI/edit#heading=h.fmgccxbr3ska)


packages/paste-core/components/slider/src/Slider.tsx
```diff
@@ -0,0 +1,142 @@
...
+  disabled?: boolean;
+  hasError?: boolean;
+  hideRangeLabels?: boolean;
+  minValue?: number;
+  maxValue?: number;
+  step?: number;
+  /**
+   * Used to adjust how the numbers are rendered and interpreted.
+   * https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat \
+   */
```
SiTaggart, line 21: praise: useful TSDoc comments

packages/paste-core/components/slider/src/Slider.tsx
```diff
@@ -0,0 +1,142 @@
...
+  numberFormatter: Intl.NumberFormat;
+  value?: number;
+  onChange?: (value: number) => void;
+  /** Fired when the slider stops moving, due to being let go. */
+  onChangeEnd?: (value: number) => void;
+}
+
+export const Slider: React.FC<SliderProps> = (props) => {
+  const inputRef = React.useRef(null);
+  const trackRef = React.useRef(null);
```
SiTaggart, line 31: question: shouldn't we be forwarding and merging at least one of these refs for a consumer?

packages/paste-core/components/slider/src/helpers.ts
```diff
@@ -0,0 +1,16 @@
...
+    return 'colorBackgroundStrong';
+  } else if (states.error) {
+    return 'colorBackgroundError';
+  } else if (states.dragging) {
+    return 'colorBackgroundPrimary';
+  } else if (states.focused) {
+    return 'colorBackgroundPrimaryStronger';
+  } else if (states.hovered) {
+    return 'colorBackgroundPrimary';
+  }
```
SiTaggart, line 14: suggestion: would these nested else if statements all be better off as a sort of map? `sliderColorToken[props.states]` kinda?

packages/paste-core/components/slider/src/SliderThumb.tsx
```diff
@@ -0,0 +1,28 @@
...
+  return (
+    <Box
+      {...props}
+      backgroundColor={getColorToken(props)}
+      boxShadow={props.dragging || props.focused ? 'shadowFocus' : 'none'}
+      borderRadius="borderRadiusCircle"
+      width="20px"
+      height="20px"
+      marginTop="space10"
+      cursor="pointer"
```
SiTaggart, line 22: suggestion: this might need to change when disabled

packages/paste-core/components/slider/src/Slider.tsx
```diff
@@ -0,0 +1,142 @@
...
+          width="100%"
+          borderRadius="borderRadius20"
+          cursor="pointer"
+        >
+          <SliderThumb {...thumbProps} {...uiStateProps} element={`${element}_THUMB`}>
+            <ScreenReaderOnly>
+              <input
+                ref={inputRef}
+                {...inputProps}
+                aria-labelledby={undefined}
```
SiTaggart, line 129: issue: this also needs to be described by the help text. It might work because of the props being spread but our examples should show and implement this

packages/paste-core/components/slider/src/Slider.tsx
```diff
@@ -0,0 +1,142 @@
...
+      disabled: isDisabled,
+      error: props.hasError,
+      hovered,
+      focused,
+      dragging: isDragging,
+    };
+  }, [isDisabled, props.hasError, hovered, focused, isDragging]);
+
+  return (
+    <Box role="group" element={element}>
```
SiTaggart, line 81: suggestion: I don't think the group or presentation role really do anything for us here. might be safer to remove them.

packages/paste-core/components/slider/src/Slider.tsx
```diff
@@ -0,0 +1,142 @@
...
+          lineHeight="lineHeight30"
+          fontWeight="fontWeightSemibold"
+          color="colorTextWeak"
+        >
+          <Box as="span" element={`${element}_RANGE_LABELS_MIN`}>
+            {props.numberFormatter.format(minValue)}
+          </Box>
+          <Box as="span" element={`${element}_RANGE_LABELS_MAX`}>
+            {props.numberFormatter.format(maxValue)}
+          </Box>
```
SiTaggart, line 99: suggestion: this is a tricky one. Right now voice over is treating the "presentation" element as the parent or group. When I navigate to it what happens is I only get one element, which announces both the min and the max value in one go. I think it would be better if they were navigable separately (maybe its because they are `span` and it treats them as inline text), along with a little screen reader text desrcibing that one is the min and one is the max values. Ideally the min and max attributes on the input itself would do the job, then you could just hide this entirely, but it doesn't do anything in VO at least.

packages/paste-core/components/slider/src/helpers.ts
```diff
@@ -0,0 +1,16 @@
+import type {UIStates, BackgroundColorTokens} from './types';
+
+export const getColorToken = (states: UIStates): BackgroundColorTokens => {
```
serifluous, line 3: question/suggestion: How hard would it be to do a subtle animation on/off hover so the color change isn't as drastic?

packages/paste-core/components/slider/src/helpers.ts
```diff
@@ -0,0 +1,16 @@
...
+  if (states.disabled) {
+    return 'colorBackgroundStrong';
+  } else if (states.error) {
+    return 'colorBackgroundError';
+  } else if (states.dragging) {
+    return 'colorBackgroundPrimary';
+  } else if (states.focused) {
+    return 'colorBackgroundPrimaryStronger';
```
serifluous, line 11: when the slider is focused and hovered, the color also needs to be colorBackgroundPrimary

packages/paste-core/components/slider/stories/index.stories.tsx
```diff
@@ -0,0 +1,205 @@
...
+};
+
+export const Disabled = (): React.ReactNode => {
+  const [value, setValue] = React.useState<number>(0.01);
+  const id = useUID();
+
+  return (
+    <Form>
+      <FormControl>
+        <Label htmlFor={id}>Volume</Label>
```
serifluous, line 52: todo: Labels should be disabled in disabled stories and examples

packages/paste-core/components/slider/src/helpers.ts
TheSisb: Can't do. The order matters, and they're props being set to true/false and not string keys I can lookup as easily.

packages/paste-core/components/slider/src/Slider.tsx
TheSisb: Fixed in the docs PR, but brought it over to this PR. You are 100% correct. Great catch

TheSisb: Input storybook doesn't implement this well, but website examples do. Fixing storybook examples for Input.

packages/paste-core/components/slider/src/helpers.ts
TheSisb: very! Done.

packages/paste-core/components/slider/src/Slider.tsx
```diff
@@ -0,0 +1,154 @@
...
+        <StyledTrack
+          {...uiStateProps}
+          element={`${element}_TRACK`}
+          left={thumbProps.style?.left}
+          height="4px"
+          width="100%"
+          borderRadius="borderRadius20"
+          transition="background 150ms ease"
+        >
+          <SliderThumb {...thumbProps} {...uiStateProps} element={`${element}_THUMB`}>
```
nkrantz, line 135: suggestion: can we shift focus to the thumb when you interact with the slider? and/or when you click on the thumb

packages/paste-core/components/slider/stories/index.stories.tsx
```diff
@@ -0,0 +1,213 @@
...
+import {CustomizationProvider} from '@twilio-paste/customization';
+import {useTheme} from '@twilio-paste/theme';
+
+import {Slider} from '../src';
+
+// eslint-disable-next-line import/no-default-export
+export default {
+  title: 'Components/Slider',
+};
+
```
nkrantz, line 15: suggestion (non-blocking): can you add a story where the currently selected value is visible somewhere on the screen?

packages/paste-core/components/slider/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,140 @@
...
+import {CustomizationProvider} from '@twilio-paste/customization';
+
+import {Slider} from '../src';
+
+const onChangeMock: jest.Mock = jest.fn();
+const onChangeEndMock: jest.Mock = jest.fn();
+
+const PercentFormatter = new Intl.NumberFormat('en-US', {style: 'percent'});
+
+describe('Slider', () => {
```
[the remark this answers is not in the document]
TheSisb: Just the ones we're checking for now afaict. aria-describedby and aria-valuetext. Did you see something that I missed?

packages/paste-core/components/slider/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,140 @@
...
+      expect(screen.queryByRole('presentation')).toBe(null);
+
+      fireEvent.keyDown(input, {key: 'ArrowDown', code: 'ArrowDown'});
+      // Disabled so unchanged in call count
+      expect(onChangeMock).toHaveBeenCalledTimes(1);
+      expect(onChangeEndMock).toHaveBeenCalledTimes(1);
+    });
+  });
+
+  describe('Customization', () => {
```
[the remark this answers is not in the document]
nkrantz: oh nice

SiTaggart (approved):

packages/paste-core/components/slider/src/Slider.tsx
```diff
@@ -0,0 +1,160 @@
...
+        onClick={onClickHandler}
+      >
+        <StyledTrack
+          {...uiStateProps}
+          element={`${element}_TRACK`}
+          left={thumbProps.style?.left}
+          height="4px"
+          width="100%"
+          borderRadius="borderRadius20"
+          transition="background 150ms ease"
```
serifluous, line 138: issue: It seems like either the color transition isn't working on the track or appears to go at a much faster speed than the thumb.

packages/paste-core/components/slider/src/helpers.ts
```diff
@@ -0,0 +1,16 @@
+import type {UIStates, BackgroundColorTokens} from './types';
+
+export const getColorToken = (states: UIStates): BackgroundColorTokens => {
+  if (states.disabled) {
+    return 'colorBackgroundStrong';
+  } else if (states.error) {
+    return 'colorBackgroundError';
```
serifluous, line 7: question: Can't remember if you mentioned that there was no hover state on error, but I just noticed in chromatic 😬  Possible to add or nah? Sorry, didn't realize it wasn't in Figma.

If it is, we can use colorBackgroundErrorStrong like our other component error states.

serifluous (approved):

PR #3395, opened 2023-08-09
fix: try fix our requireJS clash with building for the node platform and not browser

SiTaggart:
Description and useful links to discussions / issues / tickets

### Change Platform target to browser

We currently target node as the platform. That has some consequences with the way esbuild handles [conditions](https://esbuild.github.io/api/#conditions) and the way it handles export maps.

What seems to be happening is that when it targets node, a packages export map may list a `module` path, but esbuild ignores it and just grabs the default or import. Depending on the way the library is built that means an es modules build of our ackage will grab and bundle a commonJS or UMD version of the library and not the es module version.

Switching the platform to `browser` means it will respect the export map and grab the actual es module version of any dependencies.

## Checklist


packages/paste-libraries/uid/package.json
```diff
@@ -24,7 +24,7 @@
     "tsc": "tsc"
   },
   "dependencies": {
-    "react-uid": "2.3.1"
+    "react-uid": "2.3.3"
```
SiTaggart, line 27: removes tslib as a dependency.

package.json
```diff
@@ -214,7 +214,8 @@
   },
   "resolutions": {
     "csstype": "3.0.11",
-    "playwright": "1.28.1"
+    "playwright": "1.28.1",
+    "tslib": "2.6.1"
```
SiTaggart, line 218: removes all the different versions

nkrantz (approved):
👍

PR #3414, opened 2023-08-14
fix: cypress flakes

TheSisb:
Trying to fix CI cypress flakes


cypress/integration/sitemap-vrt/constants.ts
```diff
@@ -37,7 +37,6 @@ export const SITEMAP = [
   '/components/display-heading/',
   '/components/display-pill-group/',
   '/components/disclosure/',
-  '/components/editable-code-block/',
```
TheSisb, line 40: Loading Monaco editor takes a while, so this page always times out. Since we added website page VRT to cover example breakage, but we now have the storybook example view and storybook has vrt through chromatic, we don't need this page anymore and it solves our flake.

SiTaggart (approved):

PR #3425, opened 2023-08-17
fix(input, input-box): allow passing `type="url"`

TheSisb:
Fixes https://github.com/twilio-labs/paste/discussions/3422


SiTaggart (changes requested):
issue(blocking): need to add a docsite example too and update the props table

TheSisb:
I knew I was forgetting something!

SiTaggart (approved):

PR #3430, opened 2023-08-21
feat(meter): create package

nkrantz:
Adds @twilio-paste/meter package with Meter functionality, stories and tests. Designs and stories to be finalized in a future PR.


packages/paste-core/components/meter/src/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+   * They can be added back into the Paste Meter API depending on the spec & designs.
+   *
+   * low?: number;
+   * high?: number;
+   * optimum?: number;
+   */
+}
+
+const Meter = React.forwardRef<HTMLMeterElement, MeterProps>(({element = 'METER', ...props}, ref) => {
+  const {label, showValueLabel = Boolean(label), value = 0, minValue = 0, maxValue = 100} = props;
```
TheSisb, line 27:
```suggestion
  const {label, showValueLabel = label != null, value = 0, minValue = 0, maxValue = 100} = props;
```

packages/paste-core/components/meter/src/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+  // Calculate the width of the bar as a percentage
+  const percentage = (value - minValue) / (maxValue - minValue);
+  const barWidth = `${Math.round(percentage * 100)}%`;
+
+  return (
+    <Box as="div" {...meterProps} role="meter" maxWidth="size30" element={`${element}_WRAPPER`}>
+      <Box display="flex" justifyContent="space-between" element={`${element}_LABEL_WRAPPER`}>
+        {label && (
+          // @ts-expect-error Not using Label's required `htmlFor` prop, using `aria-labelledby` instead
+          <Label {...labelProps} as="div" element={`${element}_LABEL`}>
```
TheSisb, line 39: suggestion/thought: you can copy Slider in that Label is a composition and not a prop, then you wont need the ts-expect-error because you would be using htmlFor.

packages/paste-core/components/meter/src/index.tsx
```diff
@@ -0,0 +1,58 @@
...
+const Meter = React.forwardRef<HTMLMeterElement, MeterProps>(({element = 'METER', ...props}, ref) => {
+  const {label, showValueLabel = Boolean(label), value = 0, minValue = 0, maxValue = 100} = props;
+  const {meterProps, labelProps} = useMeter(props);
+
+  // Calculate the width of the bar as a percentage
+  const percentage = (value - minValue) / (maxValue - minValue);
+  const barWidth = `${Math.round(percentage * 100)}%`;
+
+  return (
+    <Box as="div" {...meterProps} role="meter" maxWidth="size30" element={`${element}_WRAPPER`}>
```
TheSisb, line 35: thought: I wonder if this element should be `METER` and then the bar is `METER_BAR` and `METER_FILL`? Especially if the Label stuff is pulled out into a composition

packages/paste-core/components/meter/src/index.tsx
nkrantz: I was under the impression that `htmlFor` was for interactive labelled elements, which is why we were using `aria-labelledby` here. Which would also mean that consumers would have to ts-ignore their own Label components if they were composed and using `aria-labelledby`. @SiTaggart did I understand you correctly about that?

SiTaggart: Almost, just a minor technicality. Labels and the for attribute are for Form elements, Slider, _technially_ is an `input` element under the hood in the DOM. 

image

Meter and Progress bar aren't, they are just divs with roles so a `label` html element can't be used. It has to be a div or span, using the label styles and applying laballedby on the element that holds the role / replaces the `input`

nkrantz: Ok, so I think it makes the most sense to keep Label in the Meter package like it is right now rather than as a composition. If it was a composition, consumers would either have to ignore the required `htmlFor` prop or style their own labels.

SiTaggart: An alternative to that might be wrapping our label and creating the `MeterLabel` component. You get the same compositional API (it's flexible and not a prop), and you get to control the attributes like the element it renders. 

You could even fake the same API as label using `htmlFor` but your wrapped label component does all the "inside baseball" that a consumer probably doesn't need to know about, like the technicality of using HTML labels

SiTaggart: So to the consumer it's similar

```
<MeterLabel htmlFor={uid}>
  my label text
</MeterLabel>
<Meter id={uid} />
<HelpText>
  my help text
</HelpText>
```

But under the hood you do all the "right" things

packages/paste-core/components/meter/src/index.tsx
nkrantz: Yeah, this is the main thing I was hoping for feedback on. Even though this element is technically the meter, it includes both labels so I named it `METER_WRAPPER` because I figured the naming for customization was purely for applying styles and would make more sense to consumers who are trying to style the whole component, including the labels, even though it's not _technically_ the meter itself. But that being said I like your suggestion too and definitely can go in that direction!

packages/paste-core/components/meter/src/index.tsx
nkrantz: What about exporting it from the Label package, since Progress Bar (and potentially other future components) will need to use it too?

Calling it htmlFor even when it's aria-labelledby under the hood seems potentially misleading even if in most cases consumers don't need to know (or even understand) the difference.

But also, all of this is to avoid one ts-ignore, right? Is it necessary?

SiTaggart: I'm less fussed about the ts ignore, but more about what it's like to be a consumer. There's kind of been an underlying principle of the system that when picking up similar components they share a familiar API, so you kind of have to do less docs searching and mental cycles.

They shouldn't have to know they the label is technically different, and I don't think people will spend the time to understand it anyway.

I also don't think the Label package should know about it or how it has to be used by another package. It shouldn't care, it's a Meter and Progress only concern. Each of those exporting it's own label is fine IMO

packages/paste-core/components/meter/src/index.tsx
SiTaggart: FWIW if you look at the shadow dom of the native element, it's not too disimilar 

image

The outer element is the `meter` and then there are sub parts including the bar.

packages/paste-core/components/meter/src/index.tsx
nkrantz: I'm not convinced that the <MeterLabel> API is a familiar API but I'm fine with it... also it will probably be more complicated than just <MeterLabel> because there is also the value label and potentially min and max labels too, those are still open questions from the kickoff. Just to call that out.

I guess the ts-ignores will be present in the package in that case too

packages/paste-core/components/meter/src/MeterLabel.tsx
```diff
@@ -0,0 +1,27 @@
...
+export interface MeterLabelProps extends HTMLPasteProps<'div'>, Pick<BoxProps, 'element'> {
+  children: string;
+  htmlFor: string;
+}
+
+const MeterLabel = React.forwardRef<HTMLLabelElement, MeterLabelProps>(
+  ({element = 'METER_LABEL', children, htmlFor, ...labelProps}, ref) => {
+    return (
+      <Box display="flex" width="fit-content" element={`${element}_WRAPPER`}>
+        <Label {...labelProps} as="div" element={element} id={`${htmlFor}${labelIdSuffix}`} ref={ref}>
```
SiTaggart, line 17: suggestion: I suspect we need some tests around ensuring our slight of hand remains working when we do our labelling, both here and in the meter itself

packages/paste-core/components/meter/src/Meter.tsx
```diff
@@ -0,0 +1,69 @@
...
+  return (
+    <Box
+      as="div"
+      {...meterProps}
+      role="meter"
+      id={id}
+      maxWidth="size30"
+      position="relative"
+      element={element}
+      aria-labelledby={`${id}${labelIdSuffix}`}
```
TheSisb, line 44: thought: by making `id` required and doing this like this, we're effectively saying you cannot label a Meter with `aria-label` or a custom `aria-labelledby`. Do we still want to allow custom values for those labeling methods?

packages/paste-core/components/meter/src/MeterLabel.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+export interface MeterLabelProps extends HTMLPasteProps<'div'>, Pick<BoxProps, 'element'> {
+  children: string;
+  htmlFor: string;
+}
+
+const MeterLabel = React.forwardRef<HTMLLabelElement, MeterLabelProps>(
+  ({element = 'METER_LABEL', children, htmlFor, ...labelProps}, ref) => {
+    return (
+      <Box display="flex" width="fit-content" element={`${element}_WRAPPER`}>
```
TheSisb, line 16: question: I wonder if `display="flex" width="fit-content"` is equivalent to just doing `display="inline"`?

nkrantz: Seems fine with `display="inline"`! These styles will likely change after design work though

packages/paste-core/components/meter/src/MeterLabel.tsx
nkrantz: added some more but let me know if there's more coverage you want to see

packages/paste-core/components/meter/src/MeterLabel.tsx
nkrantz: actually just got rid of the label wrapper for now

packages/paste-core/components/meter/src/Meter.tsx
nkrantz: updated to allow custom labeling options

TheSisb (approved):
Excellent

PR #3442, opened 2023-08-24
chore: remove prop-types

TheSisb:
Removes all prop-types in every component. Marks these changes as major - breaking, with upgrade guide updated.

Also updates `pin-external-deps` danger script to display which package has the unpinned deps.

The danger check is a false positive.


packages/paste-core/core-bundle/progress-bar/package.json
```diff
@@ -0,0 +1,8 @@
+{
```
nkrantz, line 1: sneaky rebase

TheSisb: Doh

SiTaggart (approved):

tools/plop-templates/package.hbs
```diff
@@ -46,7 +46,7 @@
     "@twilio-paste/styling-library": "^2.0.2",
     "@twilio-paste/theme": "^10.0.2",
     "@twilio-paste/types": "^5.0.0",
-    "prop-types": "^15.7.2",
+    ,
```
SiTaggart, line 49: issue: looks like a whoopsie

SiTaggart (changes requested):

tools/plop-templates/package.hbs
TheSisb: fixed, good catch.

PR #3450, opened 2023-08-28
fix(combobox): require prop now works as intended

TheSisb:
Fixes issue where passing `required` would prevent the form submit event. 

Fixes: https://github.com/twilio-labs/paste/discussions/2885


nkrantz (approved):
👍

PR #3468, opened 2023-09-01
chore: upgrade cypress version to v13

SiTaggart:


TheSisb (approved):

TheSisb:
sidebar cypres tests all failing now that cypres is running again

PR #3525, opened 2023-10-05
chore: add github discussion auto-responder

TheSisb:
Uses our specially prompted ChatGPT4 endpoint to attempt rapid responses to incoming Github discussions. This doesn't mean we won't respond. A human will follow-up with more context or by affirming the bot response. This is in order to help people get answers they need, faster.

### DIFF
--- a/.github/workflows/on_discussion_create.yml
+++ b/.github/workflows/on_discussion_create.yml
@@ -1,0 +1,40 @@
+name: Discussion auto-responder
+
+on:
+  discussion:
+    types: [created]
+
+env:
+  GITHUB_TOKEN: ${{ github.token }}
+
+jobs:
+  on_discussion_open:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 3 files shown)


.github/workflows/on_discussion_create.yml
```diff
@@ -0,0 +1,40 @@
...
+    types: [created]
+
+env:
+  GITHUB_TOKEN: ${{ github.token }}
+
+jobs:
+  on_discussion_open:
+    runs-on: ubuntu-latest
+    permissions:
+      discussions: write
```
TheSisb, line 14: Note: needed to be able to write discussion comments

SiTaggart: If we're using the service account to post as, we probably don't need this permissions line.

SiTaggart (approved):

PR #3528, opened 2023-10-06
chore: remove description list vertical margin

nkrantz:
Possible visual breaking change so it's marked as a minor with clear instructions. Removes top and bottom margin from `<dl>` and replaces the `<dd>` last of type bottom margin with a bottom margin on the `DescriptionListSet` as long as it's not the last one in the `<dl>`.

### DIFF
--- a/.changeset/thirty-dots-reply.md
+++ b/.changeset/thirty-dots-reply.md
@@ -1,0 +1,6 @@
+---
+"@twilio-paste/description-list": minor
+"@twilio-paste/core": minor
+---
+
+[Description List] Adjust spacing on Description List component. Description List was previously inheriting 14px of top and bottom margin from the HTML element. This change removes that spacing, so the component now has 0px of top/bottom margin. The same amount of spacing remains between DescriptionListTerms and DescriptionListDetails. In order to maintain designs with Description Lists, you may have to replace the removed spacing to your layout.
--- a/packages/paste-core/components/description-list/src/DescriptionListSet.tsx
+++ b/packages/paste-core/components/description-list/src/DescriptionListSet.tsx
@@ -6,7 +6,7 @@ import * as React from "react";
 
 export const StyledDescriptionListSet = styled.div(
   css({
-    "& > dd:last-of-type": {
+    ":not(:last-of-type)": {
       marginBottom: "space60",
     },
     "& > dd:not(:last-of-type)": {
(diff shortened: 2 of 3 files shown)


SiTaggart (approved):

.changeset/thirty-dots-reply.md
```diff
@@ -0,0 +1,6 @@
+---
+"@twilio-paste/description-list": minor
+"@twilio-paste/core": minor
+---
+
+[Description List] Adjust spacing on Description List component. Description List was previously inheriting 14px of top and bottom margin from the HTML element. This change removes that spacing, so the component now has 0px of top/bottom margin. The same amount of spacing remains between DescriptionListTerms and DescriptionListDetails. In order to maintain designs with Description Lists, you may have to replace the removed spacing to your layout.
```
SiTaggart, line 6: praise: lovely detailed and clear changeset

serifluous (approved):

.changeset/thirty-dots-reply.md
TheSisb: +1

packages/paste-core/components/description-list/src/DescriptionListSet.tsx
```diff
@@ -6,7 +6,7 @@ import * as React from "react";
 
 export const StyledDescriptionListSet = styled.div(
   css({
-    "& > dd:last-of-type": {
+    ":not(:last-of-type)": {
```
TheSisb, line 9: question: I'm confused by this change. Can you explain what this does please?

nkrantz:
```
<DescriptionList>
  <DescriptionListSet>
    <DescriptionListTerm>term1</DescriptionListTerm>
    <DescriptionListDetails>details1a</DescriptionListDetails>
    <DescriptionListDetails>details1b</DescriptionListDetails>
  </DescriptionListSet>
  <DescriptionListSet>
    <DescriptionListTerm>term2</DescriptionListTerm>
    <DescriptionListDetails>details2</DescriptionListDetails>
  </DescriptionListSet>
</DescriptionList>
```

Previously, spacing was being set between Description List Sets by adding a larger bottom margin to the Description Details if it was the last one in the set (aka not on details1a). However that was also adding bottom margin to the entire component (coming from details2). This change makes the spacing that we want (between each DescriptionListSet) come from any non-last DescriptionListSet. So effectively there's no top or bottom margin to the DescriptionList but we still have the inner spacing we need.

PR #3549, opened 2023-10-13
feat(progress-bar): add new props and refine styles

TheSisb:
[ProgressBar] Added new features and updated the design.

The ProgressBarLabel component has the following new props:
- `valueLabel` which displays a plain text value on the right side of the label.
- `disabled` which applies disabled styling on the label.

The ProgressBar component has updated styling and the following new props:
- `hasError` which displays error styling.
- `disabled` which displays disabled styling.

### DIFF
--- a/.changeset/early-frogs-applaud.md
+++ b/.changeset/early-frogs-applaud.md
@@ -1,0 +1,14 @@
+---
+"@twilio-paste/progress-bar": minor
+"@twilio-paste/core": core
+---
+
+[ProgressBar] Added new features and updated the design.
+
+The ProgressBarLabel component has the following new props:
+- `valueLabel` which displays a plain text value on the right side of the label.
+- `disabled` which applies disabled styling on the label.
+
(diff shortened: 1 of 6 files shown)


serifluous (dismissed):

.changeset/early-frogs-applaud.md
```diff
@@ -0,0 +1,14 @@
+---
+"@twilio-paste/progress-bar": minor
+"@twilio-paste/core": core
```
nkrantz, line 3:
```suggestion
"@twilio-paste/core": minor
```

nkrantz (commented):
just the changelog typo.

non-blocking because sarah already approved, but it's really hard for me to tell that the indeterminate story is doing anything. It's hard to see the skeleton loader on that small of a space and feels like the animation is moving too slowly to notice it

TheSisb:
Yea, I agree about indeterminate. Looks like Sarah has plans to fast-follow with a fix.

nkrantz (approved):

PR #3557, opened 2023-10-16
chore: component types i-p

nkrantz:
- I-P types updates and docs site updates
- Pagination: use unused isHover and isFocus props
- Website: adjust ComponentHeader and SidebarNavigation to allow for name and description overrides on multiselect combobox docs page

**Note: ** ProgressBar docs not included in this PR because Shadi is working on them currently


packages/paste-core/components/input/src/Input.tsx
```diff
@@ -12,28 +12,134 @@ import { safelySpreadFormControlProps } from "./utils";
...
+   * @type {"space0"}
+   * @memberof InputProps
+   */
   padding?: "space0";
+  /**
+   *
+   * @type {BoxStyleProps["paddingRight"]}
+   * @memberof InputProps
+   */
   paddingRight?: BoxStyleProps["paddingRight"];
```
SiTaggart, line 114: Wooooow, why would we even do this? 🤔

packages/paste-core/components/list/src/index.tsx
```diff
@@ -1,5 +1,5 @@
-export * from "./List";
-export * from "./ListItem";
-export * from "./OrderedList";
-export * from "./UnorderedList";
-export * from "./types";
+export { List } from "./List";
+export { ListItem } from "./ListItem";
+export { OrderedList } from "./OrderedList";
+export { UnorderedList } from "./UnorderedList";
+export type { BaseListProps, ListItemProps, OrderedListProps, UnorderedListProps } from "./types";
```
SiTaggart, line 5: suggestion: `BaseListProps` is kind of redundant information when ordered and unordered lists are documented below. Do we use it else where outside of the package? If not we can probably not export it so only real components have API docs

packages/paste-core/components/menu/src/index.tsx
```diff
@@ -1,12 +1,33 @@
...
+type MenuStateReturnProps = MenuStateReturn;
+
+export type { MenuInitialStateProps, MenuStateReturnProps };
+
+export { useMenuState } from "./useMenuState";
+export type { MenuInitialState, MenuStateReturn } from "./useMenuState";
+export { Menu } from "./Menu";
+export { MenuButton } from "./MenuButton";
+export type { MenuButtonProps } from "./MenuButton";
+export { MenuBadgeContent, MenuBadge } from "./MenuBadge";
```
SiTaggart, line 14: suggestion: I don't think `MenuBadgeContent` is a real component that needs documenting

packages/paste-core/components/minimizable-dialog/src/index.tsx
```diff
@@ -1,5 +1,17 @@
-export * from "./MinimizableDialog";
-export * from "./MinimizableDialogButton";
-export * from "./MinimizableDialogHeader";
-export * from "./MinimizableDialogContainer";
-export * from "./MinimizableDialogContent";
+import type { MinimizableDialogStateReturn, UseMinimizableDialogStateArgs } from "./MinimizableDialogContainer";
+
+// for the sake of documenting the types we rename the state hook types to append Props to the name, so we can docuemnt them and not cause a breaking change.
+type MinimizableDialogStateReturnProps = MinimizableDialogStateReturn;
+type UseMinimizableDialogStateArgsProps = UseMinimizableDialogStateArgs;
```
SiTaggart, line 5: suggestion: could use a little alignment here. renaming `MinimizableDialogStateReturnProps` is the return of `useMinimizableDialogue`, but isn't named based on the hook, like the initial state it.

Might be better to use `useMinimizableDialogInitialState` and `useMinimizableDialogReturnState`

packages/paste-core/components/list/src/index.tsx
nkrantz: All good to remove, exporting that was a mistake

packages/paste-core/components/input/src/Input.tsx
nkrantz: yeahhhh no clue

packages/paste-core/components/menu/src/index.tsx
```diff
@@ -1,12 +1,33 @@
...
-export * from "./MenuSeparator";
-export * from "./SubMenuButton";
-export * from "./types";
+import type { MenuInitialState, MenuStateReturn } from "./useMenuState";
+
+// for the sake of documenting the types we rename the state hook types to append Props to the name, so we can docuemnt them and not cause a breaking change.
+type MenuInitialStateProps = MenuInitialState;
+type MenuStateReturnProps = MenuStateReturn;
+
+export type { MenuInitialStateProps, MenuStateReturnProps };
```
SiTaggart, line 7: praise: I kinda liked how some of these packages with hooks seemed to follow a pattern of hook docs first followed by components. Was that intentional?

nkrantz: just following the pattern from your first types PR of keeping them at the top of the file

packages/paste-core/components/menu/src/index.tsx
nkrantz: Update the status package only needs MenuBadgeContent (not MenuBadgeContentProps) so I removed the types export

SiTaggart (approved):

PR #3612, opened 2023-11-13
chore: fix customization tests

nkrantz:
**Contributing to Twilio**

### DIFF
--- a/packages/paste-customization/__tests__/CustomizationProvider.test.tsx
+++ b/packages/paste-customization/__tests__/CustomizationProvider.test.tsx
@@ -105,13 +105,13 @@ describe("CustomizationProvider", () => {
     render(
       <CustomizationProvider
         baseTheme="default"
-        elements={{ CARD: { backgroundColor: "colorBackground", variant: { padded: { padding: "space20" } } } }}
+        elements={{ CARD: { backgroundColor: "colorBackground", variants: { padded: { padding: "space20" } } } }}
       >
         <ThemeConsumerElementsExample />
       </CustomizationProvider>,
     );
     expect(screen.getByTestId("elements-test").textContent).toEqual(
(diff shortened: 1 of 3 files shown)


packages/paste-customization/__tests__/CustomizationProvider.test.tsx
```diff
@@ -105,13 +105,13 @@ describe("CustomizationProvider", () => {
...
         baseTheme="default"
-        elements={{ CARD: { backgroundColor: "colorBackground", variant: { padded: { padding: "space20" } } } }}
+        elements={{ CARD: { backgroundColor: "colorBackground", variants: { padded: { padding: "space20" } } } }}
       >
         <ThemeConsumerElementsExample />
       </CustomizationProvider>,
     );
     expect(screen.getByTestId("elements-test").textContent).toEqual(
-      'Elements: {"CARD":{"backgroundColor":"colorBackground","variant":{"padded":{"padding":"space20"}}}}',
+      'Elements: {"CARD":{"backgroundColor":"colorBackground","variants":{"padded":{"padding":"space20"}}}}',
```
nkrantz, line 114: I changed the test to use the correct prop instead of changing anything about the test because the test is doing what it is supposed to be doing. It would be ideal to type check for available style selectors, or `variants`, on the CustomizationProvider, but that would be a breaking change (?)

We already test that this functionality works on the component level in those packages so I didn't see the need to test that the variant styles are being applied in this package.

SiTaggart: Suggestion: looks like the story next to this test is the only other place I can find that suffers from this. That Story is very broken. It tries to prove that styling a custom element with variants works, but it targets a `CARD` where there is only BUTTON, and uses `variant` as the key.

If you could fix that story to be a useful, working VRT test for customization as part of this PR, that would be super rad!

SiTaggart (approved):

Issue #3697, opened 2024-01-04
Chat Message clips text if no spaces were added in the chat composer component


Viridian-Inc:
### Description

When I enter a multi-line message without spaces and submit it, I expect the chat message text will wrap, but the message is clipped.

### Link to Reproduction

https://paste.twilio.design/components/chat-composer

### Steps to reproduce

1. Go to ['chat-composer'](https://paste.twilio.design/components/chat-composer)
2. Scroll down to 'With Chat Log'
3. Enter a multi-line message without spaces in the chat compose component
4. Chat message clips the text to the width of the div

### Paste Core Version

20.7.0

### Browser

Google Chrome

### Operating System

- [ ] macOS
- [ ] Windows
- [ ] Linux

### Additional Information
![image](https://github.com/twilio-labs/paste/assets/89794917/74711d2a-9517-403f-9ebc-fb4f804a7893)

_No response_

TheSisb:
The fix will roll out in the next version of Paste. It will be a version higher than current 20.7.0

SiTaggart:
A fix was released in Core v20.8.0

PR #3723, opened 2024-01-16
fix(chatlog): fix overflow on message bubbles

TheSisb:
Fixes https://github.com/twilio-labs/paste/issues/3697


nkrantz (dismissed):

packages/paste-core/components/chat-log/src/ChatBubble.tsx
```diff
@@ -45,6 +45,9 @@ export const ChatBubble = React.forwardRef<HTMLDivElement, ChatBubbleProps>(
         paddingY="space30"
         paddingX="space40"
         marginBottom="space30"
+        wordWrap="break-word"
+        maxWidth="100%"
+        minWidth={0}
```
SiTaggart, line 50: suggestion: it might be useful to have a vrt example to cover this usecase.

PR #3727, opened 2024-01-22
feat(summary-detail): add package

TheSisb:
Adds the new SummaryDetail component package

- [x] tests
- [x] stories


packages/paste-core/components/summary-detail/src/SummaryDetailContent.tsx
```diff
@@ -0,0 +1,29 @@
+import { Box, type BoxProps, safelySpreadBoxProps } from "@twilio-paste/box";
+import { DisclosurePrimitiveContent } from "@twilio-paste/disclosure-primitive";
+import * as React from "react";
+
+import { SummaryDetailContext } from "./SummaryDetailContext";
+
+export interface SummaryDetailContentProps {
+  element?: BoxProps["element"];
```
SiTaggart, line 8: issue: missing TSDoc comment here

packages/paste-core/components/summary-detail/src/SummaryToggleButton.tsx
```diff
@@ -0,0 +1,31 @@
+import { Box, type BoxProps } from "@twilio-paste/box";
+import { Button } from "@twilio-paste/button";
+import { ChevronDisclosureIcon } from "@twilio-paste/icons/esm/ChevronDisclosureIcon";
+import * as React from "react";
+
+interface SummaryToggleButtonProps {
+  element?: BoxProps["element"];
```
SiTaggart, line 7: observation: I'm sure we have this else where, but it's the first time I recall noticing it. This component is a "private" component to the package. As a consumer you don't need to know about it... _except_ for the customization element name. As we don't need to export it, we don't document its API... but there is an API just for customization.

We don't need to do anything with that information, I'm just saying it out loud becuase I don't personally know what to do with that information right now

packages/paste-core/components/summary-detail/stories/index.stories.tsx
```diff
@@ -0,0 +1,87 @@
...
+  SummaryDetailContent,
+  SummaryDetailHeading,
+  type SummaryDetailInitialState,
+  type SummaryDetailStateReturn,
+  useSummaryDetailState,
+} from "../src";
+
+// eslint-disable-next-line import/no-default-export
+export default {
+  title: "Components/SummaryDetail",
```
SiTaggart, line 17: suggestion: I have this issue where I don't click save in Jira and my update doesn't persist. But I added an example from Flex so that we could create a real use case example story for VRT'ing.

packages/paste-core/components/summary-detail/src/SummaryToggleButton.tsx
TheSisb: Fixed with a refactor to make this exposes and composable

packages/paste-core/components/summary-detail/stories/index.stories.tsx
TheSisb: Thanks for that, changed up the design and examples

nkrantz (commented):
nice 💪

TheSisb:
@nkrantz Thanks for the review and catches, I'm rust 😊 
All fixed now

nkrantz (approved):

PR #3741, opened 2024-01-26
fix(summary-detail): fixes issue in react <18

TheSisb:
See memory leak issue here https://github.com/twilio-labs/paste/actions/runs/7670180581/job/20905756481
Suddenly happening as we added more tests which reached an upper bound by github actions

Research on jest memory leaks here: https://github.com/jestjs/jest/issues/7874

Updated our package.json with a fix that worked locally.


TheSisb:
So there was a leak in jest, which I fixed in the package.json scripts. However there was also a real issue with summaryDetail in React < 18. It was using `React.useId()` which isnt available. Furthermore, the test was crashing on the usage of `waitFor` for an action that didn't need it.

With all of these fixes, we're back on track!

SiTaggart (approved):

PR #3746, opened 2024-01-30
feat(textarea): add maxRows prop

TheSisb:
**Contributing to Twilio**


packages/paste-core/components/textarea/src/TextArea.tsx
```diff
@@ -8,7 +8,7 @@ import * as React from "react";
 
 type TextAreaVariants = "default" | "inverse";
 
-export interface TextAreaProps extends HTMLPasteProps<"textarea"> {
+export interface TextAreaProps extends Omit<HTMLPasteProps<"textarea">, "maxRows"> {
```
nkrantz, line 11: question (non-blocking): just wondering... if you didn't omit `maxRows` but still added it as a prop below as you are doing, would that have the same effect? would the extended prop be overwritten or vice versa?

nkrantz (approved):

PR #3779, opened 2024-02-16
chore: wrap try/catch around airtable data fetch for better debugging

TheSisb:
We were seeing confusing error on website builds:

image

These try/catches help us assess more clearly where the failure lies. Returns something like:

```
AirtableError {
  error: 'AUTHENTICATION_REQUIRED',
  message: 'You should provide valid api key to perform this operation',
  statusCode: 401
}
```

---
When reviewing, toggle whitespaces to see a simpler view

image


nkrantz (approved):
nice

PR #3827, opened 2024-03-20
feat(side-panel): create package

nkrantz:
Introduces the Side Panel component.


packages/paste-core/components/side-panel/src/SidePanel.tsx
```diff
@@ -0,0 +1,147 @@
...
+    const sidePanelId = id || useUID();
+
+    return (
+      <>
+        {transitions(
+          (styles, item) =>
+            item && (
+              <Box
+                position="absolute"
+                role="dialog"
```
nkrantz, line 102: @TheSisb I went with `role="dialog"` but not positive that's correct

packages/paste-core/components/side-panel/stories/index.stories.tsx
```diff
@@ -0,0 +1,223 @@
...
+      theme={theme}
+      elements={{
+        SPPW: { margin: "space50" },
+        SPPCW: { paddingRight: "space150" },
+        SP: { top: "30px" },
+        ANIMATED_SP_WRAPPER: { backgroundColor: "colorBackgroundPrimaryWeaker", borderRadius: "borderRadius80" },
+        INNER_SP: { paddingBottom: "space40" },
+        SPH: { borderBottomWidth: "borderWidth40" },
+        SPHA: { justifyContent: "flex-start" },
+        SPB: { width: "75%" },
```
henrieri, line 185: If there isn't a reason for using acronyms here, I think it would be preferable to see full component names.

henrieri (dismissed):
Approving, but follow up suggestions:

1. Provide a hook for state management (hopefully can make a standard hook with consistent API with other components that can be easily reused for other components with that behaviour.

2. Make sure elements namings for CustomizationProvider are consistent and clear, preferably component names themselves fully. I see that otherwise these are in upper snakecase. So maybe that, but I'm not sure what is the exact reasoning for upper camel case. EDIT: I see these are only in stories/tests, initially thought they might be part of a potential Customization API.

3. Make sure horizontal scrollbar doesn't appear when the side panel is sliding in.

4. In the documentation highlight when this should be used instead of Side Modal for example.

5. Add tests for both <Button /> components

packages/paste-core/components/side-panel/src/SidePanelHeader.tsx
```diff
@@ -0,0 +1,77 @@
...
+const SidePanelHeader = React.forwardRef<HTMLDivElement, SidePanelHeaderProps>(
+  ({ element = "SIDE_PANEL_HEADER", children, ...props }, ref) => {
+    const { sidePanelId, isOpen, setIsOpen, i18nCloseSidePanelTitle } = React.useContext(SidePanelContext);
+    return (
+      <Box
+        width="100%"
+        padding="space70"
+        borderBottomStyle="solid"
+        borderBottomColor="colorBorderWeaker"
+        borderBottomWidth="borderWidth10"
```
serifluous, line 55: todo: Remove bottom border from Header, and add a Separator . This is so people can add Tabs to Side Panel cleanly.

packages/paste-core/components/side-panel/stories/components/SidePanelWithContent.tsx
```diff
@@ -0,0 +1,87 @@
...
+          Assistant
+        </Heading>
+        <SidePanelHeaderActions>
+          <Button variant="secondary_icon" size="reset" onClick={() => {}}>
+            <MoreIcon decorative={false} title="open menu" size="sizeIcon50" />
+          </Button>
+        </SidePanelHeaderActions>
+      </SidePanelHeader>
+      <SidePanelBody>
+        <Box width="100%">
```
serifluous, line 32: question(maybe): Can we add negative margin here, or treat this as a placeholder for now until we have the rest of the AI components?

packages/paste-core/components/side-panel/src/SidePanelBody.tsx
```diff
@@ -0,0 +1,59 @@
...
+const SidePanelBody = React.forwardRef<HTMLDivElement, SidePanelBodyProps>(
+  ({ element = "SIDE_PANEL_BODY", children, ...props }, ref) => {
+    return (
+      <Box
+        {...safelySpreadBoxProps(props)}
+        as={SidePanelBodyWrapper as any}
+        width="100%"
+        height="100%"
+        paddingLeft="space40"
+        // @ts-expect-error no 6px spacing token but we need this amount of padding to offset the scrollbar
```
serifluous, line 44: question: why didn't we need to do this with Side Modal?

packages/paste-core/components/side-panel/stories/components/SidePanelWithContent.tsx
nkrantz: 👍 I'll add this to the AI log ticket

packages/paste-core/components/side-panel/src/SidePanelBody.tsx
nkrantz: image

Side Modal doesn't have a stylized scrollbar and Side Panel (like Sidebar) does. Side Modal right padding starts after the scrollbar, so the right side of the content of SideModalBody jumps to the left when the content starts to scroll.

I think stylized scrollbar without the jumping is a nicer experience but it's a really small difference and doing it the way Side Modal is built would avoid using the hard-coded px value. If you want to change it to be like Side Modal I'll add it to the refinement ticket

serifluous (approved):

PR #3834, opened 2024-03-27
chore: fix safari spinner bug

### LINKED ISSUE #3822: Spinner component spins off-center on Safari when browser window is zoomed
### Description

When I launched Console and navigated around I expected the spinners to behave normally but instead the black spinning part was off center from the underlying grey base.

This is reproducible on the paste documentation site as well, so seems like a Paste issue rather than Console.

It only seems to happen on Safari when the browser is zoomed to something other than 100%. Zooming on Chrome doesn't produce the same issue.

### Link to Reproduction

n/a

### Steps to reproduce

1. Open a Safari browser window
2. Go to ['https://paste.twilio.design/components/spinner'](https://paste.twilio.design/components/spinner)
3. Adjust your browser zoom level to zoomed in or zoomed out from actual size (in the menus that _Safari_ > _View_ > _Zoom in_ / _Zoom out_)
4. Watch the spinner
5. Note that the grey base and black animated spinner have resized inconsistently and the spinner is now spinning off center (see image for an example)

image

### Paste Core Version

whichever is on the paste docs site

### Browser

Safari 17.3.1 (19617.2.4.11.12)

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

_No response_

nkrantz:
Fixes #3822

### DIFF
--- a/packages/paste-core/components/spinner/src/styled.tsx
+++ b/packages/paste-core/components/spinner/src/styled.tsx
@@ -22,7 +22,7 @@ export const AnimatedStyledCircle = styled.circle<{ show: boolean; isSafari: boo
   };
 });
 
-export const StyledSvg = styled.svg<{ isSafari: boolean }>((isSafari) => {
+export const StyledSvg = styled.svg<{ isSafari: boolean }>(({ isSafari }) => {
   return {
     height: "100%",
     width: "100%",
--- a/packages/paste-core/components/spinner/src/styled.tsx
+++ b/packages/paste-core/components/spinner/src/styled.tsx
@@ -8,9 +8,12 @@ export const StyledCircleTrack = styled.circle({
   opacity: 0.25,
 });
 
-export const AnimatedStyledCircle = styled.circle<{ show: boolean }>(({ show }) => {
-  // `transformOrigin: "center"` does not work on Safari, so only apply the animation on non-Safari browsers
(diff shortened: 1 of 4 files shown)


packages/paste-core/components/spinner/src/styled.tsx
```diff
@@ -1,23 +1,31 @@
...
 
-export const AnimatedStyledCircle = styled.circle<{ show: boolean }>(({ show }) => ({
-  transformOrigin: "center",
-  animation: `1.5s ease-in-out infinite both ${CircleKeyframes}`,
-  strokeDasharray: circleCircumference,
-  opacity: show ? 1 : 0,
-}));
+export const AnimatedStyledCircle = styled.circle<{ show: boolean }>(({ show }) => {
+  // `transformOrigin: "center"` does not work on Safari, so only apply the animation on non-Safari browsers
+  const isSafari = navigator.userAgent.includes("Safari") && !navigator.userAgent.includes("Chrome");
```
TheSisb, line 13: This value should be set into state in a useEffect. Because on SSR navigator is undefined.

TheSisb: We should pass it into this and StyledSVG as a prop

nkrantz: done and done

PR #3857, opened 2024-04-11
feat(tabs): make non-fitted tabs responsive

TheSisb:
I went to great lengths to use a native scrollbar solution and style it such that I wouldn't need a pure JS built scrollbar.

In order to achieve this, I needed to add two wrapper divs to the TabList component. One for the outer scrollbar, and one for the bottom border. I still need some JS to fetch the total width of the TabList in order to position the bottom border correctly cross-browser, and to offset the scrollbar position by 4px on the outside of the element visually.

I also made the Tabs component scroll to the selected tab on mount, if it is off-screen. This initial scroll happens instantly to avoid screen noise. Subsequent tab navigation will smoothly scroll to the selected tab to provide a more refined experience.

Fitted tabs do not receive any significant changes, as by definition they shouldn't overflow or scroll.
Vertical tabs have similarly remained untouched.

I tested these changes thoroughly on Safari, Firefox, and Chrome, and checked the browser support tables for the features I'm using to verify they pass our requirements.

### Chrome

image

### Firefox

image

### Safari

image

### DIFF
--- a/packages/paste-core/components/tabs/src/TabList.tsx
+++ b/packages/paste-core/components/tabs/src/TabList.tsx
@@ -13,7 +13,7 @@ import { getElementName } from "./utils";
 /**
  * This wrapper applies styles that customize the scrollbar and its track.
  */
-const SidebarNavigationWrapper = styled.div(({ theme }: { theme: ThemeShape }) => {
+const StyledTabList = styled.div(({ theme }: { theme: ThemeShape }) => {
   const { colorBackgroundStronger, colorBackgroundInverseStronger } = theme.backgroundColors;
 
   return css({
@@ -90,7 +90,7 @@ const HorizontalTabList: React.FC<React.PropsWithChildren<{ variant?: Variants; 
   const isFitted = variant === "fitted" || variant === "inverse_fitted";
 
   return (
-    <Box as={SidebarNavigationWrapper as any} element={`${element}_SCROLL_WRAPPER`}>
+    <Box as={StyledTabList as any} element={`${element}_SCROLL_WRAPPER`}>
       <Box
         element={`${element}_CONTAINER`}
         borderBottomStyle="solid"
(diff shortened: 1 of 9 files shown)


packages/paste-core/components/tabs/src/TabList.tsx
```diff
@@ -1,13 +1,58 @@
...
 import { getElementName } from "./utils";
 
+/**
+ * This wrapper applies styles that customize the scrollbar and its track.
+ */
+const SidebarNavigationWrapper = styled.div(({ theme }: { theme: ThemeShape }) => {
+  const { colorBackgroundStronger, colorBackgroundInverseStronger } = theme.backgroundColors;
+
+  return css({
+    paddingBottom: "4px",
```
serifluous, line 20: issue: This isn't adding 4px extra space below the scrollbar:

Screenshot 2024-04-11 at 2 48 54 PM

(Screenshot in Chrome)

TheSisb: The padding is for the gap between the tab list and the scrollbar.

The gap between the tab list and the tab content is built into the TabPanel component.

serifluous: welp, wherever it needs to be, there should be an extra 4px between the scrollbar and the TabPanel so they're not right against each other when TabPanel paddingTop is 0

TheSisb: Oh I see, ok

TheSisb:
This is blocked on an accessiblity issue where my solution added wrappers for the scrollbar around the individual tabs and the tab wrapper. The tabs need to be direct children of the tab wrapper, and it is causing screenreaders to not correctly link the two elements together.

A solution would be to use aria- tags to connect them, but I didn't have great success with it in my preliminary tests. This requires more experimentation.

Another solution would be to move my wrappers outside the tabWrapper rather than between it and the tabs, and then update the styling accordingly to make this solution work. I haven't explored this, and is just an idea.

PR #3891, opened 2024-05-02
chore: fix supabase in ci

TheSisb:
This should hopefully fix CI flakes


TheSisb:
This doesn't fix the broken check yet, as I believe it's API key related and that is taking some time to regain access to, but this will help with us to stop hitting our free tier quota.

Issue #3882, opened 2024-04-29
Custom tooltips do not appear on top of everything


vpavlovskii:
### Description

Check the corresponding docs [section](https://paste.twilio.design/primitives/tooltip-primitive), for instance
<img width="459" alt="Screenshot 2024-04-29 at 15 26 39" src="https://github.com/twilio-labs/paste/assets/161604982/39042538-4e57-4e19-8e93-27f40d00ffaf">

### Link to Reproduction

https://paste.twilio.design/primitives/tooltip-primitive

### Steps to reproduce

1. Go to the corresponding docs [section](https://paste.twilio.design/primitives/tooltip-primitive)
2. Hover button from the first example
3. See the error

### Paste Core Version

latest

### Browser

Google Chrome 124.0.6367.93

### Operating System

- [X] macOS
- [ ] Windows
- [ ] Linux

TheSisb:
Hi @vpavlovskii,

Thanks for reporting this issue, I appreciate you taking the time!

The page you referenced is our Primitive Tooltip component, which is an unstyled and accessible basis upon which we build [our tooltip](https://paste.twilio.design/components/tooltip). The primitive tooltip is only concerned with API, positioning, and accessibility. In the example you reference, the primitive tooltip is appearing beneath our Sidebar due to having a lower z-index value. You can see in the following image that changing the z-index fixes the issue:

<img width="938" alt="image" src="https://github.com/twilio-labs/paste/assets/1592327/025732f1-1831-41a7-adde-f1b3c0e9779b">

If you're rolling your own Tooltip for any reason, I recommend looking at our Tooltip Component implementation and using the correct z-index token to make sure it is layered correctly.

TheSisb:
As a quick follow up, I will make some improvements to that page to demonstrate how one can style the primitive Tooltip.

TheSisb:
The website now includes the fix.

vpavlovskii:
Example for Dark theme displayed now as follows

<img width="452" alt="Screenshot 2024-05-02 at 13 02 47" src="https://github.com/twilio-labs/paste/assets/161604982/8b522533-26c8-47bb-8be4-e6bc6ed9c456">

PR #3892, opened 2024-05-06
Adds text color to Tooltip Primitive example

serifluous:
Fixes follow-up comment on #3882


serifluous:
Re-opening in #3893 to fix Vercel permissions

PR #3917, opened 2024-05-22
feat(Intput): increment/decrement reset buttons focus stylings

krisantrobus:
Adds correct focus stylings onto the Increment/Decrement buttons inside of Input component.

Accounts for inverse stylings of the buttons too.

Before:
image

After:
image

- [ X ] I acknowledge that all my contributions will be made under the project's license.

### DIFF
--- a/.changeset/wise-suns-serve.md
+++ b/.changeset/wise-suns-serve.md
@@ -1,0 +1,5 @@
+---
+"@twilio-paste/input": patch
+---
+
+[Input] updated focus stylings on increment/decrement buttons for number input type on both default and inverse variants
--- a/packages/paste-core/components/input/src/DecrementButton.tsx
+++ b/packages/paste-core/components/input/src/DecrementButton.tsx
@@ -3,12 +3,14 @@ import { Button, type ButtonProps } from "@twilio-paste/button";
 import { ChevronDownIcon } from "@twilio-paste/icons/esm/ChevronDownIcon";
 import type { HTMLPasteProps } from "@twilio-paste/types";
 import * as React from "react";
+import { InputVariants } from "./Input";
 
 export interface DecrementButtonProps extends HTMLPasteProps<"button"> {
   i18nStepDownLabel?: string;
   element?: BoxProps["element"];
   // Button component restricts tabIndex values
   tabIndex?: ButtonProps["tabIndex"];
+  variant?: InputVariants;
 }
 
 export const DecrementButton = React.forwardRef<HTMLButtonElement, DecrementButtonProps>(
@@ -22,8 +24,17 @@ export const DecrementButton = React.forwardRef<HTMLButtonElement, DecrementButt
         size="reset"
         type="button"
         borderRadius="borderRadius20"
-        backgroundColor="colorBackground"
+        backgroundColor={props.variant === "inverse" ? "colorBackgroundInverseStrong" : "colorBackground"}
         marginRight="space30"
+        _focus={{
+          borderRadius: "borderRadius20",
+          borderWidth: "borderWidth10",
+          borderStyle: "solid",
+          borderColor: props.variant === "inverse" ? "colorBorderInverseStrong" : "colorBorderPrimary",
+        }}
+        borderWidth="borderWidth10"
+        borderStyle="solid"
+        borderColor="transparent"
       >
         <ChevronDownIcon
           decorative={false}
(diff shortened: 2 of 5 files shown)


packages/paste-core/components/input/src/DecrementButton.tsx
```diff
@@ -22,8 +24,17 @@ export const DecrementButton = React.forwardRef<HTMLButtonElement, DecrementButt
...
         marginRight="space30"
+        _focus={{
+          borderRadius: "borderRadius20",
+          borderWidth: "borderWidth10",
+          borderStyle: "solid",
+          borderColor: props.variant === "inverse" ? "colorBorderInverseStrong" : "colorBorderPrimary",
+        }}
+        borderWidth="borderWidth10"
+        borderStyle="solid"
+        borderColor="transparent"
```
krisantrobus, line 37: I had to add a border to the buttons and make it transparent. After changing the _focus style I saw the icons would move slightly due to the border changing for one and not the other. Before an after shown in the video:
 https://github.com/twilio-labs/paste/assets/55083528/c99f3ecd-1f9e-4b0b-93a0-98066daf7409

serifluous: ooh ok yeah i was wondering about the use of border colors here & how you got it to work.

This solution works fine, but you could also use shadows instead. We have shadow border tokens (`shadowBorderInverseStrong` and `shadowBorderPrimary`) to account for this.

I have no like...eng perspective on which solution is better, so I'll rely on @nkrantz here for that.

serifluous: ok nvm I do have an opinion. Just realized it adds an extra 1px gray border all around the buttons.

todo: Let's use the shadow border tokens instead.

.changeset/wise-suns-serve.md
```diff
@@ -0,0 +1,5 @@
+---
+"@twilio-paste/input": patch
+---
+
+[Input] updated focus stylings on increment/decrement buttons for number input type on both default and inverse variants
```
nkrantz, line 6: praise: love the detailed changeset description, that's something we strive for and helps consumers

issue(blocking): we also include "@twilio-paste/core" on all changesets that are part of the core package. for almost all changes, the core change will be the same as the specific package (so in this case, it would be a patch to core), but for new packages (like the addition of the side panel component), we do a minor to core and a major to the new package, to avoid unnecessary breaking changes to core.

packages/paste-core/components/input/src/DecrementButton.tsx
```diff
@@ -22,8 +24,17 @@ export const DecrementButton = React.forwardRef<HTMLButtonElement, DecrementButt
         size="reset"
         type="button"
         borderRadius="borderRadius20"
-        backgroundColor="colorBackground"
+        backgroundColor={props.variant === "inverse" ? "colorBackgroundInverseStrong" : "colorBackground"}
```
nkrantz, line 28: nitpick: _very_ much non-blocking but when using another prop, in most cases we add it to the list of destructured props on line 17 so that we can use `variant` rather than `props.variant` (especially if we're already destructuring other specific props in the component)

packages/paste-core/components/input/stories/input.stories.tsx
```diff
@@ -713,6 +713,32 @@ export const DefaultNumberInput = (): React.ReactNode => {
 
 DefaultNumberInput.storyName = "Number Input - Controlled";
 
+export const DefaultNumberInverseInput = (): React.ReactNode => {
```
nkrantz, line 716: praise: new story for this missed use case 👏

.changeset/wise-suns-serve.md
nkrantz: whoops I was looking at an old commit, resolving this!

nkrantz (commented):
Great catch with the inverse variant!

packages/paste-core/components/input/src/DecrementButton.tsx
krisantrobus: Great suggestion 🙏  I'll make a change and push it up before we merge

packages/paste-core/components/input/src/DecrementButton.tsx
krisantrobus: Added boxShaow instead of border and that resolved all the spacing issues. Thank you! 🙇

serifluous (approved):

PR #3918, opened 2024-05-24
on-confirm-disabled-rename

cogwizzle:
Renaming the onConfirmDisabled property in AlertDialog to be isConfirmDisabled so that the property is more transparent how it should be used. onConfirmDisabled reads more like it expects a function. isConfirmDisabled reads more like it expects a boolean. https://github.com/twilio-labs/paste/discussions/3901

**Contributing to Twilio**


nkrantz (dismissed):
Thanks for this improvement and for branching in the repo!

nkrantz (approved):

PR #3923, opened 2024-05-29
chore(SidePanel): fix a11y contrast issue

krisantrobus:
Fix issues that occur with a11y color-contrast on Side Panel

### DIFF
--- a/packages/paste-core/components/side-panel/stories/index.stories.tsx
+++ b/packages/paste-core/components/side-panel/stories/index.stories.tsx
@@ -182,7 +182,7 @@ export const ContentDemo: StoryFn = () => {
   const mainContentSkipLinkID = useUID();
   const sidePanelId = useUID();
   return (
-    <Box backgroundColor="colorBackgroundBrand10">
+    <Box>
       <SidebarWithContent topbarSkipLinkID={topbarSkipLinkID} mainContentSkipLinkID={mainContentSkipLinkID} />
       <SidebarPushContentWrapper collapsed={true} variant="compact">
         <Topbar id={topbarSkipLinkID}> </Topbar>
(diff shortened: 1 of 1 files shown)


packages/paste-core/components/side-panel/stories/index.stories.tsx
```diff
@@ -1,28 +1,36 @@
...
   SidePanel,
   SidePanelBody,
   SidePanelButton,
   SidePanelContainer,
+  SidePanelContext,
   SidePanelHeader,
   SidePanelHeaderActions,
   SidePanelPushContentWrapper,
-} from "../src";
+} from "@twilio-paste/side-panel";
```
nkrantz, line 27: question: why this change?

packages/paste-core/components/side-panel/stories/index.stories.tsx
```diff
@@ -1,28 +1,36 @@
+import { Anchor } from "@twilio-paste/anchor";
+import { Avatar } from "@twilio-paste/avatar";
```
nkrantz, line 2: suggestion(non-blocking): probably best to keep these eslint-disables at the top of the file so eslint doesn't get angry about these 2 imports

packages/paste-core/components/side-panel/stories/index.stories.tsx
krisantrobus: I reviewed the other stories and they all used the same import format. I only now realize that is for all other components except the package it is changing. I will revert this

packages/paste-core/components/side-panel/stories/index.stories.tsx
krisantrobus: Great suggestion, addressed in latest PR

krisantrobus (commented):
👏 it's resolved! Great final fix for this

nkrantz (approved):

PR #3927, opened 2024-05-31
feat(ai-chat-log): add new AI Chat Log components

krisantrobus:
## AI Chat Log

Originally branched from Shadi's original [PR](https://github.com/twilio-labs/paste/pull/3873)

Adds several new components:
- AIChatLog
- AIChatLogger
- useAIChatLogger
- AIChatMessage
- AIChatMessageAuthor
- AIChatMessageBody
- AIChatMessageActionGroup
- AIChatMessageActionCard
- AIChatMessageLoading


packages/paste-core/components/ai-chat-log/src/useAIChatLogger.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+type PushAIChat = (chat: AIPartialIDChat) => void;
+type PopAIChat = (id?: string) => void;
+
+export type UseAIChatLogger = (...initialChats: AIPartialIDChat[]) => {
+  aiChats: AIChat[];
+  push: PushAIChat;
+  pop: PopAIChat;
+  clear: () => void;
+};
```
henrieri, line 22: I'm wondering about this API. Do you know why Shadi decided on this?

1. It seems that what I would otherwise call a message is considered an "Ai Chat" entity here. Usually I would think Chat is equal to Conversation, so one Chat would contain of many Messages.

So I would usually think messages would be more appropriate prop here. Also what OpenAI uses, some other LLMs for their API.

The second thing, I'm not sure if it's good to have initialChats as a parameter that alone is a spread array.

It would narrow down anything you could pass there in the future. I would suggest having an object to pass as config where you could have initialMessages prop. Like react-hook-form has initialValues.

packages/paste-core/components/ai-chat-log/__tests__/AIChatLogger.spec.tsx
```diff
@@ -0,0 +1,35 @@
...
+
+import { AIChatLogger, AIChatMessage, AIChatMessageBody } from "../src";
+import type { AIChat } from "../src/useAIChatLogger";
+
+const chats: AIChat[] = [
+  {
+    id: "uid1",
+    variant: "bot",
+    content: (
+      <AIChatMessage variant="bot">
```
henrieri, line 12: I'm wonder if API usage wise it would be possible to avoid having to pass the variant twice.

henrieri (commented):
Overall looks good, but I'm wondering about some of the API definition.

packages/paste-core/components/ai-chat-log/__tests__/AIChatLogger.spec.tsx
krisantrobus: Hey @henrieri, when using the hook you would have to define twice but if you're just using the logs you wouldn't need to. With the hook in the previous chat log it controlled the animation direction and we've kept the same logic for this.

This is currently the same way ChatLogger works and we want to maintain consistency between the components as they are so closely related. May make these components more intuitive for developers if they are already using ChatLogger.

@nkrantz we should discuss this though because it would make sense to have something like this in future.

packages/paste-core/components/ai-chat-log/package.json
```diff
@@ -0,0 +1,79 @@
+{
+  "name": "@twilio-paste/ai-chat-log",
+  "version": "0.0.0",
+  "category": "data display",
+  "status": "production",
+  "description": "Ai chat log.",
```
krisantrobus, line 6: FYI I am updating the package desc in my docs PR

packages/paste-core/components/ai-chat-log/src/AIChatMessageLoading.tsx
```diff
@@ -0,0 +1,97 @@
...
+   */
+  i18nStopGeneratingLabel?: string;
+  /**
+   * I18n accessible screen reader text to give context to the "Stop generating" button when `onStopLoading` is passed for non-english languages. Should read as a sentence, e.g. "Stop generating AI response".
+   *
+   * @default "AI response"
+   * @type {string}
+   * @memberof AIChatMessageLoadingProps
+   */
+  i18nAIScreenReaderText?: string;
```
krisantrobus, line 47: praise: great addition

packages/paste-core/components/ai-chat-log/src/AIChatMessageLoading.tsx
```diff
@@ -0,0 +1,97 @@
...
+}
+
+export const AIChatMessageLoading = React.forwardRef<HTMLDivElement, AIChatMessageLoadingProps>(
+  (
+    {
+      onStopLoading,
+      i18nStopGeneratingLabel = "Stop generating",
+      i18nAIScreenReaderText = "AI response",
```
krisantrobus, line 55: Should the default text here match the props description? "Stop generating AI response"

packages/paste-core/components/button-group/src/ButtonGroup.tsx
```diff
@@ -79,6 +79,8 @@ export const ButtonGroup = React.forwardRef<HTMLDivElement, ButtonGroupProps>(
         ref={ref}
         element={element}
         display="inline-flex"
+        flexWrap={attached ? "nowrap" : "wrap"}
+        rowGap="space40"
```
krisantrobus, line 83: Does this add any additional space below the ButtonGroup is it's attached too? Might impact spacings on rollout

krisantrobus (commented):
Great implementation, left a few minor comments for clarification but very happy! p.s. I cannot approve because I created the PR 😆

packages/paste-core/components/ai-chat-log/src/useAIChatLogger.tsx
nkrantz: These are both good points. The reason we went with these decisions was solely to mirror the existing ChatLog API. Because many consumers are already familiar with that API, we wanted to replicate it as closely as possible with this one.

packages/paste-core/components/ai-chat-log/__tests__/AIChatLogger.spec.tsx
nkrantz: Same as above - agree but we want to mirror ChatLogger API here.

packages/paste-core/components/ai-chat-log/src/AIChatMessageLoading.tsx
nkrantz: It's kind of clunky but I made them 2 separate props - stop generating label and screen reader text. Only the stop generating label gets rendered on the screen, but the text passed into the screen reader text prop goes into a Screen Reader Only component to give screen reader users more context about what stops generating. So a sighted user will see "Stop generating" but a screen reader will read out loud "stop generating AI response"

packages/paste-core/components/button-group/src/ButtonGroup.tsx
nkrantz: rowGap should only apply spacing in between flex rows! Not above or below the flex object itself.
image

packages/paste-core/components/ai-chat-log/__tests__/AIChatLogger.spec.tsx
nkrantz: Whoops - didn't see Kristian's comment. Agreed, would be nice to ticket this change across both packages.

henrieri (dismissed):

packages/paste-core/components/ai-chat-log/src/AIChatMessageAuthor.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+    return (
+      <Box
+        {...safelySpreadBoxProps(props)}
+        ref={ref}
+        aria-label={props["aria-label"]}
+        element={element}
+        display="flex"
+        alignItems="center"
+        columnGap="space30"
```
serifluous, line 69:
```suggestion
        columnGap="space40"
```

packages/paste-core/components/ai-chat-log/src/AIChatMessageActionGroup.tsx
```diff
@@ -0,0 +1,38 @@
...
+      <Box
+        {...safelySpreadBoxProps(props)}
+        ref={ref}
+        element={element}
+        display="flex"
+        flexWrap="wrap"
+        alignItems="center"
+        columnGap="space30"
+        fontWeight="fontWeightMedium"
+        paddingTop="space40"
```
serifluous, line 30:
```suggestion
        marginTop="space40"
```

packages/paste-core/components/ai-chat-log/src/AIChatMessageActionGroup.tsx
```diff
@@ -0,0 +1,38 @@
...
+  ({ children, element = "AI_CHAT_MESSAGE_ACTION_GROUP", ...props }, ref) => {
+    return (
+      <Box
+        {...safelySpreadBoxProps(props)}
+        ref={ref}
+        element={element}
+        display="flex"
+        flexWrap="wrap"
+        alignItems="center"
+        columnGap="space30"
```
serifluous, line 28:
```suggestion
        columnGap="space40"
```

nkrantz (approved):

PR #3997, opened 2024-07-22
chore(collaboration): added script to resolve typedoc issue

krisantrobus:
Added a new script into package.json to run a series of commands to resolve the typedoc issue when forking the repo.

After investigation I found that if you run the typedoc command before running the build command it will output incorrect typedocs and cause large unintended and incorrect changes to various typedoc files. What is more interesting is that these incorrect changes are different per machine.

The resolution is to clean the local NX cache and run the full build and typedocs scrips. Tested locally by forking the repo multiple times.

If we see this issue in collab PRs we can now ask them to just run this to resolve the issue instead of granting write access to our repo


nkrantz (approved):
🔥🔥🔥 super useful improvements thank youuuu

PR #4003, opened 2024-07-26
Pill large dismiss

PixeledCode:
Fix dismiss button position for large pill size

### DIFF
--- a/.changeset/small-turkeys-change.md
+++ b/.changeset/small-turkeys-change.md
@@ -1,0 +1,6 @@
+---
+"@twilio-paste/form-pill-group": minor
+"@twilio-paste/core": minor
+---
+
+[Form Pill Group]: Fix position of dismiss button for large pill size
--- a/.changeset/small-turkeys-change.md
+++ b/.changeset/small-turkeys-change.md
@@ -1,6 +1,6 @@
 ---
-"@twilio-paste/form-pill-group": minor
-"@twilio-paste/core": minor
+"@twilio-paste/form-pill-group": patch
+"@twilio-paste/core": patch
 ---
 
 [Form Pill Group]: Fix position of dismiss button for large pill size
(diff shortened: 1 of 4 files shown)


.changeset/small-turkeys-change.md
```diff
@@ -0,0 +1,6 @@
+---
+"@twilio-paste/form-pill-group": minor
+"@twilio-paste/core": minor
```
nkrantz, line 3:
```suggestion
"@twilio-paste/form-pill-group": patch
"@twilio-paste/core": patch
```
Because we're not adding any new exports here, this can be a patch for both form-pill-group and core

nkrantz (commented):
👏 super quick turnaround

just a tiny change with the changeset, otherwise perfect

krisantrobus (approved):

packages/paste-core/components/form-pill-group/src/PillCloseIcon.tsx
```diff
@@ -49,8 +52,8 @@ export const PillCloseIcon: React.FC<React.PropsWithChildren<PillCloseIconProps>
       display="flex"
       alignItems="center"
       justifyContent="center"
-      height="sizeIcon40"
-      width="sizeIcon40"
+      height={size === "large" ? "sizeIcon50" : "sizeIcon40"}
+      width={size === "large" ? "sizeIcon50" : "sizeIcon40"}
```
roxanagomez, line 56: **issue:** PillCloseIcon for the large variant should be slightly bigger. Currently, both variants seem to be the same size.
Screenshot 2024-07-26 at 2 29 51 PM

Design sizes:

Default (PillCloseIcon): size-icon-10
Large (PillCloseIcon): size-icon-20

roxanagomez: Note: Not sure why the sizes are different in the code :'( is that expected? or am I checking in the wrong place?

nkrantz: It's on line 63 of this file 🤓 I'll bump it up to sizeIcon20

roxanagomez (approved):

nkrantz (approved):

PR #4017, opened 2024-07-31
fix(combobox): modify `MultiSelectComboboxProps.onSelectedItemsChange` typing to reflect reality

SethNuteTwilio:
The current typing for this prop is misleading and has caused some confusion on the Segment side of the house. From my code comment, the true behavior of the callback function here is as follows:
```
* If `state` prop is passed, `UseMultiSelectPrimitiveStateChange<Item>` is the type for the `newSelectedItems` callback param
*
* If `state` prop isn't passed, `any[]` is the type for the `newSelectedItems` callback param
```

**Contributing to Twilio**

### DIFF
--- a/.changeset/shy-birds-jump.md
+++ b/.changeset/shy-birds-jump.md
@@ -1,0 +1,5 @@
+---
+"@twilio-paste/combobox": patch
+---
+
+Changed type of MultiselectComboboxProps.onSelectedItemsChange.newSelectedItems to include type that is returned when not in IoC mode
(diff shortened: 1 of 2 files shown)


joetessy (approved):
👀

YiyangLi (approved):
:shipit:

nkrantz:
Hey @SethNuteTwilio, can you share a reproduction of the type error? Just want to review before we get this change in. Sorry it's taken so long to get to this, team capacity is very low right now!

PR #4026, opened 2024-08-08
Fix/button resize plus ai bug

krisantrobus:
- Addresses issue with typescript exposing props on popover reset variant
- Correctly sets decorative AI icon color
- Fixes naming and customization on AI examples

### DIFF
--- a/packages/paste-core/components/popover/src/types.ts
+++ b/packages/paste-core/components/popover/src/types.ts
@@ -13,7 +13,7 @@ type PopoverButtonBaseProps = {
 };
 
 export type PopoverButtonProps = PopoverButtonBaseProps &
-  Omit<ButtonProps, "as"> & {
+  ButtonProps & {
     /**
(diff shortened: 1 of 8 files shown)


packages/paste-core/components/popover/src/types.ts
```diff
@@ -13,7 +13,7 @@ type PopoverButtonBaseProps = {
 };
 
 export type PopoverButtonProps = PopoverButtonBaseProps &
-  Omit<ButtonProps, "as"> & {
```
krisantrobus, line 16: Adding the Omit is messing with the underlying types. With omit it will not show additional pseudo types for reset in an IDE. Removing it fixes this.

I manually tested setting the "as" prop to be something other than button and it had no effect as button is applied in the button component it renders.

serifluous (approved):

nkrantz (approved):

PR #4035, opened 2024-08-16
fix(supabase): use exact package version

krisantrobus:
Now use an exact package version which seems to work better with the CLI

### DIFF
--- a/apps/backend/supabase/schema.gen.ts
+++ b/apps/backend/supabase/schema.gen.ts
@@ -34,13 +34,340 @@ export interface Database {
   }
   public: {
     Tables: {
-      [_ in never]: never
+      assistant_threads: {
+        Row: {
+          created_at: string
+          id: number
+          openai_thread_id: string | null
+          slack_thread_ts: string | null
+        }
+        Insert: {
+          created_at?: string
+          id?: number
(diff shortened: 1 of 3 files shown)


apps/backend/supabase/schema.gen.ts
```diff
@@ -34,13 +34,340 @@ export interface Database {
   }
   public: {
     Tables: {
-      [_ in never]: never
+      assistant_threads: {
```
krisantrobus, line 37: All of this file is generated by a script. This is what is used in the GitHub action that was failing. Supabase updated their DB and the latest docker image that was being pulled and the script ran against was missing types. This came from my lacal machine with latest DB up

yarn.lock
```diff
@@ -42784,17 +42784,17 @@ resolve@^2.0.0-next.3:
   languageName: node
   linkType: hard
 
-"supabase@npm:^1.136.3":
-  version: 1.141.0
```
krisantrobus, line 42788: This was the version being used by CI and overnight batch job. Hopefully updating this one will fix it. Works locally but we won't know about the nightly job until code is merged.

nkrantz (approved):

PR #4042, opened 2024-08-21
fix(nx): remove access token. Sets via env variables

krisantrobus:
Remove the static NX token. It was a read only token for an open source project but flagged by internal security. 

The token was deleted from NX Cloud and another one is shared in team 1Password if needed for local development.

The token has been updated in GitHub variables and you can see it correctly connecting for this PR 4042:
image

### DIFF
--- a/.github/workflows/on_pull_request_package_size.yml
+++ b/.github/workflows/on_pull_request_package_size.yml
@@ -6,6 +6,9 @@ jobs:
   package-size:
     name: Check package sizes
     runs-on: ubuntu-latest
+    env:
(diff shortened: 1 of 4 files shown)


.github/workflows/on_pull_request_package_size.yml
```diff
@@ -6,6 +6,9 @@ jobs:
   package-size:
     name: Check package sizes
     runs-on: ubuntu-latest
+    env:
```
krisantrobus, line 9: Had to add this in. It was previously using the token in the nx.json but now had to use env variable

PR #4047, opened 2024-08-26
fix(form-pill): fix issues where pill text not being truncated

krisantrobus:
Addresses an issue where form pill text would wrap when it is longer than form pill group parent.

Key considerations when working with FormPillGroup. As FormPillGroup wraps the pills there should be no ability to overflow the width of the formPillGroup, only to stretch as far as the parent group. In order to achieve this I need to find the string of the children as it could potentially be be composed like our storybooks. Apply the truncate only to the string. Currently for pill we would see the same issue with long text as we were already seeing in the original bug, it is not limited to being in a popover but to any narrow container.

Before:
Screenshot 2024-08-26 at 12 57 55 PM

After:
Screenshot 2024-08-26 at 1 13 51 PM
Screenshot 2024-08-26 at 1 22 55 PM


packages/paste-core/components/form-pill-group/package.json
```diff
@@ -58,6 +59,7 @@
     "@twilio-paste/style-props": "^9.1.0",
     "@twilio-paste/styling-library": "^3.0.0",
     "@twilio-paste/theme": "^11.0.0",
```
PixeledCode, line 61: any reason why `peerDependencies` has 14.0.0 version but `devDependencies` has 14.1.0 version of `@twilio-paste/truncate`?

krisantrobus: This was the same versions for each in other components that used it to be consistent. Should resolve locally anyway

PixeledCode: cool cool

packages/paste-core/components/combobox/stories/Combobox.stories.tsx
```diff
@@ -885,3 +885,13 @@ ComboboxInModal.parameters = {
     disable: true,
   },
 };
+
+export const ComboboxErrorNarrowContainer: StoryFn = () => {
+  return (
+    <Box maxWidth="150px">
+      <Combobox items={items} labelText="Choose a component:" helpText="This is the help text" hasError />
```
serifluous, line 892: suggestion: For VRT, can you show this one with the long option selected by default?

packages/paste-core/components/combobox/stories/MultiselectCombobox.stories.tsx
```diff
@@ -688,3 +688,27 @@ export default {
...
+        }}
+        onSelectedItemsChange={(selectedItems) => {
+          // eslint-disable-next-line no-console
+          console.log(selectedItems);
+        }}
+      />
+    </Box>
+  );
+};
+MultiselectComboboxInNarrowContainer.storyName = "Narrow Contrainer";
```
serifluous, line 714:
```suggestion
MultiselectComboboxInNarrowContainer.storyName = "Narrow Container";
```

packages/paste-core/components/combobox/package.json
```diff
@@ -55,6 +55,7 @@
     "@twilio-paste/styling-library": "^3.0.0",
     "@twilio-paste/text": "^10.0.0",
     "@twilio-paste/theme": "^11.0.0",
+    "@twilio-paste/truncate": "^14.0.0",
```
nkrantz, line 58: question: does combobox need truncate since you're just using textOverflow?

nkrantz: oh jk realizing it was added as a dependency for form pill group!

packages/paste-core/components/form-pill-group/stories/index.stories.tsx
```diff
@@ -177,6 +179,157 @@ export const FormPillTreeVariant = (): JSX.Element => {
...
+              {pill}
+            </Text>
+          </FormPill>
+        ))}
+      </FormPillGroup>
+    </Box>
+  );
+};
+
+PillNarrowContainerFormattedTextContent.storyName = "Pill Narrow Container Formatted Text Content";
```
nkrantz, line 331: suggestion(non-blocking) this line isn't necessary unless you're changing the name of the story, adds errors to the file

packages/paste-core/components/form-pill-group/stories/index.stories.tsx
```diff
@@ -177,6 +179,157 @@ export const FormPillTreeVariant = (): JSX.Element => {
...
+          >
+            {index % 3 === 2 ? (
+              <Avatar
+                size={size === "large" ? "sizeIcon20" : "sizeIcon10"}
+                name="avatar example"
+                src="./avatars/avatar4.png"
+              />
+            ) : null}
+            {index % 3 === 1 ? <CalendarIcon decorative size={size === "large" ? "sizeIcon20" : "sizeIcon10"} /> : null}
+            <Text fontWeight="fontWeightBold" fontStyle="italic">
```
nkrantz, line 321: issue: Text should have an `as` prop

packages/paste-core/components/form-pill-group/stories/index.stories.tsx
```diff
@@ -177,6 +179,157 @@ export const FormPillTreeVariant = (): JSX.Element => {
...
+  React.PropsWithChildren<{
+    selected?: boolean;
+    dismissable?: boolean;
+    disabled?: boolean;
+    ariaLabel?: string;
+    size?: FormPillGroupSizeVariant;
+  }>
+> = ({ selected = false, dismissable = true, disabled = false, ariaLabel = "Basic pills:", size }) => {
+  const pillState = useFormPillState();
+  const pillState2 = useFormPillState();
```
nkrantz, line 298: suggestion(non-blocking): should this line be removed? doesn't look like it's being used

packages/paste-core/components/form-pill-group/stories/index.stories.tsx
```diff
@@ -177,6 +179,157 @@ export const FormPillTreeVariant = (): JSX.Element => {
...
+          </FormPill>
+        ))}
+      </FormPillGroup>
+    </Box>
+  );
+};
+
+PillNarrowContainerDecorationsEnd.storyName = "Pill Narrow Container Decorations End";
+
+export const PillNarrowContainerFormattedTextContent: React.FC<
```
nkrantz, line 288:
```suggestion
export const PillNarrowContainerFormattedTextContent: StoryFn<
```

could you update the stories in this file to use StoryFn (imported from storybook/react)?

nkrantz (commented):
I like this solution, still looks kind of weird if you try to format text inside of a form pill (with an ellipsis after each text element), but it's definitely an improvement from the current implementation. 
image

nkrantz (approved):

serifluous (approved):

PR #4049, opened 2024-08-27
fix: combobox inside of popover bug

nkrantz:
Fixes a bug where Comboboxes (Singleselect and Multiselect) can't be used when placed inside of Popovers. The bug was caused by the Combobox Listbox Portal interfering with the Popover Reakit Portal, and was preventing hover styles from being applied as well as closing the entire Popover on a selection of an item (Multiselect only). [See this CodeSandbox for a reproduction and more context](https://codesandbox.io/p/sandbox/combobox-in-popover-krzsgd).

This PR adds the `usePortal` prop to Combobox (which defaults to `true`). It should be used (and set to `false`) when Combobox is used in a Popover. When Combobox doesn't use a Portal, it uses an absolutely positioned Box to position the listbox and attach it to the input box. This means that the ListboxPositioner isn't used on these Comboboxes, so the listbox won't position itself above the input box when the input is at the bottom of the window. If this bug comes up it will be addressed in the future.

Examples added to Singleselect and Multiselect Combobox docs.

### DIFF
--- a/.changeset/healthy-cobras-whisper.md
+++ b/.changeset/healthy-cobras-whisper.md
@@ -1,0 +1,6 @@
+---
+"@twilio-paste/combobox": minor
+"@twilio-paste/core": minor
+---
+
+[Combobox] Add new prop `usePortal` which defaults to `true`. The prop was introduced to address a bug when Comboboxes are placed in Popovers. usePortal should be set to false when using a Combobox inside a Popover in order to retain full functionality.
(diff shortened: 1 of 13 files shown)


krisantrobus (approved):
This is such a clean and elegant solution. It works so well, I also checked the storybooks to see if the original responsive would be reintroduced and it's all working perfectly. Really great job on this. Filter patterns is saved!!

serifluous (approved):

PR #4051, opened 2024-08-28
feat(button-group): add justifyContent prop, use flex

nkrantz:
https://twilio-productivity.atlassian.net/browse/DS-445

### DIFF
--- a/packages/paste-core/components/button-group/src/ButtonGroup.tsx
+++ b/packages/paste-core/components/button-group/src/ButtonGroup.tsx
@@ -40,14 +40,12 @@ const AttachedStyled = styled.span(
 
 const UnattachedStyled = styled.span(
   css({
-    "& > button:not(:first-of-type)": {
-      marginLeft: "space40",
-    },
-    "& > button:first-of-type": {
-      marginLeft: "space0",
-    },
+    display: "flex",
+    columnGap: "space40",
   }),
 );
+
+type JustifyOptions = "flex-start" | "center" | "flex-end" | "space-between";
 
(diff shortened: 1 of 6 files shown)


serifluous (approved):

packages/paste-core/components/button-group/src/ButtonGroup.tsx
```diff
@@ -65,11 +63,19 @@ export interface ButtonGroupProps extends HTMLPasteProps<"div"> {
...
    */
   attached?: boolean;
+  /**
+   * Sets the `justify-content` CSS property. Only used for unattached button groups.
+   *
+   * @default "flex-start"
+   * @type {JustifyOptions}
+   * @memberof ButtonGroupProps
+   */
+  justifyContent?: JustifyOptions;
```
krisantrobus, line 73: question, do we need to expose this prop? For the justify storybook would we really want users to be able to use the display between. We could just update the default value and use the attached boolean to determine whether is need sot be different for the 2 supported variants

nkrantz: hmm, you mean making unattached always centered and attached always flex-start? that would be a breaking design change. maybe I'm misunderstanding

krisantrobus: I mean leave the styling to how it currently is without any change when attached an then only apply the new flex when they are unattached. No new props for consumers

krisantrobus (approved):
Happy to move forward, looks great!

PR #4067, opened 2024-09-12
feat(responsive-tabs): allow scroll on components with tabs

krisantrobus:
Added scroll functionality to components with tabs:
- InPageNavigation [Figma](https://www.figma.com/design/l2qtpQVMmXzQPVB3rEAA8i/Paste-Components-%26-Themes?node-id=38779-17203&node-type=frame&t=h6Cp7U8xK5aGa9MM-0)
- CodeBlock [Figma](https://www.figma.com/design/l2qtpQVMmXzQPVB3rEAA8i/branch/kSocgHKROkCy6b4CRacA8r/Paste-Components-%26-Themes?node-id=39068-6480&node-type=section&t=L7t0CxTwG1Jip7wW-0)
- Tabs [Figma](https://www.figma.com/design/l2qtpQVMmXzQPVB3rEAA8i/Paste-Components-%26-Themes?node-id=38779-17306&node-type=frame&t=YYpsPnHzeCh7vyjm-0)


packages/paste-core/components/code-block/src/CodeBlockTabList.tsx
```diff
@@ -1,9 +1,72 @@
...
+const ShadowLeft = styled.div(({ bgColor, left }: { bgColor: string; left: string }) => {
+  return css({
+    content: "' '",
+    position: "absolute",
+    left,
+    top: 0,
+    bottom: 0,
+    width: "0px",
+    pointerEvents: "none",
+    boxShadow: `6px 12px 12px 2px ${bgColor}`,
```
krisantrobus, line 18: Used absolute values (not design tokens) as we do not currently have anything for negative 6 like in the below example. This would also need to be a new boxShadow variable. 

This is for the edge shadow on the scroll bar so seems like it's an edge case and not standard on any other components. Happy to update design tokens if we feel the need!

packages/paste-core/components/code-block/src/CodeBlockTabList.tsx
```diff
@@ -1,9 +1,69 @@
...
 import * as React from "react";
 
+const Shadow = styled.div(({ bgColor, position }: { bgColor: string; position: "left" | "right" }) => {
+  const Styles = {
+    left: {
+      boxShadow: `6px 12px 12px 2px ${bgColor}`,
+      left: 0,
+    },
+    right: {
+      boxShadow: `-6px 12px 12px 2px ${bgColor}`,
```
krisantrobus, line 16: Used absolute values (not design tokens) as we do not currently have anything for negative 6 like in the below example. This would also need to be a new boxShadow variable.

This is for the edge shadow on the scroll bar so seems like it's an edge case and not standard on any other components. Happy to update design tokens if we feel the need!

packages/paste-core/components/code-block/src/CodeBlockTabList.tsx
```diff
@@ -17,12 +77,67 @@ export interface CodeBlockTabListProps extends Omit<TabListProps, "aria-label">
...
     return (
-      <Box paddingX="space70">
-        <TabList {...props} aria-label="label" ref={ref} element={element}>
-          {children}
-        </TabList>
-      </Box>
+      <TabPrimitiveList {...(tab as any)} as={Box} {...props} element={element} ref={ref}>
+        <Box
+          element={`${element}_CHILD_WRAPPER`}
+          marginLeft="space70"
```
krisantrobus, line 113: changed to margin instead of padding as the absolute position of the left div for left shadow would also need to use the design token. Now it can just be aligned to start of container

packages/paste-core/components/tabs/src/Tab.tsx
```diff
@@ -108,6 +108,7 @@ const getTabBoxStyles = (orientation: Orientation, variant: Variants): BoxStyleP
         borderColor: "transparent",
         borderStyle: "solid",
         borderWidth: "borderWidth10",
+        borderBottomWidth: "borderWidth20",
```
krisantrobus, line 111: original change from Shadi's PR for updated designs

packages/paste-core/components/code-block/src/CodeBlockTabList.tsx
nkrantz: hmm yeah I think adding a specific shadow token for this would be fine, just name it something specific to what we're using it for

packages/paste-core/components/code-block/src/CodeBlockTabList.tsx
nkrantz:
```suggestion
          marginX="space70"
```

Also not seeing the scrollbar

Screenshot 2024-09-12 at 8 33 27 PM

packages/paste-core/components/code-block/src/CodeBlockTabList.tsx
```diff
@@ -17,12 +77,67 @@ export interface CodeBlockTabListProps extends Omit<TabListProps, "aria-label">
...
+    const theme = useTheme();
+    //  ref to the scrollable element
+    const scrollableRef = React.useRef<HTMLDivElement>(null);
+    const [scrollShadow, setScrollShadow] = React.useState<"none" | "left" | "right" | "both">("none");
+
+    // Function to handle scroll event
+    const handleScroll = (): void => {
+      if (scrollableRef.current) {
+        // No scrollbar, so no shadow
+        if (scrollableRef.current.clientWidth === scrollableRef.current.scrollWidth) {
```
nkrantz, line 90: suggestion: can you use refElement.getBoundingClientRect() here instead? hopefully it will provide everything you need and I _think_ it's supposed to be more accurate to what's actually rendered in the DOM

https://stackoverflow.com/questions/32438642/clientwidth-and-clientheight-report-zero-while-getboundingclientrect-is-correct

packages/paste-core/components/code-block/src/CodeBlockTabList.tsx
```diff
@@ -17,12 +77,67 @@ export interface CodeBlockTabListProps extends Omit<TabListProps, "aria-label">
...
+            ref={scrollableRef}
+            display="flex"
+            flexWrap="nowrap"
+            flexShrink={0}
+            element={`${element}_CHILD`}
+            overflowX="auto"
+            overflowY="hidden"
+            onScroll={handleScroll}
+          >
+            {children}
```
nkrantz, line 131: issue: looks like this change is pushing the tab children themselves up by a pixel or so (see how the top box shadow on selected tabs is being cut off the tiniest bit)

old:
image
current:
image

packages/paste-core/components/in-page-navigation/src/InPageNavigation.tsx
```diff
@@ -95,20 +145,31 @@ const InPageNavigation = React.forwardRef<HTMLDivElement, InPageNavigationProps>
       <InPageNavigationContext.Provider value={{ variant, orientation }}>
         <Box {...safelySpreadBoxProps(props)} as="nav" ref={ref} element={element}>
           <Box
-            as="ul"
-            listStyleType="none"
-            element={`${element}_ITEMS`}
+            as={StyledScrollTabList as any}
```
nkrantz, line 148: suggestion(blocking): let's keep this a `ul` (was this change intentional?)

nkrantz (commented):
it looks SO GOOD, working really well in the stories. was it intentional to have the left/right shadows on scrollable tabs for code block but not for tabs themselves (and in page nav)? could be a nice touch there too

I didn't test in non-chrome browsers yet but let's make sure we do that before merging

packages/paste-core/components/tabs/src/index.tsx
```diff
@@ -2,6 +2,7 @@ import type { TabStateReturn } from "./Tabs";
 
 export { Tabs, useTabState } from "./Tabs";
 export type { TabsProps } from "./Tabs";
+export { TabsContext } from "./TabsContext";
```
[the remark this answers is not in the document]
krisantrobus: Shadi exported this so it can be used in CodeBlock

packages/paste-core/components/code-block/src/CodeBlockTabList.tsx
krisantrobus: It shouldn't be x as the margin is only so that the wrapper aligns with the header. ther tabs should extend all teh way to the end of the component. marginX would not allow that 
image

packages/paste-core/components/code-block/src/CodeBlockTabList.tsx
krisantrobus: Fixed in latest commit

packages/paste-core/components/in-page-navigation/src/InPageNavigation.tsx
krisantrobus: No this was probably just a copy paste error. I updated the styled component to stay `ul`. Good catch!

packages/paste-core/components/code-block/src/CodeBlockTabList.tsx
krisantrobus: Those tokens have been added in now

packages/paste-core/components/code-block/src/CodeBlockTabList.tsx
krisantrobus: Great suggestion. I made the change and is working well

packages/paste-core/components/in-page-navigation/src/InPageNavigation.tsx
krisantrobus: Actually I reverted that commit. It looks like I removed that `ul` but I actually just wrapped it. The original box is now just nested

packages/paste-core/components/code-block/src/CodeBlockTabList.tsx
nkrantz: can we confirm with design that it should be marginLeft and not marginX? looks kinda jank to have a left margin but not a right margin on scroll (like in the screenshot I posted)

cc @twilio-labs/design-systems-pd

packages/paste-core/components/code-block/stories/index.stories.tsx
```diff
@@ -125,6 +125,22 @@ export const CodeBlockTabs: StoryFn = () => (
     <CodeBlockTabList>
       <CodeBlockTab>Ruby</CodeBlockTab>
       <CodeBlockTab>PHP</CodeBlockTab>
+      <CodeBlockTab>Python</CodeBlockTab>
```
nkrantz, line 128: suggestion(non-blocking): if you have time it might be nice to throw this story in a fixed-width box so we can vrt the scroll

nkrantz: same with the in page nav overflow example

packages/paste-core/components/in-page-navigation/src/InPageNavigation.tsx
```diff
@@ -68,6 +103,21 @@ const InPageNavigation = React.forwardRef<HTMLDivElement, InPageNavigationProps>
   ) => {
     const isFullWidth = variant === "fullWidth" || variant === "inverse_fullWidth";
     const isInverse = variant === "inverse" || variant === "inverse_fullWidth";
+    const listRef = React.useRef<HTMLOListElement>(null);
+    // Scroll to the selected tab if it exists on mount
+    React.useEffect(() => {
```
nkrantz, line 108: niiiiice

packages/paste-core/components/code-block/stories/index.stories.tsx
krisantrobus: Great suggestion! I'll get that in

nkrantz (commented):
just a couple style things left!

1. tabs in codeblock styles - marginRight question and scrollbar
2. tabs themselves have changed a bit. first 2 pics are current tabs, last 2 are with these changes. the tab (and focus ring) should be sitting right on the bottom border, and the bottom border shouldn't be visible on the selected tab
Screenshot 2024-09-13 at 12 25 09 PM
Screenshot 2024-09-13 at 12 34 29 PM
image
image

packages/paste-design-tokens/tokens/global/box-shadow.yml
```diff
@@ -230,3 +230,11 @@ props:
   shadow-border-bottom-notification-stronger:
     value: "{!offset-0} {!offset-05} {!offset-0} {!palette-red-90}"
     comment: Stronger bottom shadow border for notification.
+
+  # Scrollable elements
+  shadow-scroll-right-inverse:
+    value: "{!offset-negative-30} {!offset-50} {!offset-50} {!offset-10} {!palette-gray-100}"
+    comment: Left shadow for scrollable elements on inverse backgrounds.
```
nkrantz, line 237:
```suggestion
    comment: Right shadow for scrollable elements on inverse backgrounds.
```

packages/paste-core/components/in-page-navigation/src/InPageNavigation.tsx
```diff
@@ -91,25 +200,80 @@ const InPageNavigation = React.forwardRef<HTMLDivElement, InPageNavigationProps>
       );
     }
 
+    if (isFullWidth) {
```
krisantrobus, line 203: Added a new condition as wrapping in flex broke the fitted variant

roxanagomez (commented):
but I noticed some unexpected changes in tabs used across examples in other components, templates, and on the docs site, I think we might have previously had the bottom border of tabs/in-page navigation set to fill by default. 

Screenshot 2024-10-10 at 4 22 20 PM
Screenshot 2024-10-10 at 4 23 42 PM

nkrantz:
This is totally something I could've missed a conversation about, but are you aware of this change?

Before:
Screenshot 2024-10-14 at 3 18 32 PM

In this PR:
image

(the little white line under the Tabs doesn't extend the width of the TabList anymore)

krisantrobus:
> This is totally something I could've missed a conversation about, but are you aware of this change?
> 
> Before: Screenshot 2024-10-14 at 3 18 32 PM
> 
> In this PR: image
> 
> (the little white line under the Tabs doesn't extend the width of the TabList anymore)

Good catch, fixed for both Tabs and InPageNavigation but didn't see an issue with this in chromatic

nkrantz (approved):

packages/paste-core/components/in-page-navigation/src/OverflowButton.tsx
```diff
@@ -0,0 +1,72 @@
...
+  _hover: {
+    color: "colorTextPrimaryStrongest",
+    cursor: "pointer",
+  },
+};
+
+const InverseStyles: BoxStyleProps = {
+  color: "colorTextIconInverse",
+  _hover: {
+    color: "colorTextInverseWeak",
```
serifluous, line 29:
```suggestion
    color: "colorTextInverse",
```

Needs to go to a brighter token to be 3:1 contrast with colorTextIconInverse

packages/paste-core/components/code-block/src/OverflowButton.tsx
```diff
@@ -0,0 +1,55 @@
...
+  position: "left" | "right";
+  visible?: boolean;
+  element?: BoxProps["element"];
+  showShadow?: boolean;
+}
+
+const Styles: BoxStyleProps = {
+  color: "colorTextIconInverse",
+  _hover: {
+    color: "colorTextInverseWeak",
```
serifluous, line 18:
```suggestion
    color: "colorTextInverse",
```

packages/paste-core/components/tabs/src/OverflowButton.tsx
```diff
@@ -0,0 +1,73 @@
...
+  _hover: {
+    color: "colorTextPrimaryStrongest",
+    cursor: "pointer",
+  },
+};
+
+const InverseStyles: BoxStyleProps = {
+  color: "colorTextIconInverse",
+  _hover: {
+    color: "colorTextInverseWeak",
```
serifluous, line 29:
```suggestion
    color: "colorTextInverse",
```

serifluous (dismissed):

nkrantz (approved):

serifluous (approved):

packages/paste-core/components/tabs/src/Tabs.tsx
```diff
@@ -48,13 +48,32 @@ export interface TabsProps extends TabPrimitiveInitialState {
...
+    // Scroll to the selected tab if it exists on mount
+    React.useEffect(() => {
+      if (typeof selectedId === "string") {
+        setTimeout(() => {
+          document
+            // eslint-disable-next-line unicorn/prefer-query-selector
+            .getElementById(selectedId)
+            // eslint-disable-next-line @typescript-eslint/ban-ts-comment
+            // @ts-ignore - behavior is typed incorrectly in Typescript v4, fixed in v5+
+            ?.scrollIntoView({ behavior: prevSelectedTab === undefined ? "instant" : "smooth" });
```
dmitriy-kudelko, line 70: Hi @krisantrobus,

This PR introduced some weird behavior since it applies to all `Tabs` variations including vertical ones. When used with vertical tabs `scrollIntoView` obviously scrolls both vertically and horizontally until a tab header is visible which causes an annoying experience when you need to scroll up to revert this unintended view change.

Could this be please fixed?

krisantrobus: Hey @dmitriy-kudelko apologies for this. Working on a fix and will update once we have a PR up

krisantrobus: PR is up and we are reviewing internally: https://github.com/twilio-labs/paste/pull/4143

dmitriy-kudelko: Cool. Thanks!

dmitriy-kudelko: @krisantrobus , great news. Thank you for implementing this feature and for quickly releasing a fix! 👍 
We've been waiting for it and have already applied it to some places on the UI.

BTW, there seems to be also an issue with the inability to scroll tabs till the very end. Clicking on the right arrow doesn't seem to help either. The left boundary doesn't have this problem. Not sure if it's something on our end, but we don't have any margin/padding customizations for the Tabs component which could explain that. It doesn't seem to depend on the window width either. I attached a video:

https://github.com/user-attachments/assets/3f987474-c635-46a6-b5f0-526d8d826e3c

krisantrobus: Hmm this is very strange and something we have not encountered. Could you share more details. What is the viewport width you are using and number of tabs. If you could replicate in a codesandbox that would be be very helpful. Also are you able to use the tab key to highlight a tab and use arrow keyboard buttons to get to the very last item?

krisantrobus: @dmitriy-kudelko I could not replicate it in our strorybooks when making some page compositions, however I did find it in a deploy preview for our Paste upgrade in one of our internal tools. I will be working on this again tomorrow and we should see another path. Again, so sorry for these issues

krisantrobus: Hi @dmitriy-kudelko I have validated the fix in 20.18.1 and I can confirm the issue has been resolved.

dmitriy-kudelko: Hi @krisantrobus , nice! Thank you so much!

PR #4073, opened 2024-09-17
fix(a11y): do not assign aria-labelledby attribute to icons unless there is an associated title

stern-shawn:
[See docs team ticket for context.](https://twilio-productivity.atlassian.net/browse/DOCSPLAT-268)

It looks like Paste Icons assign an `aria-labelledby` attribute to all icons, even if there is no `<title />` element that can be referenced by the matching `titleId`. This is throwing warnings in the 'wave' a11y tool mentioned in our QA ticket for the Twilio Docs but you can see the same issue in the Paste docs or anything else that uses Paste Icons and/or the Button with link functionality (which renders a button containing a `decorative` arrow icon).

My fix is to only assign this aria attribute if there will be a valid `<title />` element to reference, avoiding the a11y error/warning.

Without this change (a page with lots of Buttons as links, which render an `<ArrowForwardIcon decorative />` internally:

image

With this change made directly to the Paste icons source in `node_modules/`:

image

**Contributing to Twilio**

### DIFF
--- a/.changeset/tasty-emus-dream.md
+++ b/.changeset/tasty-emus-dream.md
@@ -1,0 +1,5 @@
+---
+"@twilio-paste/icons": patch
+---
+
+Only assign aria-labelledby attribute to icons with a title to link to
--- a/packages/paste-icons/__test__/icons.spec.tsx
+++ b/packages/paste-icons/__test__/icons.spec.tsx
@@ -23,9 +23,14 @@ import {
 
 describe("Icons", () => {
   describe("HTML attributes", () => {
-    it("should have the default element name", () => {
-      const { container } = render(<AgentIcon decorative />);
+    it("should have the default element name and attributes", () => {
+      const { container, getByTitle } = render(<AgentIcon decorative={false} title="Agent Smith" />);
+      expect(getByTitle('Agent Smith')).toBeInTheDocument();
       expect(container.querySelector('[data-paste-element="ICON"]')).toBeInTheDocument();
+    });
+    it("should not have an aria-labelledby attribute if decorative", () => {
+      const { container } = render(<AgentIcon decorative />);
+      expect(container.querySelector('svg')?.attributes).not.toContain("aria-labelledby");
     });
     it("should have a custom element name", () => {
       const { container } = render(<AgentIcon element="CUSTOM_ICON" decorative />);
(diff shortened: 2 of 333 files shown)


serifluous:
Hey @sstern ! @nkrantz looked into this yesterday and we add `aria-hidden` to all decorative icons, meaning they're not visible to screen readers since the content of the icon should be duplicated by its context.

[On the WAVE site, they explain this a bit more](https://wave.webaim.org/help):

> **How does WAVE handle "hidden" elements?**
> WAVE detects accessibility issues in all page elements, even those that are hidden using CSS, the hidden attribute, aria-hidden="true", and/or tabindex="-1". This is by design. In most cases, hidden elements are revealed to users at some point. We believe it's important that these potential end user accessibility issues in drop-down menus, tab panels, dialog boxes, carousels, and other "hidden" elements be identified.
> In most cases the WAVE interface will provide an indication of icons that are relevant to hidden elements—the icon in the sidebar will be slightly dimmed and will present an informative tooltip. These elements may be examined by turning off the Styles option in the WAVE sidebar.
> If WAVE has detected an error in an element that is hidden and that is never presented to the end user, or that is perhaps modified to address the error before being presented (something WAVE cannot detect), the WAVE error can be ignored - or you may instead choose to adapt the code to ensure the element is accessible in its hidden state.
> In short, if we err on this point, we err on the side of users with disabilities. In other words, we have chosen to identify issues in hidden content that may be presented to users, rather than suppress these errors in cases where advanced development techniques have been implemented to _potentially_ hide that content from users.

^ Let us know if ignoring the error works for you like they suggest. If not, we'll do a bit more digging into the best way forward.

stern-shawn:
That sounds good to me! I had noticed that `aria-hidden` was being applied and was surprised that WAVE was flagging the elements anyway. Instead of looking at their docs I assumed there was a code fix (which there _is_ via this change, but this change isn't necessary if the error is ignorable 😄 ).

If this is something you'd like to pursue, I'm happy to tweak this PR as needed, otherwise I can report this back to my team and close this!

serifluous:
Sounds good to me! To tweak the PR, we'd have to do more digging on whether this is the best solution, or [using `aria-label` instead](https://tink.uk/the-difference-between-aria-label-and-aria-labelledby/), or something else.

So yeah, let's just close this out. Thanks for bringing this up!

PR #4083, opened 2024-09-20
fix(nav): docs site navigation

PixeledCode:
- fixed horizontal scroll on mobile nav
- Change Paste logo to the default logo

### DIFF
--- a/packages/paste-website/src/components/site-wrapper/site-header/SiteHeaderSearch.tsx
+++ b/packages/paste-website/src/components/site-wrapper/site-header/SiteHeaderSearch.tsx
@@ -4,6 +4,7 @@ import { SearchIcon } from "@twilio-paste/icons/esm/SearchIcon";
 import { InlineCode } from "@twilio-paste/inline-code";
 import { ScreenReaderOnly } from "@twilio-paste/screen-reader-only";
 import { Text } from "@twilio-paste/text";
+import { useWindowSize } from "@twilio-paste/utils";
 import * as React from "react";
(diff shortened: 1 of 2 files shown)


packages/paste-website/src/components/site-wrapper/site-header/SiteHeaderSearch.tsx
```diff
@@ -4,13 +4,15 @@ import { SearchIcon } from "@twilio-paste/icons/esm/SearchIcon";
...
 import { Text } from "@twilio-paste/text";
+import { useWindowSize } from "@twilio-paste/utils";
 import * as React from "react";
 import { useHotkeys } from "react-hotkeys-hook";
 
 import { SiteSearch } from "../../site-search";
 
 const SiteHeaderSearch: React.FC = () => {
   const [isOpen, setIsOpen] = React.useState(false);
+  const { breakpointIndex } = useWindowSize();
```
roxanagomez, line 15: **question:** Do you know if we have a minimum window size we optimize for? I'm testing at 320px and it breaks, it’s almost like the background shrinks faster than the rest of the elements, or maybe it’s the scroll bar container? I was only able to test on this small screen in the browser, in my mobile device is working well. 
Screenshot 2024-09-24 at 1 00 46 PM

PixeledCode: yeah so with current implementation, 340px is the lower limit

roxanagomez (approved):

krisantrobus (approved):

PR #4103, opened 2024-10-14
chore(ci): add contribution labeler

krisantrobus:
Update the workflows to include a labeler that add "Contribution" to any PR that is created by someone that is not part of our GutHub teams

### DIFF
--- a/.github/workflows/on_pull_request_open.yml
+++ b/.github/workflows/on_pull_request_open.yml
@@ -2,7 +2,7 @@ name: Opened Pull Request Labeler
 
 on:
   pull_request:
-    types: [opened]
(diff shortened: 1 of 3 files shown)


.github/workflows/on_pull_request_open.yml
```diff
@@ -2,16 +2,39 @@ name: Opened Pull Request Labeler
 
 on:
   pull_request:
-    types: [opened]
+    types: [opened, synchronize]
 
 jobs:
-  pr-triage:
+  pr-context-labeller:
```
nkrantz, line 8: labeller vs labeler ? this is one of those words that I can never actually remember how to spell lol. I think it's labeler

PR #4109, opened 2024-10-16
[Inline Code]: Update children prop

PixeledCode:


.changeset/spicy-badgers-double.md
```diff
@@ -0,0 +1,6 @@
+---
+"@twilio-paste/inline-code": minor
+"@twilio-paste/core": minor
```
nkrantz, line 3: suggestion: this can be a patch since we're not adding any new props or exports

packages/paste-core/components/inline-code/stories/index.stories.tsx
```diff
@@ -94,6 +95,16 @@ export const DisplayingAPIKey: StoryFn = () => {
   );
 };
 
+export const WithTruncate: StoryFn = () => {
```
nkrantz, line 98: suggestion: could you also add a story with a truncated inline code within a block of regular paragraph text?

krisantrobus (approved):

serifluous (approved):

PR #4136, opened 2024-11-04
feat(keyboard-key): create package

krisantrobus:
This pull request introduces a new `KeyboardKey` component. It's purpose is to display visual indicators for keyboard shortcuts available to users.


.changeset/shaggy-sheep-confess.md
```diff
@@ -0,0 +1,6 @@
+---
+"@twilio-paste/core": patch
+"@twilio-paste/design-tokens": patch
```
nkrantz, line 3:
```suggestion
"@twilio-paste/core": minor
"@twilio-paste/design-tokens": minor
```

.changeset/sweet-mugs-admire.md
```diff
@@ -0,0 +1,6 @@
+---
+"@twilio-paste/keyboard-key": major
+"@twilio-paste/core": major
```
nkrantz, line 3:
```suggestion
"@twilio-paste/core": minor
```

packages/paste-core/components/keyboard-key/__tests__/hooks.spec.tsx
```diff
@@ -0,0 +1,316 @@
...
+import { Default } from "../stories/index.stories";
+
+describe("Hooks", () => {
+  it("should handle pressed styling", async () => {
+    const { getAllByText } = render(<Default />);
+
+    const controlKey = getAllByText("Control")[0];
+    const bKey = getAllByText("B")[0];
+    expect(controlKey).toBeDefined();
+    expect(controlKey).toBeDefined();
```
nkrantz, line 14: line dupe

packages/paste-core/components/keyboard-key/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,71 @@
...
+      );
+
+      const controlKey = getAllByText("Control")[0];
+      const bKey = getAllByText("B")[0];
+      expect(controlKey).toHaveAttribute("data-paste-element", "MY_CUSTOM_KEY_ONE");
+      expect(bKey).toHaveAttribute("data-paste-element", "MY_CUSTOM_KEY_TWO");
+      expect(controlKey.parentElement).toHaveAttribute("data-paste-element", "MY_CUSTOM_KEY_GROUP");
+    });
+
+    it("should customize styling", async () => {
```
nkrantz, line 40: suggestion: testing custom styles is done with vrt so these tests are redundant (unless there's a specific reason you're testing them here too)

packages/paste-core/components/keyboard-key/src/KeyboardKey.tsx
```diff
@@ -0,0 +1,103 @@
...
+    backgroundColor: "colorBackgroundStrong",
+    boxShadow: undefined,
+  },
+  inverse: {
+    backgroundColor: "colorBackgroundInverseStronger",
+    boxShadow: undefined,
+  },
+};
+
+export interface KeyboardKeyProps extends HTMLPasteProps<"div"> {
```
nkrantz, line 47:
```suggestion
export interface KeyboardKeyProps extends HTMLPasteProps<"kbd"> {
```

packages/paste-core/components/keyboard-key/src/KeyboardKeyContext.tsx
```diff
@@ -0,0 +1,38 @@
...
+  /**
+   * The aray of currently active keys
+   * @default []
+   * @type { string[] }
+   * @memberof KeyboardCombinationState
+   */
+  activeKeys?: string[];
+  /**
+   * Whether the keys should show the disbled state
+   * @default null
```
nkrantz, line 15: question: should disabled default to false?

packages/paste-core/components/keyboard-key/src/hooks.ts
```diff
@@ -0,0 +1,94 @@
...
+    if (!e.repeat) {
+      setActiveKeys((prev) => {
+        return Array.from(new Set([...prev, e.key]));
+      });
+    }
+  };
+
+  const handleKeyUp = (e: KeyboardEvent): void => {
+    /**
+     * Due to a weird behavior on macOS we need to clear the set if the user pressed down the meta key and presses another key.
```
nkrantz, line 37: 👏

tools/plop-templates/component-component.hbs
```diff
@@ -7,7 +7,7 @@ export interface {{pascalCase component-name}}Props extends HTMLPasteProps<"div"
   children?: React.ReactNode;
   /**
    * Overrides the default element name to apply unique styles with the Customization Provider
-   * @default '{constantCase component-name}'
+   * @default '{{constantCase component-name}}'
```
nkrantz, line 10: question: did you test out these 2 hbs file changes with the script? I was having trouble using double curly brackets since that's the format plop already uses for their own variables

packages/paste-core/components/keyboard-key/src/KeyboardKey.tsx
```diff
@@ -0,0 +1,103 @@
...
+   * @memberof KeyboardKeyProps
+   */
+  element?: BoxProps["element"];
+  /**
+   * Sets the key text that will be used to determine if the key has press stylings
+   * @default 'KEYBOARD_KEY'
+   * @type string
+   * @memberof KeyboardKeyProps
+   */
+  keyText?: string;
```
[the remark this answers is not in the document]
serifluous: Agree. Maybe `keystroke`? `keypress`? Was there a reason we moved away from `keyCode` (not that it's _great_, but it's still less ambiguous than `keyText`)?

packages/paste-core/components/keyboard-key/src/KeyboardKeyContext.tsx
krisantrobus: I guess the default is undefined which I see is set to a null value in other areas. I will update to default to false

packages/paste-core/components/keyboard-key/src/KeyboardKeyGroup.tsx
```diff
@@ -0,0 +1,35 @@
+import { Box } from "@twilio-paste/box";
+import type { BoxProps } from "@twilio-paste/box";
+import type { HTMLPasteProps } from "@twilio-paste/types";
+import * as React from "react";
+
+import { KeyboardCombinationState, KeyboardKeyCombinationContext } from "./KeyboardKeyContext";
+
+export type KeyboardKeyVariants = "default" | "inverse";
```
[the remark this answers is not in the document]
krisantrobus: Great suggestion!

packages/paste-core/components/keyboard-key/__tests__/index.spec.tsx
krisantrobus: No there is no particulart reason. I just looked at other test files. The main one was to check element name changes which is done above. I'll actually merge these two to make sure element name changes make the style changes too

packages/paste-core/components/keyboard-key/src/KeyboardKey.tsx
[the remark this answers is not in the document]
krisantrobus: Yea I went back and fore keyText and keyValue but felt keyValue was not explicit to be the string not integer value.

We can actually use the key integer value but it changes in firefox for some keys. You can check the mappings out [here](https://www.freecodecamp.org/news/javascript-keycode-list-keypress-event-key-codes/) and we can use any column event.which | event.key | event.code the prop

packages/paste-core/components/keyboard-key/src/KeyboardKey.tsx
```diff
@@ -0,0 +1,103 @@
...
+      activeKeys,
+      enablePressStyles,
+      variant = "default",
+    } = React.useContext(KeyboardKeyCombinationContext);
+
+    const isKeyActive =
+      !disabled && activeKeys && keyText && activeKeys.map((k) => k.toLowerCase()).includes(keyText.toLowerCase());
+
+    return (
+      <Box
```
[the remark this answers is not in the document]
krisantrobus: done!

packages/paste-core/components/keyboard-key/src/KeyboardKey.tsx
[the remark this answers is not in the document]
serifluous: `keyboardText`?

tools/plop-templates/component-component.hbs
krisantrobus: I just ran this for a before and after. In the case of the comment you are right the 2 curly braces, it makes it do some function but without both it just printed the text which was not correct. Here are a before and after of how the changes affected the output. Package name used in the script "TestComponent"

| Before   | After |
| -------- | -------- |
| @default '{constantCase component-name}'  |                 @default 'TEST_COMPONENT' |
| export TestComponent from './TestComponent' |     export { TestComponent } from './TestComponent' |

packages/paste-core/components/keyboard-key/src/KeyboardKey.tsx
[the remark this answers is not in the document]
krisantrobus: What is we aligned it to that column in the resource table.,`eventKey`. That is closer to the actual attribute that gets shared by the eventListener

packages/paste-core/components/keyboard-key/__tests__/index.spec.tsx
nkrantz: yeah you'll see it in some old test files because we used to do it that way before we realized it was pointless bc the style tests happen in vrt

packages/paste-core/components/keyboard-key/src/KeyboardKey.tsx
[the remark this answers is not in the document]
nkrantz: yess `eventKey` sounds better (super clear that it's a key code rather than visible text) and also feels more relevant to what it's actually doing

packages/paste-core/components/keyboard-key/src/KeyboardKeyContext.tsx
nkrantz: makes sense, just nice to have a default for optional props

packages/paste-core/components/keyboard-key/src/KeyboardKey.tsx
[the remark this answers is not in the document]
krisantrobus: Done!

packages/paste-core/components/keyboard-key/src/KeyboardKey.tsx
```diff
@@ -59,11 +58,11 @@ export interface KeyboardKeyProps extends HTMLPasteProps<"div"> {
    * @type string
    * @memberof KeyboardKeyProps
    */
-  keyText?: string;
+  keyEvent?: string;
```
nkrantz, line 61: suggestion (non-blocking): maybe add to the description comment here that the docs have a chart of usable keyEvents? or a link to the map that you were using at one point?

slash would it make sense to limit the type of `keyEvent` from `string` to an actual list of usable key codes? I'm guessing no for scalability, but wanted to ask

nkrantz (approved):
Very nice changes, thanks for incorporating all the feedback!

serifluous (dismissed):

packages/paste-core/components/keyboard-key/src/KeyboardKey.tsx
krisantrobus: A really great suggestion. Pushing up the comment update. We can use the `event.which` which is the code number. It's something I feel developers would be more used to but it changes based on browser as mentioned in the link to the table. It is also deprecated so browsers at some point in the future may stop returning that field.

serifluous (approved):

nkrantz (approved):

PR #4170, opened 2024-11-21
fix(progress-steps): unnecessary re-renders

krisantrobus:
Wrapped Progress steps in useMemo to avoid unnecessary re-renders. Note that this issue only occurs when a prop form the component is passed into the emotion styled.div that causes re-renders. If the option passed to this are static then there is no issue which is why only this component has changes as it is conditional on orientation.


nkrantz (approved):
👏 documentation

roxanagomez (approved):

PR #4188, opened 2024-12-13
fix(ci): supabase prod pipelines

krisantrobus:
add the supabase ci back in for prod. I had to update the dependencies to use the latest versions of Supabase which caused the generated types to update


PixeledCode (approved):

nkrantz (approved):
👏 thank you!

PR #4191, opened 2024-12-17
fix(select): Disabled Select should use color-background

PixeledCode:
https://paste-docs-git-fix-select-bg-twilio.vercel.app/components/select#disabled-select

### DIFF
--- a/packages/paste-core/components/select/src/Select.tsx
+++ b/packages/paste-core/components/select/src/Select.tsx
@@ -110,6 +110,7 @@ export const SelectElement = React.forwardRef<HTMLSelectElement, SelectProps>(
         width="100%"
         _disabled={{
           color: variant === "inverse" ? "colorTextInverseWeakest" : "colorTextWeaker",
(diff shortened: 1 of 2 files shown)


packages/paste-core/components/select/src/Select.tsx
```diff
@@ -110,6 +110,7 @@ export const SelectElement = React.forwardRef<HTMLSelectElement, SelectProps>(
         width="100%"
         _disabled={{
           color: variant === "inverse" ? "colorTextInverseWeakest" : "colorTextWeaker",
+          backgroundColor: "transparent",
```
nkrantz, line 113: do we have any way of testing this on windows? might be safer to set it to the actual desired token for disabled rather than transparent

krisantrobus (approved):

packages/paste-core/components/select/src/Select.tsx
PixeledCode: I used browerstack to test it and works for me too

PR #4200, opened 2024-12-20
chore(docs): fix sitemap

krisantrobus:
Previous deployment did not include all routes although worked locally. This PR should address the issue.

Decided to use the existing local sitemap we create for cypress as the complete page structure in the Vercel deployment cannot be used. Only the blog path can be accessed programatically.


nkrantz (approved):
nice, I like this solution. very clean

PixeledCode (approved):

PR #4213, opened 2025-01-20
fix(ci): update supabase and OpenAI version

krisantrobus:
Update version to resolve CI checks passing.

The latest version of Postgres (the underlying db) is updated to `postgres:15.8.1.026`. Any version above 15.7 has a later version of pgvector. This does not affect the hosted Supabase for stage or prod as migrations are applied based on changes made. All migrations are applied to local and CI instances as they are starting from scratch.

This change was made on Jan 14th in this [Commit](https://github.com/supabase/supabase/pull/31184/files).

Also updated the OpenAI version to use the new V2 assistant as it no longer works with V1.


packages/paste-website/src/pages/api/paste-assistant-message.ts
```diff
@@ -182,7 +182,7 @@ export default async function handler(req: NextApiRequest, res: NextApiResponse)
   /**
    * perform run on the assistant to process the newly added user message
    */
-  let run = await openai.beta.threads.runs.create(threadId, { assistant_id: assistantID });
+  let run = await openai.beta.threads.runs.create(threadId, { assistant_id: assistantID, model: "gpt-4o" });
```
krisantrobus, line 185: Updated the model, should have performance improvements

apps/backend/supabase/migrations/20230928013336_initial_schema.sql
```diff
@@ -1,4 +1,4 @@
-create extension if not exists "vector" with schema "public" version '0.5.0';
+create extension if not exists "vector" with schema "public";
```
krisantrobus, line 1: Will default to the pgvector of the DB used by the environment

nkrantz (approved):

nkrantz (approved):

serifluous (approved):

nkrantz (approved):

PR #4228, opened 2025-02-03
feat(website): update data viz docs with plugin and license info

serifluous:


packages/paste-website/src/pages/foundations/data-visualization/index.mdx
```diff
@@ -67,6 +66,8 @@ The easiest way to use the data visualization tokens in your charts is to use `u
 
 Be sure to include the [Highcharts Accessibility module](/foundations/data-visualization#adding-highcharts-accessibility-module) to ensure that your charts are accessible to all users.
 
+The design systems team maintains a Highcharts license for Twilio Console, SendGrid, Flex, Segment, and Admin. If you are a Twilio customer building with Highcharts, you'll need [your own license](https://shop.highcharts.com/).
```
krisantrobus, line 69: nit: I don't think we should mention "Admin" (which should be one admin). It's an internal tool and we may not want to expose that. We could add something like, for internal products please reach out to the team or something.

serifluous: good call—updated!

krisantrobus (approved):

nkrantz (approved):

PR #4236, opened 2025-02-07
feat(chart-provider): add component to wrap charts

krisantrobus:
Adds the new component `ChartProvider` used to manage state for the child chart implementations.

Added additional typedocs to the context attributes so developer understand the usage correctly when creating their own context providers or when using them in the custom components.

Added some storybooks and unit tests to be sure the callbacks and hooks function correctly.

[Storybooks](https://5e53448165911c0022e68c74-klmdnptvzy.chromatic.com/?path=/story/components-chartprovider--default) I am implementing a chart to fully test the context functions but the chart component is not included with this provider.


packages/paste-core/components/chart-provider/package.json
```diff
@@ -0,0 +1,61 @@
+{
+  "name": "@twilio-paste/chart-provider",
+  "version": "0.0.0",
+  "category": "data display",
```
krisantrobus, line 4: Not sure exactly what category we want to use for these charting components. What do you all think?

packages/paste-core/components/chart-provider/src/ChartProvider.tsx
```diff
@@ -0,0 +1,81 @@
...
+   * Overrides the default element name to apply unique styles with the Customization Provider
+   * @default null
+   * @type {BoxProps['element']}
+   * @memberof ChartProviderProps
+   */
+  highchartsOptions: Highcharts.Options;
+  pasteOptions?: never;
+}
+
+export type ChartProviderProps = HighchartsOptions;
```
krisantrobus, line 31: A bit redundant at this point as there is only one type but sets the groundwork for union types that will only allow users to either set `pasteOptions` or `highchartsOptions`

packages/paste-core/components/chart-provider/src/ChartProvider.tsx
```diff
@@ -0,0 +1,81 @@
...
+           * Any events we want to fire on changes to charts should be handled here
+           *
+           * Keeping track of chart type is a good way to trigger rerenders of other components inside ChartProvider
+           * There is an issue with useEffects on the chart object change only because React only does shallow
+           * comparisons and considers them the same object. Picking a value that is garuanteed to change
+           * between renders/types will overcome this.
+           */
+
+          // eslint-disable-next-line object-shorthand
+          redraw: function (this) {
```
krisantrobus, line 52: When using highcharts you have to use these verbose functions and this syntax. Tried it other ways but just doesn't work :/

packages/paste-core/components/chart-provider/src/ChartContext.tsx
```diff
@@ -0,0 +1,68 @@
...
+   * @memberof ChartContextProps
+   */
+  chartRef?: HTMLElement;
+  /**
+   * The current chart type. Used to trigger rerenders of other components inside ChartProvider.
+   *
+   * @type {string}
+   * @memberof ChartContextProps
+   */
+  chartType?: string;
```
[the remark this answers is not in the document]
krisantrobus: When working on the sandbox I found that custom components that rendered based on nested elements could not pick up changes to re-render. 

Specifically for this I had a legend that would look at the current series and see if there was a marker symbol used to show the correct marker in the legend. When switching between chart types it would no cause a re-render due to react only shallow comparing the chart context object. Picking up the change in the chart type options and listening for that fixed the issue. 

It is not needed for this release so I'm happy to leave out and investigate further if there is a better way to capture this to avoid a breaking change in future.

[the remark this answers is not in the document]
PixeledCode: I think we can leave this in since its very internal thing and probably will only cause minor change even if changed in future

packages/paste-core/components/chart-provider/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,44 @@
...
+      {children}
+    </ChartProvider>
+  );
+};
+
+describe("ChartProvider", () => {
+  it("should render", () => {
+    const { getByText, getByTestId } = render(<TestChartProvider>test</TestChartProvider>);
+    expect(getByText("test")).toBeDefined();
+    expect(getByTestId("chart-provider").getAttribute("data-paste-element")).toEqual("CHART_PROVIDER");
```
nkrantz, line 23: question: is this assertion adding anything that's not already being tested on line 41?

packages/paste-core/components/chart-provider/package.json
```diff
@@ -0,0 +1,61 @@
...
+  "status": "beta",
+  "description": "A component used to wrap an individual chart to store and share state to child charting elements.",
```
nkrantz, line 6:
```suggestion
  "description": "Chart Provider is a data visualization component used to wrap an individual chart to store and share state to child charting elements.",
```

packages/paste-core/components/chart-provider/__tests__/index.spec.tsx
```diff
@@ -0,0 +1,44 @@
...
+  it("should update chart type when options change", async () => {
+    const { getByTestId } = render(<UpdateChartTypeChangeOnRedraw />);
+
+    expect(getByTestId("chart-type-paragraph").textContent).toEqual("The chart type in the context is: line");
+
+    await act(async () => {
+      getByTestId("change-to-column-btn").click();
+    });
+
+    expect(getByTestId("chart-type-paragraph").textContent).toEqual("The chart type in the context is: column");
```
nkrantz, line 35: suggestion: in addition to these tests, maybe some tests to ensure that all the context values get correctly passed down?

nkrantz (commented):
Couple small comments, but code and stories look great! Are we leaving it as beta for now until more charting components are exported?

packages/paste-core/components/chart-provider/src/ChartContext.tsx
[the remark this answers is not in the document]
krisantrobus: I'll remove it for now, it's not being currently used. I will make the change in an upcoming commit.

packages/paste-core/components/chart-provider/__tests__/index.spec.tsx
krisantrobus: This is just making sure the children show up. A little unnecessary

packages/paste-core/components/chart-provider/__tests__/index.spec.tsx
krisantrobus: Currently chartType and options are the only ones being set. The options part is currently being tested as the redraw updates the chart type so I am testing that the modified options callback is working correctly and that the chartType gets updated.

We should keep expanding these tests though once we start adding in more options.

packages/paste-core/components/chart-provider/__tests__/index.spec.tsx
nkrantz: gotcha, non-blocking but would be nice to remove if it's redundant!

nkrantz (approved):

PixeledCode (approved):

PR #4250, opened 2025-02-19
Fix dark mode flicker docs

krisantrobus:
The current problem was that we were using a default theme in the useDarkMode hook that was the light mode. Then once the component was rendered it ran the use effect hook to pull the value from local storage and then updated the theme.

This change captures the cookies at the request level to the server and passes the theme value we store in the cookies to the page to use.

Still see the flicker but I think this is due to SSG (static generated pages at build time). I'm continuing to look at it.

image

### DIFF
--- a/packages/paste-website/src/components/site-wrapper/site-header/SiteHeaderSearch.tsx
+++ b/packages/paste-website/src/components/site-wrapper/site-header/SiteHeaderSearch.tsx
@@ -11,7 +11,8 @@ import { SiteSearch } from "../../site-search";
 const SiteHeaderSearch: React.FC = () => {
   const [isOpen, setIsOpen] = React.useState(false);
   const { breakpointIndex } = useWindowSize();
-  const isMacOS = navigator.platform.toUpperCase().includes("MAC");
+  // navigator is not available in SSR, so we need to check if it exists before using it
+  const isMacOS = typeof window !== "undefined" && navigator && navigator?.platform.toUpperCase().includes("MAC");
   const platformTriggerKey = isMacOS ? "Meta" : "Control";
 
   const onOpen = (): void => {
--- a/packages/paste-website/src/components/site-wrapper/site-header/SiteHeaderSearch.tsx
+++ b/packages/paste-website/src/components/site-wrapper/site-header/SiteHeaderSearch.tsx
@@ -11,8 +11,8 @@ import { SiteSearch } from "../../site-search";
 const SiteHeaderSearch: React.FC = () => {
   const [isOpen, setIsOpen] = React.useState(false);
   const { breakpointIndex } = useWindowSize();
-  // navigator is not available in SSR, so we need to check if it exists before using it
-  const [isMacOS, setIsMacOS] = React.useState(false);
+  // navigator is not available in SSR, settign a default until it renders to client
+  const [isMacOS, setIsMacOS] = React.useState<boolean>();
   const platformTriggerKey = isMacOS ? "Meta" : "Control";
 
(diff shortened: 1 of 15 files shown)


packages/paste-website/src/components/site-wrapper/site-header/SiteHeaderSearch.tsx
```diff
@@ -11,9 +11,14 @@ import { SiteSearch } from "../../site-search";
...
   const [isOpen, setIsOpen] = React.useState(false);
   const { breakpointIndex } = useWindowSize();
-  const isMacOS = navigator.platform.toUpperCase().includes("MAC");
+  // navigator is not available in SSR, so we need to check if it exists before using it
+  const [isMacOS, setIsMacOS] = React.useState(false);
   const platformTriggerKey = isMacOS ? "Meta" : "Control";
 
+  React.useEffect(() => {
+    setIsMacOS(typeof window !== "undefined" && navigator && navigator?.platform.toUpperCase().includes("MAC"));
+  }, []);
```
krisantrobus, line 20: This error came up after we removed the delayed rendering because window and navigator are only available when the page is running in the client. We will see the flicker from ctrl to cmd in the search but we can create a follow up ticket if it is needed in the future.

packages/paste-website/src/components/site-wrapper/site-header/SiteHeaderSearch.tsx
```diff
@@ -68,7 +73,7 @@ const SiteHeaderSearch: React.FC = () => {
               Search
             </Text>
           </Box>
-          {breakpointIndex === 0 ? null : (
+          {breakpointIndex === 0 || isMacOS === undefined ? null : (
```
krisantrobus, line 76: Do not render the keyboard shortcuts in the search bar until we know what OS they are using

packages/paste-website/src/hooks/useDarkMode.tsx
```diff
@@ -34,14 +53,10 @@ export const useDarkMode = (): UseDarkModeReturn => {
   };
 
   React.useEffect(() => {
-    const localTheme = SimpleStorage.get("theme") as ValidThemeName;
+    const cookieTheme = getCookie(themeCookieKey) as ValidThemeName;
 
-    if (window.matchMedia?.("(prefers-color-scheme: dark)").matches && !localTheme) {
+    if (window.matchMedia?.("(prefers-color-scheme: dark)").matches && !cookieTheme) {
       setMode(ValidThemes.DARK);
```
krisantrobus, line 59: As we are setting the default value from the cookie we don't need local storage anymore. We also don't want to use teh document cookie in this hook as we will still get the flickering issue. The reason this works now is that itis inside a useEffect hook and in functions so we know it would already be rendered in the client.

packages/paste-website/src/pages/patterns/object-status/index.mdx
```diff
@@ -139,7 +139,7 @@ export const getStaticProps = async () => {
         </Heading>
         <Paragraph marginBottom="space0">
           Create a composition with <Anchor href="/primitives/box">Box</Anchor>,{' '}
-          <Anchor href="/components/icon">Icon</Anchor>, and <Anchor href="/components/text">Text</Anchor>.
+          <Anchor href="/components/icon">Icon</Anchor>, and <Anchor href="/primitives/text">Text</Anchor>.
```
krisantrobus, line 142: For some reason the link test started failing, probably because it saw the 404 error now that we don't delay rendering

PR #4265, opened 2025-03-07
feat(storybook): working stories

krisantrobus:
Upgraded the version of Storybook to get it running again. required some additional dependency updates


.storybook/addons/google-analytics/register.js
```diff
@@ -1,4 +1,5 @@
-const { addons } = require("@storybook/addons");
+const { addons } = require("@storybook/manager-api");
```
krisantrobus, line 1: They deprecated the other package

.storybook/main.ts
```diff
@@ -57,5 +60,9 @@ const config: StorybookConfig = {
       },
     });
   },
+
+  docs: {
+    autodocs: false,
+  },
```
krisantrobus, line 66: Kept this option in as we can set true to create docs. It's pretty cool, creates a prop tables and gives a preview of all storybooks. I disabled though because they could be pretty long depending on how many stories a component has and could potentially hit the chromatic length limit and fail our tests. We can investigate in future but we didn't have it origionally so not required

.storybook/preview.tsx
```diff
@@ -281,7 +281,6 @@ export const decorators = [
 
 export const parameters = {
   isTestEnvironment,
-  actions: { argTypesRegex: "^on[A-Z].*" },
```
krisantrobus, line 284: deprecated

package.json
```diff
@@ -121,17 +121,16 @@
     "@octokit/graphql-schema": "^14.39.1",
     "@percy/cli": "^1.10.1",
     "@percy/cypress": "^3.1.2",
-    "@storybook/addon-a11y": "7.6.4",
-    "@storybook/addon-actions": "7.6.4",
-    "@storybook/addon-essentials": "7.6.4",
-    "@storybook/addon-interactions": "7.6.4",
-    "@storybook/addon-links": "7.6.4",
-    "@storybook/addons": "7.6.4",
```
krisantrobus, line 129: depricated

package.json
```diff
@@ -121,17 +121,16 @@
...
-    "@storybook/cli": "7.6.4",
-    "@storybook/react": "7.6.4",
-    "@storybook/react-vite": "7.6.4",
-    "@storybook/test-runner": "0.16.0",
-    "@storybook/testing-library": "0.2.2",
```
krisantrobus, line 134: depricated

packages/paste-core/components/sidebar/stories/full-composition/docs.stories.tsx
```diff
@@ -24,17 +24,11 @@ import {
...
 import { SearchBox } from "./components/SearchBox";
-import { SupportMenu } from "./components/SupportMenu";
-import { UpgradeBadge } from "./components/UpgradeBadge";
-import { UserDialogExample } from "./components/UserDialogSegment";
-import { WorkspaceSwitcherMenu } from "./components/WorkspaceSwitcher";
 
 // eslint-disable-next-line import/no-default-export
 export default {
-  title: "Components/Sidebar/FullCompositions",
+  title: "Components/Sidebar/Docs",
```
krisantrobus, line 31: We already had another storybook with the same name and the new version failed to compile with it. This is for the docs site so more appropriately named

PR #4292, opened 2025-03-31
chore(chat-composer): adjust spacing, make icon optional

nkrantz:
**Contributing to Twilio**


nkrantz (commented):
Technically this is a breaking change, removing the Media Object from the Chat Composer file. If anyone is using the Customization Provider to target "CHAT_COMPOSER_MEDIA_OBJECT" that will break. I think it's safe to mark this as a patch based on the likelihood of that being used, but leaving it open for discussion.

krisantrobus (approved):
LGTM, we're not pulling out anything form the SPI spec so while it may break customization the UI styling does feel like a patch. Thinking CustomizationProvider changes sit outside of the standard. 🚀

PixeledCode (approved):

PR #4295, opened 2025-04-02
Add form pattern

PixeledCode:


packages/paste-website/src/pages/patterns/form/index.mdx
```diff
@@ -0,0 +1,714 @@
...
+  <ListItem>
+    For <Anchor href="https://docs.google.com/document/d/1sYcP_i_YTwYWWdjKDQfkIdMj1NqyZV5c_8k4BxrEYc0/edit?pli=1&tab=t.0#heading=h.yi4naptd6s8d">Multi-step Forms</Anchor> and <Anchor href="/page-templates/wizard">Wizards</Anchor>, keep the primary button right-aligned, placed after secondary buttons to indicate the direction of the form. Refer to our <Anchor href="/page-templates/wizard#actions-footer">Wizards action footer</Anchor> guidelines for additional guidance.
+  </ListItem>
+</UnorderedList>
+
+## Behavior
+
+### Validation
+
+#### Validation on submit
```
jhanvi1399, line 831: Suggestion: Can we keep the demo left aligned and remove the excess margin here? The outer 24 px margin is fine and consistent with our other demos so thats okay. 

same suggestion goes for 
- inline validation demo (top, left, right margin)
- conditional field, main address, demo (top margin 48)

image

packages/paste-website/src/pages/patterns/form/index.mdx
```diff
@@ -0,0 +1,714 @@
...
+  </ListItem>
+  <ListItem>
+    If validation fails, do not display an error message while the user is typing. Keep the form field in a neutral state and consider using a checklist-style inline validation that marks rules as complete as the user types.
+  </ListItem>
+  <ListItem>
+    Use positive inline validation once the user has entered the data again.
+  </ListItem>
+</UnorderedList>
+
+### Errors
```
jhanvi1399, line 897: Issue(blocking): We discourage using placeholders. Should we avoid using them in our demos?

image

jhanvi1399, line 897: Issue: the anchors text should be the same as the form field name.
image

packages/paste-website/src/pages/patterns/form/index.mdx
```diff
@@ -0,0 +1,714 @@
...
+
+Use an [error Callout](/components/callout#error-callout) to communicate errors for a particular section of a page, or that applies to the whole page. Within the Callout, list errored fields as anchors and shift focus to the first errored field in the form. For additional guidance on how to compose error messages, refer to the [error state pattern](/patterns/error-state).
+
+### Confirmation and deletion
```
jhanvi1399, line 930: Issue: In the demo here of confirmation and deletion, the action "change" should be an anchor because I was thinking that should technically redirect the user to that field and allow them to edit. We don't need the redirection in demo but just that it should be an anchor instead of link-style button.

jhanvi1399 (changes requested):

packages/paste-website/src/pages/patterns/form/index.mdx
```diff
@@ -0,0 +1,714 @@
...
+    Check accessibility for each form element component—for example, follow Time Picker accessibility guidelines. You can also refer to the WCAG Forms Tutorial for more details.
+  </CalloutText>
+</Callout>
+
+## Ingredients
+
+### What’s in a form?
+
+The different parts of a form are:
+
```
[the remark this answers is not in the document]
PixeledCode: thats default margin bottom in paragraph component, it's not just before lists but everywhere

packages/paste-website/src/pages/patterns/form/index.mdx
jhanvi1399: Suggestion: Can we match the order of the error callout anchors to the sequence of the form fields? For example, after the country anchor, it should be city—not state.
image

krisantrobus (approved):

jhanvi1399 (approved):

PR #4304, opened 2025-04-09
feat(card): add fillWidth

krisantrobus:
Add the fillWidth prop to force stretch in flex containers


packages/paste-website/src/pages/components/card/index.mdx
```diff
@@ -124,6 +124,17 @@ In scenarios where there is a requirement for the card to fill the height of the
   code={FillHeightExample}
 />
 
+### Fill width
+
+Card width should fill the available space of the parent container. However, when wrapped in a flex container, the Card will shrink to fit the available space. To prevent this, you can use the `fillWidth` prop.
```
serifluous, line 129: suggestion/question: Should "flex container" be "flexbox" instead? I feel like the way it's worded right now could get confusing because we have Flex the product and Flex the component that's an implementation of flexbox.

packages/paste-website/src/component-examples/CardExamples.ts
```diff
@@ -123,6 +123,111 @@ export const FillHeightExample = `const CardExample = () => {
 
 render(<CardExample />);`.trim();
 
+export const FillWidthExample = `const CardExample = () => {
+  return (
+    <Grid gutter="space40" rowGap="space70" equalColumnHeights>
```
serifluous, line 128: suggestion: Can you make the Cards for this example also not fill height? (I think by removing `equalColumnHeights`? Just to make it super clear that this is a fill-width example.

packages/paste-website/src/pages/components/card/index.mdx
serifluous, line 129:
```suggestion
By default, Card width fills the available space of the parent container. However, when wrapped in a flexbox, Card shrinks to fit the available space. To prevent this, use the `fillWidth` prop.
```

krisantrobus: Very good point! I'll make the change

packages/paste-website/src/component-examples/CardExamples.ts
krisantrobus: I tried that and then it doesn't show the issue. Adding the `equalColumnHeight` is the thing that triggers the columns to use flex so without it both cards stretch the full width which we don't want

serifluous: ohhhh. ok yeah just ignore this suggestion then.

PixeledCode (approved):

serifluous (approved):

roxanagomez (approved):

PR #4400, opened 2026-06-26
chore: upgrade next to v16

nkrantz:
Upgrade next to version 16


packages/paste-theme-designer/package.json
```diff
@@ -16,7 +16,7 @@
     "@datadog/browser-rum": "^5.0.0",
     "@twilio-paste/core": "^21.0.0",
     "@twilio-paste/icons": "^13.0.0",
-    "next": "^15.2.0",
+    "next": "^16.2.9",
```
cvita, line 19: Is there a safe `v15` patch version we could try instead of the major version bump?

